PR #121, opened 2012-07-28
Fix CSHARP-541: Make BsonClassMap<TClass>.MapMember Typesafe

optimiz3:
Fix for CSHARP-541: Make BsonClassMap<TClass>.MapMember Typesafe
- Rebased on top latest changes from master
- Added a strongly typed BsonMemberMap
- Made MapMember variants return a strongly typed BsonMemberMap
  - This made it simpler to write member mapping code
    - Before: BsonClassMap.RegisterClassMap<MyType>(classMap => { classMap.AutoMap(); classMap.MapMember(myType => myType.MyField).SetShouldSerializeMethod(obj => !((MyType)obj).MyField != null); });
    - After: BsonClassMap.RegisterClassMap<MyType>(classMap => { classMap.AutoMap(); classMap.MapMember(myType => myType.MyField).SetShouldSerializeMethod(myType => myType.MyField != null); });
- Deprecated MapField, MapProperty, MapIdField, MapIdProperty, MapExtraElementsField, MapExtraElementsProperty, UnmapField, and UnmapProperty
  - In .Net it is forbidden to have have a property and field with the same name; having both *Field and *Property versions when the API was capable of determining what to do needlessly complicated the API.
  - The lambda versions were semantically incorrect as they were calling MapMember, which does not distinguish between fields and properties.
  - All of this code was refactored to a few small versions, which reduced bloat (for example, all versions called MapMember, which performs argument validation and freeze checking).
- Fixed a small validation bug in SetSerializeDefaultValue
- Eliminated the custom GetDefaultValue code in BsonMemberMap and replaced it with default(TMember)

### DIFF
--- a/Bson/Serialization/BsonClassMap.cs
+++ b/Bson/Serialization/BsonClassMap.cs
@@ -639,6 +639,22 @@ namespace MongoDB.Bson.Serialization
         }
 
         /// <summary>
+        /// Gets a member map (only considers members declared in this class).
+        /// </summary>
+        /// <param name="memberInfo">The member info.</param>
+        /// <returns>The member map (or null if the member was not found).</returns>
+        public BsonMemberMap GetMemberMap(MemberInfo memberInfo)
+        {
+            if (memberInfo == null)
+            {
+                throw new ArgumentNullException("memberInfo");
+            }
+
+            // can be called whether frozen or not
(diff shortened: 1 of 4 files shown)


IanWhalen:
Closing as per comment on https://jira.mongodb.org/browse/CSHARP-541

PR #122, opened 2012-07-31
Improve Bson Deserialization Performance by 10-15% (Refactor Discriminator Code)

alexnagy:
1. Ongoing profiling of the driver revealed that refactoring discriminator handling code was a strong candidate for performance improvement.
    * A fair bit of time was being spent in Type verification, DiscriminatorConvention lookup, Serializer lookup, and the associated locking code.
    * Of note, discriminator code was being executed during deserialization regardless of whether it was appropriate for the Type.
        * Specifically, only unsealed class types can have useful discriminators in .Net as only these types can be inherited.
        * Additionally, the implementation of the StandardDiscriminator was always evaluating for primitive types, however the only nominal type that is assignable at the CLR layer is System.Object, making much of the computation unnecessary.
2. To achieve performance and maintainability improvements the following was done:
    * Factor out discriminator code into a DiscriminatorSerializer wrapper serializer.
    * Only wrap serializers that can be discriminated.
    * Factored out previous performance related serializer caching work into a SerializerHolder class
        * Replaced lots of duplicate discriminator lookup -> serializer lookup -> serializer caching code with this class.
        * Also applied this class to generic collections that had been missed in previous rounds of performance improvement as they were not covered by benchmarks.
        * In several cases this reduced class size as these holders could be made static relative to the generic class types.
3. To improve the performance of discriminator lookups the following was done:
    * Combine BsonSerializer.__discriminatedTypes and BsonSerializer.__typesWithRegisteredKnownTypes into a single dictionary.
    * Make accesses to the new dictionary only require a lock for updating
4. Another minor performance improvement to reduce the overhead of BsonClassMapSerializer.DeserializeMember (one of the hottest methods)
    * Move the try/catch setup to outside the member loop in BsonClassMapSerializer.Deserialize, this way the try/catch only needs to be setup once per instance rather than once per member.
5. Another minor performance improvement to reduce the overhead of DictionaryGenericSerializer
    * Call dictionary.Add(key, value) instead of .Add(keyValuePair) as the second requires cloning a struct on the stack as the struct is passed by value
6. Net is that classes with primitive properties (i.e. almost all) benefit from this, as well as all generic collection types.
7. Finally, care was taken not to disturb any publicly facing APIs, as the aim was to make improvements turnkey.

Raw profiling data, I can provide hard VS profiler results and benchmark code as well if needed.

Iterations	master	optimiz3	Improvement
25000	10.3441722	9.3571325	10%
25000	10.4369589	9.4071719	10%
25000	10.5356442	9.4646407	10%
25000	10.4389251	9.409648367	10%

### DIFF
--- a/Bson/Bson.csproj
+++ b/Bson/Bson.csproj
@@ -150,6 +150,9 @@
     <Compile Include="Serialization\BsonSerializationInfo.cs" />
     <Compile Include="Serialization\Serializers\CollectionGenericSerializers.cs" />
     <Compile Include="Serialization\Serializers\CollectionSerializers.cs" />
+    <Compile Include="Serialization\Serializers\NullSerializer.cs" />
+    <Compile Include="Serialization\Serializers\SerializerHolder.cs" />
+    <Compile Include="Serialization\Serializers\DiscriminatorSerializer.cs" />
     <Compile Include="Serialization\Serializers\KeyValuePairSerializer.cs" />
     <Compile Include="Serialization\Serializers\DictionaryGenericSerializer.cs" />
     <Compile Include="Serialization\Serializers\DictionarySerializer.cs" />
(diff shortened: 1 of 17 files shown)


craiggwilson:
Hi Alex,

Our 2.0 driver refactored the BSON library and we believe this has been optimized there as well. We won't be doing anything to 1.x driver related to this.

Thanks so much for the PR.

PR #150, opened 2013-02-12
CSHARP-650 Support LINQ All by converting to !Any

darkiri:


craiggwilson:
Hi,

We've added interfaces underneath the existing BsonReader and BsonWriter classes in the 2.0 driver. This should satisfy the needs here.

Craig

jyemin:
We no longer support the original LINQ implementation that this is targeting, and CSHARP-650 has already been closed.

PR #152, opened 2013-02-28
CSHARP-692: Add support for the $comment meta operator

einaregilsson:
Adds support for setting the $comment option on a MongoCursor, and to use it in a LINQ query. This can be very useful during profiling to tag queries with a comment that shows up in system.profile.

Example usage:
```csharp
var result = collection.FindAll().SetComment("querytag").ToList();
```
and for LINQ:

```csharp
var result = collection.AsQueryable()
                       .Where(o => o.Field == 1)
                       .WithComment("querytag")
                       .ToList();
```

Our use case for this is basically that we have a lot of different clients accessing the same database. In some cases we see slow queries and we'd like to know which app, and which class is generating the query. We'd like to just use the logged in user name to distinguish between apps, but we go through mongos, and because of [SERVER-7538](https://jira.mongodb.org/browse/SERVER-7538) we can't get that info. So, instead we'd like to just tag all our queries with a $comment to be able to trace where they're coming from.


craiggwilson:
Hi Einar,

Thanks for the PR. Our 2.0 LINQ support (against the new API) will have a more standard way of attaching behaviors like this. We won't be doing anything to the old linq provider.

Thanks for the PR.

PR #176, opened 2014-02-28
CSHARP-919 MongoCursor.Query setter

garethbudden:
Not sure if/where to create a test for this, it seems a very simple change?


craiggwilson:
Hi Gareth,

Thanks so much for the PR. We've added this ability in the new 2.0 API, but won't be supporting it in the old API or 1.x driver.

Craig

PR #179, opened 2014-03-21
CSHARP-926 LINQ queries with properties of type byte enum

lshako:
if (unaryExpression.Type == Enum.GetUnderlyingType(enumType)) is false in case of byte, short etc, because you convert these type to int


jyemin:
CSHARP-926 was closed long ago.  Closing this out as well.

PR #186, opened 2014-08-12
CSHARP-1029: Making properties virtual for easier testing

sloncho:
Make properties virtual, so they can be mocked for unit testing IBsonSerializer implementation.


craiggwilson:
Hi,

We've added interfaces underneath the existing BsonReader and BsonWriter classes in the 2.0 driver. This should satisfy the needs here.

Craig

PR #187, opened 2014-08-21
CSHARP-1046: Added Update<MyDocument>.SetPositional

sloncho:


craiggwilson:
Hi,

This solution really only helps in set, but the positional operator is useful in many other update operators as well. We've solved this in the 2.0 driver by supporting a -1 indexer or ElementAt(-1) method call.

Thanks for the PR.

PR #190, opened 2014-10-06
CSHARP-1080: fix unexpected change non id field name in BSON document

ExM:
https://jira.mongodb.org/browse/CSHARP-1080

I want to serialize following class:

``` csharp
public class TestModel_CustomNotId
{
    [BsonElement("id")]
    public ObjectId Id { get; set; }
}
```

I expected Id field to be serialized as "id",
but the result BSON document is containing "_id" field.


jyemin:
This is no longer fresh enough to merge.  We will re-evaluate a fix if CSHARP-1080 is ever scheduled, but currently there are no votes for that issue.

PR #191, opened 2014-10-08
Merge pull request

Oger-Me:
Merge pull request


craiggwilson:
Hi,

I can';t completely tell what this is, but it appears as though we should ignore null queries. In that case, we are not going to do that as, most likely, this is an error in your application and would produce unexpected results for certain users.

Craig

PR #192, opened 2014-10-08
Add NUUID as an alternative to CSUUID

Eibwen:
Robomongo defaults to UUIDs being called NUUID, even though it does not support that class, it needs this script
The toNUUID function is exactly the same as toCSUUID except the string concat, so they could easily call a common function to convert the binary, but I neglected to make that change.


jyemin:
This is likely no longer relevant, so closing out.  If this is still of interest to anyone, please open a Jira issue to re-start discussion.

PR #193, opened 2014-10-28
The query builder was generating invalid queries when ANDing two or more

carducci:
conditions that both exist in a in an array of objects.

Assuming a document like this
{
    _objectid:...,
    Item:555,
    MetaData: [{
        Foo: Bar
    },
    {
        Baz: Bat
    }]...
}
Query.And with two conditons in that array would generate invalid syntax.
I found the solution was to promote to $and everytime. There is certainly
a more elegant solution, but this solved my immediate problem with no ill
effects.

### DIFF
--- a/src/MongoDB.Driver/Builders/QueryBuilder.cs
+++ b/src/MongoDB.Driver/Builders/QueryBuilder.cs
@@ -893,7 +893,7 @@ namespace MongoDB.Driver.Builders
                     }
                     else
                     {
-                        query.Add(clause);
+                        PromoteQueryToDollarAndForm(query, clause);
                     }
                 }
             }


craiggwilson:
Hi,

I'm still not sure what bug we are actually fixing. Your example above doesn't give anything but a sample document, but not example of the invalid query. In addition, there are no tests. If you'd like us to consider this, I need more information. Also, if this is truly a bug, please file a JIRA ticket under the CSHARP project at jira.mongodb.org.

Craig

Thieund:
I encountered the same problem with driver versions 2.1.x and 2.2.0. It worked with 2.0.1.
The following code will result in "Specified cast is not valid":
            var builder = Builders<Foo>.Filter;
            Expression<Func<Foo, bool>> firstFilter = (x => x.Prop== 2067);
            var finalFilter = builder.Eq(x => x.IsDeleted, false) & builder.Where(filter);
            return await Collection.Find(finalFilter).FirstOrDefaultAsync();

craiggwilson:
Thanks Thiend, we'll see if we can repro.

rstam:
I am unable to reproduce this using either 2.1.1 or 2.2.0. I had to make some assumptions to get the sample code you provided to compile. I used this class:

    public class C
    {
        public int Id;
        public int Prop;
        public bool IsDeleted;
    }

And this code to test it:

    var client = new MongoClient("mongodb://localhost");
    var database = client.GetDatabase("test");
    var collection = database.GetCollection<C>("test");

    await collection.DeleteManyAsync(new BsonDocument());
    await collection.InsertOneAsync(new C { Id = 1, IsDeleted = false, Prop = 2067 });

    var builder = Builders<C>.Filter;
    Expression<Func<C, bool>> firstFilter = x => x.Prop == 2067;
    var finalFilter = builder.Eq(x => x.IsDeleted, false) & builder.Where(firstFilter);
    var document = await collection.Find(finalFilter).FirstOrDefaultAsync();

With both 2.1.1 and 2.2.0 the Find/FirstOrDefaultAsync returned the matching document.

Thieund:
Sorry Rstam, now I'm clear about what happening, and I'm not sure if it is a bug. Using the class C you wrote above where Prop is an int, if instead of "Expression<Func<C, bool>> firstFilter = x => x.Prop == 2067;" I use this:
long prop = 2067;
Expression<Func<C, bool>> firstFilter = x => x.Prop == prop;
then the exception {"Specified cast is not valid."} is thrown.

But it actually did work in version 2.0.1. Now it only works when I explicitly cast prop to int as following:
Expression<Func<C, bool>> firstFilter = x => x.Prop == (int)prop;

So it's the casting problem, not the AND. Sorry about that.

rstam:
Thanks for the additional information. I've created a JIRA ticket for this:

https://jira.mongodb.org/browse/CSHARP-1509

The issue is a result of an implicit conversion from Int32 to Int64. It can be more minimally reproduced with the following query:

    var builder = Builders<C>.Filter;
    var filter = builder.Where(x => x.Prop == 2067L); // note: x.Prop is an int
    var document = await collection.Find(filter).FirstOrDefaultAsync();

Since the issue turned out not to be related to And after all I'm going go close the pull request. Thanks for finding this issue though!

PR #195, opened 2014-12-01
Added the support of the OfType linq method

phenixdotnet:
Added the support of the OfType linq method to support automatically filtering on subdocument type
All advices and/or remarks are welcome :)


craiggwilson:
Thanks pheonixdotnet. We are going to get our 2.0 out the door and then will be able to include this into LINQ for 2.1.

PR #196, opened 2015-01-21
replace GeoNearOptions.null with a null and avoid the compiler warning

DavidHoerster:
removed the `IMongoGeoNearOptions.Null` reference as this was causing a "warning as error" result during local build.  Used `null` instead.


craiggwilson:
Hi,

As this applies to the 1.x driver, we won't be making a change like this.

Thanks.

luizstudy:
Hello could someone please help me !? Using Expand Object I can read an xml file complex. Must now read this Expando Object and turn again in a xml again.
Thanks for listening!

craiggwilson:
I'm sorry Luiz, this is very much no the right place to discuss that. Perhaps you should try stackoverflow.com

luizstudy:
I`m sorry! Thanks!

PR #197, opened 2015-02-01
CSHARP-1178: Treat capped collection max size as long instead of int.

I3arnon:


craiggwilson:
Awesome. I've thrown this into code review.

I3arnon:
Very cool...
So, is that enough to enter the contributors list, or is that too small?

craiggwilson:
Yeah. I just forgot. It's done.

PR #201, opened 2015-04-08
add null checks to ClusterBuilder

scottcarr:
Hello, I'm working an a static analysis tool that can find missing null checks.  It suggested you add these checks.  There are a couple methods in ClusterBuilder that take a function (called configurator) as the argument.  ConfigureServer and ConfigureSsl don't check if configurator is null, but the others do.


jyemin:
This hasn't caused issues in practice, so while it's technically correct it seems unnecessary.

PR #202, opened 2015-04-18
Fixes most issues with binding.

Mohsen7s:
This fixes most issues like System.InvalidOperationException while trying to convert ObjectId to string and vice-versa with binding MongoDb objects in ASP.NET data controls.


Mohsen7s:
The conversation is done by data controls like GridView in ASP.NET not user code.
You can replicate it with following class:
    public class MongoDbObj1
    {
        public ObjectId Id { get; set; }
        public string Field1{ get; set; }
    }
Now simply try to bind the class with ObjectDataSource and specify Select & Update method.
As you know in web environment user inputs are assumed as strings and then converted to their real type. So GridView fail when tries to translate a string back to ObjectId to save it.

PR #208, opened 2015-07-12
Overload methods added for Find, CountAsync, DeleteManyAsync

sergun:
There is not so convenient to make calls like `colleciton.Find(new BsonDocument())` and generates load of redundant code. So to make calls cleaner I've added overloads for `Find()`, `CountAsync()` and `DeleteManyAsync()` methods.

### DIFF
--- a/src/MongoDB.Driver/IMongoCollectionExtensions.cs
+++ b/src/MongoDB.Driver/IMongoCollectionExtensions.cs
@@ -18,6 +18,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Threading;
 using System.Threading.Tasks;
+using MongoDB.Bson;
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Linq;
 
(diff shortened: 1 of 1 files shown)


craiggwilson:
Hi sergun,

Thanks for the PR. All the drivers have made a very explicit decision to require you to indicate that you actually want to find/count/delete/etc... all documents. While this is useful in testing, in practice, it's outside the norm to want to do that. As such, I'm going to close the PR.

Thanks again,
Craig

sergeyzwezdin:
Hi @craiggwilson 
Thanks, I understand your decision, but did you see how production code is changed?
Load of mess now like this:

```await database.Find(new BsonDocument()).ToListAsync()```

It makes code difficult to read and maintain.

I guess there will be some 3rd party helper then. But if so, why not to include this into base API?

craiggwilson:
I completely understand where you're coming from, but as I alluded to, most production code won't have that at all because most production code doesn't find/delete/count *everything* in the database. There will almost always be an actual filter present. For this reason, we want to make it a little more difficult to do what might be damaging to your data or to system performance.

In the next update, there will be a Builders<T>.Filter.Empty so you don't have to new up a new BsonDocument all the time. It certainly doesn't make it easier to read though, although it would be more clear to a maintainer.

Hope that makes sense.

sergeyzwezdin:
Ok, thanks for explanation.

PR #209, opened 2015-07-24
First example should get collection from db.

IonoclastBrigham:


rstam:

While attempting to merge this pull request I discovered that this fix had already been made.

PR #211, opened 2015-08-13
Moving "Bulk Write" documentation to reference

phber:
The only documentation on "Bulk Writing" is currently under "quick tour" which makes it difficult to find. It would be more accesible in the "Writing" section under reference.


craiggwilson:
Hi phber,

Thanks for pointing this out. It was an oversight.  Rather than move the documentation, for now it'd be best to just copy it. I've created a [ticket](https://jira.mongodb.org/browse/CSHARP-1377) to improve this. If you'd like to resubmit with the content copied, we can pull it in.

Craig

phber:
Added commit with a temporary copy instead.

/Phber

PR #212, opened 2015-08-13
Update reference/bson/mapping/index.md to fix a grammatical error.

adamaverycole:
Fixed what appears to be a grammatical mistake.  "...it value is..." to "...its value is..."


rstam:
Rebased on master and merged.

PR #215, opened 2015-09-06
Bug Report

AlexGeller:
As I known, MongoDB collection's update function has four parameters, it just like this:

db.users.update({"birthday" : "10/13/1978"}, {"$set" : {"gift" : "Happy Birthday!"}}, false, true)

but in the Driver code, it put the third and fourth parameters in UpdateFlags enum. I think this is wrong to put two irrelevant parameters in the same enum, it is difficult to distinguish their function. In addition, MongoCollection‘s Update function below：

public virtual WriteConcernResult Update(IMongoQuery query, IMongoUpdate update, MongoUpdateOptions options)

It has a Bug too. It use UpdateFlags to initinal "isMulti" and "isUpsert" options in the UpdateRequest class.

            var isMulti = (options.Flags & UpdateFlags.Multi) == UpdateFlags.Multi; //Bug。
            var isUpsert = (options.Flags & UpdateFlags.Upsert) == UpdateFlags.Upsert; //Bug。
            var writeConcern = options.WriteConcern ?? _settings.WriteConcern ?? WriteConcern.Acknowledged;

            var request = new UpdateRequest(UpdateType.Unknown, queryDocument, updateDocument)
            {
                IsMulti = isMulti,
                IsUpsert = isUpsert
            };

There is impossible to set update multi exist documents in MongoDB, or insert a new document into Collection. because options "IsMulti" and "IsUpsert " are always just one to be true.


craiggwilson:
Hi Alex,

Bug reports should be filed here: https://jira.mongodb.org. 

In this particular case, I believe our code is correct. [UpdateFlags|https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Legacy/UpdateFlags.cs] is a [Flags enum|https://msdn.microsoft.com/en-us/library/system.flagsattribute(v=vs.110).aspx], which means both options can be set to it at the same time. Checking to see if any one option is set is done with a bitwise AND of the correct flag and checking it's equality. As such, this is what we've done. In order to pass both flags in at the same time, you would call Update(query, update, UpdateFlags.Multi | UpdateFlags.Upsert);

I don't totally disagree regarding your assessement. In fact, the API for this is totally different in the non-legacy code of 2.0. However, this exists in the legacy package of our 2.x codebase (and in the 1.x branch) and therefore has been technically deprecated. We will not be making improvements on this part of the code. 

If you still feel this is a bug, please file a ticket at https://jira.mongodb.org under the CSHARP project and we'll happily take a look.

Thanks,
Craig

PR #217, opened 2015-09-30
CSHARP-837: Add support for CompareTo in the legacy predicate translator

Nashenas88:
I'm aware that [CSHARP-837](https://jira.mongodb.org/browse/CSHARP-837) has already closed, but I would like to add it to the legacy version of the driver since it's useful not just to `string`, but also to all types that use `CompareTo`, e.g. `Guid`.

### DIFF
--- a/src/MongoDB.Driver.Legacy/Linq/Translators/PredicateTranslator.cs
+++ b/src/MongoDB.Driver.Legacy/Linq/Translators/PredicateTranslator.cs
@@ -288,6 +288,12 @@ namespace MongoDB.Driver.Linq
                 return query;
             }
 
+            query = BuildCompareToQuery(variableExpression, operatorType, constantExpression);
+            if (query != null)
+            {
+                return query;
+            }
+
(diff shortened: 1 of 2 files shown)


craiggwilson:
Thanks @Nashenas88.  We'll discuss whether or not we want to make these types of adjustments to the legacy API. You're PR looks great though, and thanks for including tests!

Nashenas88:
I fixed the tests to reference the right variable. There are other errors in the commit  that seem completely unrelated to my change.

@craiggwilson Thank you! For what it's worth, we have some concerns with the API of the 2.0 non-legacy version of the driver, (mainly [this one](https://jira.mongodb.org/browse/CSHARP-1428)). Not having the `WriteResult` really makes the code for handling errors and logging more complicated.

jyemin:
Hi @Nashenas88 sorry to leave this sitting for such a long time.  At this point we don't plan to make any further changes in the legacy LINQ implementation.

PR #219, opened 2015-10-15
Adding a new BSON serialization provider to be able to determine that…

mfidemraizer:
Because some requirements in my project, I had to create a base class derived by some domain objects which implement `IDictionary<TKey, TValue>`, but these domain objects are still POCOs and I want them to be serialized using regular BSON class map serializer. 

When I ran some test I found some issue: my domain object wasn't getting an auto-assigned GUID. After clonning the driver source code, I debugged the code step by step and I found that my domain objects were serialized as collections rather than POCOs, because they inherit a base class which implements `IDictionary<TKey, TValue>` as I mentioned above.

Then, I thought, *what about a way to configure that certain types must be serialized as class maps to solve my issue?*, and I implemented a serialization provider `ForceAsBsonClassMapSerializationProvider` to fix my problem.

For example, this new provider may be used as follows:

    public class SomeImpl : IDictionary<string, object> 
    {
        // Imagine that I've implemented all IDictionary<string, object> members
    }

    BsonSerializer.RegisterSerializationProvider(new ForceAsBsonClassMapSerializationProvider(typeof(IDictionary<string, object>)));

It can be also registered a concrete type to be forced as a BSON class map:

    BsonSerializer.RegisterSerializationProvider(new ForceAsBsonClassMapSerializationProvider(typeof(SomeImpl)));

Or also both:

    // Order doesn't matter...
    BsonSerializer.RegisterSerializationProvider(new ForceAsBsonClassMapSerializationProvider(typeof(IDictionary<string, object>), typeof(SomeImpl)));

Because I feel that other developers might encounter the same limitation, I believe that it would be a good point merging this simple serialization provider. 

Thank you in advance and I'll look forward for your comments, suggestions...!

### DIFF
--- a/src/MongoDB.Bson/MongoDB.Bson.csproj
+++ b/src/MongoDB.Bson/MongoDB.Bson.csproj
@@ -187,6 +187,7 @@
     <Compile Include="Serialization\CreatorMapDelegateCompiler.cs" />
     <Compile Include="Serialization\ExpressionVisitor.cs" />
     <Compile Include="Serialization\BsonSerializerRegistry.cs" />
+    <Compile Include="Serialization\ForceAsBsonClassMapSerializationProvider.cs" />
     <Compile Include="Serialization\IBsonDictionarySerializer.cs" />
     <Compile Include="Serialization\IBsonPolymorphicSerializer.cs" />
     <Compile Include="Serialization\IBsonSerializerRegistry.cs" />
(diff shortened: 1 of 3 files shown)


rstam:
This is an interesting approach and the code looks OK to me.

We're not sure at this point whether this is something that should be in the official driver or not. This might be the kind of thing that makes more sense to have in your application.

mfidemraizer:
@rstam Hey, thank you for your comments.

BTW, now I'm going to drop this pull request and add a new one which also includes another feature, and I believe that it will add more value to the current one, and if you want, we can continue this discussion in the new pull request.

mfidemraizer:
@rstam  See #220 . If you want, you can copy-paste your comment, and we can continue the discussion there :+1:

mfidemraizer:
@rstam Hey, please review #221 . I'm sorry but I had to close the other pull request because I was contributing in the wrong way. Now I've forked and created a feature branch to keep things simpler and be able to contribute more in the future. You'll also note that #221 provides an additional feature compared to #219.

Thanks in advance and I'll look forward for your comments there...

PR #223, opened 2015-10-27
ToHexString sb.Append(b.ToString("x2"));

HellDeveloper:
This method is faster


craiggwilson:
Thanks. I have a couple of questions:

1. Have you measured it? If so, how much faster? 
2. Do the tests pass after making the change? 
3. Is there any reason to leave the comment in there?

jyemin:
Closing as we never heard back from the submitter.

PR #224, opened 2015-10-29
CSHARP-1460: added UseLocalTime option in JsonWriterSettings

stefanocastriotta:
I added an option on JsonWriterSettings, to serialize DateTimeFields using LocalTime instead of UTC. I wrote a unit test to make sure it doesn't break anything and seems ok.

https://jira.mongodb.org/browse/CSHARP-1460


jyemin:
As mentioned on CSHARP-1460, we have a more general solution to this now, so I'm going to close this out.

PR #225, opened 2015-11-15
CSHARP-1478: BsonClassMap.Reset() is not consistent with default constructor

bursauxa:
- fixed BsonClassMap.Reset() method to reflect default constructor behavior regarding IgnoreExtraElements property.
- updated associated test to compare results of the Reset() method with a fresh instance, rather than predefined results.

### DIFF
--- a/src/MongoDB.Bson.Tests/Serialization/Attributes/BsonAttributeTests.cs
+++ b/src/MongoDB.Bson.Tests/Serialization/Attributes/BsonAttributeTests.cs
@@ -14,7 +14,6 @@
 */
 
 using System.Linq;
-using MongoDB.Bson;
 using MongoDB.Bson.Serialization;
 using MongoDB.Bson.Serialization.Attributes;
 using MongoDB.Bson.Serialization.IdGenerators;
@@ -26,7 +25,6 @@ namespace MongoDB.Bson.Tests.Serialization.Attributes
     public class BsonAttributeTests
     {
         [BsonDiscriminator("discriminator", Required = true)]
-        [BsonIgnoreExtraElements(false)]
         public class Test
         {
             [BsonDefaultValue("default1")]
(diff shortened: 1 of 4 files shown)


bursauxa:
I had a look at the test failures (I must admit I had only run the ones in the Bson assembly on my box). They give the impression that, before the change in BsonClassMap, these tests were working thanks to a weird side effect, as the anonymous classes throughout the fixture do not have an _id field. Why the serializer was fine with no _id field in anonymous classes despite requiring it in regular classes, I do not know yet.

The issue seems too complicated to grasp at once, I will have a deeper look this evening.

bursauxa:
... and how deeper it was !

I have identified the following points, please correct me if I am mistaken:

- my initial bug report was inaccurate : the common BsonClassMap constructor calls Reset() as its last line, so by definition Reset() has results identical to calling the constructor.

- what does *not* have identical results is calling the static methods LookupClassMap(...), or RegisterClassMap(...). Transparent creations, such as when instantiating a MongoCollection, are done this way. Indeed, this call ends up applying class-defined conventions. Since IgnoreExtraElementsConvention(flase) is present by default, any instance of BsonClassMap obtained in the "normal" way will have this set to false.

- on the other hand -- and here we are coming to the failed build! -- there is one case in which BsonClassMap instances are created separately: in SerializerBuilder.BuildProjectedSerializer. There is a nice comment block explaining the rationale behind not using the global registry, and it also makes sense that conventions would not be applied for this type of objects.

**Summary**

The existing implementation has BsonClassMap.IgnoreExtraElements default to true, and then being overriden to false by the default convention pack.

The only scenario I identified where defaulting to true serves a purpose is for the translation of projection expressions.

In my opinion, this should be inverted: the defaults of BsonClassMap and the default convention pack should be aligned, with translations being treated as separate, specific exceptions.

PR #226, opened 2015-11-19
The GroupBy query operator is not supported

jakey188:
Exception error： MongoDB.Driver.Legacy.dll 
The GroupBy query operator is not supported.

my code:
 public void TestMethod1()
        {
            var db = new MongoDbContext();
            var query = (from c in db.GetCollection<UserLoginLog>().AsQueryable()
                select c).GroupBy(c => c);
            var list = query.ToList(); // execute query
        }

It's why?

### DIFF
--- a/Docs/landing/data/releases.toml
+++ b/Docs/landing/data/releases.toml
@@ -1,6 +1,6 @@
 current = "2.0.1"
 [[versions]]
-  version = "2.1.0-rc1"
+  version = "2.1.0"
   status = "rc"
   docs = "http://mongodb.github.io/mongo-csharp-driver/2.1/"
   api = "http://api.mongodb.org/csharp/2.1"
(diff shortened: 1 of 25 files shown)


craiggwilson:
@jakey188 This isn't a great place to ask questions. You've created a pull request somehow... 

GroupBy is certainly supported. However, not this version of GroupBy. Why? Because MongoDB doesn't do grouping the same way that this operator is works. Hence, you'll need to do something else with it, like project a transformation. Our documentation gives an example: http://mongodb.github.io/mongo-csharp-driver/2.1/reference/driver/crud/linq/#group.

craiggwilson:
@jakey188 I missed your stack trace. GroupBy is not supported in the Legacy API. You'll need to use the new API and it's LINQ support to use GroupBy.

PR #227, opened 2015-12-04
Fix for NullReferenceException for seekable download stream:

rquackenbush:
Some potential fixes for seekable download stream.
- Checking for _chunk == null before starting.
- added logic for cacluating the the last chunk number.

https://jira.mongodb.org/browse/CSHARP-1489


rstam:
I ran into some trouble merging this because line feeds weren't consistent.

Did you set core.autocrlf to true before cloning the repo to your machine?

Just a heads up for any subsequent pull requests.

Thanks!

rquackenbush:
Apologies - I didn't do that. We run with a different configuration for our environment by default. 

I'll change that if I submit any more pull requests.

Thanks for accepting my pull!

PR #229, opened 2015-12-11
CSHARP-1503: Mark IMongoQueryProvider as public

potz:
Would it be possible to have the [IMongoQueryProvider interface ](https://github.com/mongodb/mongo-csharp-driver/blob/v2.2.0/src/MongoDB.Driver/Linq/IMongoQueryProvider.cs) marked as public?

We are building a Framework that has built-in Async LINQ support, and for our mongodb data adapter it would make for extremely cleaner code if we could access ExecuteAsync<T> directly, instead of having to rewrite all the CountAsync, SumAsync etc (that are already coded inside the core of our framework).

This interface is well documented, and does not have any properties of types that are not exposed, so it seems to me that it can be easily changed to public. Of course that means keeping backwards compatibility on the API from now on, but I think it can be useful for users of the C# driver in general to have access to it.

### DIFF
--- a/src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
+++ b/src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
@@ -24,7 +24,7 @@ namespace MongoDB.Driver.Linq
     /// <summary>
     /// An implementation of <see cref="IQueryProvider" /> for MongoDB.
     /// </summary>
-    internal interface IMongoQueryProvider : IQueryProvider
+    public interface IMongoQueryProvider : IQueryProvider
     {
         /// <summary>
         /// Gets the collection namespace.


tetious:
This would also be helpful for mocking IMongoQueryable for unit tests that hit AsQueryable(). It is extremely painful to do this right now. :(

tetious:
Is anyone going to comment on this one way or another? This has come up again for me. It is a requirement for mocking many of the linq operators that cast to this interface. (such as Any() or Count()).

I'll make and maintain a fork if you aren't going to do this, but it would be nice to at least get a response.

potz:
Hi @tetious, 

You might want to follow the issue on Jira, where they did answer my original question:

https://jira.mongodb.org/browse/CSHARP-1503

But basically, they're considering but it won't happen before v2.3.0, if at all.

craiggwilson:
Sorry for not replying. Got lost in the shuffle.

I actually don't understand how this would be useful. We don't expose this anywhere (obviously), hence mocking it shouldn't be necessary. I would assume that you could simply mock IQueryProvider and be fine. Perhaps you guys could provide an example of what exactly you are doing that would require this change?

tetious:
Thanks @craiggwilson. I'm happy to explain. It is possible I'm doing something dumb. ;)

I have mocked IQueryProvider, and in most cases that works well. In a few cases, though, with some of the *Async operators, you are leaking IMongoQueryProvider, and needfully as far as I can tell, as they rely on ExecuteAsync.

Example from MongoQueryable:
```csharp
        public static Task<bool> AnyAsync<TSource>(this IMongoQueryable<TSource> source, CancellationToken cancellationToken = default(CancellationToken))
        {
            return ((IMongoQueryProvider)source.Provider).ExecuteAsync<bool>(
```

This means I have to use the non-async versions in order for them to be testable. Does this make sense?

craiggwilson:
Ah... I see what you're doing. Ok, fair enough. Well, as @potz  said, we are tracking this in the JIRA ticket. 2.3 will be the earliest it can come out as it's a public API change. 2.3 is tied to support for .NET Core, so.

tetious:
Just pinging on this again, as 2.3 came out and this may have been missed. It would be really helpful to have this to improve testability when using AsQueryable, and the change is so simple and already provided in a helpful PR. Please consider applying.

danielbiloglav:
Any update on this issue? I'm also in need of this in order to fully unit test async operations.

daviddesmet:
Is this still going to be considered? I'm asking since I hit the same situation regarding the testability when using `AsQueryable`.

DmitryLukyanov:
Hi, @daviddesmet , 

Thank you for the provided suggestion, but we have decided not to accept it. Mocking a third-party library - such as the MongoDB .NET Driver - is considered an anti-pattern because it hides changes in the mocked behaviour. The generally accepted solution is to encapsulate the dependency behind your own interface, which you can then mock and integration test. There has been much written on this topic, but this blog post is a good start with references to additional resources on the topic: 

https://dev.to/satansdeer/dont-mock-what-you-dont-own-cd6

If there is a particular problem that you are trying to solve that can't be solved using the encapsulation technique, please elaborate further as there may be other solutions to consider.

Sincerely,
MongoDB .NET Driver Team

daviddesmet:
Hi @DmitryLukyanov,

In my use case I'm trying to test the DAL. Here's a simple example of a method:

```csharp
// UserManager<TUser>

public virtual async Task<TUser> FindByPhoneNumberAsync(string phoneNumber, CancellationToken cancellationToken = default)
{
    return await Store.AsQueryable().Where(x => x.PhoneNumber == phoneNumber).SingleOrDefaultAsync(cancellationToken);
}
```

As you can see, this is a simple query which finds a User by the associated phone number (if there's any).
Trying to mock the `Store` which is a `IMongoCollection<TUser>` throws an exception on the `Where` clause and also on the `SingleOrDefaultAsync` one if I move the expression to there due to `IMongoQueryable<T>` and `IMongoQueryProvider` respectively.

A workaround, as already stated is to use non-async versions so we can have it testable which is not desirable.

I have quite a lot of repositories which ain't tested due to this limitation.
It would be nice if there's an option for us to be able to test them, perhaps maybe an In-Memory only store like EF offers so we don't have to mock them.

DmitryLukyanov:
Hey @daviddesmet ,
you still can mock the repository methods themselves and write integration tests based on this. Mocking a LINQ provider itself can hide details whether actual LINQ expression can be successfully translated into MQL or no.
So, we still think that it's better to avoid mocking 3rd party dependencies like Mongo driver.

PR #230, opened 2015-12-31
Bugfix CreateCollection doesn't return and throws NullReferenceException

i3arnon:
`CreateCollectionAsync` returns a `Task` in these cases.
I guess when creating the sync version the return was dropped as there's no longer anything to return.
This causes a `NullReferenceException` when `options == null`


rstam:
I've created a JIRA ticket for this bug:

https://jira.mongodb.org/browse/CSHARP-1524

PR #232, opened 2016-01-28
CSHARP-1447: Allow serialization of generic IDictionary implementations with non-public constructors when serializing to an interface definition

jhadwen:
- Fixes regression, now allows serialization of dictionaries that do not have public parameterless constructors

### DIFF
--- a/src/MongoDB.Bson/Serialization/Serializers/DictionaryInterfaceImplementerSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/DictionaryInterfaceImplementerSerializer.cs
@@ -13,6 +13,7 @@
 * limitations under the License.
 */
 
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using MongoDB.Bson.Serialization.Options;
@@ -164,7 +165,7 @@ namespace MongoDB.Bson.Serialization.Serializers
         DictionarySerializerBase<TDictionary, TKey, TValue>,
         IChildSerializerConfigurable,
         IDictionaryRepresentationConfigurable<DictionaryInterfaceImplementerSerializer<TDictionary, TKey, TValue>>
-            where TDictionary : class, IDictionary<TKey, TValue>, new()
+            where TDictionary : class, IDictionary<TKey, TValue>
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="DictionaryInterfaceImplementerSerializer{TDictionary, TKey, TValue}"/> class.
@@ -271,7 +272,7 @@ namespace MongoDB.Bson.Serialization.Serializers
         /// <returns>The instance.</returns>
         protected override TDictionary CreateInstance()
         {
-            return new TDictionary();
+            return Activator.CreateInstance<TDictionary>();
         }
 
         // explicit interface implementations
--- a/src/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
+++ b/src/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
@@ -14,7 +14,9 @@
 */
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -41,6 +43,19 @@ namespace MongoDB.Bson.Tests.Serialization.DictionaryGenericSerializers
             public IDictionary<object, object> ID { get; set; }
             public SortedDictionary<object, object> SD { get; set; }
             public SortedList<object, object> SL { get; set; }
+        }
+
+        public class RO<TKey, TValue> : ReadOnlyDictionary<TKey, TValue>
+        {
+            private RO(IDictionary<TKey, TValue> dictionary)
+                : base(dictionary)
+            {
+            }
+
+            public static RO<TKey, TValue> ConstructorReplacement(IDictionary<TKey, TValue> dictionary)
+            {
+                return new RO<TKey, TValue>(dictionary);
+            }
         }
 
         [Test]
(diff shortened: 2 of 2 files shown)


src/MongoDB.Bson/Serialization/Serializers/DictionaryInterfaceImplementerSerializer.cs
```diff
@@ -164,7 +165,7 @@ public class DictionaryInterfaceImplementerSerializer<TDictionary, TKey, TValue>
         DictionarySerializerBase<TDictionary, TKey, TValue>,
         IChildSerializerConfigurable,
         IDictionaryRepresentationConfigurable<DictionaryInterfaceImplementerSerializer<TDictionary, TKey, TValue>>
-            where TDictionary : class, IDictionary<TKey, TValue>, new()
+            where TDictionary : class, IDictionary<TKey, TValue>
```
rstam: I don't think we should make this change. For this serializer to work, TDictionary *really* needs to satisfy the new() type constraint (i.e. have a public no-argument constructor).

All this change is doing is moving the error from compile time to runtime, becase Activator.CreateInstance on line 275 below will throw a MissingMethodException if the TDictionary type does not have a public no-argument constructor.

jhadwen: Actually I don't believe the statement is true that the new() constraint is utilised or "caught" anywhere at compile time, see the unit test for proof that this fails at runtime. Effectively new() in this case is actually just being used as syntactic sugar for Activator.CreateInstance. The unit test also proves that he serializer works correctly with my change, and none of the other existing tests break because of this change.

This is definitely a regression, the old versions of the library allowed this scenario and allowed us to serialise and deserialise correctly.

rstam: The reason it works for you in your tests is that your saved instance is *not* being deserialized as an instance of RO&lt;K,V> but rather as an instance of Dictionary&lt;K,V>. You even have an assertion checking that the RO did not round trip:

https://github.com/mongodb/mongo-csharp-driver/pull/232/files#diff-8547b37a13284a844ea81aa11321d170R422

This is because your "nominal" type is IDictionary&lt;K, V> (an interface), and we are simply making an assumption that you don't care what the actual type of your deserialized instance is (because you used an interface), so during deserialization we arbitrarily picked the simplest class we know of that implements IDictionary&lt;K,V>, namely Dictionary&lt;K,V>.

If you had a class that looked like this Activator.CreateInstance would fail at runtime during deserialization:

    public class C
    {
        public RO<string, int> F;
    }

Because now it really would try to deserialize an instance of RO&lt;string, int> and it wouldn't be able to create that instance (Activator.CreateInstance would throw).

It's quite possible that older versions of the driver weren't enforcing the new() type constraint at compile time, but the constraint would still be there at runtime.

jhadwen: Absolutely aware that it is not a round trip, but the behaviour of specifying that the member as IDictionary is a contract that the states that I don't care what the concrete implementation that is deserialized is. The driver behaviour therefore explicitly allows non-round trips and there are other tests in this file that enforce that behaviour to be allowed.

What changed in the latest drivers is:
- I'll allow a non-round trip from A:I -> B:I (where B is the drivers choice of implementation) but only if I could also do a round trip A:I -> A:I

Therefore if my interface stated do not do a round trip I'm now constrained by a separate use case that performs a round-trip.

Could you please clarify what you mean in relation to compile time constraint? From what I see of the calls to this class, the type parameter is constructed dynamically and therefore it is not a compile time constraint but a runtime generic type constraint. If anything all I can be accused of in this fix is mutating the location of a (when round-tripping) runtime error from serialization to deserialization, given that was the behaviour before I don't see that as a problem.

Either the change needs to be here, or there needs to be a serializer that when generating a class map understands correctly that IDictionary member definition is not a round trip and should therefore do no inspection on the implementation class.

rstam: Thanks for all your thoughtful comments. Probably the easiest way to clarify some of my comments is with some code. Please see the latest commit on my branch:

https://github.com/rstam/mongo-csharp-driver/commits/csharp1447

I've added two failing tests.

The first test illustrates that removing the new() constraint from the DictionaryInterfaceImplementerSerializer simply moves the error from the instantiation of the serializer to a later point in time when Deserialize is called. From your latest comments I realize that you understand this.

The second test illustrates the compile time error you were asking me about. Prior to the change made in this pull request line 74 would have been a compile time error. I also realize that this is only a compile time error if you actually write code like this. If the serializer is created at run time then it is a runtime error.

I don't necessarily agree that declaring a field as an interface type automatically implies that the actual type of the value is not relevant. In some cases you might be happy to only partially serialize the actual instance knowing that it is going to be deserialized as something else. In other cases you might care deeply about preserving the actual type of the instance. So you can't say that there is a hard and fast rule (a "contract") that says that just because an interface was used that one automatically does not care about the actual type.

Nonetheless, I understand that in your case you intend for the actual type to be ignored.

Seems like what is actually intended here is:

1. When serializing an RO<K,V> value ignore the type and serialize it like an IDictionary<K,V>
2. When deserializing the value you don't care about the resulting type, as long as it implements IDictionary<K,V>

We don't currently have a way to say "serialize an actual instance partially using whatever serialization is implied by some interface". And that whole approach only makes sense for Serialize anyway, where Deserialize is involved one must commit to an actual type.

Given that we don't currently have any better way to do what you are wanting to do, removing the new() constraint as you have done is probably the best short term workaround we can come up.

Can you please rebase your commit on master (the pull request is based on a very old commit) and re-submit?

Thanks for your perseverance.

jhadwen:
Commit rebased.

Thanks Robert for your time on this.

jyemin:
Since CSHARP-1447 has been fixed and released, I'm going to close this PR.

PR #233, opened 2016-02-01
CSHARP-1552: Tests under CSharp515Tests fail when assemblies are signed

jhadwen:
Possible alternate is to use the same or similar code as in TypeNameDiscriminator.GetDiscriminator but I've chosen this implementation for independence.


rstam:
I pulled this pull request, rebased on master and made some minor changes.

FYI to all contributors: make sure your core.autocrlf setting is set to true to ensure that source files are committed to the repository with the proper line endings.

Thanks for the pull request!

PR #235, opened 2016-03-17
CHARP-1542 Abstract Types Linq

scbond:
This attempts to fix the regression from the Driver Version 1.92 for handling Abstract Types within the Linq Statement.

I've added a new Serialization Expression Called AsType. This is inserted when visiting the MemberExpression and wraps around the abstract type with the correct serializer.

When the AsType Expression is encountered during processing it returns a FieldExpression with the concrete type Serializer.


craiggwilson:
Thanks Stephen... I'm going to run some tests to see if this might already be fixed. There were 2 related tickets that might have fixed this (1604 is currently in code review) that might make this unnecessary. If it turns out it's still needed, I'll take a look. Regardless, thank you so much for your effort here. It takes a lot to understand a LINQ implementation and contribute.

craiggwilson:
Hi Stephen, I believe piggy-backing on the fix for [CSHARP-1604](https://github.com/mongodb/mongo-csharp-driver/commit/a06d0319514f213c5b0350d26c0bdd03cef1e6c7) made this simple(https://github.com/mongodb/mongo-csharp-driver/commit/de410dcf378433383e5198265f993fe0328e7a8a). Please let me know if this fixes the issues you were addressing here.

scbond:
Hi Craig,

Thanks for looking at this, that certainly does fix the issue much appreciated. 

Thanks

Stephen

PR #236, opened 2016-04-07
Remove reference to removed class in documentation.

mindless2112:
`MongoConnectionStringBuilder` no longer exists. Don't suggest that people can use it.


rstam:
Rebased on master and merged.

PR #237, opened 2016-05-12
[CSHARP-1664] Allow dynamically created Expression<Func<>> to be used in a Mongo Linq query

alex687:


jyemin:
As CSHARP-1664 has already been fixed and released, I'm going to close this PR.

PR #238, opened 2016-05-20
CS-30851:  Adds CreatorClassMapConvention and associated tests

joeenzminger:
This feature allows consumers of the driver to inject their own object creation mechanism into the BsonClassMap through the ConventionRegistry .  This is useful in scenarios where object creation is delegated to an IOC Container (such as Autofac, etc).

### DIFF
--- a/src/MongoDB.Bson.Tests/MongoDB.Bson.Tests.csproj
+++ b/src/MongoDB.Bson.Tests/MongoDB.Bson.Tests.csproj
@@ -142,6 +142,7 @@
     <Compile Include="Serialization\Conventions\CamelCaseElementNameConventionsTests.cs" />
     <Compile Include="Serialization\Conventions\ConventionPackTests.cs" />
     <Compile Include="Serialization\Conventions\ConventionRunnerTests.cs" />
+    <Compile Include="Serialization\Conventions\CreatorClassMapConventionTests.cs" />
     <Compile Include="Serialization\Conventions\DelegateClassMapConventionTests.cs" />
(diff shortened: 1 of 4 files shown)


craiggwilson:
Hi Joe. Thanks so much for this.

My problem with this, as is, is that it unconditionally uses the function. As you can see, this is relatively simple to write for those users who need something like this. In addition, it's not widely needed. So I'm leaning towards not doing anything here. However, we probably should enhance our documentation or provide an example of how to do this.

Thoughts?
Craig

joeenzminger:
Craig - not sure what you mean by "unconditionally uses the function" (I may be missing something obvious).  It will only use the creator supplied in the convention under the same conditions that it would use the compiled expression version in BsonClassMap GetCreator() (which appears to be if there are no creatorMaps defined on the classMap), and if the ConventionPack the Convention is registered with is registered with a filter that maps to the specific type requested.

However, I do agree that this is just a more specific version of the DelegateClassMapConvention and really just serves to make it a little more clear how to do this.  However, that fact is also true for all of the other IClassMapConvention implementations.  They could all be replaced with DelegateClassMapConvention at the expense of making it a little less clear how to configure BsonClassMap. 

Documentation or a 'how to integrate an IOC container' would, of course, make everything clearer as well.  Would be happy to write something up.  Where would be the best place to contribute that?

craiggwilson:
Ah yes... you are right about the ConventionPack itself having a type filter.

What I was indicating is that this particular class is extremely easy to write, as you noticed, once you know what hooks are necessary. I'm going to close this pull request. I've opened up a JIRA ticket for the documentation example: https://jira.mongodb.org/browse/CSHARP-1672.

PR #239, opened 2016-05-22
CSHARP-1675: Fixed ObjectSerializer with DateTime issue

MaksimSimkin:


craiggwilson:
Hi @MaksimSimkin,

I do see the issue. This is only a partial fix though. 
1. The serialize method is still incorrect and doesn't respect the registered serializer.
2. You've illuminated a whole portion of code that disregards a registered serializer (for instance, if someone registered a custom Int32 serializer).

Also, I don't believe that serializerRegistry.GetSerializer can ever return null. So, we'll always do this.

I need to talk with my colleague and see what we think about this. Hopefully, users don't encounter this all that often. May I ask why you are using type object at all?
Craig

MaksimSimkin:
Hi @craiggwilson 

Yes, i see, that solution i proposed is just a local fix and should be changed.  Sorry for it, it was really try to solve the problem quick and dirty.  I'l think about better way. 

Short description about howe we are using it:

We are storing in our db the collection of different Metadata objects  (strings, integeres and datetimes)

`    public class AttributeValue
    {
        public string AttributeKey { get; set; }
        public string AttributeName { get; set; }
        public string ValueTypeName { get; set; }
        public object Value { get; set; }
    }` 

So we have figured out, that what  we are storing in the db is not always the same, we are getting out. Now it's not a production problem,  we have a workaround, using overriden ObjectSerializer:

`    public class CustomObjectSerializer : ObjectSerializer
    {
        public override object Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
        {

            var bsonType = bsonReader.GetCurrentBsonType();

            if (bsonType == BsonType.DateTime)
            {
                return BsonSerializer.SerializerRegistry.GetSerializer(typeof (DateTime)).Deserialize(context, args);
            }

            return base.Deserialize(context, args);
        }
    }`

craiggwilson:
Ok. I'm going to close this PR as it doesn't fix the whole problem. Thanks so much for pointing this out and having a solution at all. It has pointed out an issue we have. Also, your use case sounds fine. I didn't envision anyone doing something like this, but now we know.

Craig

PR #244, opened 2016-07-01
Added aggregate options for AsQueryable

dkendrick:
Submitting around as build timed out


jyemin:
As this change has already been applied in another commit, I'm going to close this out.  Sorry for leaving it sitting for so long.

PR #245, opened 2016-07-02
Small changes in MongoUtils.cs

Oscarbralo:
Changes:

I've changed the var's, because is not good regarding the performance use var, because you know perfectly what thos functions are returning, so there is no need to let the program, to infere. And return that directly, there is no need to store the hash in a string and then return.

About the CamelCase, is the string value was an empty string, the program will crash with an ArgumentOutOfRangeException, so a simple check will fix that part.

And at the end, only to standarized, it's cleaner to standarized all the "" to String.Empty.

### DIFF
--- a/src/MongoDB.Driver/MongoUtils.cs
+++ b/src/MongoDB.Driver/MongoUtils.cs
@@ -35,7 +35,7 @@ namespace MongoDB.Driver
         public static string Hash(string text)
         {
             MD5 md5 = MD5.Create();
-            bytes[] bytes = md5.ComputeHash(Encoding.UTF8.GetBytes(text));
+            byte[] bytes = md5.ComputeHash(Encoding.UTF8.GetBytes(text));
             return BitConverter.ToString(bytes).Replace("-", "").ToLowerInvariant();
         }
 


src/MongoDB.Driver/MongoUtils.cs
```diff
@@ -34,10 +34,9 @@ public static class MongoUtils
         /// <returns>The MD5 hash.</returns>
         public static string Hash(string text)
         {
-            var md5 = MD5.Create();
-            var bytes = md5.ComputeHash(Encoding.UTF8.GetBytes(text));
-            var hash = BitConverter.ToString(bytes).Replace("-", "").ToLowerInvariant();
-            return hash;
+            MD5 md5 = MD5.Create();
```
rstam: In general we prefer the use of var. There is no performance penalty to using var.

src/MongoDB.Driver/MongoUtils.cs
```diff
@@ -34,10 +34,9 @@ public static class MongoUtils
...
+            byte[] bytes = md5.ComputeHash(Encoding.UTF8.GetBytes(text));
+            return BitConverter.ToString(bytes).Replace("-", "").ToLowerInvariant();
```
rstam: It's true that the intermediate variable (hash) is not required here, but it also does no harm.
 
In some cases (though probably not this one) the intermediate variable facilitates debugging.

src/MongoDB.Driver/MongoUtils.cs
Oscarbralo: That ok then if you prefer that ;) 
And about the other change? The onw about the validation os the string is not null?

src/MongoDB.Driver/MongoUtils.cs
Oscarbralo: If the purpose is debugging, then is not bad to let the variable the way it is.

src/MongoDB.Driver/MongoUtils.cs
```diff
@@ -71,7 +70,7 @@ internal static string ToInsecureString(SecureString secureString)
         {
             if (secureString == null || secureString.Length == 0)
             {
-                return "";
+                return String.Empty;
```
rstam: The choice between "" and String.Empty is mostly a stylistic choice.

As a general rule we prefer "" to String.Empty. "" is a compile time constant rather than a property access and it is shorter.

Oscarbralo: It's a good explanation Robert. 
as you said, it's only stylistic choice ;)

src/MongoDB.Driver/MongoUtils.cs
```diff
@@ -57,7 +56,7 @@ public static TimeSpan TimeSpanFromMicroseconds(long microseconds)
         /// <returns>The camel cased string.</returns>
         public static string ToCamelCase(string value)
         {
-            return value.Substring(0, 1).ToLower() + value.Substring(1);
+            return !String.IsNullOrWhiteSpace(value) ? value.Substring(0, 1).ToLower() + value.Substring(1) : String.Empty;
```
rstam: Good catch.

You could submit a new pull request with this fix.

We generally prefer avoiding negative conditions, so this could be written as:

    return value.Length == 0 ? "" : value.Substring(0, 1).ToLower() + value.SubString(1);

rstam:

See the comments for why it is being closed.

PR #246, opened 2016-07-02
AggregateFluent.cs update

Oscarbralo:
Some changes in style to make the code clean.

### DIFF
--- a/src/MongoDB.Driver/AggregateFluent.cs
+++ b/src/MongoDB.Driver/AggregateFluent.cs
@@ -28,54 +28,62 @@ namespace MongoDB.Driver
 {
     internal class AggregateFluent<TDocument, TResult> : AggregateFluentBase<TResult>
     {
+        // consts
+        private const string OPERATOR_GROUP_NAME = "$group";
+        private const string OPERATOR_LOOKUP_NAME = "$lookup";
+        private const string OPERATOR_MATCH_NAME = "$match";
+        private const string OPERATOR_PROJECT_NAME = "$project";
+        private const string OPERATOR_SORT_NAME = "$sort";
+        private const string OPERATOR_UNWIND_NAME = "$unwind";
+        
         // fields
         private readonly IMongoCollection<TDocument> _collection;
         private readonly AggregateOptions _options;
         private readonly List<IPipelineStageDefinition> _stages;
+        
 
         // constructors
         public AggregateFluent(IMongoCollection<TDocument> collection, IEnumerable<IPipelineStageDefinition> stages, AggregateOptions options)
         {
-            _collection = Ensure.IsNotNull(collection, nameof(collection));
-            _stages = Ensure.IsNotNull(stages, nameof(stages)).ToList();
-            _options = Ensure.IsNotNull(options, nameof(options));
+            this._collection = Ensure.IsNotNull(collection, nameof(collection));
+            this._stages = Ensure.IsNotNull(stages, nameof(stages)).ToList();
+            this._options = Ensure.IsNotNull(options, nameof(options));
         }
 
         // properties
         public override AggregateOptions Options
         {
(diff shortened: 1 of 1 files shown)


src/MongoDB.Driver/AggregateFluent.cs
```diff
@@ -28,54 +28,62 @@ namespace MongoDB.Driver
...
         private readonly List<IPipelineStageDefinition> _stages;
+        
 
         // constructors
         public AggregateFluent(IMongoCollection<TDocument> collection, IEnumerable<IPipelineStageDefinition> stages, AggregateOptions options)
         {
-            _collection = Ensure.IsNotNull(collection, nameof(collection));
-            _stages = Ensure.IsNotNull(stages, nameof(stages)).ToList();
-            _options = Ensure.IsNotNull(options, nameof(options));
+            this._collection = Ensure.IsNotNull(collection, nameof(collection));
```
rstam: The style guidelines for our project are to only use "this." when absolutely required.

Our naming convention for member field names is to start them with an "_" so it is already obvious that the name is a member.

src/MongoDB.Driver/AggregateFluent.cs
```diff
@@ -206,14 +209,13 @@ public override IAggregateFluent<TNewResult> Unwind<TNewResult>(FieldDefinition<
...
-            const string operatorName = "$unwind";
             var stage = new DelegatedPipelineStageDefinition<TResult, TNewResult>(
-                operatorName,
+                OPERATOR_UNWIND_NAME,
                 (s, sr) =>
                 {
                     var newResultSerializer = options.ResultSerializer ?? (s as IBsonSerializer<TNewResult>) ?? sr.GetSerializer<TNewResult>();
 
-                    var fieldName = "$" + field.Render(s, sr).FieldName;
+                    string fieldName = String.Concat("$", field.Render(s, sr).FieldName);
```
rstam: "+" is more readable and as far as I know there is no performance advantage to using Concat instead.

src/MongoDB.Driver/AggregateFluent.cs
Oscarbralo: If your style guidelines are that way, then perfectly understood for future commits.

src/MongoDB.Driver/AggregateFluent.cs
```diff
@@ -28,54 +28,62 @@ namespace MongoDB.Driver
...
     internal class AggregateFluent<TDocument, TResult> : AggregateFluentBase<TResult>
     {
+        // consts
+        private const string OPERATOR_GROUP_NAME = "$group";
+        private const string OPERATOR_LOOKUP_NAME = "$lookup";
+        private const string OPERATOR_MATCH_NAME = "$match";
+        private const string OPERATOR_PROJECT_NAME = "$project";
+        private const string OPERATOR_SORT_NAME = "$sort";
+        private const string OPERATOR_UNWIND_NAME = "$unwind";
+
```
rstam: I believe we chose to make these local constants inside the methods that use them because they are never referenced anywhere else.

rstam:

See comments.

src/MongoDB.Driver/AggregateFluent.cs
Oscarbralo: Check the benchmark performance here: for concatenate string using + and String.Concat.  Maybe it's useful
http://www.dotnetperls.com/string-concat

Search this line and check the example:
B:  281 ms (Concat 4 strings, all at once)

src/MongoDB.Driver/AggregateFluent.cs
Oscarbralo: Yeah, almost all of them, but there are one or two that are being references in two different methods, maybe you can only let that ones here and the others inside the method

PR #247, opened 2016-07-02
Small fix check string not empty

Oscarbralo:


rstam:
Could you submit a new or revised pull request based on current master?

Note the message "from unknown repository" above.

rstam:
Rebased on master and merged.

PR #248, opened 2016-07-02
Update ClusterRegistry.cs

Oscarbralo:
Just in case those lists comes null. Only to prevent crashes


jyemin:
This hasn't been an issue in practice, so I'm going to close this.  If anyone is still interested in this, please open a Jira issue to re-open discussion.

PR #249, opened 2016-07-02
Fix some IsNotNull problems

Oscarbralo:
I've checked that some IsNotNull checks were missing, so I've added.


jyemin:
At this point there are too many merge conflicts to take this further, and in practice I'm not sure the null checks will help in practice.  Closing this out.

PR #251, opened 2016-07-21
SecureString cleanups

pieceofsummer:
It all started with one block of code in AuthenticationHelper, when I stumbled upon a GCHandle, which was not needed there. 
Then I've seen different encodings (BSTR/Unicode) when marshalling unmanaged strings… and then I just couldn't stop until I refactored it all.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Authentication/AuthenticationHelper.cs
+++ b/src/MongoDB.Driver.Core/Core/Authentication/AuthenticationHelper.cs
@@ -63,11 +63,7 @@ namespace MongoDB.Driver.Core.Authentication
         public static string MongoPasswordDigest(string username, SecureString password)
         {
             var prefixBytes = Utf8Encodings.Strict.GetBytes(username + ":mongo:");
-            return MongoPasswordDigest(prefixBytes, password);
-        }
 
-        public static string MongoPasswordDigest(byte[] prefixBytes, SecureString password)
-        {
             using (var md5 = MD5.Create())
             {
                 var passwordChars = new char[password.Length];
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver.Core/Core/Authentication/AuthenticationHelper.cs
```diff
@@ -62,63 +62,47 @@ public static void Authenticate(IConnection connection, ConnectionDescription de
...
-                    unmanagedPassword = Marshal.SecureStringToBSTR(password);
-                    var passwordChars = new char[password.Length];
-                    GCHandle passwordCharsHandle = new GCHandle();
+                    Marshal.Copy(unmanagedPassword, passwordChars, 0, passwordChars.Length);
+
+                    var passwordBytesCount = Utf8Encodings.Strict.GetByteCount(passwordChars);
+                    var buffer = new byte[prefixBytes.Length + passwordBytesCount];
                     try
                     {
-                        passwordCharsHandle = GCHandle.Alloc(passwordChars, GCHandleType.Pinned);
```
rstam: Pinning the data is important for security reasons. If a garbage collection were to occur while this method was executing things could be moved around in memory and copies of the password could be left behind in memory.

pieceofsummer:
@rstam this makes sense, but not the way it is currently implemented. 

Lines like 

    passwordBytesHandle = GCHandle.Alloc(passwordBytesHandle, GCHandleType.Pinned);

guard nothing, and the final buffer was never guarded to begin with.

pieceofsummer:
Returned pinning back, with all buffers being properly pinned now.

rstam:
I wasn't able to rebase your commit on master because we had done lots of work on a branch that recently got merged into master and we had both edited the same areas of code which resulted in merge conflicts.

While I wasn't able to take your commit verbatim I did make somewhat similar changes based on the work you had done (in particular, using Unicode instead of BSTR unmanaged strings and making sure all the buffers are properly pinned and zeroed out).

I also added your name to our contributors list. Thank you very much for your work.

My changes are on this branch and commit if you want to comment on them:

https://github.com/rstam/mongo-csharp-driver/commits/csharp1727

https://github.com/rstam/mongo-csharp-driver/commit/bc75e1229e1d614a17984fbb7662e9977dd40dec

Thanks again!

PR #253, opened 2016-07-27
Better support for dates and strings in LINQ

pieceofsummer:
While playing around with some custom serializers and complex classes, I've found a few errors and a bunch of not implemented functions, so I spent an evening trying to make the world a better place :) 

Hope it will be helpful.

### DIFF
--- a/src/MongoDB.Driver/Linq/Translators/AggregateLanguageTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Translators/AggregateLanguageTranslator.cs
@@ -86,6 +86,8 @@ namespace MongoDB.Driver.Linq.Translators
                     return TranslateOperation((BinaryExpression)node, "$multiply", true);
                 case ExpressionType.New:
                     return TranslateNew((NewExpression)node);
+                case ExpressionType.NewArrayInit:
+                    return TranslateNewArrayInit((NewArrayExpression)node);
                 case ExpressionType.Not:
                     return TranslateNot((UnaryExpression)node);
                 case ExpressionType.NotEqual:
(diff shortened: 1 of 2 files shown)


craiggwilson:
Hi @pieceofsummer,

Thanks so much for the PR. Couple things I noticed. 
1. For code consistency, we always use braces with conditional statements, even one liners.
2. There aren't any tests. While I'm sure your implementations are perfect, we really need tests to ensure we don't break them in the future.

If you could do both those things, we'll look at pulling this in.
Thanks,
Craig

pieceofsummer:
Hi @craiggwilson,

1) No problem. I like short one-liners, but it's not too hard to sacrifice them for the sake of greater good :)

2) I highly doubt they're perfect, but I already was too sleepy to think of how to set up test cluster early in the morning :) I'll add some unit tests later.

There's also that recent issue with .NET Core when connecting to DNS endpoints on Unix, which makes even existing tests to fail for me… :(

pieceofsummer:
@craiggwilson, I'm also a bit confused with the current implementation of math functions. 

Docs say almost all math functions (except **$pow** and **$log**, for obvious reasons) take just a single number, but the driver gives them an array (with single number inside). Furthermore, Mongo server accepts both dialects, while being rather strict about other functions' arguments.

Can you please clarify this a little?

craiggwilson:
Formally, the server wants an array as the argument. For single-valued args, they accept just a value without being in an array. I forgot what we currently do, but might as well follow whatever pattern we are currently using.

pieceofsummer:
Added some unit tests, hope it's ok now :)

craiggwilson:
This is awesome. We'll discuss and see if there is anything left to do. Might ask you to squash and rebase on master just before the pull to ensure you get credit properly. Will let you know.

pieceofsummer:
Should I rebase or close this? 

I kinda see $dateToString implemented in 79853e2b37ad8c87501799379bb9f7a2a31d9086, do you plan to add other date functions as well?

elliott-pd:
I have made an updated version of this PR: https://github.com/mongodb/mongo-csharp-driver/pull/351 which is based off current master.

jyemin:
Closing as superseded by #351

PR #256, opened 2016-08-26
Fix EnumRepresentationConvention with nullable enum properties

mattwhetton:
This fix ensures the EnumRepresentationConvention works as expected onNullable enum properties.

Ref: https://jira.mongodb.org/browse/CSHARP-1839


rstam:
A different approach to fix this was committed here:

https://github.com/mongodb/mongo-csharp-driver/commit/4bf62804c4faa566a9242cf99f0240aa62f532b3

Thanks for the bug report and the pull request.

PR #257, opened 2016-10-23
CSHARP-5554: Improved ObjectId parsing and ToString() performance

pieceofsummer:
While writing an `ObjectId`-like class for another project (where I needed a time-based identifier, but didn't want to bring dependency on a whole `MongoDB.Bson` package just for `ObjectId`), I was curious about performance of my implementation compared to the reference one.

My test was like serializing a million of identifiers (as strings), and then parsing them back. Here are my results:

|  | `ObjectId.ToString()` | `ObjectId.Parse()` |
| --- | --: | --: |
| MongoDB.Bson (v2.3.0) | 3475ms | 897ms |
| My (humble) implementation | 87ms | 81ms |

So this probably looks like a valid reason to make a pull request and improve things a little :) 

I'm not sure if it affects other cases where `ObjectId` is serialized, like performance of database responses' deserialization. But shouldn't hurt at least.

### DIFF
--- a/src/MongoDB.Bson/ObjectModel/ObjectId.cs
+++ b/src/MongoDB.Bson/ObjectModel/ObjectId.cs
@@ -44,16 +44,27 @@ namespace MongoDB.Bson
         /// <summary>
         /// Initializes a new instance of the ObjectId class.
         /// </summary>
+        /// <param name="a">First 32 bits</param>
+        /// <param name="b">Second 32 bits</param>
+        /// <param name="c">Third 32 bits</param>
+        private ObjectId(int a, int b, int c)
+        {
+            _a = a; _b = b; _c = c;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the ObjectId class.
+        /// </summary>
         /// <param name="bytes">The bytes.</param>
         public ObjectId(byte[] bytes)
         {
             if (bytes == null)
             {
-                throw new ArgumentNullException("bytes");
+                throw new ArgumentNullException(nameof(bytes));
             }
             if (bytes.Length != 12)
             {
-                throw new ArgumentException("Byte array must be 12 bytes long", "bytes");
+                throw new ArgumentException("Byte array must be 12 bytes long", nameof(bytes));
             }
 
             FromByteArray(bytes, 0, out _a, out _b, out _c);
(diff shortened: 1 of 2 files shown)


damieng:
Please ignore the previous comments/closure, was meant for a different PR.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -471,14 +513,7 @@ public bool Equals(ObjectId rhs)
...
         {
-            if (obj is ObjectId)
-            {
-                return Equals((ObjectId)obj);
-            }
-            else
-            {
-                return false;
-            }
+            return obj is ObjectId && Equals((ObjectId)obj);
```
ArielTipalti, line 516:
```suggestion
            return obj is ObjectId objId && Equals(objId );
```

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -541,7 +576,24 @@ public void ToByteArray(byte[] destination, int offset)
         /// <returns>A string representation of the value.</returns>
         public override string ToString()
         {
-            return BsonUtils.ToHexString(ToByteArray());
+            var buffer = new char[24];
```
ArielTipalti, line 579:
```suggestion

public override string ToString()
{
    Span<char> buffer = stackalloc char[24]; 

    int a = _a;
    int b = _b;
    int c = _c;

    for (int i = 7; i >= 0; i--)
    {
        var r = (a & 0x0F);
        buffer[i] = (char)(r < 10 ? (r + '0') : (r - 10 + 'a'));
        a >>= 4;

        r = (b & 0x0F);
        buffer[i + 8] = (char)(r < 10 ? (r + '0') : (r - 10 + 'a'));
        b >>= 4;

        r = (c & 0x0F);
        buffer[i + 16] = (char)(r < 10 ? (r + '0') : (r - 10 + 'a'));
        c >>= 4;
    }

    return new string(buffer); 
}

```

BorisDog:
Thank you for your contribution @pieceofsummer.
Closing this as it has gone stale.
Related improvements to `ObjectId` done in https://github.com/mongodb/mongo-csharp-driver/pull/1960 and https://github.com/mongodb/mongo-csharp-driver/pull/2043.

Please submit a Jira ticket and a new PR for any follow-ups.

PR #258, opened 2016-11-07
Add debug info to exception if deserialization fails

Sebmaster:
In case of an unknown BSON type (for example: Undefined), the driver would crash with no way to debug this case if there're a lot of result records.

This change now provides the (root-)property that failed deserialization and the previously decoded values (mainly _id) of the document to easily find it.


jyemin:
Hi @Sebmaster I apologize for letting this sit for so long with no response.  Typically we require a Jira issue to be created before looking at a PR and perhaps that's why we missed this.  I'm going to close this out because there are now merge conflicts, but if anyone is interested in pursuing this please open a Jira issue to restart discussion.

PR #259, opened 2016-11-23
CSHARP-1836: Added additional check for default 'VersionReturned' property value and added unit tests.

MaKCbIMKo:
The initial problem is that property is nullable.

It could be solved by moving from using nullable fields (not sure why it needs to be nullable).
Almost all overloads of FindAndModify method work properly because specify which version need to be returned.

Obsolete version 'FindAndModifyResult FindAndModify(IMongoQuery query, IMongoSortBy sortBy, IMongoUpdate update)' pretty easy to modify, but it won's solve the problem, because the problem is in 'FindAndModifyArgs' class.


rstam:
Rebased on master and merged manually.

Thanks for the pull request.

MaKCbIMKo:
I was glad to contribute to make things better.

PR #260, opened 2016-11-28
Preserve element name if its characters are all in capital letters in camel-case convention

mfidemraizer:
Hi!

It's a very simple improvement.

Let's say we need to serialize a class like this:

```
public class Customer
{
       public string Id { get; set; }
       public string TIN { get; set; }
}
```

Current `CamelCaseElementNameConvention` will serialize `Customer.TIN` as `tIN` while it should be just `TIN`.

In summary, I would say that if a property is in capital letters **it should mean that it's an *acronym*** (or leaving a property in capital letters is the intended behavior).

## Solution

My proposed solution is a new `else if` in the whole convention code:

```
else if (memberName == memberName.ToUpper())
{
        return memberName;
}
```

---

Thank you in advance and I'll look forward for your comments!

### DIFF
--- a/src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
+++ b/src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
@@ -23,6 +23,24 @@ namespace MongoDB.Bson.Serialization.Conventions
     /// </summary>
     public class CamelCaseElementNameConvention : ConventionBase, IMemberMapConvention
     {
+        /// <summary>
+        /// Default constructor
+        /// </summary>
+        /// <param name="skipWhenAllCaps">
+        /// When set to true, this convention will skip members in capital letters (for example: "ALLCAPS")
+        /// in order to preserve acronyms or member names designed to be in upper-case.
+        /// </param>
+        public CamelCaseElementNameConvention(bool skipWhenAllCaps = false)
+        {
+            SkipWhenAllCaps = skipWhenAllCaps;
+        }
+
+        /// <summary>
+        /// Gets if this convention will skip members in capital letters (for example: "ALLCAPS")
+        /// in order to preserve acronyms or member names designed to be in upper-case.
+        /// </summary>
+        private bool SkipWhenAllCaps { get; }
+
         // public methods
         /// <summary>
         /// Applies a modification to the member map.
@@ -42,11 +60,15 @@ namespace MongoDB.Bson.Serialization.Conventions
             {
                 return "";
             }
+            else if (SkipWhenAllCaps && memberName == memberName.ToUpper())
+            {
+                return memberName;
+            }
             else if(memberName.Length == 1)
             {
                 return Char.ToLowerInvariant(memberName[0]).ToString();
             }
-            else 
+            else
             {
                 return Char.ToLowerInvariant(memberName[0]) + memberName.Substring(1);
             }
--- a/tests/MongoDB.Bson.Tests/Serialization/Conventions/CamelCaseElementNameConventionsTests.cs
+++ b/tests/MongoDB.Bson.Tests/Serialization/Conventions/CamelCaseElementNameConventionsTests.cs
@@ -27,6 +27,7 @@ namespace MongoDB.Bson.Tests.Serialization.Conventions
             public int Age { get; set; }
             public string _DumbName { get; set; }
             public string lowerCase { get; set; }
+            public string ALLCAPS { get; set; }
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


rstam:
This is a good idea, but the problem is that it breaks compatibility with existing code (or data).

For your purposes you could either write your own convention or occasionally annotate a property using [BsonElement("TIN")] to override the default element name when necessary.

mfidemraizer:
@rstam  Uhm, yes. You're right about the backwards compatibility.

BTW, I can update this pull request with a `CamelCaseElementNameConvention` constructor to preserve previous behavior:

`public CamelCaseElementNameConvention(bool leaveAllCaps = false)`

And the whole `else if ` would look as follows:

```
else if (LeaveAllCaps && memberName == memberName.ToUpper())
{
        return memberName;
}
```

Would be this more acceptable? Let me know what's your opinion to work on the modification.

mfidemraizer:
@rstam Well, I've already pushed these changes. See the code to review them.

mfidemraizer:
AFAIK, this feature will be rejected and for that reason I'm going to close it. For now, I've duplicated this convention class into a custom one with my addition in my own project... BTW, I still feel that this should come out-of-the-box in the driver. No one expects to get property name in upper-case as `tIN`...

craiggwilson:
I wouldn't close this. It seems like a good solution. Arguably, and acronym in all caps should be all lower-case, hence "tin" instead of "TIN".

Robert can comment further about anything that needs to be done for the PR though.

mfidemraizer:
@craiggwilson I've re-opened again... ok.

Let's see what would this change require to get merged!

src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
```diff
@@ -23,6 +23,24 @@ namespace MongoDB.Bson.Serialization.Conventions
...
     public class CamelCaseElementNameConvention : ConventionBase, IMemberMapConvention
     {
+        /// <summary>
+        /// Default constructor
+        /// </summary>
+        /// <param name="skipWhenAllCaps">
+        /// When set to true, this convention will skip members in capital letters (for example: "ALLCAPS")
+        /// in order to preserve acronyms or member names designed to be in upper-case.
+        /// </param>
+        public CamelCaseElementNameConvention(bool skipWhenAllCaps = false)
```
rstam: It's not clear that members that are all upper case should be left all upper case. It's just as likely that they should all be converted to lower case.

How about instead of a bool parameter here we use an enum like this:

```
public enum AllUpperCaseHandling
{
    ConvertFirstCharacterToLowerCase = 0, // the default for backward compatibility
    LeaveUpperCase,
    ConvertAllToLowerCase
}
```

That would allow either use case to be supported.

src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
```diff
@@ -42,11 +60,15 @@ private string GetElementName(string memberName)
             {
                 return "";
             }
+            else if (SkipWhenAllCaps && memberName == memberName.ToUpper())
```
rstam: Is ToUpper compatible with all possible Unicode strings?

rstam:
If we're going to do this we're going to need a JIRA ticket for it.

https://jira.mongodb.org/browse/CSHARP-883 seems related.

src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
mfidemraizer: @rstam  Hey, it sounds good. Taking the idea of an enumeration, maybe it's just about defining the `enum` identifier in a more generic way:

```
public enum CamelCaseElementNameHandling
{
        Default,
        PreserveUpperCase,
        LowerCase
}
```

I believe that whatever handling you choose to use, every handling should share some rules and conventions like the related issue you've found: if we get an element name like `IOStream` it should be serialized as `ioStream`. 

BTW, it's not easy to solve this problem, since I guess we're not going to implement a semantic detection to do something like `IOStream` to `ioStream`.

My proposal would be that we can implement a `CamelCaseConventionSettings` class to attach some settings there:

```
public class CamelCaseConventionSettings
{
        public CamelCaseElementNameHandling ElementNameHandling { get; set; }
        public Func<string, string> ElementNameHandlers { get; set; }
}
```

`ElementNameHandler` would be useful to even customize more the convention:

```
settings.ElementNameHandler = name =>
{
         if(name.StartsWith("IO"))
         {
                 return name.replace("IO", "io");
         }
        else
        {
                return name;
        }
};
```

And `CamelCaseConvention` should accept an optional parameter of type `ConventionSettings`. 

What's your opinion about this proposal?

src/MongoDB.Bson/Serialization/Conventions/CamelCaseElementNameConvention.cs
mfidemraizer: @rstam  I'm not 100% sure, but docs says:

> This method uses the casing rules of the current culture to convert each character in the current instance to its uppercase equivalent. If a character does not have an uppercase equivalent, it is included unchanged in the returned string.

mfidemraizer: They also state:

> To convert a string to uppercase by using the casing conventions of the current culture, call the ToUpper(CultureInfo) method overload with a value of CultureInfo.CurrentCulture for its culture parameter.

So we should use `ToUpper(CultureInfo)` overload.

jyemin:
I'm going to close this PR since there are merge conflicts and no Jira issue was ever created to continue the discussion.  If anyone is still interested in pursuing this, please open a new Jira issue to continue the discussion.

PR #265, opened 2017-01-11
CSHARP-1875: Steer clear of code that needs ntdll.dll

MarLoe:
As few changes as possible to accomplish the same functionality without using those parts of System.Runtime.InteropServices.RuntimeInformation that will reference ntdll.dll

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
 using MongoDB.Bson;
 
 namespace MongoDB.Driver.Core.Connections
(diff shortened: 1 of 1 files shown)


src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -69,44 +69,117 @@ internal static BsonDocument CreateDriverDocument(string driverVersion)
...
             string osVersion;
-            var match = Regex.Match(osName, @" (?<version>\d+\.\d[^ ]*)");
-            if (match.Success)
-            {
-                osVersion = match.Groups["version"].Value;
+
+#if NET45
+            if (Type.GetType("Mono.Runtime") != null)
+            {
+                Console.WriteLine("Hello Mono!!!");
```
derickr: Were you meant to leave this in here?

MarLoe: Absolutely not. I must have blinked... ;)

PR has been updated

G3N7:
This is a blocker for any folks using Mono, any ETR for getting this merged in and released?

tedblad:
Agree with @G3N7, this is a blocker. A merge and release would be very much appreciated.

bondjames12:
Simple change and 1 less dependency. Fixes mono compatibility (even if its not officially tested)

Zaid-Ajaj:
After succesfully setting up docker with mono, I am stuck with this problem. Merging any time soon?

rstam:
It has been merged. See: https://jira.mongodb.org/browse/CSHARP-1875

We have not yet determined when the next release will be so it the change is not yet available on Nuget.

Zaid-Ajaj:
@rstam Thanks for the quick response! For the people still blocked on this issue: just roll back MongoDb driver to 2.3.0 where there is no need for System.RuntimeInformation and delete all packages the depend on System.RuntimeInformation and it works (at least on my ubuntu 16.04 + mono 4.6.1.5 inside docker)

PR #267, opened 2017-01-18
Query is incorrect if type downcasting is present in expression

SadKirK:
In current project we store additional properties of entity in Dictionary<string, object> 
when try to search through collection with expression for instance: x => (int) x.W["one"] == 1
it's being converted to query: "{'one': 1}"
but should be: "{'W.one': 1}"
The root cause is type casting.

It works on driver's version 2.2.3, but since 2.2.4 we have issue, described above


jyemin:
Hi @SadKirK 

First of all, I apologize for letting this PR sit so long unattended.

If you're still interested in pursuing this issue, the next step wold be to open a Jira issue at https://jira.mongodb.org/projects/CSHARP to discuss whether this is a change we should consider.  If you decide to do so, please include a link to this PR.  Until then, I'm going to close this out.

PR #270, opened 2017-02-28
fix the misspelled word from adminstrative to administrative

chucklu:
correct the misspelled word


rstam:
Rebased on master and merged.

PR #271, opened 2017-03-07
CSHARP-1914 Remove cloning of the certificates

alexdawes:
Remove cloning of the certificates, as this was problematic as the private key is not included in the RawData of a cert and a cert cannot always be cloned in a deep manner (if it is non exportable)


alexdawes:
I've tested this on both a net462 client running on Windows and a netstandard1.6 client running on Linux. Both worked fine against a mongodb with certificates.

PR #276, opened 2017-04-12
fixed typeInfo properties selection

sad-whale:
Getting only the public properties of an instance, declared on current hierarchy level.
Otherwise, properties of base class will be selected and mapped, which leads to exception when call AutoMap on inherited classes


jyemin:
Hi @sad-whale 

First of all, I apologize for letting this set unattended for so long.  

If you're still interested in pursuing this change, please open a Jira issue at https://jira.mongodb.org/projects/CSHARP to discuss, referencing this PR.  Until then, I'm going to close this out.

PR #277, opened 2017-04-24
adding missing parentesis in polymorphism examples

salaros:
probably a typo


rstam:
Rebased on master and merged.

PR #278, opened 2017-04-29
CSHARP-1958: Removed document serialization when Linq Any() is called

ahydrax:
Hello there,

I've changed invocation of method Take by method Count. This will prevent from unnecessary document deserialization.


ahydrax:
Any updates?

rstam:
Thank you for submitting the request. It seems like this change requires counting all the documents that match which not be efficient. Therefore we are closing the pull request but keeping the ticket open.

PR #280, opened 2017-06-19
OO refactoring of BsonTrie --> deserialization performance improvement

dnickless:
Hi there

This pull request doesn't change an awful lot in a sense that the executing code is still almost identical to before. Except, with this change, there is less branching going on in the `GetChild(byte keyByte)` method in the `BsonTrie` type. This results in a small performance improvement during deserialization. Also, the longer the field names are that someone deals with the more this improvement will be noticeable.

So what has been changed? The old `BsonTrie` node implementation had a bunch of `if` checks in `GetChild(byte keyByte)` and `AddChild(BsonTrieNode<TValue> child)` which would always differentiate between three cases:
1) a node that does not have any child (a leaf node)
2) a node that has exactly one child (introduced as a performance improvement)
3) a node with more than one child.

While the `AddChild(BsonTrieNode<TValue> child)` method is not relevant in terms of performance since it doesn't get called anymore once the class maps are frozen, the `GetChild(byte keyByte)` method is absolutely crucial during deserialization as it gets called once per every field found in the input stream.

In order to get rid of those `if` statements, I have now extracted the existing code into an OO layout where we now have
1) a `BsonTrieLeafNode` type (representing case 1),
2) a `BsonTrieSingleChildNode` type (representing case 2) and
3) a `BsonTrieMultipleChildrenNode` (representing case 3).

Within the new equivalent of each particular `GetChild` method inside each of these three types (now called `TryGetChild(byte keyByte, out BsonTrieNodeBase<TValue> child)`, another performance improvement introduced in the second check-in) there is next to no branching happening anymore which is why we are faster during deserialization.

As a little gift on top, the memory footprint of the BsonTrie gets smaller in total, too, since each of the three new types only has the fields that it really needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ Please see our [guidelines](CONTRIBUTING.md) for contributing to the driver.
 * Jacob Jewell              jacobjewell@eflexsystems.com
 * Danny Kendrick            https://github.com/dkendrick
 * Brian Knight              brianknight10@gmail.com  
+* Anatoly Koperin           https://github.com/ExM
 * Nik Kolev                 nkolev@gmail.com
 * Oleg Kosmakov             kosmakoff@gmail.com
 * Maksim Krautsou           https://github.com/MaKCbIMKo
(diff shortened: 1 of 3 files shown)


dnickless:
Riiiight, wait a second. Something still seems to be wrong. I'll fix that and see if the performance still looks so much better.

dnickless:
Damn, I just fixed the issue in the code and there goes my and utterly naive hope of improving performance majorly... Still, performance and also memory usage are a little bit better than before plus we have OO code now instead of if/else if/else Spaghetti - but probably a bit harder to understand, too.

PR #281, opened 2017-06-21
Reduction of reflection code --> minor deserialization performance gain

dnickless:
This change is too trivial to even write about it. ;)

There was some reflection code in the `Deserialize` method which would get called repeatedly. But since the ClassMap is frozen (gets checked in the constructor) it should be sufficient to execute this code once upon construction of the Serializer and then use the cached value.


dnickless:
Git confuses the hell out of me. Will try again getting this request right...

PR #282, opened 2017-06-21
Strategy pattern for BsonClassMapSerializer

dnickless:
The `BsonClassMapSerializer.DeserializeClass(BsonDeserializationContext context)` method contained code to handle two basic cases depending on the value of `_classMap.HasCreatorMaps`: The deserialization using a `Dictionary<string, object>` and the "standard" deserialization. By introducing a strategy pattern I could reduce branching (minor performance gain in theory but potentially wiped out due to additional method calls, I get varying results...) and refactor the code into a more OO kind of layout.


dnickless:
I just rebased and optimized the suggested refactoring. Also, there now is some performance-relevant change (less repeated usage of reflection) included which - should you wish not to merge the refactoring part - could be applied to the current code base separately.

BorisDog:
Hi @dnickless, thank you for this architectural improvement suggestion.
We are going to make some major refactoring in this area in scope of improving serialization epic [CSHARP-3230](https://jira.mongodb.org/browse/CSHARP-3230), and unfortunately the proposed changes don't align with the current roadmap.
Therefore we are closing this PR.

PR #283, opened 2017-06-21
Performance: Some more caching and reduced null checks on hot path in BsonSerializer

dnickless:
Again just some algorithmic optimization to make things faster. Mind you, this will have a really tiny effect only... I just avoided some checks that weren't required in the LookupActualType method and cached one reflection call.

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializer.cs
@@ -360,8 +360,6 @@ namespace MongoDB.Bson.Serialization
         /// <returns>A discriminator convention.</returns>
         public static IDiscriminatorConvention LookupDiscriminatorConvention(Type type)
         {
-            var typeInfo = type.GetTypeInfo();
-
             __configLock.EnterReadLock();
             try
             {
@@ -382,6 +380,7 @@ namespace MongoDB.Bson.Serialization
                 IDiscriminatorConvention convention;
                 if (!__discriminatorConventions.TryGetValue(type, out convention))
                 {
+                    var typeInfo = type.GetTypeInfo();
                     if (type == typeof(object))
                     {
                         // if there is no convention registered for object register the default one
@@ -398,8 +397,8 @@ namespace MongoDB.Bson.Serialization
                     {
                         // inherit the discriminator convention from the closest parent (that isn't object) that has one
                         // otherwise default to the standard hierarchical convention
-                        Type parentType = type.GetTypeInfo().BaseType;
-                        while (convention == null)
+                        Type parentType = typeInfo.BaseType;
+                        while (true)
                         {
                             if (parentType == typeof(object))
                             {
@@ -438,7 +437,6 @@ namespace MongoDB.Bson.Serialization
         /// <returns>An IdGenerator for the Id type.</returns>
         public static IIdGenerator LookupIdGenerator(Type type)
         {
-            var typeInfo = type.GetTypeInfo();
             __configLock.EnterReadLock();
             try
             {
@@ -459,6 +457,7 @@ namespace MongoDB.Bson.Serialization
                 IIdGenerator idGenerator;
                 if (!__idGenerators.TryGetValue(type, out idGenerator))
                 {
+                    var typeInfo = type.GetTypeInfo();
                     if (typeInfo.IsValueType && __useZeroIdChecker)
                     {
                         var iEquatableDefinition = typeof(IEquatable<>);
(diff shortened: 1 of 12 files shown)


src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -318,25 +317,32 @@ public static Type LookupActualType(Type nominalType, BsonValue discriminator)
...
                 {
                     actualType = TypeNameDiscriminator.GetActualType(discriminator.AsString); // see if it's a Type name
-                }
 
-                if (actualType == null)
-                {
-                    string message = string.Format("Unknown discriminator value '{0}'.", discriminator);
-                    throw new BsonSerializationException(message);
-                }
+                    if (actualType == null)
```
BorisDog, line 333: Seems that the semantics of raising an exception when actualType == null and discriminator.IsString == false is lost here.

BorisDog (changes requested):

src/MongoDB.Bson/Serialization/BsonSerializer.cs
dnickless: Very correct observation... I reverted this bit as it wasn't helping a lot anyway so now, the changes are very digestible.

BorisDog (approved):

DmitryLukyanov (approved):
nit: this value https://github.com/mongodb/mongo-csharp-driver/pull/283/files#diff-ac6526d3ae431d24b51f464a5746bca97fbf64612afeeb0da6079f6b4d9de2c5R520 can be used in line 542. Also, it would be good to quickly check this file on similar changes. Other than this, LGTM

BorisDog:
Thanks @dnickless, 
In addition @DmitryLukyanov comment, same change can be applied to lines [401](https://github.com/mongodb/mongo-csharp-driver/blob/5672935b662a10d85cc4eb24e7b7a3e34803bc98/src/MongoDB.Bson/Serialization/BsonSerializer.cs#L401) -[363](https://github.com/mongodb/mongo-csharp-driver/blob/5672935b662a10d85cc4eb24e7b7a3e34803bc98/src/MongoDB.Bson/Serialization/BsonSerializer.cs#L363)

DmitryLukyanov (approved):
LGTM, thanks @dnickless

BorisDog:
Jira ticket: CSHARP-3324

src/MongoDB.Bson/Serialization/Conventions/AttributeConventionPack.cs
```diff
@@ -120,8 +120,10 @@ private bool AllowsDuplicate(Type type)
 
             private void OptInMembersWithBsonCreatorMapModifierAttribute(BsonClassMap classMap)
             {
+                TypeInfo classTypeInfo = classMap.ClassType.GetTypeInfo();
```
rstam, line 123: nit: use var

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -318,9 +318,15 @@ public static Type LookupActualType(Type nominalType, BsonValue discriminator)
                             }
                         }
                     }
+
+                    // no need for additional checks, we found the right type
+                    if (actualType != null)
+                    {
+                        return actualType;
+                    }
```
rstam, line 326: This is a bug.

This prevents us from detecting ambiguous discriminators above.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -318,9 +318,15 @@ public static Type LookupActualType(Type nominalType, BsonValue discriminator)
...
                 }
 
-                if (actualType == null && discriminator.IsString)
+                if (discriminator.IsString)
```
rstam, line 329: Removing the check for `actualType == null` would be a bug depending on what we do with the above comment.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -392,8 +397,8 @@ public static IDiscriminatorConvention LookupDiscriminatorConvention(Type type)
                     {
                         // inherit the discriminator convention from the closest parent (that isn't object) that has one
                         // otherwise default to the standard hierarchical convention
-                        Type parentType = type.GetTypeInfo().BaseType;
-                        while (convention == null)
+                        Type parentType = typeInfo.BaseType;
+                        while (true)
```
rstam, line 401: Interesting change.

Perhaps `while (parentType != null)` would be a more appropriate change?

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonSerializer.cs
rstam: Correction: I misinterpreted the nesting level of this early return and thought it was exiting the foreach loop early.

Sorry.

rstam (changes requested):

BorisDog:
Thanks @dnickless! Merged [here](https://github.com/mongodb/mongo-csharp-driver/commit/d74848d44e045e95b4f46e48bd05ef13b49b3001)

PR #284, opened 2017-06-27
CSHARP-2009: Correct discriminator when using OfType<>

MarLoe:
Render correct discriminator when using OfType<>

### DIFF
--- a/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
+++ b/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
@@ -13,7 +13,6 @@
 * limitations under the License.
 */
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -27,13 +26,15 @@ namespace MongoDB.Driver
     {
         // private fields
         private readonly FilterDefinition<TDocument> _filter;
+        private readonly UpdateDefinition<TDocument> _update;
         private readonly IMongoCollection<TDocument> _wrappedCollection;
 
         // constructors
-        public FilteredMongoCollectionBase(IMongoCollection<TDocument> wrappedCollection, FilterDefinition<TDocument> filter)
+        public FilteredMongoCollectionBase(IMongoCollection<TDocument> wrappedCollection, FilterDefinition<TDocument> filter, UpdateDefinition<TDocument> update)
         {
             _wrappedCollection = wrappedCollection;
             _filter = filter;
+            _update = update;
         }
 
         // public properties
@@ -55,6 +56,11 @@ namespace MongoDB.Driver
         public FilterDefinition<TDocument> Filter
         {
             get { return _filter; }
+        }
+
+        public UpdateDefinition<TDocument> Update
+        {
+            get { return _update; }
         }
 
         public override IMongoIndexManager<TDocument> Indexes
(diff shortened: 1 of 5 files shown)


MarLoe:
This is an attempt to address the bug reported here: [CSHARP-2009](https://jira.mongodb.org/browse/CSHARP-2009)

Please pay attention to the two added unit tests that exposes the problem.

src/MongoDB.Driver/FilteredMongoCollectionBase.cs
```diff
@@ -26,13 +26,15 @@ internal abstract class FilteredMongoCollectionBase<TDocument> : MongoCollection
     {
         // private fields
         private readonly FilterDefinition<TDocument> _filter;
+        private readonly UpdateDefinition<TDocument> _update;
         private readonly IMongoCollection<TDocument> _wrappedCollection;
 
         // constructors
-        public FilteredMongoCollectionBase(IMongoCollection<TDocument> wrappedCollection, FilterDefinition<TDocument> filter)
+        public FilteredMongoCollectionBase(IMongoCollection<TDocument> wrappedCollection, FilterDefinition<TDocument> filter, UpdateDefinition<TDocument> update)
```
craiggwilson: This is a breaking change. It will be necessary to create another ctor and deprecate this one. I doubt anyone is using this directly, but it's still part of the public API.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -369,7 +369,16 @@ public override IFilteredMongoCollection<TDerivedDocument> OfType<TDerivedDocume
...
             var ofTypeFilter = new BsonDocumentFilterDefinition<TDerivedDocument>(renderedOfTypeFilter);
 
-            return new OfTypeMongoCollection<TDocument, TDerivedDocument>(this, derivedDocumentCollection, ofTypeFilter);
+            UpdateDefinition<TDerivedDocument> derivedUpdate = null;
+            var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(typeof(TDocument));
+            var discriminatorValue = discriminatorConvention?.GetDiscriminator(typeof(TDocument), typeof(TDerivedDocument));
+            if (discriminatorValue?.IsBsonArray == true)
+            {
+                var builder = new UpdateDefinitionBuilder<TDerivedDocument>();
+                derivedUpdate = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
```
craiggwilson: Nice! I wonder if it's possible for a user to create a discriminator convention that is based on something other than the type hierarchy, making it possible for this to change at runtime. That would be kinda weird.

src/MongoDB.Driver/OfTypeMongoCollection.cs
```diff
@@ -27,8 +27,9 @@ internal class OfTypeMongoCollection<TRootDocument, TDerivedDocument> : Filtered
         public OfTypeMongoCollection(
             IMongoCollection<TRootDocument> rootDocumentCollection,
             IMongoCollection<TDerivedDocument> derivedDocumentCollection,
-            FilterDefinition<TDerivedDocument> ofTypeFilter)
-            : base(derivedDocumentCollection, ofTypeFilter)
+            FilterDefinition<TDerivedDocument> ofTypeFilter,
+            UpdateDefinition<TDerivedDocument> derivedUpdate)
```
craiggwilson: Same here with needing a separate ctor.

craiggwilson: Nevermind... this is an internal class.

src/MongoDB.Driver/FilteredMongoCollectionBase.cs
craiggwilson: Nevermind... this is an internal class.

src/MongoDB.Driver/MongoCollectionImpl.cs
MarLoe: @craiggwilson, you're right. However, this was "inspired" from DocumentOfTypeFilterDefinition.Render ([#1925](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/FilterDefinitionBuilder.cs#L1925)) where a special handling case for arrays is put in place.
So I guess that other things will break as well if a non-array custom discriminator convention is put in place.

MarLoe:
@craiggwilson This addresses the "corner-case" where an upsert leads to an insert and the full discriminator convention is in use. I tried to explain this in the bug report I filed ([CSHARP-2009](https://jira.mongodb.org/browse/CSHARP-2009)). But it was files prior to my attempt of actually trying to fix it.

ondrejpialek:
Just hit this bug and honestly and I am having a hard time rewriting our collection code not to use `OfType<>`. We use upserts everywhere and read back the modified document, so type safety is quite useful in this case.

Any chance of this PR making it into the product any time soon?

BorisDog:
@MarLoe thanks for bringing this issue to our attention!
We've fixed this issue [here](https://github.com/mongodb/mongo-csharp-driver/commit/2de35a4c9e6f5bcb76ec1bc63a242c4333c4d276).

PR #286, opened 2017-07-18
Update README.md

kosmakoff:
Replace my email with GitHub profile URL


kosmakoff:
Basically, just got an email from weird guy who decided it's clever thing to just fetch all email addresses from this project's readme and do mass mailing with questions. With this change I hope to minimize such occurrences in future.

PR #287, opened 2017-08-10
CSHARP-2029: maintenance interval used every time

ExM:
fix by https://jira.mongodb.org/browse/CSHARP-2029


jyemin:
I've been reviewing open PRs, and realize that this duplicates a fix that has since been made in driver.  See https://jira.mongodb.org/browse/CSHARP-2029 for details.

ExM:
This is not a duplicate.
There MinConnectionPoolSize was set to 1 to indicate that the maintenance interval is not being used
with other MinConnectionPoolSize values the effect is preserved, but in this case more parallel connections are required to demonstrate.

PR #288, opened 2017-08-10
CSHARP-2030: Added allowDuplicateElementNames to BsonDocument.Parse

ar7z1:
In this PR I fixed [[CSHARP-2030] Add allowDuplicateElementNames to BsonDocument.Parse](https://jira.mongodb.org/browse/CSHARP-2030).


rstam:
See the JIRA ticket for the reasoning behind why we don't think this feature should be added.

Thanks for your suggestion.

PR #289, opened 2017-08-23
Solve issue in .net core 2.0

kimiko88:
NU1605 issue could happen when you test a project that use mongoDB.Driver.Core library

### DIFF
--- a/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
+++ b/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
@@ -36,9 +36,8 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Core" />
-    <Reference Include="System.Runtime.InteropServices.RuntimeInformation, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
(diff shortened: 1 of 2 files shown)


BigFishVN:
Can you please merge this pull request to fix NU1605 issue? I can't publish my app which was use net core 2 with MongoDB.Driver 2.4.4

rstam:
Can you please explain how this change resolves the problem?

The existing dependency on System.Runtime.InteropServices.RuntimeInformation is on version >= 4.0.0. Why does that not already work?

BigFishVN:
@rstam I created [a sample project](https://github.com/BigFishVN/mongo-netcore2-error) to show the errors when used _dotnet --publish -r (Windows RID)_. The NU1605 happens after I referenced [Microsoft.AspNetCore.Authentication](https://github.com/aspnet/Security) **(2.0.0)**. Hope this sample may help to solve the problem.

BigFishVN:
I found the problem. With Visual Studio 2017, project.json is [no longer available](https://docs.microsoft.com/en-us/dotnet/core/tools/cli-msbuild-architecture). The project properties treat warnings as errors contain NU1605 as default, which you can see in the image below:
capture
I can bypass this error by select "None" and can publish the app. Hope the next version of MongoDB.Driver will remove this warning as default.

jyemin:
Hi @kimiko88 I think we've fixed this in a subsequent release, but if not please open a Jira issue to discuss it further.

PR #290, opened 2017-09-01
Fixing method summaries to be coherent

alexandrevribeiro:
Fixing the summary of both "GetSerializer" methods to be coherent with what they actually do. 
They not only gets a registered serializer, but they also ADDs it when it wasn't registered before (behavior I disagree BTW). And it took me a while to find out this was the problem I was facing.

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
@@ -40,7 +40,8 @@ namespace MongoDB.Bson.Serialization
 
         // public methods
         /// <summary>
-        /// Gets the serializer (explicitly registered or default) for the specified <paramref name="type" />.
+        /// Gets the serializer for the specified <paramref name="type" />.
+        /// If none is already registered, the serialization providers will be used to create a serializer and it will be automatically registered.
         /// </summary>
         /// <param name="type">The type.</param>
         /// <returns>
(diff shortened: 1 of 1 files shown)


BorisDog (approved):

JamesKovacs (approved):
LGTM

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -40,7 +40,7 @@ public BsonSerializerRegistry()
 
         // public methods
         /// <summary>
-        /// Gets the serializer for the specified <paramref name="type" />.
+        /// Gets the serializer (explicitly registered or default) for the specified <paramref name="type" />.
```
rstam, line 43: This change is misleading. There is no such thing as a "default" serializer.

What actually happens is that if no serializer is already registered the serialization providers are used to get one and it is automatically registered.

I would suggest a better change would be:

    /// Gets the serializer for the specified <paramref name="type" />. If none is already registered, the serialization providers will be used to create a serializer and it will be automatically registered.

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -63,7 +63,7 @@ public IBsonSerializer GetSerializer(Type type)
         }
 
         /// <summary>
-        /// Gets the serializer for the specified <typeparamref name="T" />.
+        /// Gets the serializer (explicitly registered or default) for the specified <typeparamref name="T" />.
```
rstam, line 66: See previous comment.

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
BorisDog: Thanks, done.

JamesKovacs (approved):
LGTM

rstam (approved):
LGTM

BorisDog:
[Pushed](https://github.com/mongodb/mongo-csharp-driver/commit/7945ed58c4ddae5f26ef7fb3b6ebff7f69d1490c)

PR #292, opened 2017-10-06
CSHARP-2043: Added SuppressEnsureIndexes option to GridFSBucketOption…

jg11jg:
CSHARP-2043: Added SuppressEnsureIndexes option to GridFSBucketOptions.  By default is false and has no impact on any part of the system.  HOWEVER, IT CAN BE A USEFUL OPTION TO HAVE AVAILABLE IN A CLUSTER WHERE GRIDFS COLLECTIONS ARE BEING SHARDED.  

If set to true, then suppresses the call to EnsureIndexes which occurs during every GridFS upload call.  Benefit of this suppression is that: 1) EnsureIndexes [effectively] does a count against the fs.files collection, which will a) add latency in a cluster with global shards, b) requires find permission which may not be deisrable in a write-only repository and most importantly c) cause entire cluster to become unavailable if any single shard is unavailable, even if data being saved is not housed in the unavailable shard.  2) Less importantly, EnsureIndexes includes a list index command which implies a higher privilege than is really needed.  If developer is a) confident that GridFS subsystem is properly set up and b) wants to robustify sharded GridFS against one shard being down c) wants keep client permissions to a minimum

### DIFF
--- a/src/MongoDB.Driver.GridFS/GridFSBucket.cs
+++ b/src/MongoDB.Driver.GridFS/GridFSBucket.cs
@@ -814,16 +814,19 @@ namespace MongoDB.Driver.GridFS
             {
                 if (!_ensureIndexesDone)
                 {
-                    var isFilesCollectionEmpty = IsFilesCollectionEmpty(binding, cancellationToken);
-                    if (isFilesCollectionEmpty)
+                    if (!_options.SuppressEnsureIndexes)
                     {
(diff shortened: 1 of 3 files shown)


rstam:
See comments in JIRA ticket:

https://jira.mongodb.org/browse/CSHARP-2043

papafe:
Superseded by https://github.com/mongodb/mongo-csharp-driver/pull/2047

PR #296, opened 2017-10-13
Adjust collection and index read preference

SnailDev:
The default readprefence is primary so that I list collections and indexs on slave database and collection will be fail.
I think collection and index list operation could be just read operation that can be special setting. please consider it, thank you so much.


jyemin:
While this is a reasonable idea, the behavior has been specified this way for all of our drivers, and we're unlikely to change it now.  It could also be consider backwards breaking.

My apologies for letting this sit so long unattended.

PR #297, opened 2017-10-17
Fixed [CSHARP-2067]

alkampfergit:
Added correct async call when connecting socket asynchronously.

Actually I have a project where with parallel unit testing (xunit) the code has timeout connection. The original code is not so good because it will use a Task.Run to execute the connection, when there is the connectAsync() socket method. Moreover without using the ConfigureAwait(false) there is the possibilty of a deadlock.


alkampfergit:
I noticed that ALL pull requests are failing builds, you should check and fix the build I suppose.

andrii-litvinov:
I experience same issue after migrating project to .net core 2.1. Any idea when this PR can be accepted?

vincentkam:
Hello Gian Maria,
Thank you so much for your pull request! 

Hello Andrii Litvinov and Gian Maria,
We currently have a fix for this as part of another PR (https://github.com/vincentkam/mongo-csharp-driver/pull/18) that will hopefully be merged no later than the end of the month.

PR #298, opened 2017-10-19
Fix MongoClientSettings/MongoServerSettings equality comparer

jbatonnet:
This PR should fix an issue where the following code this trigger a MongoException "MongoServer.Create has already created XXX servers which is the maximum number of servers allowed."
The Equals method of MongoClientSettings/MongoServerSettings does return false even if the settings are the same accross several calls.

```cs
class Test
{
    static void ConfigureCluster(ClusterBuilder builder)
    {
        // Do something cool
    }

    void Test()
    {
        for (int i = 0; i < 200; i++)
        {
            var settings = MongoClientSettings.FromUrl(MongoUrl.Create("XXX"));
            settings.ClusterConfigurator = ConfigureCluster;

            var client = new MongoClient(settings);
            var server = client.GetServer();
        }
    }
}
```

## ClusterConfigurator delegate comparison

ClusterConfigurator is a delegate. According to C# spec 7.10.8, the == operator performs some additional checks than ReferenceEquals.
If we use the same static method as callback, Equals should return true.
The sample above illustates this very case.

## WriteConcern comparison

WriteConcern is a class that does not implement == operator.
We should use object.Equals method in order to compare them, just like ReadConcern.

## Testing consideration

The test is faulty here. It does use the Clone function to perform various Equality tests. This Clone method does copy all fields references instead of a deep copy.
Since Clone is part of the public interface, we can't modify it to deep copy as it would change the current behavior.
I added a few lines to deep clone the MongoClientSettings/MongoServerSettings in order to make the first test fail without my PR.

### DIFF
--- a/src/MongoDB.Driver.Legacy/MongoServerSettings.cs
+++ b/src/MongoDB.Driver.Legacy/MongoServerSettings.cs
@@ -718,7 +718,7 @@ namespace MongoDB.Driver
             var rhs = (MongoServerSettings)obj;
             return
                 _applicationName == rhs._applicationName &&
-                object.ReferenceEquals(_clusterConfigurator, rhs._clusterConfigurator) &&
+                _clusterConfigurator == rhs._clusterConfigurator &&
                _connectionMode == rhs._connectionMode &&
                _connectTimeout == rhs._connectTimeout &&
                _credentials == rhs._credentials &&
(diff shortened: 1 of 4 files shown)


jbatonnet:
AppVeyor CI seems to fail itself, not related to this PR:

image

Also, please note that this PR would need to be backported to all maintenance branches.
I use 2.2.4 driver as NuGet package.

jyemin:
First of all, I apologize for leaving this PR sitting unattended for so long.  Unfortunately, it's no longer mergeable due to conflicts.

If anyone is still interested in pursuing this change, please open a Jira issue at https://jira.mongodb.org/projects/CSHARP to start a discussion on whether we would want to consider this.

PR #301, opened 2017-11-01
Documentation improvements for Skunkworks

dgolub:
This pull request fills in most of the missing examples in the C# driver documentation and also fixes a batch of typos that I spotted.

@shannonbradshaw

### DIFF
--- a/Docs/reference/content/reference/driver/crud/linq.md
+++ b/Docs/reference/content/reference/driver/crud/linq.md
@@ -491,25 +491,6 @@ var result = collection.AsQueryable().Select(p => p.Age).Max();
 ]
 ```
 
-
-#### Sum
-
-All forms of [`Sum`]({{< msdnref "system.linq.queryable.Sum" >}}) are supported.
-
-```csharp
-var result = collection.AsQueryable().Sum(p => p.Age);
-
-// or
-
-var result = collection.AsQueryable().Select(p => p.Age).Sum();
-```
-```json
-[
-    { $group: { _id: 1, __result: { $sum: "$Age" } } }
-]
-```
-
-
 #### OfType
 
 All forms of [`OfType`]({{< msdnref "system.linq.queryable.oftype" >}}) are supported.
--- a/Docs/reference/content/reference/driver/crud/linq.md
+++ b/Docs/reference/content/reference/driver/crud/linq.md
@@ -307,8 +307,7 @@ var query = from p in collection.AsQueryable()
 ]
 ```
 
-.. note::
-   An anonymous type, as above, has a constructor with two parameters as required.
+{{% note %}}An anonymous type, as above, has a constructor with two parameters as required.{{% /note %}}
 
 Sometimes, the compiler will also generate this two-parameter anonymous type transparently. Below is an example of this with a custom projection:
 
@@ -506,7 +505,7 @@ var result = collection.AsQueryable().Select(p => p.Age).Sum();
 ```
 ```json
 [
-    { $group: { _id: 1, __result: { $Sum: "$Age" } } }
+    { $group: { _id: 1, __result: { $sum: "$Age" } } }
 ]
 ```
 
--- a/Docs/reference/content/reference/driver/expressions.md
+++ b/Docs/reference/content/reference/driver/expressions.md
@@ -272,12 +272,6 @@ Find(p => p.Age % 2 == 0)
 
 #### $regex
 
-```csharp
-Find(p => p.Name.IndexOf("John") == 0)
-```
-```json
-{ Name: /^(?!.{0,-1}John).{0}John/s }
-```
 ---
 ```csharp
 Find(p => p.Name[0] == 'a')
--- a/Docs/reference/content/reference/driver/expressions.md
+++ b/Docs/reference/content/reference/driver/expressions.md
@@ -248,6 +248,15 @@ Find(p => p.Children.ContainsKey("John"))
 { "Children.John": { $exists: true } }
 ```
 
+#### $type
+
+```csharp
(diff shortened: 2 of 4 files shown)


BorisDog (approved):

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -251,55 +255,68 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#evaluatio
...
-
+---
 ```csharp
-// no example yet
+Find(p => p.Name.ToLower().Contains("John"))
+```
+```json
+{ Name: /john/is }
 ```
+{{% note %}}`ToLowerInvariant`, `ToUpper`, `ToUpperInvariant`, `Trim`, `TrimStart`, and `TrimEnd` are also supported for the first function call. `StartsWith` and `EndsWith` are also supported for the second function call.{{% /note %}}
```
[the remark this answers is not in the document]
JamesKovacs: 👍

Docs/reference/content/reference/driver/crud/linq.md
```diff
@@ -492,25 +491,6 @@ var result = collection.AsQueryable().Select(p => p.Age).Max();
...
-
-var result = collection.AsQueryable().Select(p => p.Age).Sum();
-```
-```json
-[
-    { $group: { _id: 1, __result: { $Sum: "$Age" } } }
-]
-```
-
-
```
JamesKovacs, line 513: Why are we removing `$sum`? We should use a better example than summing ages (which makes little to no sense), but we should be removing the operator example entirely.

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -235,13 +242,10 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#element"
...
 
 ```csharp
-// no example yet
+Find(p => p.Children.ContainsKey("John"))
 ```
-
-#### $type
-
-```csharp
-// no example yet
```
JamesKovacs, line 244: The `$type` example is being removed. We should include it and provide an example.

JamesKovacs (changes requested):
Overall looks good, but we are losing examples/heading for `$sum` and `$type`. We should include examples of those operators.

Docs/reference/content/reference/driver/crud/linq.md
BorisDog: It's a duplicate, removing one of the instances :)

Docs/reference/content/reference/driver/expressions.md
BorisDog: Thanks for catching this. Added an example

JamesKovacs (approved):
LGTM

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -251,14 +264,60 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#evaluatio
...
 ```
 
 #### $regex
 
 ```csharp
-// no example yet
+Find(p => p.Name.IndexOf("John") == 0)
+```
+```json
+{ Name: /^(?!.{0,-1}John).{0}John/s }
```
rstam, line 279: This one is surprising to say the least!

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -251,14 +264,60 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#evaluatio
...
+```
+---
+```csharp
+Find(p => p.Name[0] == 'a')
+```
+```json
+{ Name: /^.{0}a/s }
```
rstam, line 286: Another surprising one...

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -251,14 +264,60 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#evaluatio
...
+```
+```json
+{ Name: /^.{5,}$/s }
+```
+---
+```
+Find(p => p.Name.ToLower() == "john")
+```
+```json
+{ Name: /^john$/i }
```
rstam, line 300: While it's true that this is what we are currently returning is it correct?

What about locales or collations?

rstam: If we're not sure this is correct should we be documenting it?

Same question for several other surprising results...

rstam (changes requested):
My only potential change request is where I comment that we might be documenting things we don't want to document here.

rstam (changes requested):
Putting back in request changes status because there are unanswered comments.

Docs/reference/content/reference/driver/expressions.md
BorisDog: Removed

Docs/reference/content/reference/driver/expressions.md
BorisDog: Works as expected

Docs/reference/content/reference/driver/expressions.md
BorisDog: This example of `'ToLower'` would work as expected regardless of collations.

rstam: I'm not sure that's a true statement, but it will  work with most values (at least with the standard collation).

It should work with the value `john` if that's what you mean.

rstam (approved):
LGTM

BorisDog:
[Pushed](https://github.com/BorisDog/mongo-csharp-driver/commit/9e538e2cd6e2ea5f1e1ecf1358ebfe7e963d596a)

PR #303, opened 2017-11-13
CSHARP-1999: Support for readonly properties that contain instances

dnickless:
This PR implements https://jira.mongodb.org/browse/CSHARP-1999.

´BsonMemberMap´ now supports a `SetUseExistingInstance(bool useExistingInstance)` method (there's also a convention and an attribute) which can be used in order to tell the driver to deserialize a document directly into an existing instance (either straight from the root level or individually for a configured property).

This allows for some greater level of flexibility around entity design:

- Readonly auto-initialized properties can now be populated by the driver which wasn't possible before:
  - `public List<int> Foo { get; } = new List<int>();`
  - `public SomeType Bar { get; } = new SomeType();`)
- Also, in cases of heavy constructors where the instantiation of an entity type `A` through the driver would immediately trigger the creation of some contained nested instances `B` (and `C` and so on...) as part of a top-level constructor call, superfluous constructor calls can be avoided by reusing the already created instances.
- As a side-product of the implementation, existing instances can now be partially/incrementally populated by the driver (e.g. through `BsonSerializer.Deserialize(bsonReader, existingInstance: existingInstance);`) which may be an interesting feature in some slightly more arcane scenarios:
  - e.g. for entities that need to hold data from multiple source (read: collections/queries)
  - or when a user should be able to populate an entity partially which, however, then needs to be enriched with some fields that are stored in the database.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonUseExistingInstanceAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonUseExistingInstanceAttribute.cs
@@ -1,4 +1,4 @@
-﻿/* Copyright 2010-2014 MongoDB Inc.
+﻿/* Copyright 2010-present MongoDB Inc.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


fes3507:
I'm waiting for this change to be applied in release. Why it is not marged for such a long time?

JamesKovacs:
Thank you for the PR. Unfortunately the PR changes the internal serialization semantics and could introduce subtle bugs. Additionally there are supported ways of using readonly properties and fields via the use of overloaded constructors. Please see [our analysis and recommendations](https://jira.mongodb.org/browse/CSHARP-1999?focusedCommentId=3464070&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3464070) in the associated JIRA ticket.

PR #304, opened 2017-11-13
Reduction of reflection calls --> minor deserialization performance gain

dnickless:
There was some reflection code in the `Deserialize` method which would get called repeatedly. But since the ClassMap is frozen (gets checked in the constructor) it should be sufficient to execute this code once upon construction of the `BsonClassMapSerializer` and then use the cached value.


BorisDog:
Thanks @dnickless for your contribution. This file has changed a few times since this PR, and this fix is not relevant anymore.

PR #305, opened 2017-11-15
CSHARP-2096: Make EnumRepresentationConvention also affect collections of Enums

dnickless:
The entire issue is documented here:

https://jira.mongodb.org/browse/CSHARP-2096

 With this fix, the EnumRepresentationConvention should now work as expected.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
+++ b/src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
@@ -50,49 +50,65 @@ namespace MongoDB.Bson.Serialization.Conventions
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
-            var memberType = memberMap.MemberType;
-            var memberTypeInfo = memberType.GetTypeInfo();
-
-            if (memberTypeInfo.IsEnum)
+            var serializer = memberMap.GetSerializer();
+            var reconfiguredSerializer = Apply(serializer);
+            if (reconfiguredSerializer != null)
             {
-                var serializer = memberMap.GetSerializer();
+                memberMap.SetSerializer(reconfiguredSerializer);
+            }
+        }
+
+        private IBsonSerializer Apply(IBsonSerializer serializer)
+        {
+#if NETSTANDARD1_5 || NETSTANDARD1_6
+            var type = serializer.ValueType.GetTypeInfo();
+#else
+            var type = serializer.ValueType;
+#endif
+
+            if (type.IsEnum)
+            {
                 var representationConfigurableSerializer = serializer as IRepresentationConfigurable;
                 if (representationConfigurableSerializer != null)
                 {
                     var reconfiguredSerializer = representationConfigurableSerializer.WithRepresentation(_representation);
-                    memberMap.SetSerializer(reconfiguredSerializer);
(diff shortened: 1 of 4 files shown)


narciero:
Any progress on getting this merged?

vincentkam:
Hi @narciero, 
Thank you for your interest in this ticket and PR! We are currently focusing on 4.2 features and will get back to investigating this ticket and PR as soon as 4.2 work slows down.

GDreyV:
@vincentkam, any updates on that? it was created in 2017!

JesApp:
Just ran into this problem as well. You are already at 4.4 and still no news on this issue?

Please think about prioritising this. The fact that enums in a collection do not get serialized as strings, despite the `EnumRepresentationConvention(BsonType.String)` being set, is extremely counterintuitive. This pull request shouldn't be considered a feature - or "improvement" as [CSHARP-2096](https://jira.mongodb.org/browse/CSHARP-2096) calls it - it should be considered a bugfix. Alternatively, this should be closed and the `EnumRepresentationConvention(BsonType.String)` should be removed from the codebase, so the users know that this driver can't serialize enums as strings.

JamesKovacs:
We acknowledge that this PR has been sitting idle for a considerable amount of time and appreciate everyone's continued interest in these improvements to the MongoDB .NET Driver. We are investing in improvements to make serialization behaviour in the driver more intuitive and less error prone. This is an example where the driver does not behave as expected necessarily.

If you enable the `EnumRepresentationConvention` or annotate an enum property/field with `[BsonRepresentation(BsonType.String)]`, the enum is serialized as its string representation rather than its integer representation as expected. This PR is meant to address the case where you are serializing a collection of enums and wish each one serialized as a string. Unfortunately the PR only works for `List<T>` and `Dictionary<K,V>`. It does not work for arrays of enums (`T[]`), other collection types (queues, stacks, linked lists, etc.), nor collection interfaces (`IList<T>`, `IDictionary<K,V>`, `IEnumerable<T>`, `IQueryable<T>`, `ICollection<T>`, etc.). Even if we supported all these collection types and interfaces, it would not work for custom collection types.

Rather than focusing on enums, let's consider the more general case of serialization conventions on collection types. The .NET Driver does not currently support the following either:

```
[BsonRepresentation(BsonType.String)]
public IEnumerable<ObjectId> ObjectIdCollection { get; set; }
```

What is the expected result? Should each item in the collection be serialized as a string or should the entire collection be serialized as a single comma-separated string?

Another case to consider is that of `Dictionary<K,V>`:

```
[BsonRepresentation(BsonType.String)]
public Dictionary<ObjectId, int> SomeDictionary { get; set; }
```

Should both keys and values be serialized as strings? Should we add a `Target` parameter to specify what it applies to? Do we allow applying two `BsonRepresentationAttributes`, one for `Key` and another for `Value`?

The original problem of serializing collections of enums as their string representation can be solved either with a custom serializer on the containing type or for the enum itself. The other two mentioned cases can be solved similarly.

We realize that this is not an ideal solution and we would like to provide better serialization semantics in the driver. We are reviewing serialization feature requests and bugs in an attempt to implement a more consistent and holistic approach to address problems such as this. We appreciate your patience and continued feedback on [CSHARP-2096](https://jira.mongodb.org/browse/CSHARP-2096).

brunomello123:
I just ran in this problema today...  2024

PR #306, opened 2017-11-27
CSHARP-2114: MongoIndexManager reads list of indexes not only from primary

ar7z1:
Issue: https://jira.mongodb.org/browse/CSHARP-2114.

[Indexes_List_should_execute_a_ListIndexesOperation](https://github.com/mongodb/mongo-csharp-driver/blob/1c20602fb9e95e393c452e91109caa467c541c84/tests/MongoDB.Driver.Tests/MongoCollectionImplTests.cs#L1721) (test that checks "list indexes" operation) works after my changes.


jyemin:
As the linked Jira issue CSHARP-2114 has been closed works-as-designed, we don't plan to pursue this change.

PR #308, opened 2017-12-10
CSHARP-2123 Change maxTimeMS set from double to int

byTimo:
Issue: https://jira.mongodb.org/browse/CSHARP-2123


rstam:
Thank you for the pull request.

Merged and then modified slightly, primarily to round up instead of down and to add more testing.

PR #310, opened 2017-12-14
CSHARP-2127: Fix for EnumSerializer to support overflowing UInt32s

dnickless:
Details are here:
JIRA: https://jira.mongodb.org/browse/CSHARP-2127
StackOverflow: https://stackoverflow.com/questions/47781059/c-sharp-mongodb-tries-to-convert-enumuint32-to-int32-and-crashes

This change enables the use of the `AllowOverflow` property of the `BsonRepresentation` attribute on Enum properties.

And, yes, having UInt32s as the underlying type of an Enum certainly seems like an unusual case to start with. The fix is pretty little invasive, though, so I thought I'd create this PR for it.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Serializers/EnumSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/EnumSerializer.cs
@@ -14,12 +14,7 @@
 */
 
 using System;
-using System.IO;
-using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


BorisDog:
Thanks @dnickless! Fixed [here](https://github.com/mongodb/mongo-csharp-driver/commit/3592cf28b7de3c007cb74370ea764d0e2a5c3f6d).

PR #311, opened 2017-12-19
CSHARP-2132: Fixed issue with duplicate property and added unit tests.

CyAScott:


rstam:
Thanks for the pull request, but it turns out that this change is not needed. 

See the comments in:

https://jira.mongodb.org/browse/CSHARP-2132

PR #312, opened 2018-01-10
BsonType.Timestamp support in ObjectSerializer

udda:
### Scenario

Calling `IMongoDatabase.RunCommandAsync<object>` fails when response contains values with BSON type *Timestamp*. This happens at every request when using replica sets.

A `FormatException` is thrown saying: *ObjectSerializer does not support BSON type Timestamp*.  
Specialized serializers (like `BsonTimestampSerializer`) can't be used here, because the anonymous response type doesn't allow class mapping.

### Solution

Add an entry to `ObjectSerializer` type recognition code, in order to support *Timestamp* deserialization.

That's already done for other BSON types (binary, int, decimal, array, string, objectId, etc.)

### Notes

[CSHARP-1162](https://jira.mongodb.org/browse/CSHARP-1162) says that Timestamp class is for internal use only, this may be the reason why it's not supported. This PR does not aim to change that.

Nevertheless, `RunCommandAsync` (and possibly other methods) just fail with collections that contain a Timestamp value. This is often out of programmers control, and the proposed change may be a painless solution.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
@@ -129,6 +129,9 @@ namespace MongoDB.Bson.Serialization.Serializers
                 case BsonType.String:
                     return bsonReader.ReadString();
 
+                case BsonType.Timestamp:
+                    return bsonReader.ReadTimestamp();
+
                 default:
                     var message = string.Format("ObjectSerializer does not support BSON type '{0}'.", bsonType);
                     throw new FormatException(message);


dnickless:
We are also experiencing the exact same bug - our clients cannot run our application against a replica set. Please merge this PR!

dnickless:
I created a JIRA ticket: [CSHARP-2668](https://jira.mongodb.org/browse/CSHARP-2668)

jyemin:
@udda 

What command are you attempting to execute?

udda:
@jyemin I don't really remember since it happened over a year ago, but I think that *every* command (not queries) caused the error. There was a field named like `t` or `_ts`, of type `Timestamp`, containing the operation time. Not sure whether it's still there in MongoDB 4.

jyemin:
@udda yes, you're right.  The field is indeed `operationTime` (as well as `$clusterTime.clusterTime`)

idanbd:
I'm experiencing the same issue, wondering what's stopping this from being merged?

BorisDog:
Hello @idanbd, 
Unfortunately this PR won't fix the failing `IMongoDatabase.RunCommandAsync<object>`, as` BsonType.Binary` is not supported by `ObjectSerializer` as well. The reason for this is that `ObjectSerializer` initially was not designed to support types which don't have direct mapping to .NET types.
Using `IMongoDatabase.RunCommandAsync<BsonDodocument>` instead should resolve this issue.

JamesKovacs:
As noted above, simply adding `BsonTimestamp` support (which is intended as an internal type) does not fully address this issue. The recommended approach is to use `IMongoDatabase.RunCommandAsync<BsonDocument>`. Please comment on [CSHARP-2668](https://jira.mongodb.org/browse/CSHARP-2668) if you have additional input. Thanks!

PR #313, opened 2018-01-26
CSHARP-2155: Fixed ObjectSerializer serializing Decimal as Discriminated Value and added unit test

Dyedan:


rstam:
We can't make this change because .NET's Decimal type and BSON's Decimal128 type are not the same.

See https://jira.mongodb.org/browse/CSHARP-2155 for a fuller explanation.

PR #314, opened 2018-01-30
CSHARP-1378 BulkWrite enumerates requests argument multiple times

M-frankied:
`BulkWrite` currently enumerates its `requests` param 3 times:
- to check whether it is empty
- to build the bulkwrite operation
- to copy `requests` into the result object

That behaviour can be very costly if `requests` represents e.g. a remote db table.

This pull requests contains 2 commits:

- **CSHARP-1378 enumerate bulkwrite request list once**: Simply copies the enumerable to a list once and uses that list everywhere the enumerable was used. Includes a unit test that fails for the case where `BulkWrite`/`BulkWriteAsync` enumerates the enumerable multiple times. 

- **CSHARP-1378 remove extra list copy**: With the above commit `BulkWrite` does one more list copy than in the baseline case. This second commit optimizes that extra copy away because I don't want to worsen performance for cases where `requests` is cheap to enumerate. I put it in a separate commit because you might find avoiding the extra list copy not worth the hassle because "computers are fast now".

Some tests fail in my ad-hoc environment, but then they also fail there without these changes, so that should be ok.

### DIFF
--- a/src/MongoDB.Driver/BulkWriteResult.cs
+++ b/src/MongoDB.Driver/BulkWriteResult.cs
@@ -113,10 +113,10 @@ namespace MongoDB.Driver
         /// <param name="processedRequests">The processed requests.</param>
         protected BulkWriteResult(
             int requestCount,
-            IEnumerable<WriteModel<TDocument>> processedRequests)
+            List<WriteModel<TDocument>> processedRequests)
             : base(requestCount)
         {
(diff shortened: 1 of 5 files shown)


sanych-sun:
Thank you for your contribution. This PR was very helpful and excellent starting point for me to fix the original issue in another PR #1298.

PR #315, opened 2018-02-20
CSHARP-2011

otemnov:


BorisDog:
Hi @otemnov 
We have been unable to reproduce this issue and the driver's networking code has evolved since this ticket was filed. Most notably maxConnecting limits the number of simultaneous connection attempts to a destination node. Please reopen this ticket with a comment if you observe this problematic behavior in recent driver releases along with steps to reproduce.

PR #316, opened 2018-03-10
CSHARP-1144 Fix MongoWaitQueueFullException

galamai:
This fix solves the problem of MongoWaitQueueFullException on requests from many threads.


rstam:
Thank you for submitting this pull request.

Server selection limits the number of threads/tasks that can be waiting for server selection to complete as a protective measure to prevent an unbounded number of threads/tasks from queueing up. This change attempts to circumvent that by putting an unbounded queue in front of server selection (indirectly, since it's actually gating access to StartImplicitSession).

Instead, you can configure a larger value for MaxServerSelectionWaitQueueSize if the default value of 500 is too small for your load.

PR #317, opened 2018-03-12
Support for 64 bit skip and limit in querys

ravenpride:
Hello,

at the moment I'm working on a project that makes use of a MongoDB that might span multiple MongoDB nodes (depending on the customer's needs). The database will store many small documents, so there may come the day where the number of documents exceeds an 32 bit integer value. This is a limitation the C# MongoDB driver introduces.

I've adapted the code to support querys to skipping 64 bit distances and fetching up to 2^63-1 documents. The size of a batch is still an 32 bit integer (due to the limitation of .NET collections). The driver provides an option to fetch all data in a single batch. That option works as before for document counts up to 2^31-1. If there are more documents to fetch, multiple batches á max. 2^31-1 documents are returned, so there should not be a problem with existing installations.

I would appreciate to see the change in your codebase.

PS: I'd like to have the unit tests run, but VS 2017 seems to have a problem with the referenced version of xunit. Updating to the most recent version of xunit threw a few thousand errors due to severe syntactic changes, so I hope everything is fine anyway ;-)


ravenpride:
Could the development team please make a brief statement as to whether the change is of interest or whether the limitation is by design? I'd love to work with big result sets. This is probably a borderline case, but I'd like to cover it.

rstam:

I have a couple concerns about these changes.

The first is that this would be a breaking change since some data types are changing.

The second is that large skip values are inherently inefficient and should not be encouraged. The server still has to do all (or most) of the work to read the documents and count them to skip over the initial documents before starting to return the rest of the documents. Therefore skip values should be relatively small.

Large limits would seem to be a bit more reasonable, but an easy work around is for an application to simply stop reading when it has received as many documents as it wishes to receive. Usually when a limit is used it is set to a relatively low value. It is particularly useful when the limit is very small, as it prevents the server from returning more documents than necessary in the first batch.

ravenpride:
Thank you very much for the detailed explanation. The reason for this decision is very understandable. I will adapt our repository implementation accordingly. I didn't realize that long skip distances meant so much effort for the database engine.

PR #319, opened 2018-04-05
CSHARP-2071: Using Linq Aggregation to project the root document

mbinic:
Hello,

Here's a first, naive, change to translate `{document}` to `$$ROOT` for `$push` in a `$group` aggregation stage.
I say "naive" because I'm not sure it doesn't break existing functionality, being new to this part of the code base. I have run all tests in `AggregateGroupTranslatorTests` and they are green.

This is by no means a complete solution. I'd be happy to add tests and extend any dependent functionality if you think this is a the step in the right direction.

Cheers,
Marko

CSHARP-2071

### DIFF
--- a/src/MongoDB.Driver/Linq/Translators/AggregateLanguageTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Translators/AggregateLanguageTranslator.cs
@@ -117,6 +117,8 @@ namespace MongoDB.Driver.Linq.Translators
                                 return TranslateArrayIndex((ArrayIndexExpression)node);
                             case ExtensionExpressionType.Concat:
                                 return TranslateConcat((ConcatExpression)node);
+                            case ExtensionExpressionType.Document:
+                                return TranslateDocument((DocumentExpression)node);
                             case ExtensionExpressionType.Except:
                                 return TranslateExcept((ExceptExpression)node);
                             case ExtensionExpressionType.FieldAsDocument:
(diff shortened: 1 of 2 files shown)


shiya-kohn:
This solution is faulty, but the issue has been open for years and no one is paying attention, the community's lack of care to basic serialization and translation issues is appalling. It makes it hard to use the type system without falling back to `BsonDocument` and json all over. Why is no one paying attention to issues like this?

mbinic:
Hi @shiya-kohn,

I understand your frustration, the impact of the community here does seem to be significantly less than advertised. My team has seen it on other (larger) PRs from 2017, and that's why I decided to go for a small one here. Essentially testing the water before investing any more effort.

That said, there's not much we can do about it, and the best advice I can offer (if you're working on an enterprise product) would be to "bury" the usages of types coming from this library under an abstraction, i.e. use the repository and specification patterns instead of using it directly.

Good luck!

AlseinX:
It is not a feature but a bug. It was rediculous to keep this bug for many years. I don't know why they still don't care about it.

slimjack:
This feature would allow wider use of Automapper instead of mapping hardcoded in a query.

cypressious:
Seriously hoping this PR gets some traction soon.

jyemin:
Hi everyone, we've scheduled time for someone to look into this.  I apologize for letting it sit so long unattended.

DmitryLukyanov:
Thanks @mbinic for your PR! 
Unfortunately, the suggested changes lead to generating unsupported queries in some of the cases. 
Since this logic is already covered in a new LINQ (v3) implementation which currently is `in-progress`, we reject this PR. We will check the root cause of this PR when the new LINQ implementation will be ready.

PR #320, opened 2018-04-17
CSHARP-2162: GridFS MD5 digest must be optional

vincentkam:
Add disableMD5 flag and associated tests while deprecating the checksum field


vincentkam:
Withdrawn.

PR #323, opened 2018-04-28
Fix for prefix-lookalike fields like Name and NameOfPet

igor-betin:
Or Connection and ConnectionType as it was in my case
In case projection looks like 
(item => new Result { Connection =  item.Connection, ConnectionType = item.ConnectionType} )
the method in question returns only one field - Connection
It results in FindProjectionTranslator.VisitField not finding ConnectionType  in fields list and it mistakenly sets _fullDocument = true, wich results in using ProjectingDeserializer that deserializes whole (very large) document and slows our system like x10 times.


igor-betin:
Jira is down - can't open issue

PR #324, opened 2018-05-03
CSHARP-2159: When a MongoNodeIsRecoveringException occurs the server type must be set to Unknown.

rstam:
The fix is trivial.

There is no automated test for this, but I manually verified that I got the same exception as the help ticket:

https://jira.mongodb.org/browse/HELP-6419

And then I verified that the exception no longer occurs after the fix was made.


rstam:
Withdrawing because the pull request was created in the wrong repo.

PR #328, opened 2018-06-14
CSHARP-1991: Add support for OP_COMPRESSED

AlexEndris:
This is the implementation to https://jira.mongodb.org/browse/CSHARP-1991. I tried to follow the existing code in style as closely as possible and took some inspiration from the mongo-java-driver.

Please provide feedback on the changes. I've successfully tested it both through the unit and integration tests against a local MongoDB instance as well as our proprietary application that uses an Atlas MongoDB cluster.

Unfortunately, .net does not provide an implementation of zlib with suitable headers. Only Deflate. So I had to include a 3rd party package to achieve the compression. Please provide feedback on this as well, as I'm not at all knowledgeable in licensing stuff.

Also please ignore double commits. They are the result from wrongly updating fork.

Thanks!

### DIFF
--- a/src/MongoDB.Driver.Core.Dotnet/MongoDB.Driver.Core.Dotnet.csproj
+++ b/src/MongoDB.Driver.Core.Dotnet/MongoDB.Driver.Core.Dotnet.csproj
@@ -1,5 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
-
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <VersionPrefix>2.7.0</VersionPrefix>
     <VersionSuffix>beta1</VersionSuffix>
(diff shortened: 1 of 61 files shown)


sachin5284:
Can you confirm two things;
* this driver support snappy compression
* this driver support compression while Query/ Reading from Mongo

AlexEndris:
@sachin5284 
1. No, I've not been able to find a nuget package of snappy compression, that worked well and also supported net-standard. Plus, quite frankly, we (the company I work for) didn't need that and I had a deadline.
2. Yes, absolutely, it's actually both ways. Messages to and from a mongo server will be compressed if enabled. Also, we (the company I work for) have been using this version of Mongo in production code for 4 months now. I've yet to see one issue.

Feel free to expand on this, maybe adding snappy. Unfortunately, I won't be able to develop this any further. I'm also waiting to have this either approved or a different implementation is pushed into master. But seeing that in the corresponding JIRA ticket the version and date gets pushed back constantly, I highly doubt this will be happening this year. That's why we decided to help develop at least the zlib compression part.

sachin5284:
@AlexEndris 
Sure, I will try to dig deep into code and maybe can add snappy compression support in it. 
I have just gone through few test cases and found NotSupportedException in CompressedMessageBinaryEncoder.cs file for ReadMessage() method, which leads me to ask about reading/Query part but it seems like I have to go through code first.
 Thanks for the quick reply.

AlexEndris:
Oh, don't worry about that part. This is a little quirk.

Basically, the "ReadMessage()" method is only used when actually reading the contents of a message that comes from the network stream. But due to reasons I can't remember, this isn't possible with the compressed message. The compressed message is only created at the out skirts of the network stuff.. Was a bit complicated. Feel also free to improve on that!

If you've got any questions just ask. I'd be happy to help you along a little.

AlexEndris:
Mistakenly closed

sherry-ummen:
Congrats! Happy that this is merged. Do you know which version will have this fix?

AlexEndris:
> Congrats! Happy that this is merged. Do you know which version will have this fix?

Take this with a grain of salt, as I'm not at all in the loop of release cycles. But it could be part of version 2.9. Timewise, I have no idea.

vincentkam:
The fix is currently slated for [2.9.0](https://jira.mongodb.org/browse/CSHARP-1991)! :)

PR #333, opened 2018-07-29
CSHARP-2345: Include get only properties on object serialization via Conventions

lopes-felipe:


IanWhalen:
Closing as per the discussion in https://jira.mongodb.org/browse/CSHARP-2345.

PR #334, opened 2018-08-05
CSHARP-2349: Support empty field names, since MongoDB supports them

admilazz:
This change fixes the problem described in https://jira.mongodb.org/browse/CSHARP-2349

The problem is that MongoDB supports empty field names but the driver doesn't.


sanych-sun:
BSON supports empty field names, but not null. That said, it is not recommended. We don't foresee usage accordingly of this and hence closing this ticket since its not a priority.

PR #335, opened 2018-08-05
CSHARP-2264: Add support for min() and max() to find calls

admilazz:
This change fixes the problem identified in https://jira.mongodb.org/browse/CSHARP-2264

It adds MinKey and MaxKey properties to the FindOptions and adds Min and Max to the FindFluent interface, allowing queries like db.c.find().min(x).max(y) to be executed using the driver.

I'm not sure if they should also be added to MapReduceOptions, etc.


jyemin:
As CSHARP-2264 has been closed, as well as CSHARP-1896, I'm going to close out this PR.

PR #336, opened 2018-08-07
relative link

xeptore:
hi. in Projection section, below the Projection Definition Builder there's a link to _tests_ github page. i cannot understand the md used syntax but in documentation page, it works as a relative link instead of an absolute link and i guess this change will work.
it worth mentioning that all other similar links have the same problem, at least in this page.

thanks.


jyemin:
The link at https://mongodb.github.io/mongo-csharp-driver/2.11/reference/driver/definitions/#projection-definition-builder works for me.  If you still see issues, we can re-open this, but for now I'm going to close this out.

PR #340, opened 2018-09-06
CSHARP-2379 Make evergreen compile use diagnostic output for cake compile

markbenvenuto:
In order to debug the issue in [CSHARP-2371](https://jira.mongodb.org/browse/CSHARP-2379), I needed to increase the verbosity of Cake to Diagnostic. This should be the default since it will aid in debugging similar problems in the future, and does not have a significant impact on the verbosity of the build.


vincentkam:
Thanks Mark! Rebased, merged, and pushed.

PR #342, opened 2018-09-13
CSHARP-2383: Provide overloads for WriteBytes/ReadBytes methods that …

XiCode:
…allow buffer reuse.

I've created overloads for reading and writing buffers that are larger than the data they hold. With this new api one can limit the size of the used buffer with a parameter. Now it is possible the reuse one large enough buffer and reduce fragmentation of memory.

The overloads in JsonWriter will clone buffers that are too large and reduce them to the given size.


adelinowona:
Hi @XiCode, thanks for your contribution. We have plans on updating our codebase to using Span in the near future and that will involve/void such work as this. Thus I'll be closing this PR.

PR #344, opened 2018-09-21
Change ClusterTime to return BsonTimestamp instead of BsonDocument

jsmurphy:


jyemin:
I don't think this is correct.  Cluster time really is represented as a BsonDocument:

```
$clusterTime" : {
		"clusterTime" : Timestamp(0, 0),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	}
```

Though it contains a timestamp, it also include a signature field.  Together it's a BsonDocument.

I'm going to close this out, but let me know if you have any further questions.

PR #347, opened 2018-10-05
CSHARP-2450: Performance: Reduced lock contention in BsonSerializer.LookupActualType

dnickless:
BsonSerializer.LookupActualType is on the hot path during deserialization and has proven to be a nasty bottleneck in my setup:

image

I managed to remove the need for the ReaderWriterLockSlim by switching to a (unfortunately not generic) Hashtable type instead of a HashSet<T>. Hashtable is safe for concurrent reads (as per https://docs.microsoft.com/en-us/dotnet/api/system.collections.hashtable?redirectedfrom=MSDN&view=netframework-4.7.2#thread-safety) and the amount of casting introduced is very small.

The HashSet<T> instances that are tracked inside the Hashtable get replaced instead of updated upon a write which allows consumers to safely iterate over them.

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializer.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


IanWhalen:
Hi @dnickless, thanks for opening this pull request, and my apologies for the delay in response here.  The MongoDB .NET team relies primarily on [our JIRA project](http://jira.mongodb.org/browse/CSHARP) to track and prioritize our work.

In order to continue the conversation, can you please open a ticket at the above link so we can discuss your proposal there?

dnickless:
Hi @IanWhalen, kindly see the updated JIRA ticket. This change appears to indicate some massive (!) improvement.

IanWhalen:
Cool, thanks!  We're still working through a lot of effort to get the driver ready for release concurrent to 4.2 and all of the new features that will be available there, but this is still on our radar when we get to handling PRs and other subsequent work.

dnickless:
Kindly have a look at my second attempt which isn't really better than this one here in my opinion but certainly less drastic and still effective: #433

PR #348, opened 2018-10-11
CSHARP-2414: Enum Deserializer Issue with BsonType.String

sunyutj:
Use Enum.TryParse<TEnum> ( ) instead of Enum.Parse ( )


rstam:
Thank you for submitting a pull request. However, I am closing the pull request without merging it. See comments in:

https://jira.mongodb.org/browse/CSHARP-2414

PR #350, opened 2018-10-22
CSHARP-2405 Update ImmutableTypeClassMapConvention.cs

rbugginsvia:
Upate to fix problem with static members in F# not allowing serialisation
https://jira.mongodb.org/browse/CSHARP-2405

The above limits the GetProperties to just instance public properties - ignoring static properties.


DmitryLukyanov:
Already merged into master.

PR #360, opened 2018-11-27
CSHARP-2442: Increase StringBuilder capacity

Achilles1515:
This StringBuilder object is being created because the string named "value" needs escaping. It is unknown at this point exactly how many characters need escaping, but AT MINIMUM there is one. This means the new escaped string length will be larger than original string length by at least one character, and the StringBuilder constructor capacity parameter should reflect this logic. Without this change, the StringBuilder object is GUARANTEED to allocate and chain together a new StringBuilder object under the hood as it increases it's capacity to handle the larger escaped string.

### DIFF
--- a/src/MongoDB.Bson/IO/JsonWriter.cs
+++ b/src/MongoDB.Bson/IO/JsonWriter.cs
@@ -710,7 +710,7 @@ namespace MongoDB.Bson.IO
                 return value;
             }
 
-            var sb = new StringBuilder(value.Length);
+            var sb = new StringBuilder(value.Length + 1);
 
             foreach (char c in value)
             {


Achilles1515:
> Hi @Achilles1515 as per our [contributing guide](https://github.com/mongodb/mongo-csharp-driver/blob/master/CONTRIBUTING.md), please open an issue at jira.mongodb.org in the CSHARP project and link it to the pull request.
> 
> Thanks,
> Jeff

Done. Sorry about that.

dnickless (changes requested):
How about something like this even:

            int replacements = value.Count(c => NeedsEscaping(c));
            if (replacements == 0)
            {
                return value;
            }

            var sb = new StringBuilder(value.Length + replacements);

Also, NeedsEscaping should be made static as we currently have a lambda closure over "this" here which allocates no matter if we end up escaping things or not.
image
image

rstam:
This PR was included as part of PR https://github.com/mongodb/mongo-csharp-driver/pull/1290

Thank you for submitting it.

PR #370, opened 2019-04-09
CSHARP-2579: Allow overriding cached serializers in registry.

Chips100:
This change makes the BsonSerializerRegistry accept registrations of serializers for types that have been cached before. Manual (explicit) registrations will now override the previously created default serializer.

About the implementation: I did not use a HashSet for storing which types have been registered explicitly because of concurrency issues as described here:
https://stackoverflow.com/questions/18922985/concurrent-hashsett-in-net-framework/18923091

This is my first pull request, so I would not mind if my changes are reviewed especially closely.


Chips100:
Per discussion in the corresponding ticket, this was a flawed approach.

PR #373, opened 2019-06-21
Convention based String Id as ObjectId handling

nurih:
There is no current convention allowing for a string id member on a class represented as an ObjectId with an ObjectId generator. You can achive this with attributes, but this requires the class to reference external libraries which is not always desirable. The issue is further described here: https://jira.mongodb.org/browse/CSHARP-2644 

This PR introduces a convention-only based solution. Using this convention there is no need for any attribute decoration or explicit BsonClassMap configuration.


jyemin:
The commit from this PR has been merged on the command line some time ago, so closing this out.

PR #377, opened 2019-08-13
CSHARP-2592: Adding netstandard2.0 support to libraries

jbogard:
This also removes the `System.Reflection.Emit.Lightweight` dependency in favor of expression tree compilation.

I left the test projects alone since they target `netcoreapp2.1`. In order to test against `netstandard1.5` you'd have to add `netcoreapp1.x` target....which is deprecated.

### DIFF
--- a/src/MongoDB.Bson/Exceptions/BsonException.cs
+++ b/src/MongoDB.Bson/Exceptions/BsonException.cs
@@ -14,7 +14,7 @@
 */
 
 using System;
-#if NET452
+#if NET452 || NETSTANDARD2_0
 using System.Runtime.Serialization;
 #endif
 
@@ -23,7 +23,7 @@ namespace MongoDB.Bson
     /// <summary>
     /// Represents a BSON exception.
     /// </summary>
-#if NET452
+#if NET452 || NETSTANDARD2_0
     [Serializable]
 #endif
     public class BsonException : Exception
@@ -66,7 +66,7 @@ namespace MongoDB.Bson
         {
         }
 
-#if NET452
+#if NET452 || NETSTANDARD2_0
         /// <summary>
         /// Initializes a new instance of the BsonException class (this overload used by deserialization).
         /// </summary>
--- a/src/MongoDB.Bson/Exceptions/BsonInternalException.cs
+++ b/src/MongoDB.Bson/Exceptions/BsonInternalException.cs
@@ -14,7 +14,7 @@
 */
 
 using System;
-#if NET452
+#if NET452 || NETSTANDARD2_0
 using System.Runtime.Serialization;
 #endif
 
@@ -23,7 +23,7 @@ namespace MongoDB.Bson
     /// <summary>
     /// Represents a BSON internal exception (almost surely the result of a bug).
     /// </summary>
-#if NET452
+#if NET452 || NETSTANDARD2_0
     [Serializable]
 #endif
     public class BsonInternalException : BsonException
@@ -56,7 +56,7 @@ namespace MongoDB.Bson
         {
         }
 
-#if NET452
+#if NET452 || NETSTANDARD2_0
         /// <summary>
         /// Initializes a new instance of the BsonInternalException class (this overload used by deserialization).
         /// </summary>
--- a/src/MongoDB.Bson/Exceptions/BsonSerializationException.cs
+++ b/src/MongoDB.Bson/Exceptions/BsonSerializationException.cs
@@ -14,7 +14,7 @@
 */
 
 using System;
-#if NET452
+#if NET452 || NETSTANDARD2_0
 using System.Runtime.Serialization;
 #endif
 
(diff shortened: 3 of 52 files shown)


Digi59404:
This looks good! Thanks James!

jbogard:
Thoughts? Comments? Concerns?

vincentkam:
Thanks for your PR! The .NET Driver team is considering adding support for .NET Standard 2.0 soon

horeaper:
> Thanks for your PR! The .NET Driver team is considering adding support for .NET Standard 2.0 soon

Is there an time line for that? Currently when publishing a .net core 3.0 app with MongoDB driver, error NU1605 will block the process. It can be ignored but that requires modifing every project file that uses the driver, which is quite tedious for large solutions.

adamhathcock:
Any chance to move on to the better netstandard?

jbogard:
@adamhathcock I'd fix the merge conflicts but I don't know if this PR is still under review for merge 🤷‍♂

AlseinX:
Looking forward to it. But as it shows, a conflict in `src/MongoDB.Driver.Core/Core/Compression/SnappyCompressor.cs` still exists.
When will it be solved and merged?

jbogard:
@AlseinX same comment, don't know if this PR is under review. I didn't get any feedback so I don't want to go through the work of fixing merge conflicts if there's no point

AlseinX:
This PR is really helpful and the removal of  `Emit`ting brings much more compatibility to some runtime implementations.
It sounds disappointing that the maintainers of an open source project ignores a helpful PR for months.
Commonly a PR is reviewed in several days when it is ready for review, either merged, rejected, requested changes, or given a reason why it is being on hold, as long as the project is still active, otherwise the readme file would contains "This repo is developed by the official team only and contribution is not welcomed".

rstam:
I'm sorry we haven't gotten to this yet. We are definitely planning to target .NET Standard 2.0 as soon as possible. The main issues that have stopped us from doing it sooner are:

1. Uncertainty regarding whether it's OK to drop support for .NET Standard 1.5 and just replace it with .NET Standard 2.0
2. If we keep support for .NET Standard 1.5 how do we test both the netstandard15 and netstandard20 artifacts

We are researching whether any of our users would be impacted by removing support for .NET Standard 1.5.

Note that once we target .NET Standard 2.0 we also need to review the #if's in the entire code base to see what features can now be re-enabled once we target .NET Standard 2.0 (and are no longer constrained by the more limited APIs available in .NET Standard 1.5). We plan to do that as a separate effort.

Can someone elaborate on why removing use of Emit is helpful? Where is Emit not supported?

Thanks!

Digi59404:
Hey @rstam  I can confirm that Emit is not supported on Xamarin iOS Devices or Devices that do not allow dynamic generation of code.  Microsoft has released a hotfix for this by enabling an interpreter to run on iOS. But this affects performance and causes other issues. 

https://github.com/mongodb/mongo-csharp-driver/pull/377
https://devblogs.microsoft.com/xamarin/introducing-xamarin-ios-interpreter/

rstam:
@Digi59404 thanks for that information. That's helpful.

AlseinX:
As for the issues:

1. Actually the .NET Implementations that do not support netstandard2.0 but support netstandard1.5 are out of support and deprecated, and the projects that are built against netstandard1.5 only (ASP.NET Core 1.x/2.x targets netstandard rather than netcoreapp) are also compatible to netstandard2.0 by design. The ancient projects that never intend to upgrade the target platform are basically inactive, no longer maintained and more unlikely to update the mongodb driver.

2. Even if it is really necessary to keep support for netstandard1.5, and test for both targets separately are needed. Assuming that the test would pass on netstandard1.5 and fail on netstandard2.0 (by design, this is impossible unless there is conditional compilation to behave specifically, or mongodb driver is designed to work with a bug that only occurs on netstandard1.5 implementations), the projects that run on a netstandard2.0 implementation would not work as well even if the mongodb packages pass the test on a .netstandard1.5 implementation. Currently the mongodb csharp driver is tested on netstandard1.5 implementations but actually used on netstandard2.0+ implementations. If the assumption is a truth, it is broken already.

jbogard:
I've fixed merge conflicts here.

jbogard:
It looks like just adding `netstandard2.0` support breaks some tests around `ISupportInitialize`:

image

Even if I go back to using the `Reflection.Emit` approach, these tests break. This will need some investigation into why `netstandard2.0` breaks but `netstandard1.5` works

jbogard:
@rstam @vincentkam I've fixed the test failures by adding the `NETSTANDARD2_0` in the places where it conditionally compiles `NET452`. Build is green now.

jbogard:
I rebased and squashed this PR because it had a bunch of useless old commits

adamhathcock:
Somehow, there’s already conflicts.  Sorry @jbogard 

I still think out of support versions of net standard should be dropped.

jbogard:
Merged from upstream, which incorporated CSHARP-2394

jbogard:
@adamhathcock not anymore ;)

DmitryLukyanov:
Thanks @jbogard for your participation, the changes are already in master. They will be available to consume in the next release

PR #378, opened 2019-08-15
CSHARP-2692: Fixed a performance issue during deserialization

dnickless:
Less byte buffer allocations during deserialization as per https://jira.mongodb.org/browse/CSHARP-2692.


BorisDog:
Thanks @dnickless, this issue was addressed [here](https://github.com/mongodb/mongo-csharp-driver/commit/f0bece570fdd3c8ede73981adaa89a225caa978c)

PR #385, opened 2019-09-27
CSHARP-2758 : Add Span<T> support to ObjectId

jasonkuo41:
This adds the System.Memory package to the driver.

I hope that we could use this to further boost the performance of the driver with the new concept of Span<T> and Memory<T> introduced in .Net Core 2.1.

I'll add Span support to ObjectId for now, note that it may be possible to also add such support (and many other concepts introduced in System.Memory) towards BsonStream making it more performant in the future.

### DIFF
--- a/src/MongoDB.Bson/BsonUtils.cs
+++ b/src/MongoDB.Bson/BsonUtils.cs
@@ -52,6 +52,16 @@ namespace MongoDB.Bson
         }
 
         /// <summary>
+        /// Get's the hex string's equivalent byte array size
+        /// </summary>
+        /// <param name="s"></param>
+        /// <returns></returns>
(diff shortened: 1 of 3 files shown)


jasonkuo41:
As per: https://jira.mongodb.org/browse/CSHARP-2758

BorisDog:
Thanks for submitting this ticket  @jasonkuo41 

We are going to introduce support for Span<T> in our serialization in [CSHARP-5125](https://jira.mongodb.org/browse/CSHARP-5125).
It's scheduled after 3.0 release where we plan to introduce some architectural changes.
Closing this ticket as won't as do, as this will be part of [CSHARP-5125](https://jira.mongodb.org/browse/CSHARP-5125).

PR #388, opened 2019-10-25
CSHARP-1525 - Add DiscardBufferedData method to JsonReader

alistairjevans:
Added a method to JsonReader that allows a consumer to clear the memory use of the reader when reading very large JSON documents.

Came across this problem myself, found it in JIRA and thought I'd contribute!

### DIFF
--- a/src/MongoDB.Bson/IO/JsonReader.cs
+++ b/src/MongoDB.Bson/IO/JsonReader.cs
@@ -111,7 +111,7 @@ namespace MongoDB.Bson.IO
         {
         }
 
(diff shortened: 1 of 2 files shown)


JamesKovacs:
Based on our investigation of this issue, the `JsonBuffer` will grow to fit the largest JSON document, but the buffer contents will be reset after each top-level document is read. Please re-open [CSHARP-1525](https://jira.mongodb.org/browse/CSHARP-1525) with a repro of the issue demonstrating the issue if this is still a problem with the latest version.

PR #395, opened 2020-03-23
ValueType properties should be deserializable from null if supported by their serializer

Corniel:
Currently, if a struct/ValueType has its own converter registered it can be deserialized from null. However if the same struct/ValueType is property (or field) of a class, deserialization fails. Given that there is no test failing if I change that behaviour, I assume that it is a bug. Nevertheless, given example below, I think it makes sense to support it anyway.

``` C#
public readonly struct HouseNumber
{
    private readonly int val;
    public HouseNumber(int val) => this.val = val;
    public override string ToString() => val < 1 ? "" : val.ToString();
    public object ToJson() => val < 1 ? null : (object)val;
}

internal class Address
{
    public HouseNumber Number { get; set; }
}

public class HouseNumberSerializer : SerializerBase<HouseNumber>
{
    public override HouseNumber Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
    {
        var bsonType = context.Reader.GetCurrentBsonType();

        return bsonType == BsonType.Int32
            ? new HouseNumber(context.Reader.ReadInt32())
            : default;
    }

    public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, HouseNumber value)
    {
        var val = value.ToJson();
        if(val is int num)
        {
            context.Writer.WriteInt32(num);
        }
        else
        {
            context.Writer.WriteNull();
        }
    }
}
```

### DIFF
--- a/src/MongoDB.Bson/IO/JsonReader.cs
+++ b/src/MongoDB.Bson/IO/JsonReader.cs
@@ -199,7 +199,7 @@ namespace MongoDB.Bson.IO
                 // in JSON the top level value can be of any type so fall through
                 State = BsonReaderState.Type;
             }
-            if (State != BsonReaderState.Type)
+            if (State != BsonReaderState.Type && State != BsonReaderState.Value)
             {
                 ThrowInvalidState("ReadBsonType", BsonReaderState.Type);
             }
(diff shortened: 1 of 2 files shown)


vincentkam:
Created https://jira.mongodb.org/browse/CSHARP-3023 to track this 😸

Corniel:
@vincentkam What is the SLA on this kind of bug-reports/features requests?!

rstam:
See the explanation in the JIRA ticket for why no driver changes are needed for this to work. The JIRA ticket shows two small changes needed for the new tests in this pull request to pass.

https://jira.mongodb.org/browse/CSHARP-3023

Corniel:
@rstam The implementation is written to prove my point. In practice I registered a separate converter for that type. And that was is never called, and it should. It would be extremely inconvenient to have to write a specific converter per object, because the default implementation does not work with value types. So please re-open.

Corniel:
@rstam Thanx for this suggestion. Do you think a different error, could help explaining what goes wrong?!

PR #398, opened 2020-04-21
CSHARP-2596: AsQueryable with session (usable during transaction).

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/5e9e5706850e6113494ce0f7

12 failed tasks
7 timeouts
2 drop collection in admin database
2 reIndex not supported
1 change stream _id

src/MongoDB.Driver/IMongoCollectionExtensions.cs
```diff
@@ -69,11 +69,24 @@ public static IAggregateFluent<TDocument> Aggregate<TDocument>(this IMongoCollec
         /// <param name="aggregateOptions">The aggregate options</param>
         /// <returns>A queryable source of documents.</returns>
         public static IMongoQueryable<TDocument> AsQueryable<TDocument>(this IMongoCollection<TDocument> collection, AggregateOptions aggregateOptions = null)
+        {
+            return AsQueryable(collection, session: null, aggregateOptions);
```
rstam, line 73: As I write this comment I realize that we don't usually allow session to be null in the public API.

rstam:
Wrong repo.

PR #405, opened 2020-08-13
Removed Unnecessary Task.Yeld CSHARP-3183

alkampfergit:
This PR is related to https://jira.mongodb.org/browse/CSHARP-3183?filter=-2

Actually when I run tests in project https://github.com/ProximoSrl/NStore XUnit runner creates an high number of threads and the connection deadlock.

I've found debugging the code that a Thread enter in MonitorServerAsync, wait for Task.Yield() but never exit from the wait, because an high number of thread are waiting in line 368 of Cluster.cs

image

That particular wait is waiting for three tasks, a cancellation a timeout and a **task that is completed when the cluster changes description**. When too many thread at the same time enter that wait, probably all thread pool threads are in waiting, and for some reason MonitorServerAsync never exit awaiting from the Task.Yeld and everything is freezed.

image

Removing the await Task.Yield() solves the problem. That yield was recently added in commit e0568c0f894e2eaafcb894287f33c0b14aa7624b

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -189,7 +189,7 @@ namespace MongoDB.Driver.Core.Servers
 
         private async Task MonitorServerAsync()
         {
-            await Task.Yield(); // return control immediately
+            //await Task.Yield(); // return control immediately
 
             var metronome = new Metronome(_serverMonitorSettings.HeartbeatInterval);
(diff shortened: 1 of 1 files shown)


DmitryLukyanov:
Hey @alkampfergit , thanks for your PR, We've already noticed this issue. The patch release will be created asap

alkampfergit:
Perfect, thanks.

DmitryLukyanov:
Thanks @alkampfergit , the solution for this problem has been released in 2.11.1. Let me know if you have any questions.

alkampfergit:
Perfetct, thanks.

PR #407, opened 2020-08-19
CSHARP-1467 CSHARP-1771 Add support for IQueryable features used by OData (namely IIF expression)

Pzixel:
I'm trying to connect OData provider to MongoDB using this driver. And I faced several problems with it. For example, OData nullability handling forces it to generate queries like `foo == null ? null : foo.bar` that cannot be handled by current driver. The other problem is `Expression.Parameter` allows to make parameter anonymous, while mongo driver expects it to always have a name, thus it fails to generate a query that otherwise is fine to execute. There are also several smaller fixes.

Although I did it to make my OData query work they are general improvements that allow to use IQueryable interface better.

I'm forced into using fork until this gets somehow adopted in upstream so I hope this will eventually go into master.

Related to https://jira.mongodb.org/browse/CSHARP-1771 , https://jira.mongodb.org/browse/CSHARP-1467

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonMemberMap.cs
+++ b/src/MongoDB.Bson/Serialization/BsonMemberMap.cs
@@ -470,7 +470,7 @@ namespace MongoDB.Bson.Serialization
             {
                 throw new ArgumentNullException("serializer");
             }
-            if (serializer.ValueType != _memberType)
+            if (serializer.ValueType != _memberType && serializer.ValueType.DeclaringType != _memberType)
             {
                 var message = string.Format("Value type of serializer is {0} and does not match member type {1}.", serializer.ValueType.FullName, _memberType.FullName);
                 throw new ArgumentException(message, "serializer");
--- a/src/MongoDB.Driver/Linq/Expressions/SelectExpression.cs
+++ b/src/MongoDB.Driver/Linq/Expressions/SelectExpression.cs
@@ -28,7 +28,7 @@ namespace MongoDB.Driver.Linq.Expressions
         public SelectExpression(Expression source, string itemName, Expression selector)
         {
             _source = Ensure.IsNotNull(source, nameof(source));
-            _itemName = Ensure.IsNotNull(itemName, nameof(itemName));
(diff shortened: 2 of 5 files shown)


src/MongoDB.Bson/Serialization/BsonMemberMap.cs
```diff
@@ -470,7 +470,7 @@ public BsonMemberMap SetSerializer(IBsonSerializer serializer)
             {
                 throw new ArgumentNullException("serializer");
             }
-            if (serializer.ValueType != _memberType)
+            if (serializer.ValueType != _memberType && serializer.ValueType.DeclaringType != _memberType)
```
Pzixel, line 473: This check was too harsh. For example, in OData `serializer.ValueType` is `Microsoft.AspNet.OData.Query.Expressions.PropertyContainer+AutoSelectedNamedProperty`1[System.Nullable`1[System.Guid]]`, while `memberType` is `Microsoft.AspNet.OData.Query.Expressions.PropertyContainer`.

They are 100% compatible so lifting this check doesn't make bad things happening in runtime but they allow to write actual queries using OData.

src/MongoDB.Driver/Linq/Expressions/SelectExpression.cs
```diff
@@ -28,7 +28,7 @@ internal sealed class SelectExpression : ExtensionExpression, ISourcedExpression
         public SelectExpression(Expression source, string itemName, Expression selector)
         {
             _source = Ensure.IsNotNull(source, nameof(source));
-            _itemName = Ensure.IsNotNull(itemName, nameof(itemName));
+            _itemName = itemName;
```
Pzixel, line 31: Is not required, `itemName` can legitimately be null.

Pzixel:
Ping?

Pzixel:
Anyone?

rstam:
Thank you for submitting this pull request. We are working on a new implementation of our LINQ translator and will make sure IIF works properly in the new implementation.

PR #410, opened 2020-09-03
CSHARP-3196: fix for 'invalid nonce' error when making concurrent con…

blushingpenguin:
These feels like a hack on a hack but it does fix the problem.


DmitryLukyanov:
Thanks, @blushingpenguin. We've implemented a bit different solution which will be available in the next 2.11.2 patch release.

PR #413, opened 2020-10-02
CSHARP-3218: Add [Preserve] to MongoDB.Bson for AOT compiled targets

nirinchev:
On AOT compiled targets (such as Xamarin.iOS/Xamarin.Mac), the linker will remove all members that are not statically linked by the application. This obviously becomes an issue when members are invoked via reflection or objects are dynamically instantiated. The MongoDB.Bson assembly deals with de/serialization, so it has quite a lot of that. Ultimately, it would be great if we selectively applied `[Preserve]` only to members that absolutely have to be preserved and allowed the linker to strip parts of the assembly that are not needed. Since that is a massive undertaking, a stopgap measure is to preserve the entire assembly.

More info here: https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/linker.

### DIFF
--- a/src/MongoDB.Bson/Properties/AssemblyInfo.cs
+++ b/src/MongoDB.Bson/Properties/AssemblyInfo.cs
@@ -16,8 +16,14 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using MongoDB.Bson.Internals;
 
 [assembly: CLSCompliant(true)]
 [assembly: ComVisible(false)]
+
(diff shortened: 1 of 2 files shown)


DmitryLukyanov:
This change has been merged and will be available in the next patch release

tkrafael:
Does this same approach will work with native aot? I've been testing native aot support and it crashes right before collection instantiantion due to serializer registry using reflection based construction of those objects.
Tried to not use automap and it crashes in another point.
Dunno that's correct way to make driver compatible with native aot.
Is mongodb team looking to support native AOT at some time?

PR #415, opened 2020-10-21
Add BsonArray.Add(string, BsonValue) overload

anuith:
For simpler BsonArray initialization by taking advantage of C# collection initializer.
So we can simplify BsonArray initializing from this :
```
new BsonArray
{
    new BsonDocument("a", 1),
    new BsonDocument("b", 2),
};
```
To this :
```
new BsonArray
{
    { "a", 1 },
    { "b", 2 },
};
```


anuith:
There's one thing I'm not sure I should do is that I made a reference to `BsonDocument`. I'm not sure that it will cause any problem or does it violates any rules/guideline or not.

Please clarify if I do.

PR #418, opened 2020-11-16
CSHARP-3239: Unescape AWS session token when parsing from connection string.

JamesKovacs:
`authMechanismProperties` in the connection string have to be URL escaped, but we did not unescape them when parsing the connection string. Notably this led to `AWS_SESSION_TOKEN` containing `%2F` rather than `/` as expected.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
@@ -593,9 +593,7 @@ namespace MongoDB.Driver.Core.Configuration
             }
 
             var txtRecords = _dnsResolver.ResolveTxtRecords(host, cancellationToken);
-            var options = GetOptionsFromTxtRecords(txtRecords);
-
-            var resolvedOptions = GetResolvedOptions(options);
+            var resolvedOptions = GetResolvedOptionsFromTxtRecords(txtRecords);
 
             return BuildResolvedConnectionString(resolvedScheme, hosts, resolvedOptions);
         }
@@ -643,9 +641,7 @@ namespace MongoDB.Driver.Core.Configuration
             }
 
             var txtRecords = await _dnsResolver.ResolveTxtRecordsAsync(host, cancellationToken).ConfigureAwait(false);
-            var options = GetOptionsFromTxtRecords(txtRecords);
-
-            var resolvedOptions = GetResolvedOptions(options);
+            var resolvedOptions = GetResolvedOptionsFromTxtRecords(txtRecords);
 
             return BuildResolvedConnectionString(resolvedScheme, hosts, resolvedOptions);
         }
@@ -692,9 +688,9 @@ namespace MongoDB.Driver.Core.Configuration
             mergedOptions.AddRange(
                 resolvedOptions
                 .AllKeys
-                .SelectMany(x => resolvedOptions
-                    .GetValues(x)
-                    .Select(y => $"{x}={Uri.EscapeDataString(y)}")));
+                .SelectMany(key => resolvedOptions
+                    .GetValues(key)
+                    .Select(value => $"{key}={Uri.EscapeDataString(value)}")));
 
             if (mergedOptions.Count > 0)
             {
@@ -769,8 +765,10 @@ namespace MongoDB.Driver.Core.Configuration
             foreach (Capture option in match.Groups["option"].Captures)
             {
                 var parts = option.Value.Split('=');
-                _allOptions.Add(parts[0], parts[1]);
-                ParseOption(parts[0].Trim(), Uri.UnescapeDataString(parts[1].Trim()));
+                var name = parts[0].Trim();
+                var value = Uri.UnescapeDataString(parts[1].Trim());
+                _allOptions.Add(name, value);
+                ParseOption(name, value);
             }
         }
 
@@ -1294,23 +1292,20 @@ namespace MongoDB.Driver.Core.Configuration
             return hosts;
         }
 
-        private List<string> GetOptionsFromTxtRecords(List<TxtRecord> txtRecords)
+        private NameValueCollection GetResolvedOptionsFromTxtRecords(IReadOnlyCollection<TxtRecord> txtRecords)
         {
             if (txtRecords.Count > 1)
             {
                 throw new MongoConfigurationException("Only 1 TXT record is allowed when using the SRV protocol.");
             }
 
-            return txtRecords.Select(tr => tr.Strings.Aggregate("", (acc, s) => acc + Uri.UnescapeDataString(s))).ToList();
-        }
+            var txtRecord = txtRecords.FirstOrDefault();
 
-        private NameValueCollection GetResolvedOptions(List<string> options)
-        {
             // Build a dummy connection string in order to parse the options
             var dummyConnectionString = "mongodb://localhost/";
-            if (options.Count > 0)
+            if (txtRecord != null)
             {
-                dummyConnectionString += "?" + string.Join("&", options);
+                dummyConnectionString += "?" + string.Join("", txtRecord.Strings);
             }
             var dnsConnectionString = new ConnectionString(dummyConnectionString);
             ValidateResolvedOptions(dnsConnectionString.AllOptionNames);
--- a/tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
@@ -21,6 +21,7 @@ using System.Net.Sockets;
 using System.Threading.Tasks;
 using FluentAssertions;
 using MongoDB.Bson;
+using MongoDB.Bson.TestHelpers.XunitExtensions;
 using MongoDB.Driver.Core.Clusters;
 using MongoDB.Driver.Core.Compression;
 using Xunit;
@@ -65,6 +66,9 @@ namespace MongoDB.Driver.Core.Configuration
         [InlineData("mongodb://test5.test.build.10gen.cc", "mongodb://test5.test.build.10gen.cc", true)]
         [InlineData("mongodb+srv://test5.test.build.10gen.cc", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=thisDB&tls=true", false)]
         [InlineData("mongodb+srv://test5.test.build.10gen.cc", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=thisDB&tls=true", true)]
+        [InlineData("mongodb+srv://test5.test.build.10gen.cc/?authSource=$external", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=%24external&tls=true", false)]
+        [InlineData("mongodb+srv://test5.test.build.10gen.cc/?authSource=%24external", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=%24external&tls=true", false)]
+        [InlineData("mongodb+srv://test5.test.build.10gen.cc/?authSource=$external&authMechanism=MONGODB-AWS", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=%24external&authMechanism=MONGODB-AWS&tls=true", false)]
         public void Resolve_should_return_expected_result(string connectionString, string expectedResult, bool async)
         {
             var subject = new ConnectionString(connectionString);
@@ -81,6 +85,27 @@ namespace MongoDB.Driver.Core.Configuration
 
             result.IsResolved.Should().BeTrue();
             result.ToString().Should().Be(expectedResult);
+        }
+
+        [Theory]
+        [ParameterAttributeData]
+        public void Resolve_against_mongodb_aws_session_token_should_return_the_expected_aws_session_token([Values(false, true)] bool escapeToken)
+
+    {
+            const string authMechanism = "MONGODB-AWS";
+            const string username = "AKIAIOSFODNN7EXAMPLE";
+            const string password = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY";
+            const string awsSessionToken = "AQoEXAMPLEH4aoAH0gNCAPyJxz4BlCFFxWNE1OPTgk5TthT+FvwqnKwRcOIfrRh3c/LTo6UDdyJwOOvEVPvLXCrrrUtdnniCEXAMPLE/IvU1dYUg2RVAJBanLiHb4IgRmpRV3zrkuWJOgQs8IZZaIv2BXIa2R4OlgkBN9bkUDNCJiBeb/AXlzBBko7b15fjrBs2+cTQtpZ3CYWFXG8C5zqx37wnOE49mRl/+OtkIKGO7fAE";
+
+            var preparedToken = escapeToken ? Uri.EscapeDataString(awsSessionToken) : awsSessionToken;
+            var uri = $"mongodb+srv://{username}:{Uri.EscapeDataString(password)}@test5.test.build.10gen.cc/test?authSource=$external&authMechanism={authMechanism}&authMechanismProperties=AWS_SESSION_TOKEN:{preparedToken}";
+            var connectionString = new ConnectionString(uri);
+
+            var result = connectionString.Resolve();
+
+            result.AuthMechanism.Should().Be(authMechanism);
+            result.Username.Should().Be(username);
+            result.AuthMechanismProperties["AWS_SESSION_TOKEN"].Should().Be(awsSessionToken);
         }
 
         [Theory]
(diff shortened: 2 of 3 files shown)


tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
```diff
@@ -704,6 +704,24 @@ public void TestFromUrlResolving(string connectionString, ConnectionStringScheme
             result.Scheme.Should().Be(expectedScheme);
         }
 
+        [Fact]
+        public void TestFromUrlWithMongoDBAWS_should_parse_credentials_correctly()
+        {
+            const string authMech = "MONGODB-AWS";
+            const string user = "AKIAIOSFODNN7EXAMPLE";
+            const string pwd = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY";
```
rstam, line 712: avoid abbreviations, consider `password` as the variable name

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
```diff
@@ -704,6 +704,24 @@ public void TestFromUrlResolving(string connectionString, ConnectionStringScheme
             result.Scheme.Should().Be(expectedScheme);
         }
 
+        [Fact]
+        public void TestFromUrlWithMongoDBAWS_should_parse_credentials_correctly()
+        {
+            const string authMech = "MONGODB-AWS";
```
rstam, line 710: avoid abbreviations (possible exception of `auth` due to extensive use already).

Consider `authMechanism` as the variable name.

rstam (changes requested):
Possible minor changes requested. Will LGTM as is if you object to the changes.

src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -834,7 +834,7 @@ public static MongoClientSettings FromUrl(MongoUrl url)
                     }
                     else
                     {
-                        credential = credential.WithMechanismProperty(property.Key, property.Value);
+                        credential = credential.WithMechanismProperty(property.Key, Uri.UnescapeDataString(property.Value ?? string.Empty));
```
rstam, line 837: We use `""` instead of `string.Empty`

I realize some people prefer `string.Empty` but there are lots of reasons not to use it. Most are a consequence of the difference between a read-only property (which `string.Empty` is) and a compile time constant (which `""` is).

rstam: Another statement I would make: `""` is always valid but `string.Empty` is sometimes not.

rstam (changes requested):
I missed the use of `string.Empty` in the first review.

src/MongoDB.Driver/MongoClientSettings.cs
JamesKovacs: No strong opinion on this. I used `string.Empty` because that was what I've used previously. Fixed.

src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -834,7 +834,7 @@ public static MongoClientSettings FromUrl(MongoUrl url)
                     }
                     else
                     {
-                        credential = credential.WithMechanismProperty(property.Key, property.Value);
+                        credential = credential.WithMechanismProperty(property.Key, Uri.UnescapeDataString(property.Value ?? ""));
```
rstam, line 837: Thinking more about this...

I think the call to `Uri.UnescapeDataString` should be in `ConnectionString.cs` when parsing the URI.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -686,15 +686,22 @@ private ConnectionString BuildResolvedConnectionString(ConnectionStringScheme re
...
             var mergedOptions = new List<string>();
             mergedOptions.AddRange(
                 resolvedOptions
                 .AllKeys
-                .SelectMany(x => resolvedOptions
-                    .GetValues(x)
-                    .Select(y => $"{x}={Uri.EscapeDataString(y)}")));
+                .SelectMany(key => resolvedOptions
+                    .GetValues(key)
+                    .Select(value => $"{key}={value}")));
```
rstam, line 704: It's OK to rename the variables, but we still need to call `Uri.EscapeDataString`

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -686,15 +686,22 @@ private ConnectionString BuildResolvedConnectionString(ConnectionStringScheme re
                 }
             }
 
+            foreach (var key in resolvedOptions.AllKeys)
+            {
+                resolvedOptions[key] = Uri.EscapeDataString(resolvedOptions[key]);
+            }
+
+            // _allOptions are the escaped options from the original connection string.
+            // Add them after escaping the resolvedOptions from the TXT record to avoid double-escaping.
```
rstam, line 695: These lines should not be added.

rstam (changes requested):
These are the principles we should follow for consistent URI escaping/unescaping:

* values are only escaped when they are on a connection string
* values should *not* be escaped when they are in memory
* values need to be re-escaped when building a new connection string
* options from txtRecords should not be prematurely unescaped

github PRs don't let me comment on lines that haven't changed, so instead please refer to my branch for the full set of suggested changes:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3239-1123

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -83,6 +86,24 @@ public void Resolve_should_return_expected_result(string connectionString, strin
...
+
+            var uri = $"mongodb+srv://{username}:{Uri.EscapeDataString(password)}@test5.test.build.10gen.cc/test?authSource=$external&authMechanism={authMechanism}&authMechanismProperties=AWS_SESSION_TOKEN:{Uri.EscapeDataString(awsSessionToken)}";
+            var connectionString = new ConnectionString(uri);
+
+            var result = connectionString.Resolve();
+
+            result.AuthMechanism.Should().Be(authMechanism);
+            result.Username.Should().Be(username);
+            result.AuthMechanismProperties["AWS_SESSION_TOKEN"].Should().Be(awsSessionToken);
+        }
```
MikalaiMazurenka, line 105: Since currently specifying unescaped token works, I suggest to add a test case for that (same for `MongoClientSettingsTests`):
```
[Theory]
[ParameterAttributeData]
public void Resolve_against_mongodb_aws_session_token_should_return_the_expected_aws_session_token(
	[Values(false, true)] bool escapeToken)
{
	const string authMechanism = "MONGODB-AWS";
	const string username = "AKIAIOSFODNN7EXAMPLE";
	const string password = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY";
	const string awsSessionToken = "AQoEXAMPLEH4aoAH0gNCAPyJxz4BlCFFxWNE1OPTgk5TthT+FvwqnKwRcOIfrRh3c/LTo6UDdyJwOOvEVPvLXCrrrUtdnniCEXAMPLE/IvU1dYUg2RVAJBanLiHb4IgRmpRV3zrkuWJOgQs8IZZaIv2BXIa2R4OlgkBN9bkUDNCJiBeb/AXlzBBko7b15fjrBs2+cTQtpZ3CYWFXG8C5zqx37wnOE49mRl/+OtkIKGO7fAE";

	var preparedToken = escapeToken ? Uri.EscapeDataString(awsSessionToken) : awsSessionToken;
	var uri = $"mongodb+srv://{username}:{Uri.EscapeDataString(password)}@test5.test.build.10gen.cc/test?authSource=$external&authMechanism={authMechanism}&authMechanismProperties=AWS_SESSION_TOKEN:{preparedToken}";
	var connectionString = new ConnectionString(uri);

	var result = connectionString.Resolve();

	result.AuthMechanism.Should().Be(authMechanism);
	result.Username.Should().Be(username);
	result.AuthMechanismProperties["AWS_SESSION_TOKEN"].Should().Be(awsSessionToken);
}
```

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
MikalaiMazurenka, line 695: I'd suggest to remove the comments as they are too verbose.

MikalaiMazurenka (approved):
The solutions looks fine to me, with just a few minor comments which are not essential.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
JamesKovacs: Done.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
JamesKovacs: Changed the logic to always use unescaped options internally. So this comment is no longer necessary.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -1294,23 +1292,20 @@ private List<string> GetHostsFromSrvRecords(IEnumerable<SrvRecord> srvRecords)
...
 
-        private NameValueCollection GetResolvedOptions(List<string> options)
-        {
             // Build a dummy connection string in order to parse the options
             var dummyConnectionString = "mongodb://localhost/";
-            if (options.Count > 0)
+            if (txtRecord != null)
             {
-                dummyConnectionString += "?" + string.Join("&", options);
+                dummyConnectionString += "?" + string.Join("", txtRecord.Strings);
```
rstam, line 1308: Don't the multiple `txtRecord.Strings` need to be separated by `&` characters?

We probably don't have any tests failing because in our tests there is only one `txtRecord.Strings`.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -63,8 +64,17 @@ public void constructor_should_throw_when_isResolved_is_invalid()
         [Theory]
         [InlineData("mongodb://test5.test.build.10gen.cc", "mongodb://test5.test.build.10gen.cc", false)]
         [InlineData("mongodb://test5.test.build.10gen.cc", "mongodb://test5.test.build.10gen.cc", true)]
-        [InlineData("mongodb+srv://test5.test.build.10gen.cc", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=thisDB&tls=true", false)]
-        [InlineData("mongodb+srv://test5.test.build.10gen.cc", "mongodb://localhost.test.build.10gen.cc:27017/?replicaSet=repl0&authSource=thisDB&tls=true", true)]
+        [InlineData("mongodb+srv://test5.test.build.10gen.cc",
```
rstam, line 67: Hard to tell what actually changed when arbitrary formatting changes are introduced at the same time...

rstam: Hmmm.... looks like NOTHING changed?

Am I missing something?

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -83,6 +93,27 @@ public void Resolve_should_return_expected_result(string connectionString, strin
...
+        [ParameterAttributeData]
+        public void Resolve_against_mongodb_aws_session_token_should_return_the_expected_aws_session_token([Values(false, true)] bool escapeToken)
+
+    {
+            const string authMechanism = "MONGODB-AWS";
+            const string username = "AKIAIOSFODNN7EXAMPLE";
+            const string password = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY";
+            const string awsSessionToken = "AQoEXAMPLEH4aoAH0gNCAPyJxz4BlCFFxWNE1OPTgk5TthT+FvwqnKwRcOIfrRh3c/LTo6UDdyJwOOvEVPvLXCrrrUtdnniCEXAMPLE/IvU1dYUg2RVAJBanLiHb4IgRmpRV3zrkuWJOgQs8IZZaIv2BXIa2R4OlgkBN9bkUDNCJiBeb/AXlzBBko7b15fjrBs2+cTQtpZ3CYWFXG8C5zqx37wnOE49mRl/+OtkIKGO7fAE";
+
+            var preparedToken = escapeToken ? Uri.EscapeDataString(awsSessionToken) : awsSessionToken;
```
rstam, line 106: I don't understand the purpose of not always escaping.

Shouldn't the `awsSessionToken` be invalid if it is not escaped? It will only be valid IFF it happens to not contain any special characters than need to be escaped.

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
```diff
@@ -704,6 +705,26 @@ public void TestFromUrlResolving(string connectionString, ConnectionStringScheme
...
+        [Theory]
+        [ParameterAttributeData]
+        public void TestFromUrlWithMongoDBAWS_should_parse_credentials_correctly([Values(false, true)] bool escapeToken)
+        {
+            const string authMechanism = "MONGODB-AWS";
+            const string username = "AKIAIOSFODNN7EXAMPLE";
+            const string password = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYzEXAMPLEKEY";
+            const string awsSessionToken = "AQoEXAMPLEH4aoAH0gNCAPyJxz4BlCFFxWNE1OPTgk5TthT+FvwqnKwRcOIfrRh3c/LTo6UDdyJwOOvEVPvLXCrrrUtdnniCEXAMPLE/IvU1dYUg2RVAJBanLiHb4IgRmpRV3zrkuWJOgQs8IZZaIv2BXIa2R4OlgkBN9bkUDNCJiBeb/AXlzBBko7b15fjrBs2+cTQtpZ3CYWFXG8C5zqx37wnOE49mRl/+OtkIKGO7fAE";
+
+            var preparedToken = escapeToken ? Uri.EscapeDataString(awsSessionToken) : awsSessionToken;
```
rstam, line 717: Same question as above.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
JamesKovacs: We are concatenating together `txtRecord.Strings`, not TXT records themselves. We have  a guard clause (lines 1297-1300) on the method asserting that there can be at most 1 TXT record. According to [our SRV spec](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst), there can be only a single TXT record, but it can contain multiple strings which are to be concatenated together without a separator.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
JamesKovacs: This was Rider trying to be "helpful" in wrapping long lines. I thought I'd reverted all its changes. Guess I missed one. Reverted.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
JamesKovacs: According to RFC3986 (URI syntax), the query component is delimited by the first `?` and ends with the first `#`. The query component can contain additional `?` characters as well as `/` characters. So the `awsSessionToken` may be escaped or it may not be. I think it worthwhile to test both cases. Typically what made the `awsSessionToken` invalid was the erroneous double-escaping of `/` characters in the token.

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
JamesKovacs: See response above.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
MikalaiMazurenka: > I don't understand the purpose of not always escaping.

I've requested to add this case because it previously worked. So my intention was to preserve backward compatibility.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
rstam: Got it. I was confused by the fact that his line was concatenating (sligtly preprocessed) TXT records *before*, but *now* it's concatenating the strings in one TXT record.

Looks correct now that I understand that change.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
rstam: OK... but **not** escaping **only** works because the particular test value used doesn't contain any special characters that need to be escaped.

rstam (approved):
LGTM

JamesKovacs:
Squashed and merged with `master`.

JamesKovacs:
Re-opening to fix PR merge.

PR #421, opened 2020-11-24
CSHARP-3266: Added support string.Equals with StringComparison argument to predicate translator

exileDev:
Reported in https://jira.mongodb.org/browse/CSHARP-3266

### DIFF
--- a/src/MongoDB.Driver/Linq/Translators/PredicateTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Translators/PredicateTranslator.cs
@@ -624,44 +624,27 @@ namespace MongoDB.Driver.Linq.Translators
 
         private FilterDefinition<BsonDocument> TranslateEquals(MethodCallExpression methodCallExpression)
         {
-            var arguments = methodCallExpression.Arguments.ToArray();
+            var arguments = methodCallExpression.Arguments.ToList();
+
+            if (arguments.Count == 0)
+                return null;
 
             // assume that static and instance Equals mean the same thing for all classes (i.e. an equality test)
-            Expression firstExpression = null;
-            Expression secondExpression = null;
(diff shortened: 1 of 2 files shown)


JamesKovacs:
Thank you for taking the time to create a PR for this issue.

Reviewing the PR, it is not always correct to translate `InvariantCultureIgnoreCase` and `OrdinalIgnoreCase` to `/^<<STRING>>$/i`. This is due to the classic Turkish `i` problem. In certain non-English languages such as Turkish, casing rules are different. For example in Turkish, "I" is not the capital form of "i", "İ" is. Similarly the lowercase of "I" is "ı" (i without a dot). To correctly handle ordinal versus invariant cultures, you must use a supporting index with a collation.

We are currently working on a new implementation of our LINQ translator and will consider better `StringComparison` support in the context of that work.

exileDev:
> We are currently working on a new implementation of our LINQ translator and will consider better StringComparison support in the context of that work.

@JamesKovacs I don't understand where I can see this implementation progress? Can you clarify what it means?

JamesKovacs:
@exileDev - We are actively working on the implementation of the new LINQ translator. While we can execute many LINQ operations already with the new provider, it is not yet complete enough to be practically useful. It is currently in a private fork and not ready for public consumption. Once it is complete enough to replace for our current LINQ provider, we will push the code to our public repo. Thank you for your patience.

exileDev:
@JamesKovacs @rstam Do you have any estimate on when it could be ready?
I'd also like to share information that we're planning to use it in our open-source solution that currently powers more than 60.000+ websites. So MongoDB also could get a lot of new users

JamesKovacs:
@exileDev We plan to include the new LINQ provider in a future minor release of the driver, but we do not have a release date for it yet.

PR #425, opened 2020-12-02
CSHARP-3271 Docs updates

BorisDog:
- Various docs cleanups
- Minor styling change
- Addressed [209](https://github.com/mongodb/mongo-csharp-driver/pull/290) and [301](https://github.com/mongodb/mongo-csharp-driver/pull/301) PRs

Updated docs can be viewed here: https://borisdog.github.io/mongo-csharp-driver/2.12/reference/


Docs/reference/content/reference/driver/crud/linq.md
```diff
@@ -492,25 +491,6 @@ var result = collection.AsQueryable().Select(p => p.Age).Max();
...
-
-var result = collection.AsQueryable().Select(p => p.Age).Sum();
-```
-```json
-[
-    { $group: { _id: 1, __result: { $Sum: "$Age" } } }
-]
-```
-
-
```
JamesKovacs, line 513: Why is the documentation for `Sum` being removed entirely? If we decide to keep `Sum`, we should find a better example than summing up people's ages. That's a strange example.

Docs/reference/content/reference/driver/expressions.md
```diff
@@ -235,13 +242,10 @@ See the [MongoDB documentation]({{< docsref "reference/operator/query/#element"
...
 
 ```csharp
-// no example yet
+Find(p => p.Children.ContainsKey("John"))
 ```
-
-#### $type
-
-```csharp
-// no example yet
```
JamesKovacs, line 244: The example below is for `$exists` but we lose the placeholder for `$type`.

JamesKovacs (changes requested):
A few minor fixes. One thing I don't like is that this PR is a grab bag of documentation fixes, XML comment fixes, and code fixes. The code fixes are admittedly minor in terms of putting modifiers in standard order. We can either split this into multiple PRs or if we accept it as-is, I would suggest that we are careful in squashing and re-ordering commits. I would keep each type of change in its own commit and only squash together related ones.

rstam:
I think we should abort this PR and start over with several smaller ones. There is too much unrelated stuff in this single PR.

Also, when making changes that start with a community submitted PR, we should always start by at least rebase the (usually old) PR on master. There might be merge conflicts to resove, and they should be resolved *before* we begin reviewing. There might be further changes required, and they also should be done after rebasing on master.

rstam (changes requested):
Change requested: open new smaller PRs that don't contain unrelated changes coming from multiple community PRs.

BorisDog:
Abandoning, addressing in few separate PRs

PR #428, opened 2020-12-28
CSHARP-2693/CSHARP-2692: Encoding.GetBytes buffer caching optimization

BorisDog:
CSHARP-2693/CSHARP-2692: Encoding.GetBytes buffer caching optimization

### DIFF
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -370,7 +370,7 @@ namespace MongoDB.Bson.IO
             ThrowIfDisposed();
 
             var length = ReadInt32();
-            var bytes = BufferCache.GetBuffer(length);
+            var bytes = ThreadStaticBuffer.GetBuffer(length);
             this.ReadBytes(bytes, 0, length);
             if (bytes[length - 1] != 0)
             {
@@ -459,7 +459,7 @@ namespace MongoDB.Bson.IO
             const int legacyCachedBufferLength = 128;
             if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= legacyCachedBufferLength)
             {
-                bytes = BufferCache.GetBuffer(legacyCachedBufferLength);
+                bytes = ThreadStaticBuffer.GetBuffer(legacyCachedBufferLength);
                 length = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
             }
             else
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -462,7 +462,7 @@ namespace MongoDB.Bson.IO
                 using var rentedBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
 
                 var length = CStringUtf8Encoding.GetBytes(value, rentedBuffer.Bytes, 0, Utf8Encodings.Strict);
-                SetBytes(rentedBuffer.Bytes, length);
+                WriteBytes(rentedBuffer.Bytes, length);
             }
             else
             {
@@ -473,10 +473,10 @@ namespace MongoDB.Bson.IO
                     throw new ArgumentException("A CString cannot contain null bytes.", "value");
                 }
 
-                SetBytes(segment.Array, segment.Count);
+                WriteBytes(segment.Array, segment.Count);
             }
 
-            void SetBytes(byte[] bytes, int length)
+            void WriteBytes(byte[] bytes, int length)
             {
                 _stream.Write(bytes, 0, length);
                 _stream.WriteByte(0);
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -370,7 +370,9 @@ namespace MongoDB.Bson.IO
             ThrowIfDisposed();
 
             var length = ReadInt32();
-            var bytes = ThreadStaticBuffer.GetBuffer(length);
+            using var rentedBuffer = ThreadStaticBuffer.GetBuffer(length);
+            var bytes = rentedBuffer.Buffer;
+
             this.ReadBytes(bytes, 0, length);
             if (bytes[length - 1] != 0)
             {
@@ -452,30 +454,33 @@ namespace MongoDB.Bson.IO
             }
             ThrowIfDisposed();
 
-            byte[] bytes;
-            int length;
 
             // Compare to 128 to preserve original behavior
             const int maxLengthToUseCStringUtf8EncodingWith = 128;
             if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= maxLengthToUseCStringUtf8EncodingWith)
             {
-                bytes = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
-                length = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
+                using var rentedBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
+
+                var length = CStringUtf8Encoding.GetBytes(value, rentedBuffer.Buffer, 0, Utf8Encodings.Strict);
+                SetBytes(rentedBuffer.Buffer, length);
             }
             else
             {
-                var segment = Utf8Encodings.Strict.GetBytesUsingThreadStaticBuffer(value);
-                bytes = segment.Array;
-                length = segment.Count;
-
-                if (Array.IndexOf<byte>(bytes, 0, 0, length) != -1)
+                using var rentedSegment = Utf8Encodings.Strict.GetBytesUsingThreadStaticBuffer(value);
+                var segment = rentedSegment.Segment;
+                if (Array.IndexOf<byte>(segment.Array, 0, 0, segment.Count) != -1)
                 {
                     throw new ArgumentException("A CString cannot contain null bytes.", "value");
                 }
+
+                SetBytes(segment.Array, segment.Count);
             }
 
-            _stream.Write(bytes, 0, length);
-            _stream.WriteByte(0);
+            void SetBytes(byte[] bytes, int length)
+            {
+                _stream.Write(bytes, 0, length);
+                _stream.WriteByte(0);
+            }
         }
 
         /// <inheritdoc/>
@@ -547,7 +552,8 @@ namespace MongoDB.Bson.IO
             }
             ThrowIfDisposed();
 
-            var segment = encoding.GetBytesUsingThreadStaticBuffer(value);
+            using var rentedSegment = encoding.GetBytesUsingThreadStaticBuffer(value);
+            var segment = rentedSegment.Segment;
 
             WriteInt32(segment.Count + 1);
             _stream.Write(segment.Array, 0, segment.Count);
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -455,16 +455,16 @@ namespace MongoDB.Bson.IO
             byte[] bytes;
             int length;
 
-            // Compare to 128 to preserve original behaviour
-            const int legacyCachedBufferLength = 128;
-            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= legacyCachedBufferLength)
+            // Compare to 128 to preserve original behavior
+            const int maxLengthToUseCStringUtf8EncodingWith = 128;
+            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= maxLengthToUseCStringUtf8EncodingWith)
             {
-                bytes = ThreadStaticBuffer.GetBuffer(legacyCachedBufferLength);
+                bytes = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
                 length = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
             }
             else
             {
-                var segment = Utf8Encodings.Strict.GetBytesCachedBuffer(value);
+                var segment = Utf8Encodings.Strict.GetBytesUsingThreadStaticBuffer(value);
                 bytes = segment.Array;
                 length = segment.Count;
 
@@ -547,7 +547,7 @@ namespace MongoDB.Bson.IO
             }
             ThrowIfDisposed();
 
-            var segment = encoding.GetBytesCachedBuffer(value);
+            var segment = encoding.GetBytesUsingThreadStaticBuffer(value);
 
             WriteInt32(segment.Count + 1);
             _stream.Write(segment.Array, 0, segment.Count);
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -15,7 +15,6 @@
 
 using System;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -32,7 +31,6 @@ namespace MongoDB.Bson.IO
         private bool _ownsStream;
         private readonly Stream _stream;
         private readonly byte[] _temp = new byte[12];
-        private readonly byte[] _tempUtf8 = new byte[128];
 
         // constructors
         /// <summary>
@@ -52,7 +50,7 @@ namespace MongoDB.Bson.IO
             _ownsStream = ownsStream;
         }
 
-        // properties        
+        // properties
         /// <summary>
         /// Gets the base stream.
         /// </summary>
@@ -372,7 +370,7 @@ namespace MongoDB.Bson.IO
             ThrowIfDisposed();
 
             var length = ReadInt32();
-            var bytes = length <= _tempUtf8.Length ? _tempUtf8 : new byte[length];
+            var bytes = BufferCache.GetBuffer(length);
             this.ReadBytes(bytes, 0, length);
             if (bytes[length - 1] != 0)
             {
@@ -457,19 +455,23 @@ namespace MongoDB.Bson.IO
             byte[] bytes;
             int length;
 
-            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= _tempUtf8.Length)
+            // Compare to 128 to preserve original behaviour
+            const int legacyCachedBufferLength = 128;
+            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= legacyCachedBufferLength)
             {
-                bytes = _tempUtf8;
-                length = CStringUtf8Encoding.GetBytes(value, _tempUtf8, 0, Utf8Encodings.Strict);
+                bytes = BufferCache.GetBuffer(legacyCachedBufferLength);
+                length = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
             }
             else
             {
-                bytes = Utf8Encodings.Strict.GetBytes(value);
-                if (Array.IndexOf<byte>(bytes, 0) != -1)
+                var segment = Utf8Encodings.Strict.GetBytesCachedBuffer(value);
+                bytes = segment.Array;
+                length = segment.Count;
+
+                if (Array.IndexOf<byte>(bytes, 0, 0, length) != -1)
                 {
                     throw new ArgumentException("A CString cannot contain null bytes.", "value");
                 }
-                length = bytes.Length;
             }
 
             _stream.Write(bytes, 0, length);
@@ -545,22 +547,10 @@ namespace MongoDB.Bson.IO
             }
             ThrowIfDisposed();
 
-            byte[] bytes;
-            int length;
+            var segment = encoding.GetBytesCachedBuffer(value);
 
-            if (encoding.GetMaxByteCount(value.Length) <= _tempUtf8.Length)
-            {
-                bytes = _tempUtf8;
-                length = encoding.GetBytes(value, 0, value.Length, _tempUtf8, 0);
-            }
-            else
-            {
-                bytes = encoding.GetBytes(value);
-                length = bytes.Length;
-            }
-
-            WriteInt32(length + 1);
-            _stream.Write(bytes, 0, length);
+            WriteInt32(segment.Count + 1);
+            _stream.Write(segment.Array, 0, segment.Count);
             _stream.WriteByte(0);
         }
     }
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -371,7 +371,7 @@ namespace MongoDB.Bson.IO
 
             var length = ReadInt32();
             using var rentedBuffer = ThreadStaticBuffer.GetBuffer(length);
-            var bytes = rentedBuffer.Buffer;
+            var bytes = rentedBuffer.Bytes;
 
             this.ReadBytes(bytes, 0, length);
             if (bytes[length - 1] != 0)
@@ -461,8 +461,8 @@ namespace MongoDB.Bson.IO
             {
                 using var rentedBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
 
-                var length = CStringUtf8Encoding.GetBytes(value, rentedBuffer.Buffer, 0, Utf8Encodings.Strict);
-                SetBytes(rentedBuffer.Buffer, length);
+                var length = CStringUtf8Encoding.GetBytes(value, rentedBuffer.Bytes, 0, Utf8Encodings.Strict);
+                SetBytes(rentedBuffer.Bytes, length);
             }
             else
             {
--- a/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
+++ b/src/MongoDB.Bson/IO/BsonStreamAdapter.cs
@@ -370,7 +370,7 @@ namespace MongoDB.Bson.IO
             ThrowIfDisposed();
 
             var length = ReadInt32();
-            using var rentedBuffer = ThreadStaticBuffer.GetBuffer(length);
+            using var rentedBuffer = ThreadStaticBuffer.RentBuffer(length);
             var bytes = rentedBuffer.Bytes;
 
             this.ReadBytes(bytes, 0, length);
@@ -459,7 +459,7 @@ namespace MongoDB.Bson.IO
             const int maxLengthToUseCStringUtf8EncodingWith = 128;
             if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= maxLengthToUseCStringUtf8EncodingWith)
             {
-                using var rentedBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
+                using var rentedBuffer = ThreadStaticBuffer.RentBuffer(maxLengthToUseCStringUtf8EncodingWith);
 
                 var length = CStringUtf8Encoding.GetBytes(value, rentedBuffer.Bytes, 0, Utf8Encodings.Strict);
                 WriteBytes(rentedBuffer.Bytes, length);
--- a/src/MongoDB.Bson/IO/BufferCache.cs
+++ b/src/MongoDB.Bson/IO/BufferCache.cs
@@ -25,6 +25,7 @@ namespace MongoDB.Bson.IO
     {
         private const int MinSize = 16;
         private const int MaxSize = 8192;
+        private const int MaxAllocationSize = 1024 * 1024 * 1024; // 1GB
         private const int MaxThreads = 1024;
 
         // private static fields
@@ -35,6 +36,11 @@ namespace MongoDB.Bson.IO
 
         public static byte[] GetBuffer(int size)
         {
+            if (size <= 0 || size > MaxAllocationSize)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), "Invalid requested buffer size");
+            }
+
             if (size > MaxSize ||
                 __buffer == null && Interlocked.Increment(ref __buffersCount) >= MaxThreads)
             {
--- a/src/MongoDB.Bson/IO/BufferCache.cs
+++ b/src/MongoDB.Bson/IO/BufferCache.cs
@@ -1,0 +1,67 @@
+﻿/* Copyright 2020-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Threading;
+
+namespace MongoDB.Bson.IO
+{
+    /// <summary>
+    /// Represents a class that provides reusable buffer per thread.
+    /// </summary>
+    internal static class BufferCache
+    {
+        private const int MinSize = 16;
+        private const int MaxSize = 8192;
+        private const int MaxThreads = 1024;
+
+        // private static fields
+        [ThreadStatic]
+        private static byte[] __buffer;
+
+        private static int __buffersCount = 0;
+
+        public static byte[] GetBuffer(int size)
+        {
+            if (size > MaxSize ||
+                __buffer == null && Interlocked.Increment(ref __buffersCount) >= MaxThreads)
+            {
+                return new byte[size];
+            }
+
+            if (!(__buffer?.Length >= size))
+            {
+                var newSize = size <= MinSize ? MinSize : CeilPower2(size);
+                __buffer = new byte[newSize];
+            }
+
+            return __buffer;
+        }
+
+        // private methods
+        private static int CeilPower2(int number)
+        {
+            number--;
+            number |= number >> 1;
+            number |= number >> 2;
+            number |= number >> 4;
+            number |= number >> 8;
+            number |= number >> 16;
+            number++;
+
+            return number;
+        }
+    }
+}
--- a/src/MongoDB.Bson/IO/ByteBufferStream.cs
+++ b/src/MongoDB.Bson/IO/ByteBufferStream.cs
@@ -32,7 +32,6 @@ namespace MongoDB.Bson.IO
         private readonly bool _ownsBuffer;
         private int _position;
         private readonly byte[] _temp = new byte[12];
-        private readonly byte[] _tempUtf8 = new byte[128];
 
         // constructors
         /// <summary>
@@ -529,7 +528,7 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                var bytes = length <= _tempUtf8.Length ? _tempUtf8 : new byte[length];
+                var bytes = BufferCache.GetBuffer(length);
                 this.ReadBytes(bytes, 0, length);
                 if (bytes[length - 1] != 0)
                 {
@@ -568,20 +567,25 @@ namespace MongoDB.Bson.IO
             }
             else
             {
+                // Compare to 128 to preserve original behaviour
+                const int legacyCachedBufferLength = 128;
+
                 byte[] bytes;
-                if (maxLength <= _tempUtf8.Length)
+                if (maxLength <= legacyCachedBufferLength)
                 {
-                    bytes = _tempUtf8;
+                    bytes = BufferCache.GetBuffer(legacyCachedBufferLength);
                     actualLength = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
                 }
                 else
                 {
-                    bytes = Utf8Encodings.Strict.GetBytes(value);
-                    if (Array.IndexOf<byte>(bytes, 0) != -1)
+                    var segmentEncoded = Utf8Encodings.Strict.GetBytesCachedBuffer(value);
+                    bytes = segmentEncoded.Array;
+                    actualLength = segmentEncoded.Count;
+
+                    if (Array.IndexOf<byte>(bytes, 0, 0, actualLength) != -1)
                     {
                         throw new ArgumentException("A CString cannot contain null bytes.", "value");
                     }
-                    actualLength = bytes.Length;
                 }
 
                 _buffer.SetBytes(_position, bytes, 0, actualLength);
@@ -715,17 +719,9 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                byte[] bytes;
-                if (maxLength <= _tempUtf8.Length)
-                {
-                    bytes = _tempUtf8;
-                    actualLength = encoding.GetBytes(value, 0, value.Length, bytes, 0);
-                }
-                else
-                {
-                    bytes = encoding.GetBytes(value);
-                    actualLength = bytes.Length;
-                }
+                var segmentEncoded = encoding.GetBytesCachedBuffer(value);
+                var bytes = segmentEncoded.Array;
+                actualLength = segmentEncoded.Count;
 
                 var lengthPlusOneBytes = BitConverter.GetBytes(actualLength + 1);
 
--- a/src/MongoDB.Bson/IO/ByteBufferStream.cs
+++ b/src/MongoDB.Bson/IO/ByteBufferStream.cs
@@ -528,12 +528,14 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                var bytes = ThreadStaticBuffer.GetBuffer(length);
+                using var rentedBuffer = ThreadStaticBuffer.GetBuffer(length);
+                var bytes = rentedBuffer.Buffer;
                 this.ReadBytes(bytes, 0, length);
                 if (bytes[length - 1] != 0)
                 {
                     throw new FormatException("String is missing terminating null byte.");
                 }
+
                 return Utf8Helper.DecodeUtf8String(bytes, 0, length - 1, encoding);
             }
         }
@@ -569,27 +571,33 @@ namespace MongoDB.Bson.IO
             {
                 // Compare to 128 to preserve original behavior
                 const int maxLengthToUseCStringUtf8EncodingWith = 128;
-
-                byte[] bytes;
+          
                 if (maxLength <= maxLengthToUseCStringUtf8EncodingWith)
                 {
-                    bytes = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
-                    actualLength = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
+                    using var acquiredBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
+                    actualLength = CStringUtf8Encoding.GetBytes(value, acquiredBuffer.Buffer, 0, Utf8Encodings.Strict);
+
+                    SetBytes(acquiredBuffer.Buffer, actualLength);
                 }
                 else
                 {
-                    var segmentEncoded = Utf8Encodings.Strict.GetBytesUsingThreadStaticBuffer(value);
-                    bytes = segmentEncoded.Array;
+                    using var rentedSegmentEncoded = Utf8Encodings.Strict.GetBytesUsingThreadStaticBuffer(value);
+                    var segmentEncoded = rentedSegmentEncoded.Segment;
                     actualLength = segmentEncoded.Count;
 
-                    if (Array.IndexOf<byte>(bytes, 0, 0, actualLength) != -1)
+                    if (Array.IndexOf<byte>(segmentEncoded.Array, 0, 0, actualLength) != -1)
                     {
                         throw new ArgumentException("A CString cannot contain null bytes.", "value");
                     }
+
+                    SetBytes(segmentEncoded.Array, actualLength);
                 }
 
-                _buffer.SetBytes(_position, bytes, 0, actualLength);
-                _buffer.SetByte(_position + actualLength, 0);
+                void SetBytes(byte[] bytes, int lenght)
+                {
+                    _buffer.SetBytes(_position, bytes, 0, actualLength);
+                    _buffer.SetByte(_position + actualLength, 0);
+                }
             }
 
             SetPositionAfterWrite(_position + actualLength + 1);
@@ -719,9 +727,9 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                var segmentEncoded = encoding.GetBytesUsingThreadStaticBuffer(value);
-                var bytes = segmentEncoded.Array;
-                actualLength = segmentEncoded.Count;
+                using var rentedSegmentEncoded = encoding.GetBytesUsingThreadStaticBuffer(value);
+                var bytes = rentedSegmentEncoded.Segment.Array;
+                actualLength = rentedSegmentEncoded.Segment.Count;
 
                 var lengthPlusOneBytes = BitConverter.GetBytes(actualLength + 1);
 
--- a/src/MongoDB.Bson/IO/ByteBufferStream.cs
+++ b/src/MongoDB.Bson/IO/ByteBufferStream.cs
@@ -528,7 +528,7 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                var bytes = BufferCache.GetBuffer(length);
+                var bytes = ThreadStaticBuffer.GetBuffer(length);
                 this.ReadBytes(bytes, 0, length);
                 if (bytes[length - 1] != 0)
                 {
@@ -573,7 +573,7 @@ namespace MongoDB.Bson.IO
                 byte[] bytes;
                 if (maxLength <= legacyCachedBufferLength)
                 {
-                    bytes = BufferCache.GetBuffer(legacyCachedBufferLength);
+                    bytes = ThreadStaticBuffer.GetBuffer(legacyCachedBufferLength);
                     actualLength = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
                 }
                 else
--- a/src/MongoDB.Bson/IO/ByteBufferStream.cs
+++ b/src/MongoDB.Bson/IO/ByteBufferStream.cs
@@ -567,18 +567,18 @@ namespace MongoDB.Bson.IO
             }
             else
             {
-                // Compare to 128 to preserve original behaviour
-                const int legacyCachedBufferLength = 128;
+                // Compare to 128 to preserve original behavior
+                const int maxLengthToUseCStringUtf8EncodingWith = 128;
 
                 byte[] bytes;
-                if (maxLength <= legacyCachedBufferLength)
+                if (maxLength <= maxLengthToUseCStringUtf8EncodingWith)
                 {
-                    bytes = ThreadStaticBuffer.GetBuffer(legacyCachedBufferLength);
+                    bytes = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
                     actualLength = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
                 }
                 else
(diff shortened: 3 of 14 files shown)


BorisDog:
[EG Patch](https://evergreen.mongodb.com/version/5fea4cdf3e8e862d05ffc5dd)  [InternalLink]

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
```diff
@@ -372,7 +370,7 @@ public override string ReadString(UTF8Encoding encoding)
             ThrowIfDisposed();
 
             var length = ReadInt32();
-            var bytes = length <= _tempUtf8.Length ? _tempUtf8 : new byte[length];
+            var bytes = BufferCache.GetBuffer(length);
```
rstam, line 373: Initial reaction: this feels like it could be very dangerous.

The main concern is that the `_tempUtf8` buffer was owned by this **class**. Now we're switching to using a buffer owned by the **thread**.

The danger is that two different classes might both try to use the same buffer in a scenario where one class calls the other and both try to use the same buffer. In this scenario the second class will reuse and overwrite the contents of the one and only cached buffer which will corrupt the caller's data. We've had bugs like this before where the same buffer was used simultaneously for two different purposes which were very hard to find.

While we can carefully examine our code to ensure that we don't let this scenario happen, it feels dangerous to use a buffer caching technique that could easily result in unexpected bugs.

Other possible issues to consider:

* Is there potential for a memory leak? What happens to the `[ThreadStatic] __buffer` when the tread that caused it to be allocated no longer exists. One would hope it got garbage collected but this would have to be verified.
* A single large enough allocation can expand the cached buffer to 8K after which it never shrinks. Will this result in an eventual steady state in which all cached buffers are 8K in size? Is that OK?

I'd like to hear what other reviewers thinking about the potential dangers of using a single cached buffer per thread.

Until that question is resolved I'm going to pause reviewing this PR further.

rstam (changes requested):
Note: we don't usually combine multiple JIRA tickets in a single PR.

I'm waiting to see what others thing about the potential dangers I've identified of using a single cached buffer (per thread) across multiple classes before reviewing further.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
```diff
@@ -457,19 +455,23 @@ public override void WriteCString(string value)
...
 
-            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= _tempUtf8.Length)
+            // Compare to 128 to preserve original behaviour
+            const int legacyCachedBufferLength = 128;
+            if (CStringUtf8Encoding.GetMaxByteCount(value.Length) <= legacyCachedBufferLength)
             {
-                bytes = _tempUtf8;
-                length = CStringUtf8Encoding.GetBytes(value, _tempUtf8, 0, Utf8Encodings.Strict);
+                bytes = BufferCache.GetBuffer(legacyCachedBufferLength);
+                length = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
```
rstam, line 463: Consider keeping `_tempUtf8` as is (possibly increasing its size to make it usable more often) and instead ONLY change the `else` clause to use a buffer from a shared pool instead of allocating a new one.

We should consider using the standard .NET `MemoryPool<byte>.Shared` pool from the `System.Buffers` namespace.

Note that this buffer pool does NOT associated buffers with threads, which is a good thing.

rstam (changes requested):

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
rstam: Actually we probably would use `ArrayPool<byte>` instead of `MemoryPool<byte>`.

Note that while we have NOT used these classes before, we ALREADY have  an indirect dependency  on the `System.Buffers.dll` Nuget package because `DnsClient.dll` uses it.

rstam (changes requested):

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
BorisDog: My preference would be to keep things simple on the hot path. MemoryPool and ArrayPool are great solutions, but introduce more complex mechanisms and involve synchronization. In this particular case as we just need a local buffer for a quick operation. MemoryPool and ArrayPool would be an overkill.

The core goal of this PR is to eliminate the need to maintain local buffers. I don’t see any good reason to keep using _tempUtf8.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
BorisDog: I understand your concerns. There are always certain dangers when using any programming technique. But this should not prevent us from using any development tool when we are convinced in its correctness.
In this case, this trivial caching technique is intended to be used in very short and self-contained scenarios, like the current example.

Memory leak: The buffers are stored in TLS, and collected when thread is disposed. In any case, accumulative buffers size is limited to 8MB for the whole application lifecycle.

8K: Yes, I believe that this is ok. Notice that this trivial caching is **intended** to be very simple and efficient. I don’t think we need anything more sophisticated at this point.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
rstam: > The core goal of this PR is to eliminate the need to maintain local buffers.

I thought the core goal of this PR was to avoid allocating new buffers when the `__tempUtf8` buffer isn't big enough, by using the thread cached buffer (when the required buffer size is <= 8K).

rstam (changes requested):
More partial review comments.

This is an interesting PR.

I'd like us to learn more about the performance characteristics of using a ThreadStatic cached buffer. It's possible that the new `GetBuffer` method adds overhead and that it would still be beneficial to keep the `_tempUtf8` field to get better performance when the requried buffer size is small enough.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
BorisDog: Memory leak: Yes, while I haven't not found any specific documentation about TLS garbage collection process yet, I've verified that objects finalizers are being called.

Larger initial size: Absolutely, good point.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
BorisDog: Eliminating branching: It's definitely a good idea. But given that CStringUtf8Encoding.GetBytes is different from Utf8Encodings.Strict.GetBytes we need to consider this carefully, and maybe address this issue in separate PR.

Performance: Good point as well. From by benchmarks, accessing thread static fields vs local field adds about 5-8% to running time. But given that the _tempUtf8 is allocated with **every** instance of the enclosing object, accessing thread static buffer is faster overall, and contributes to reducing memory pressure. In similar scenario benchmark (2 resizes), thread static approach reduces overall time by 50%, while using ArrayPool increases the running time by 100%.

src/MongoDB.Bson/IO/BufferCache.cs
```diff
@@ -0,0 +1,73 @@
...
+
+using System;
+using System.Threading;
+
+namespace MongoDB.Bson.IO
+{
+    /// <summary>
+    /// Represents a class that provides reusable buffer per thread.
+    /// </summary>
+    internal static class BufferCache
```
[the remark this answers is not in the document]
BorisDog: Good point, tnx.

src/MongoDB.Bson/IO/BufferCache.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public static byte[] GetBuffer(int size)
+        {
+            if (size <= 0 || size > MaxAllocationSize)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), "Invalid requested buffer size");
+            }
+
+            if (size > MaxSize ||
+                __buffer == null && Interlocked.Increment(ref __buffersCount) >= MaxThreads)
```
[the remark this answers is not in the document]
BorisDog: It's not decremented. The main idea is to provide a good-enough and very simple heuristics to optimize the majority of uses cases while not degrading the rest. This optimization will fit well with TPL and other ThreadsPools scenarios, which I believe would be vast majority of the use cases.

src/MongoDB.Bson/IO/BufferCache.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                var newSize = size <= MinSize ? MinSize : CeilPower2(size);
+                __buffer = new byte[newSize];
+            }
+
+            return __buffer;
+        }
+
+        // private methods
+        private static int CeilPower2(int number)
```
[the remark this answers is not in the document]
BorisDog: Tnx!

src/MongoDB.Bson/IO/BufferCache.cs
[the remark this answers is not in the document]
rstam: Or even `ThreadStaticBuffer` (seems pretty accurate!).

src/MongoDB.Bson/IO/BufferCache.cs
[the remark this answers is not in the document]
rstam: I don't think you should limit this technique to the first 1024 threads.

That means that after 1024 threads have used the thread static buffer we will fall back to allocating a NEW buffer EVERY time, which will be specially bad if you also eliminate the `_tempUtf8` field.

We can't stop applications from creating more than 1024 threads. Some applications create a new thread to process each incoming request. Applications that work that way will blow by the 1024 limit quickly.

I say just remove the `__buffersCount` field and let all threads that need it have a thread static buffer.

rstam (changes requested):

src/MongoDB.Bson/IO/BufferCache.cs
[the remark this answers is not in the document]
BorisDog: Done, tnx.

src/MongoDB.Bson/IO/BufferCache.cs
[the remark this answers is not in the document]
BorisDog: Done, tnx.

src/MongoDB.Bson/IO/BufferCache.cs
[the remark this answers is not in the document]
BorisDog: Done.

src/MongoDB.Bson/IO/EncodingHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    internal static class EncodingHelper
+    {
+        private static readonly ArraySegment<byte> __emptySegment = new ArraySegment<byte>(new byte[0]);
+
+        public static ArraySegment<byte> GetBytesCachedBuffer(this Encoding encoding, string value)
+        {
+            if (encoding == null)
+            {
+                throw new ArgumentNullException(nameof(encoding), "Value cannot be null.");
```
rstam, line 32: It is sufficient to use the single argument form. The default message is fine.

     throw new ArgumentNullException(nameof(encoding));

src/MongoDB.Bson/IO/EncodingHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value), "Value cannot be null.");
```
rstam, line 37: It is sufficient to use the single argument form. The default message is fine.

     throw new ArgumentNullException(nameof(value));

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), "Invalid requested buffer size");
+            }
+
+            if (size > MaxSize)
+            {
+                return new byte[size];
+            }
+
+            if (!(__buffer?.Length >= size))
```
rstam, line 45: I suppose this is correct, but I really don't want to jump through all the mental hoops of how nullable ints are compared and what that negated condition means.

I recommend a slightly longer but much easier to read if statement:

    if (__buffer == null || __buffer.Length < size)

because that's close to the English description of "if the buffer hasn't been created yet or is too small" then "create a new buffer".

src/MongoDB.Bson/PowerOf2.cs
```diff
@@ -21,14 +21,20 @@ internal static class PowerOf2
...
+            return result;
         }
 
         public static int RoundUpToPowerOf2(int n)
         {
-            if (n < 0 || n > 0x40000000)
+            if (IsPowerOf2(n))
             {
-                throw new ArgumentOutOfRangeException("n");
+                return n;
```
rstam, line 37: This isn't changing the results of this method right? Isn't this what the calculation below already did?

No harm in adding this, but I want to confirm that it's some sort of optimization/clarification and not a change in logic.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using FluentAssertions;
+using MongoDB.Bson.IO;
+using Xunit;
+
+namespace MongoDB.Bson.Tests.IO
+{
+    public class EncodingHelperTests
+    {
+        [Fact]
+        public void TestEncodingHelper_invalid_encoding_shouldthrow()
```
rstam, line 29: We use the following naming convention for our unit test methods:

    Method_should_something
    Method_should_something_when_something

Which in this case would be:

    GetBytesUsingThreadStaticBuffer_should_throw_when_encoding_is_null

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            var exception = Record.Exception(() => EncodingHelper.GetBytesCachedBuffer(null, "asd"));
```
rstam, line 31: A typical test method consists of three sections: Arrange, Act, Assert, separated by one blank line. In this case there is no Arrange, just Act and Assert.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            exception.Should().BeOfType<ArgumentNullException>();
+            exception.Message.Should().Contain("encoding");
```
rstam, line 33: We have  a pattern we use for testing these exceptions:

            var e = exception.Should().BeOfType<ArgumentNullException>().Subject;
            e.ParamName.Should().Be("encoding");

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
```
rstam, line 34: Putting the above comments together results in a test method conforming to our conventions that looks like:

        [Fact]
        public void GetBytesUsingThreadStaticBuffer_should_throw_when_encoding_is_null()
        {
            var exception = Record.Exception(() => EncodingHelper.GetBytesUsingThreadStaticBuffer(null, "asd"));

            var e = exception.Should().BeOfType<ArgumentNullException>().Subject;
            e.ParamName.Should().Be("encoding");
        }

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [Fact]
+        public void TestEncodingHelper_invalid_string_shouldthrow()
```
rstam, line 37: Rename to `GetBytesUsingThreadStaticBuffer_should_throw_when_value_is_null`

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            var exception = Record.Exception(() => EncodingHelper.GetBytesCachedBuffer(Encoding.ASCII, null));
+            exception.Should().BeOfType<ArgumentNullException>();
+            exception.Message.Should().Contain("value");
+        }
+
+        [Fact]
+        public void TestEncodingHelper_empty_string_shouldbe_default_buffer()
```
rstam, line 45: Rename to `GetBytesUsingThreadStaticBuffer_should_throw_when_value_is_null`

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
rstam: Apply this pattern to the other similar test methods also.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            segmentA.Array.Length.Should().Be(0);
+            segmentA.Array.Should().BeSameAs(segmentB.Array);
+        }
+
+        [Theory]
+        [InlineData(60)]
+        [InlineData(127)]
+        [InlineData(511)]
+        public void TestEncodingHelper_should_reuse_instance(int maxStringSize)
```
rstam, line 58: Rename to `GetBytesUsingThreadStaticBuffer_should_return_same_instance_when_possible`

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+                previousInstance = previousInstance ?? segment.Array;
+
+                encodedActual.ShouldAllBeEquivalentTo(encodedExpected);
+                segment.Array.Should().BeSameAs(previousInstance);
+            }
+        }
+
+        [Fact]
+        public async Task TestEncodingHelperMultithreaded_should_encode_correctly()
```
rstam, line 82: Rename to `GetBytesUsingThreadStaticBuffer_should_return_expected_result_when_multiple_threads_are_used`

rstam: I'm not really sure what this method is testing. It appears to be a multi threaded stress test that tries to do a whole bunch of tests in several threads at once.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            const int threadsCount = 10;
+            const int maxIterations = 10;
```
rstam, line 85: This isn't the "max" iterations, it IS the number of iterations. Consider renaming to `numberOfIterations`

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            const int maxSize = 1024;
+
+            var random = new Random();
+            var encoding = Utf8Encodings.Strict;
+
+            var tasks = Enumerable.Range(0, threadsCount).Select(i => Task.Run(() =>
```
rstam, line 91: I myself am usually reluctant to use Task when what I'm trying to test is something Thread related.

It's not that I don't understand the relationship between a Task and the thread(s) it might run on. But when using Task you give up control of which Thread gets used. Given enough Tasks you might even end up running two or more Tasks on the same thread (one after the other) depending on how/when they get scheduled, which might result in spurious failures.

What do you think? Would it be safer and more explicit to use Threads to test multi-threading?

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                    for (int j = 0; j < maxIterations; j++)
+                    {
+                        var sizeCurrent = random.Next(8, maxSize);
+                        var str = GetString(sizeCurrent);
+
+                        var segment = encoding.GetBytesCachedBuffer(str);
+                        var encodedExpected = encoding.GetBytes(str);
+                        var encodedActual = segment.ToArray();
+
+                        encodedActual.ShouldAllBeEquivalentTo(encodedExpected);
```
rstam, line 102: If this throws an exception it will only cause the Task to fault. It won't directly cause the test to fail.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                })).ToArray();
+
+            await Task.WhenAll(tasks);
```
rstam, line 106: This test will never fail no matter what because we aren't checking whether any of the completed tasks are faulted.

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        // static fields
+        private static readonly int[] __sizes = { 1, 2, 3, 4, 7, 8, 9, 15, 16, 17, 127, 128, 129, 8191, 8192 };
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(-1)]
+        [InlineData(-10)]
+        [InlineData(1024 * 1024 * 1024 + 1)]
+        public void TestBuffer_invalid_buffer_size_should_throw(int size)
```
rstam, line 36: Rename to `GetBuffer_should_throw_when_size_is_invalid`

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            var exception = Record.Exception(() => ThreadStaticBuffer.GetBuffer(size));
+            exception.Should().BeOfType<ArgumentOutOfRangeException>();
+        }
+
+        [Fact]
+        public void TestBufferSizeIncreasing_expected_powerof2()
```
rstam, line 43: Rename to something along the lines of `GetBuffer_should_resize_buffer_to_next_power_of_2_when_resizing_buffer`

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            ExecuteOnNewThread(1, _ =>
+            {
+                foreach (var requestedSize in __sizes)
+                {
+                    var expectedSize = Math.Max(16, 1 << (int)Math.Ceiling(Math.Log(requestedSize, 2)));
```
rstam, line 49: I try to avoid logic in tests, because of the risk that the logic in the test and in the subject under test will have the same bug (since I wrote them both).

Logic like this can be replace with explicit values using `InlineData`

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            });
+        }
+
+        [Theory]
+        [InlineData(16384)]
+        [InlineData(16385)]
+        [InlineData(16386)]
+        [InlineData(32767)]
+        [InlineData(32769)]
+        public void TestBufferSizeGreaterThanMaxSize_expected_exact_size(int requestedSize)
```
rstam, line 63: Rename `GetBuffer_should_return_expected_length_segment` or something like that

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            var buffer = ThreadStaticBuffer.GetBuffer(requestedSize);
+            buffer.Length.Should().Be(requestedSize);
+        }
+
+        [Fact]
+        public void TestBufferMultiThreaded_expected_unique_instance_per_thread()
```
rstam, line 70: Rename to `GetBuffer_should_return_different_instance_in_different_threads`

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                buffer[0] = (byte)i;
+                allBuffers.Add(buffer);
+            });
+
+            var buffersOrdered = allBuffers.OrderBy(b => b[0]).ToArray();
+            var buffersDistinct = buffersOrdered.Distinct().ToArray();
+
+            buffersOrdered.Length.Should().Be(threadsCount);
+            buffersOrdered.ShouldAllBeEquivalentTo(buffersDistinct);
```
rstam, line 99: I haven't figured out why you need to order the buffers for testing.

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,121 @@
...
+                thread.Start();
+
+                return thread;
+            }).ToArray();
+
+
+            foreach (var thread in threads)
+            {
+                if (!thread.Join(10000))
+                    throw new TimeoutException();
```
rstam, line 117: It is our practice to always use braces even for single line then clauses, like:

                if (!thread.Join(10000))
                {
                    throw new TimeoutException();
                }

rstam (changes requested):

JamesKovacs (commented):
Taking a step back from the detailed discussion...

If the problem is that in the hot path `_tempUtf8` buffer isn't sufficiently large enough and we often allocate temporary byte buffers > 128 bytes which must then be GC'd, have we considered the simpler solution of simply increasing the size of `_tempUtf8` buffer to a larger and more reasonable value?

Something else to consider... What is the lifetime of a `BsonStreamAdapter`? What I'm thinking is that if we need a large byte buffer, rather than allocating it for the one `ReadString` operation, we make the larger byte buffer the cached one:

```
var length = ReadInt32();
if (length > _tempUtf8.Length)
{
    _tempUtf8 = new byte[PowerOf2.RoundUpToPowerOf2(length)];
}
var bytes = _tempUtf8;
```

Just some ideas I wanted to throw out there for discussion.

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
```diff
@@ -457,19 +455,23 @@ public override void WriteCString(string value)
...
+                length = segment.Count;
+
+                if (Array.IndexOf<byte>(bytes, 0, 0, length) != -1)
                 {
                     throw new ArgumentException("A CString cannot contain null bytes.", "value");
                 }
-                length = bytes.Length;
             }
 
             _stream.Write(bytes, 0, length);
```
rstam, line 477: This is an example of the potential danger of using the `ThreadStaticBuffer`. We are passing a reference to the `ThreadStaticBuffer` to the `Write` method which for all we know *might* itself attempt to use the `ThreadStaticBuffer` for its own purposes. If it did, the `ThreadStaticBuffer` would get corrupted.

In order to use the `ThreadStaticBuffer` without corruption we MUST understand the FULL code path of any method we pass a reference to the `ThreadStaticBuffer` to ENSURE that only one part of the driver is using the `ThreadStaticBuffer` at the same time.

rstam (commented):
Overall I think this PR is looking OK.

Please read the warning in one of the comments about using the `ThreadStaticBuffer` without corruption.

rstam:
One other concern about using `ThreadStaticBuffer` has just come to mind.

It could  totally fall apart if a reference to the `ThreadStaticBuffer` is ever used in an async `Task` because execution of the `Task` *might* jump to a different thread later on.

BorisDog:
@rstam Interesting point re. switching threads in Tasks.
Yes, it's unsafe to use ThreadStatic when execution of a method is not guaranteed to occur on single thread. In context of Tasks, that would happen when a Task consists of multiple sub-tasks, in which case ThreadStatic should not be used. In case of a Task with no sub-tasks it's safe. Added a warning in ThreadStaticBuffer.

rstam:
> @rstam Interesting point re. switching threads in Tasks.
> Yes, it's unsafe to use ThreadStatic when execution of a method is not guaranteed to occur on single thread. In context of Tasks, that would happen when a Task consists of multiple sub-tasks, in which case ThreadStatic should not be used. In case of a Task with no sub-tasks it's safe. Added a warning in ThreadStaticBuffer.

The subtasks might not even be visible to the caller. Consider:

    var buffer = EncodingHelper.GetBytesUsingThreadStaticBuffer(...);
    await stream.WriteAsync(buffer);

If the buffer is big enough `WriteAsync` might have to write it in several passes, and subsequent passes might execute on a different thread.

So even though the two line snippet "looks" safe it is not.

We just have to be super careful how we use the `ThreadStaticBuffer`. It's a dangerous tool.

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
```diff
@@ -0,0 +1,115 @@
...
+                    bufferCurrent.Should().BeSameAs(buffer);
+                }
+
+                allBuffers.Add(buffer);
+            });
+
+            var buffersDistinct = allBuffers.Distinct().ToArray();
+
+            allBuffers.Count.Should().Be(threadsCount);
+            buffersDistinct.Length.Should().Be(threadsCount);
```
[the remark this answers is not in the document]
rstam: I understand. But it is redundant. The whole point of this test is to check that every thread gets its own buffer.

Line 111 is only testing the test... we already know that `allBuffers.Count == threadsCount`.

BorisDog:
@rstam Yes, if we have code like 
```
async Task F() {
  1. syncronious code...
  2. await T
}
```
`await T` is clearly a subtask.
I agree that we should always carefully examine our async code.

tests/MongoDB.Bson.Tests/IO/EncodingHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        public void GetBytesUsingThreadStaticBuffer_should_throw_when_value_is_null()
+        {
+            var exception = Record.Exception(() => EncodingHelper.GetBytesUsingThreadStaticBuffer(Encoding.ASCII, null));
+
+            var e = exception.Should().BeOfType<ArgumentNullException>().Subject;
+            e.ParamName.Should().Be("value");
+        }
+
+        [Fact]
+        public void GetBytesUsingThreadStaticBuffer_empty_string_should_return_default_buffer()
```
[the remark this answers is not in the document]
BorisDog: Done.

tests/MongoDB.Bson.Tests/IO/ThreadStaticBufferTests.cs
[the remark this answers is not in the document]
BorisDog: Done. Added assertion in ThreadingUtilities.

JamesKovacs (commented):
Overall I like the buffer caching technique and implementation. I see two main risks:

1. If we call `ThreadStaticBuffer.GetBuffer()` twice from the same thread in two different methods, then we will very likely get the same buffer and corrupt it. If the second buffer we happen to request is larger than the currently allocated buffer, then we would end up with two different buffers. While this wouldn't result in corruption because we by happenstance are using two different buffers, this would make debugging buffer corruption problems confusing.
2. `async` code is particularly dangerous. If we get a buffer, `await`, and then use the buffer, we will likely be executing on a different thread. The original thread might be scheduled to do other work that requires a buffer and we now have two threads accessing the same thread-static buffer.

While we can document these two limitations in the comments for `ThreadStaticBuffer`, there are no guarantees that our future selves will remember to read those comments. Are there ways to enforce these constraints in the code itself? Can we make `ThreadStaticBuffer` a private nested class to avoid misuse? (This wouldn't address the async problem though.) Should we consider a buffer pool impl instead? Other ideas?

~Another concern... Our max cached buffer size is configured to be 1GB. This means that under certain workloads that require large buffers, the memory usage of the thread static buffers will gradually increase to ThreadCount x 1GB. This will appear to be a memory leak, but will in fact simply be the thread static buffers slowly growing as each thread happens to require larger and larger buffers.~ Scratch that. Our maximum temporary allocation size is 1GB. Our maximum cached buffer size is 16KB. Maximum memory usage of the cached buffers would be at most ThreadCount x 16KB, which isn't concerning.

BorisDog:
Thanks @JamesKovacs, it's important to be extra careful with multithreaded programming .
Like with most other programming tools like Threads, TPL or synchronization primitives, we can not really use them until we understand them in full. I find that `ThreadStatic` feature technique is one of the simplest tools in multithreading tools arsenal, so we just need to get used to it. While we can always find scenarios where this tool is not applicable this should not prevent us from valid usages.

1. Yes, we should not call twice `ThreadStaticBuffer.GetBuffer()` in same thread in nested methods. Exactly like we should not use local cached buffer in same thread in nested methods.
2. Yes, as mentioned before, in some/most cases `ThreadStatic` is not suitable to use in tasks that spin off other tasks.

rstam:
> @JamesKovacs said: Are there ways to enforce these constraints in the code itself?

We could adopt a checkout/checkin protocol for the ThreadStaticBuffer

1. Attempting to check out the buffer twice would throw an exception
2. Attempting to check in a buffer from the wrong thread would throw an exception

JamesKovacs:
I like @rstam's suggested approach of throwing an exception if we violate our constraints. Let's implement that.

rstam (changes requested):
Please re-request our review after you have added the check-out/check-in protocol for the ThreadStaticBuffer.

rstam:
I'm not sure I full know the reasons why, but we avoid force pushing to PRs.

Was the force push necessary? 

Are we wrong to be avoiding force pushes?

BorisDog:
@rstam I had to rebase (to use  newer C#) and could only force push to feature branch.
Maybe there is a more graceful way?

src/MongoDB.Bson/IO/BsonStreamAdapter.cs
```diff
@@ -454,26 +454,33 @@ public override void WriteCString(string value)
...
                     throw new ArgumentException("A CString cannot contain null bytes.", "value");
                 }
-                length = bytes.Length;
+
+                SetBytes(segment.Array, segment.Count);
             }
 
-            _stream.Write(bytes, 0, length);
-            _stream.WriteByte(0);
+            void SetBytes(byte[] bytes, int length)
```
rstam, line 479: I like this local helper method but I think you should rename it to `WriteBytes`.

src/MongoDB.Bson/IO/ByteBufferStream.cs
```diff
@@ -568,24 +569,35 @@ public override void WriteCString(string value)
...
-                byte[] bytes;
-                if (maxLength <= _tempUtf8.Length)
+                // Compare to 128 to preserve original behavior
+                const int maxLengthToUseCStringUtf8EncodingWith = 128;
+          
+                if (maxLength <= maxLengthToUseCStringUtf8EncodingWith)
                 {
-                    bytes = _tempUtf8;
-                    actualLength = CStringUtf8Encoding.GetBytes(value, bytes, 0, Utf8Encodings.Strict);
+                    using var acquiredBuffer = ThreadStaticBuffer.GetBuffer(maxLengthToUseCStringUtf8EncodingWith);
```
rstam, line 577: Can you rename this variable to `rentedBuffer` for consistency?

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Represents a class that provides reusable buffer per thread.
+    /// Use this technique ONLY when:
+    ///     1. Buffer is not shared across multiple threads.
+    ///     2. No nested methods invocations use the same buffer.
+    /// Advised to limit the usage scope to a single method.
+    /// </summary>
+    internal static class ThreadStaticBuffer
+    {
+        public struct RentableBuffer : IDisposable
```
rstam, line 29: I like this helper class. Can we rename to `RentedBuffer`?

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            public byte[] Buffer { get; private set; }
+
+            public void Dispose()
+            {
+                if (!__isBufferInUse)
+                    throw new InvalidOperationException("Thread static buffer is not in use");
```
rstam, line 36: Note: we always use braces for `if` even if it's only one line.

rstam: This checks that we are actually using the buffer that we are trying to return.

We also need to check whether we are trying to return the buffer from the wrong thread.

See suggested change below.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                __isBufferInUse = false;
+            }
+
+            public static implicit operator RentableBuffer(byte[] buffer) =>
+                new RentableBuffer() { Buffer = buffer };
+        }
```
rstam, line 43: Requested refactoring for this class:

    public struct RentedBuffer : IDisposable
    {
        private readonly byte[] _buffer;
        private readonly byte[] _bytes; // usually the same as _buffer unless the request was too big

        public RentedBuffer(byte[] buffer, byte[] bytes)
        {
            _buffer = buffer;
            _bytes = bytes;
        }

        public void Dispose()
        {
            if (!__isBufferRented)
            {
                throw new InvalidOperationException("Thread static buffer is not in use");
            }
            if (_buffer != __buffer)
            {
                throw new InvalidOperationException("Attempt to return thread static buffer from the wrong thread.");
            }

            __isBufferRented = false;
        }

        public byte[] Bytes => _bytes;
    }

rstam: Notes:

We need to track both the `_buffer` *and* the `_bytes` because sometimes `RentBuffer` returns a newly allocated buffer that is *not* the thread static buffer.

We usually use explicit backing fields instead of auto properties.

I renamed the public property from `Buffer` to `Bytes` because it's not always the buffer.

Auto properties would be OK for very simple classes (this one almost qualifies), but personally I prefer explicit properties (with the leading underscore naming convention) for consistency.

What we should never do is mix explicit backing fields and auto properties in the same class (which of course you were not doing here, just making a comment for the future).

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+        private const int MinSize = 256;
+        private const int MaxSize = 16384;
+        private const int MaxAllocationSize = 1024 * 1024 * 1024; // 1GB
+
+        // private static fields
+        [ThreadStatic]
+        private static byte[] __buffer;
+
+        [ThreadStatic]
+        private static bool __isBufferInUse;
```
rstam, line 54: For consistency rename to `__isBufferRented`

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public static RentableBuffer GetBuffer(int size)
```
rstam, line 56: For consistency rename to `RentBuffer`

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+            if (size <= 0 || size > MaxAllocationSize)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), "Invalid requested buffer size");
+            }
+
+            __isBufferInUse = true;
+
+            if (size > MaxSize)
+            {
+                return new byte[size];
```
rstam, line 72: With suggested refactoring of `RentedBuffer` class this line would become:

    return new RentedBuffer(__buffer, new byte[size]);

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            if (__buffer == null || __buffer.Length < size)
+            {
+                var newSize = size <= MinSize ? MinSize : PowerOf2.RoundUpToPowerOf2(size);
+                __buffer = new byte[newSize];
+            }
+
+            return __buffer;
```
rstam, line 81: With suggested refactoring of `RentedBuffer` class this line would become:

    return new RentedBuffer(__buffer, __buffer);

rstam (changes requested):
Looks good with some relatively minor requested changes.

src/MongoDB.Bson/IO/EncodingHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Text;
+
+namespace MongoDB.Bson.IO
+{
+    /// <summary>
+    /// Represents a class that has some helper methods for <see cref="Encoding"/>.
+    /// </summary>
+    internal static class EncodingHelper
+    {
+        public struct DisposableSegment : IDisposable
```
rstam, line 26: I recommend that types that implement `IDisposable` be a `class`, not a `struct`.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// Advised to limit the usage scope to a single method.
+    /// </summary>
+    internal static class ThreadStaticBuffer
+    {
+        public interface IRentedBuffer : IDisposable
+        {
+            public byte[] Bytes { get; }
+        }
+
+        private struct RentedBuffer : IRentedBuffer
```
rstam, line 34: I recommend that types that implement `IDisposable` be a `class`, not a `struct`.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+        // private static fields
+        [ThreadStatic]
+        private static byte[] __buffer;
+
+        [ThreadStatic]
+        private static bool __isBufferRented;
+
+        private static object ThreadIdentifier => __buffer ?? (__buffer = new byte[0]);
+
+        public static IRentedBuffer GetBuffer(int size)
```
rstam, line 76: Looks like you missed renaming this to `RentBuffer`.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+    ///     1. Buffer is not shared across multiple threads.
+    ///     2. No nested methods invocations use the same buffer.
+    /// Advised to limit the usage scope to a single method.
+    /// </summary>
+    internal static class ThreadStaticBuffer
+    {
+        public interface IRentedBuffer : IDisposable
+        {
+            public byte[] Bytes { get; }
+        }
```
rstam, line 32: I understand what you are doing here by using an interface, but it may be a unnecessary. In part because the containing class is already `internal`, and also because the interface isn't really hiding anything (if you make the `RentedBuffer` constructor private then the `RentedBuffer` class has exactly the same public members as the proposed interface).

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+        private const int MaxAllocationSize = 1024 * 1024 * 1024; // 1GB
+
+        // private static fields
+        [ThreadStatic]
+        private static byte[] __buffer;
+
+        [ThreadStatic]
+        private static bool __isBufferRented;
+
+        private static object ThreadIdentifier => __buffer ?? (__buffer = new byte[0]);
```
rstam, line 74: I understand what you are doing here, but the implication of something called a `ThreadIdentifier` is that the thread would have a stable identifier, and using the `__buffer` as a pseudo-identifier is confusing because it can change over time.

The originally proposed check

    if (_buffer != __buffer)

made intuitive sense to me... is the rented buffer we are returning the right one for the current thread?

I assume you went down this path because you realized that it was theoretically possible for `__buffer` to still be null if all the initial calls to `RentBuffer` are for sizes that are so big they result in the allocation of a temporary buffer. One approach would be to ignore that, it's unlikely and it's temporary. Another approach would be to simply ensure that `__buffer` is not null in the `RentBuffer` method.

I realize that your code in fact ensures that `__buffer` is never null... but the fact that it does so as a hidden side effect of a property called `ThreadIdentifier` is rather surprising.

If you prefer comparing thread identifiers to verifying that the buffer being returned matches the thread static buffer, maybe it would be clearer to use a `Guid` as the `ThreadIdentifier` so that it behaves like you expect an identifier to behave, in the sense that it is stable over time.

rstam (changes requested):

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Unfortunately we can't prevent creation of structs with private constructors.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Thanks, it's a good idea to make this owner-returner thread check clear and readable. Initially I thought taking the `__buffer` shortcut would do, but looks like we should take more traditional approach.
Refactored using `Thread.ManagedThreadId` property.

src/MongoDB.Bson/IO/EncodingHelper.cs
BorisDog: Addressed this in `RentedBuffer `comment.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Thanks, done

src/MongoDB.Bson/IO/EncodingHelper.cs
rstam: I recommend (in general) that types that implement `IDisposable` be a `class`, not a `struct`.

This was discussed in Slack.

I completely understand and endorse the desire to use a `struct` to avoid a memory allocation. A `struct` that implements `IDisposable` can sometime behave in very unpredictable ways because value types can result in multiple copies of the `struct`.

In this particular case it appears to be safe, so I think it's OK to use `struct` if that's what you prefer.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
rstam: See previous comments (and Slack discussion) for reasons why combining `struct` with `IDisposable` can be problematic.

In this case using a `struct` has **not** prevented a memory allocation, because the return value of `RentBuffer` is the interface `IRentedBuffer` and the `struct` must be boxed to convert it to the interface type.

To avoid boxing you would have to get rid of the `IRentedBuffer` interface and return the `RentedBuffer` `struct` directly.

rstam (changes requested):

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Good point!
Removed `IRentedBuffer` interface.

Another approach would be to keep the interface, and hold internally same instance of `RenterBuffer `**class**, which would be disposed and reset each time. Not sure what is better, allowing creation of `RenterBuffer` internally outside of `ThreadStaticBuffer`, on disposing the same object repeatedly.

src/MongoDB.Bson/IO/EncodingHelper.cs
rstam: I like your idea of using `readonly struct` for `RentedBuffer`.

Maybe make this `struct` readonly also?

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+        private const int MaxSize = 16384;
+        private const int MaxAllocationSize = 1024 * 1024 * 1024; // 1GB
+
+        // private static fields
+        [ThreadStatic]
+        private static byte[] __buffer;
+        [ThreadStatic]
+        private static bool __isBufferRented;
+        [ThreadStatic]
+        private static int _threadId;
```
rstam, line 69: Use two leading underscores for static field names:

    __threadId

rstam (changes requested):

rstam:
@JamesKovacs when you review this again can you review the tests carefully also? I focused on the driver changes.

src/MongoDB.Bson/IO/EncodingHelper.cs
BorisDog: Done

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Tnx, done.

tests/MongoDB.Bson.TestHelpers/MongoDB.Bson.TestHelpers.csproj
```diff
@@ -48,4 +48,10 @@
     <ProjectReference Include="..\..\src\MongoDB.Bson\MongoDB.Bson.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
+    <PackageReference Include="System.Threading.Thread">
+      <Version>4.3.0</Version>
+    </PackageReference>
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
BorisDog: Right, since this was added this in `MongoDB.Bson` in recent commit.

tests/MongoDB.Bson.TestHelpers/ThreadingUtilities.cs
```diff
@@ -0,0 +1,69 @@
...
+            var exceptions = new ConcurrentBag<Exception>();
+
+            var threads = Enumerable.Range(0, threadsCount).Select(i =>
+            {
+                var thread = new Thread(_ =>
+                {
+                    try
+                    {
+                        action(i);
+                        actionsExecutedCount++;
```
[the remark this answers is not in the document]
BorisDog: Good catch, done.

rstam (approved):
LGTM

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+            public void Dispose()
+            {
+                if (_ownerThreadId != __threadId)
+                {
+                    throw new InvalidOperationException("Attempt to return thread static buffer from the wrong thread.");
+                }
+
+                if (!__isBufferRented)
+                {
+                    throw new InvalidOperationException("Thread static buffer is not in use.");
```
JamesKovacs, line 50: This should only happen if we attempt to dispose of the `RentedBuffer` twice. Rephrase this to "Attempt to return a thread static buffer that is not in use."

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -0,0 +1,104 @@
...
+        public static RentedBuffer RentBuffer(int size)
+        {
+            if (__isBufferRented)
+            {
+                throw new InvalidOperationException("Thread static buffer is already in use.");
+            }
+
+            if (size <= 0 || size > MaxAllocationSize)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size), "Invalid requested buffer size.");
```
JamesKovacs, line 80: `$"Requested buffer size of {size} bytes. Maximum buffer size is {MaxAllocationSize} bytes."`

tests/MongoDB.Bson.TestHelpers/ThreadingUtilities.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
+using FluentAssertions;
+
+namespace MongoDB.Bson.TestHelpers
+{
+    public static class ThreadingUtilities
+    {
+        public static void ExecuteOnNewThread(int threadsCount, Action<int> action, int timeoutMilliseconds = 10000)
```
JamesKovacs, line 26: Since your first argument is `threadsCount`, rename the method to `ExecuteOnNewThreads` (plural).

JamesKovacs (changes requested):
LGTM. Requesting a few minor changes to clarify method naming and error message text.

tests/MongoDB.Bson.TestHelpers/ThreadingUtilities.cs
BorisDog: Done

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
BorisDog: Done

JamesKovacs (approved):
LGTM

BorisDog:
Squashed and merged

PR #430, opened 2020-12-30
CSHARP-3297: Handle incomplete regex patterns in JsonScanner

rkyser:


rkyser:
Link to MongoDB Jira issue: https://jira.mongodb.org/projects/CSHARP/issues/CSHARP-3297

tests/MongoDB.Bson.Tests/IO/JsonScannerTests.cs
```diff
@@ -405,6 +407,32 @@ public void TestMinusZeroExponentMinusTwelve()
...
+
+            // The completedTask should be the getNextTokenTask
+            Assert.Equal(getNextTokenTask, completedTask);
+
+            // Unwrap the FormatException from completedTask
+            var formatException = await Assert.ThrowsAsync<FormatException>(() => completedTask).ConfigureAwait(false);
+
+            Assert.StartsWith("Invalid JSON regular expression", formatException.Message);
+        }
+
```
rstam, line 435: I would name the test method something more specific than "invalid regular expression".

I think the test should be simplified to not have to deal with detecting infinite loops (now that the bug is fixed there is no infinite loop).

The test can be simply:

    [Theory]
    [InlineData("/")]
    [InlineData("/pattern")]
    [InlineData("/pattern\\/")]
    public void TestMissingClosingSlash(string jsonRegex)
    {
        var buffer = new JsonBuffer(jsonRegex);

        var exception = Record.Exception(() => JsonScanner.GetNextToken(buffer));

        exception.Should().BeOfType<FormatException>();
        exception.Message.Should().StartWith("Invalid JSON regular expression");
    }

rstam: Note: if we don't want to use FluentAssertions in a file that was first written before we started using FluentAssertions the last two lines can be rewritten using xUnit's `Assert`.

rstam (changes requested):
Good catch!

tests/MongoDB.Bson.Tests/IO/JsonScannerTests.cs
rkyser: Thank you for the review @rstam. I'll get these changes pushed in the next couple days.

rkyser: Changes have been pushed.

I opted to use the xUnit assertions to be consistent with the other tests in `JsonScannerTests`.

rkyser:
@rstam Is there anything else you need from me? If so, let me know.

rstam (approved):
LGTM

Needs a second reviewer

tests/MongoDB.Bson.Tests/IO/JsonScannerTests.cs
```diff
@@ -13,6 +13,8 @@
 * limitations under the License.
 */
 
+using System;
+using System.Threading.Tasks;
```
BorisDog, line 17: minor: looks like there is no need for `System.Threading.Tasks`

BorisDog (changes requested):

tests/MongoDB.Bson.Tests/IO/JsonScannerTests.cs
rkyser: Good catch--change pushed.

BorisDog:
LGTM
thanks!

BorisDog (approved):

rstam:
Squashed, rebased and merged to master.

PR #433, opened 2021-01-06
CSHARP-2450 (second attempt): Performance: Reduced lock contention in BsonSerializer.LookupActualType

dnickless:
Fixes https://jira.mongodb.org/browse/CSHARP-2450.

Kindly review my individual check-ins as all 3 check-ins affect the same file and will be confusing in a single comparison.

The most recent check-in contains the most important change: This is where I trade a tiny little bit of increasing complexity (cloning a small HashSet) on the way less likely write-scenario for a substantially better performing read-scenario.

PR #434, opened 2021-01-06
Performance/(de-)serialization: Added two caches to avoid reflection calls

dnickless:
`TypeNameDiscriminator` is a static helper class that kicks in during serialization and deserialization when using type names as discriminators. There currently is quite some reflection and string-fiddling inside two crucial methods: 

`public static Type GetActualType(string typeName)`

and

`public static string GetDiscriminator(Type type)`.

This simple PR introduces two basic caches that allow for concurrent lock-free reads which appears to be a healthy choice given the low change frequency of either of the caches.

### DIFF
--- a/src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
+++ b/src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


JamesKovacs (changes requested):
Now that PR #482 has been merged into `master`, this PR should be rebased on top of it. Perf tests before and after the addition of the two caches should be made to demonstrate the performance gains from adding the two caches.

You may also want to consider evaluating the shortening of the read lock in `LookupActualType` from PR #433. Perf testing in the context of PR #482 didn't show a measurable improvement, which is why it was not included. We would be happy to consider it if additional perf tests demonstrate an improvement.

In perf testing, we recommend focusing on deserializing a simple class consisting only of an `ObjectId` and no other properties to stress the type lookup portion of the deserialization path as opposed to parsing the BSON and creating child objects. This will give us a best-case improvement scenario where we don't spend any time on-the-wire and minimize time in other areas of the deserialization code.

Please let us know if you have any questions. We look forward to reviewing the results of these further performance enhancements. And thank you once again for the collaboration.

JamesKovacs:
I've forked this work to a new JIRA ticket, https://jira.mongodb.org/browse/CSHARP-3513.

adelinowona:
Hi @dnickless, thanks for this PR. It has been a while since we got any feedback from you regarding the possible performance improvements in this PR so we'll close the PR and you are free to open a new PR if you choose to renew this work.

PR #435, opened 2021-01-13
CSHARP-3302: SDAM deadlock when invalidating former primary

BorisDog:
CSHARP3302 Addressing two issues:
1. Deadlock during heartbeat:  lock eliminated at ServerMonitor.RequestHeartbeat.
2. Eliminated concurrent Cluster.RapidHeartbeatTimerCallback invocations.

[EG](https://evergreen.mongodb.com/version/5fff23a1d1fe070b37068d4f)

Detailed deadlock flow described in [Jira ticket](https://jira.mongodb.org/browse/CSHARP-3302)
This change allows invalidating (Server.Invalidate) servers in MultiServerCluster.ProcessReplicaSetChange without acquiring ServerMonitor.Lock which might be already held by other thread.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -76,7 +76,7 @@ namespace MongoDB.Driver.Core.Clusters
         private readonly ICoreServerSessionPool _serverSessionPool;
         private readonly ClusterSettings _settings;
         private readonly InterlockedInt32 _state;
-        private readonly InterlockedInt32 _heartbeatState;
+        private readonly InterlockedInt32 _rapidHeartbeatTimerCallbackState;
 
         private readonly Action<ClusterDescriptionChangedEvent> _descriptionChangedEventHandler;
         private readonly Action<ClusterSelectingServerEvent> _selectingServerEventHandler;
@@ -90,7 +90,7 @@ namespace MongoDB.Driver.Core.Clusters
             _serverFactory = Ensure.IsNotNull(serverFactory, nameof(serverFactory));
             Ensure.IsNotNull(eventSubscriber, nameof(eventSubscriber));
             _state = new InterlockedInt32(State.Initial);
-            _heartbeatState = new InterlockedInt32(State.Initial);
+            _rapidHeartbeatTimerCallbackState = new InterlockedInt32(RapidHeartbeatTimerCallbackState.Idle);
 
             _clusterId = new ClusterId();
             _description = CreateInitialDescription();
@@ -230,25 +230,23 @@ namespace MongoDB.Driver.Core.Clusters
 
         private void RapidHeartbeatTimerCallback(object args)
         {
-            if (!_heartbeatState.TryChange(State.Initial, State.Open))
+            // Avoid requesting heartbeat concurrently
+            if (_rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.Idle, RapidHeartbeatTimerCallbackState.Running))
             {
-                // Avoid requesting heartbeat concurrently
-                return;
-            }
-
-            try
-            {
-                RequestHeartbeat();
-            }
-            catch
-            {
-                // TODO: Trace this
-                // If we don't protect this call, we could
-                // take down the app domain.
-            }
-            finally
-            {
-                _heartbeatState.TryChange(State.Initial);
+                try
+                {
+                    RequestHeartbeat();
+                }
+                catch
+                {
+                    // TODO: Trace this
+                    // If we don't protect this call, we could
+                    // take down the app domain.
+                }
+                finally
+                {
+                    _rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.Idle);
+                }
             }
         }
 
@@ -627,6 +625,12 @@ namespace MongoDB.Driver.Core.Clusters
             public const int Initial = 0;
             public const int Open = 1;
             public const int Disposed = 2;
+        }
+
+        private static class RapidHeartbeatTimerCallbackState
+        {
+            public const int Idle = 0;
+            public const int Running = 1;
         }
     }
 }
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -232,7 +232,7 @@ namespace MongoDB.Driver.Core.Clusters
         {
             if (!_heartbeatState.TryChange(State.Initial, State.Open))
             {
-                // Avoid requesting hearbeat concurrently
+                // Avoid requesting heartbeat concurrently
                 return;
             }
 
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -75,6 +75,7 @@ namespace MongoDB.Driver.Core.Clusters
         private readonly ICoreServerSessionPool _serverSessionPool;
         private readonly ClusterSettings _settings;
         private readonly InterlockedInt32 _state;
+        private readonly InterlockedInt32 _heartbeatState;
 
         private readonly Action<ClusterDescriptionChangedEvent> _descriptionChangedEventHandler;
         private readonly Action<ClusterSelectingServerEvent> _selectingServerEventHandler;
@@ -88,6 +89,7 @@ namespace MongoDB.Driver.Core.Clusters
             _serverFactory = Ensure.IsNotNull(serverFactory, nameof(serverFactory));
             Ensure.IsNotNull(eventSubscriber, nameof(eventSubscriber));
             _state = new InterlockedInt32(State.Initial);
+            _heartbeatState = new InterlockedInt32(State.Initial);
 
             _clusterId = new ClusterId();
             _description = CreateInitialDescription();
@@ -227,6 +229,12 @@ namespace MongoDB.Driver.Core.Clusters
 
         private void RapidHeartbeatTimerCallback(object args)
         {
+            if (!_heartbeatState.TryChange(State.Initial, State.Open))
+            {
+                // Avoid requesting hearbeat concurrently
+                return;
+            }
+
             try
             {
                 RequestHeartbeat();
@@ -236,6 +244,10 @@ namespace MongoDB.Driver.Core.Clusters
                 // TODO: Trace this
                 // If we don't protect this call, we could
                 // take down the app domain.
+            }
+            finally
+            {
+                _heartbeatState.TryChange(State.Initial);
             }
         }
 
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -90,7 +90,7 @@ namespace MongoDB.Driver.Core.Clusters
             _serverFactory = Ensure.IsNotNull(serverFactory, nameof(serverFactory));
             Ensure.IsNotNull(eventSubscriber, nameof(eventSubscriber));
             _state = new InterlockedInt32(State.Initial);
-            _rapidHeartbeatTimerCallbackState = new InterlockedInt32(RapidHeartbeatTimerCallbackState.Idle);
+            _rapidHeartbeatTimerCallbackState = new InterlockedInt32(RapidHeartbeatTimerCallbackState.NotRunning);
 
             _clusterId = new ClusterId();
             _description = CreateInitialDescription();
@@ -230,8 +230,8 @@ namespace MongoDB.Driver.Core.Clusters
 
         private void RapidHeartbeatTimerCallback(object args)
         {
-            // Avoid requesting heartbeat concurrently
-            if (_rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.Idle, RapidHeartbeatTimerCallbackState.Running))
+            // avoid requesting heartbeat reentrantly
+            if (_rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.NotRunning, RapidHeartbeatTimerCallbackState.Running))
             {
                 try
                 {
@@ -245,7 +245,7 @@ namespace MongoDB.Driver.Core.Clusters
                 }
                 finally
                 {
-                    _rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.Idle);
+                    _rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.NotRunning);
                 }
             }
         }
@@ -629,7 +629,7 @@ namespace MongoDB.Driver.Core.Clusters
 
         private static class RapidHeartbeatTimerCallbackState
         {
-            public const int Idle = 0;
+            public const int NotRunning = 0;
             public const int Running = 1;
         }
     }
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -37,7 +37,7 @@ namespace MongoDB.Driver.Core.Clusters
     {
         #region static
         // static fields
-        private static readonly TimeSpan __minHeartbeatInterval = TimeSpan.FromMilliseconds(500);
+        private static readonly TimeSpan __minHeartbeatIntervalDefault = TimeSpan.FromMilliseconds(500);
         private static readonly SemanticVersion __minSupportedServerVersion = new SemanticVersion(2, 6, 0);
         private static readonly IServerSelector __randomServerSelector = new RandomServerSelector();
         private static readonly Range<int> __supportedWireVersionRange = new Range<int>(2, 9);
@@ -61,6 +61,7 @@ namespace MongoDB.Driver.Core.Clusters
         #endregion
 
         // fields
+        private readonly TimeSpan _minHeartbeatInterval = __minHeartbeatIntervalDefault;
         private readonly IClusterClock _clusterClock = new ClusterClock();
         private readonly ClusterId _clusterId;
         private CryptClient _cryptClient = null;
@@ -199,7 +200,7 @@ namespace MongoDB.Driver.Core.Clusters
 
                 if (++_serverSelectionWaitQueueSize == 1)
                 {
-                    _rapidHeartbeatTimer.Change(TimeSpan.Zero, __minHeartbeatInterval);
+                    _rapidHeartbeatTimer.Change(TimeSpan.Zero, _minHeartbeatInterval);
                 }
             }
         }
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -144,7 +144,8 @@ namespace MongoDB.Driver.Core.Servers
         {
             ThrowIfNotOpen();
 
-            // _heartbeatDelay is created under _lock and is safe for concurrent access outside _lock
+            // CSHARP-3302: Accessing _heartbeatDelay inside _lock can lead to deadlock when processing concurrent heartbeats from old and new primaries.
+            // Accessing _heartbeatDelay outside of _lock avoids the deadlock and will at worst reference the previous delay
             _heartbeatDelay?.RequestHeartbeat();
         }
 
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -143,6 +143,8 @@ namespace MongoDB.Driver.Core.Servers
         public void RequestHeartbeat()
         {
             ThrowIfNotOpen();
+
+            // _heartbeatDelay is created under _lock and is safe for concurrent access outside _lock
             _heartbeatDelay?.RequestHeartbeat();
         }
 
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -143,10 +143,7 @@ namespace MongoDB.Driver.Core.Servers
         public void RequestHeartbeat()
         {
             ThrowIfNotOpen();
-            lock (_lock)
-            {
-                _heartbeatDelay?.RequestHeartbeat();
-            }
+            _heartbeatDelay?.RequestHeartbeat();
         }
 
         // private methods
--- a/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
@@ -619,5 +619,11 @@ namespace MongoDB.Driver.Core.Clusters
     internal static class ClusterReflector
     {
         public static InterlockedInt32 _state(this Cluster cluster) => (InterlockedInt32)Reflector.GetFieldValue(cluster, nameof(_state));
+
+        public static TimeSpan __minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(__minHeartbeatInterval), BindingFlags.Static | BindingFlags.NonPublic);
+        public static void __minHeartbeatIntervalSet(this Cluster cluster, TimeSpan timeSpan)
+        {
+            Reflector.SetFieldValue(cluster, nameof(__minHeartbeatInterval), timeSpan, BindingFlags.Static | BindingFlags.NonPublic);
+        }
     }
 }
--- a/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
@@ -620,10 +620,10 @@ namespace MongoDB.Driver.Core.Clusters
     {
         public static InterlockedInt32 _state(this Cluster cluster) => (InterlockedInt32)Reflector.GetFieldValue(cluster, nameof(_state));
 
-        public static TimeSpan __minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(__minHeartbeatInterval), BindingFlags.Static | BindingFlags.NonPublic);
-        public static void __minHeartbeatIntervalSet(this Cluster cluster, TimeSpan timeSpan)
+        public static TimeSpan _minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(_minHeartbeatInterval));
+        public static void _minHeartbeatIntervalSet(this Cluster cluster, TimeSpan timeSpan)
         {
-            Reflector.SetFieldValue(cluster, nameof(__minHeartbeatInterval), timeSpan, BindingFlags.Static | BindingFlags.NonPublic);
+            Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);
         }
     }
 }
--- a/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
@@ -621,9 +621,6 @@ namespace MongoDB.Driver.Core.Clusters
         public static InterlockedInt32 _state(this Cluster cluster) => (InterlockedInt32)Reflector.GetFieldValue(cluster, nameof(_state));
 
         public static TimeSpan _minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(_minHeartbeatInterval));
-        public static void _minHeartbeatIntervalSet(this Cluster cluster, TimeSpan timeSpan)
-        {
-            Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);
-        }
+        public static void _minHeartbeatInterval(this Cluster cluster, TimeSpan timeSpan) => Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);
     }
 }
--- a/tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
@@ -82,24 +82,17 @@ namespace MongoDB.Driver.Core.Tests.Jira
                 }
             }
 
-            var serverDescriptionDisconnected = new ServerDescription(
+            var serverDescription = new ServerDescription(
                 __serverId1,
                 __endPoint1,
                 type: ServerType.ReplicaSetPrimary,
                 state: ServerState.Disconnected,
                 replicaSetConfig: new ReplicaSetConfig(new[] { __endPoint1 }, "rs", __endPoint1, null));
 
-            var serverDescriptionConnected = new ServerDescription(
-              __serverId1,
-              __endPoint1,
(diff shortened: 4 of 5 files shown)


src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -61,6 +61,7 @@ internal abstract class Cluster : ICluster
         #endregion
 
         // fields
+        private readonly TimeSpan _minHeartbeatInterval = __minHeartbeatIntervalDefault;
```
BorisDog, line 64: This change is to enable reflection in unittests

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -75,6 +76,7 @@ internal abstract class Cluster : ICluster
         private readonly ICoreServerSessionPool _serverSessionPool;
         private readonly ClusterSettings _settings;
         private readonly InterlockedInt32 _state;
+        private readonly InterlockedInt32 _heartbeatState;
```
rstam, line 79: This name isn't quite accurate. The intent of this field is to protect against reentrant calls of the `RapidHeartbeatTimerCallback`. 

I recommend more precise names:

        private readonly InterlockedInt32 _rapidHeartbeatTimerCallbackState;

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -88,6 +90,7 @@ protected Cluster(ClusterSettings settings, IClusterableServerFactory serverFact
             _serverFactory = Ensure.IsNotNull(serverFactory, nameof(serverFactory));
             Ensure.IsNotNull(eventSubscriber, nameof(eventSubscriber));
             _state = new InterlockedInt32(State.Initial);
+            _heartbeatState = new InterlockedInt32(State.Initial);
```
rstam, line 93: Let's not try to reuse the existing `State` values (which are meant for a different purpose).

I propose:

            _rapidHeartbeatTimerCallbackState = new InterlockedInt32(RapidHeartbeatTimerCallbackState.NotRunning);

where the purpose specific new state class is:

        private static class RapidHeartbeatTimerCallbackState
        {
            public const int NotRunning = 0;
            public const int Running = 1;
        }

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -237,6 +246,10 @@ private void RapidHeartbeatTimerCallback(object args)
                 // If we don't protect this call, we could
                 // take down the app domain.
             }
+            finally
+            {
+                _heartbeatState.TryChange(State.Initial);
+            }
```
rstam, line 252: What you have above seems completely correct. 

I have a preference for refactoring as:

        private void RapidHeartbeatTimerCallback(object args)
        {
            // avoid running callback reentrantly
            if (_rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.NotRunning, RapidHeartbeatTimerCallbackState.Running))
            {
                try
                {
                    RequestHeartbeat();
                }
                catch
                {
                    // TODO: Trace this
                    // If we don't protect this call, we could
                    // take down the app domain.
                }
                finally
                {
                    _rapidHeartbeatTimerCallbackState.TryChange(RapidHeartbeatTimerCallbackState.NotRunning);
                }
            }
        }

but if you don't like the proposed refactoring you can keep the current version (but do apply the renamings).

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
```diff
@@ -619,5 +619,11 @@ private void AddOrRemoveServers(ClusterDescription clusterDescription)
     internal static class ClusterReflector
     {
         public static InterlockedInt32 _state(this Cluster cluster) => (InterlockedInt32)Reflector.GetFieldValue(cluster, nameof(_state));
+
+        public static TimeSpan _minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(_minHeartbeatInterval));
+        public static void _minHeartbeatIntervalSet(this Cluster cluster, TimeSpan timeSpan)
+        {
+            Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);
+        }
```
rstam, line 627: We don't usually tack `Set` on to the end of this reflection helper method. We use the presence or absence of a value parameter to distinguish between the two overloads.

Also, `=>` works with `void` methods also, so this could be refactored as:

        public static void _minHeartbeatInterval(this Cluster cluster, TimeSpan timeSpan) => Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);

rstam (changes requested):
Upon initial review this looks very reasonable.

I will review the tests later.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
BorisDog: c

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
Ideally we would use InterlockedEnum<T> and enums.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
BorisDog: Done

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
BorisDog: Done. Good point, not clear why "Set" was used.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,336 @@
...
+            }
+
+            var serverDescriptionDisconnected = new ServerDescription(
+                __serverId1,
+                __endPoint1,
+                type: ServerType.ReplicaSetPrimary,
+                state: ServerState.Disconnected,
+                replicaSetConfig: new ReplicaSetConfig(new[] { __endPoint1 }, "rs", __endPoint1, null));
+
+            var serverDescriptionConnected = new ServerDescription(
```
BorisDog, line 92: Not used, will be removed.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -227,15 +230,23 @@ public virtual void Initialize()
 
         private void RapidHeartbeatTimerCallback(object args)
         {
-            try
+            // Avoid requesting heartbeat concurrently
```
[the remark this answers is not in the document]
rstam: nit: we don't usually capitalize the first word in a comment (unless it's a doc comment that's input to the generator for the API documentation).

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+#pragma warning restore CS0618 // Type or member is obsolete
+        private readonly static ClusterId __clusterId = new ClusterId();
+        private readonly static bool? __directConnection = null;
+        private readonly static EndPoint __endPoint1 = new DnsEndPoint("localhost", 27017);
+        private readonly static EndPoint __endPoint2 = new DnsEndPoint("localhost", 27018);
+        private readonly static TimeSpan __heartbeatInterval = TimeSpan.FromMilliseconds(200);
+        private readonly static ServerId __serverId1 = new ServerId(__clusterId, __endPoint1);
+        private readonly static ServerId __serverId2 = new ServerId(__clusterId, __endPoint2);
+
+        private static HashSet<ServerId> s_primaries = new HashSet<ServerId>();
```
DmitryLukyanov, line 53: we don't use abbreviation anymore. Also, static fields are prefixed by `__`

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -143,10 +143,9 @@ public void Initialize()
         public void RequestHeartbeat()
         {
             ThrowIfNotOpen();
-            lock (_lock)
-            {
-                _heartbeatDelay?.RequestHeartbeat();
-            }
+
+            // _heartbeatDelay is created under _lock and is safe for concurrent access outside _lock
+            _heartbeatDelay?.RequestHeartbeat();
```
rstam, line 148: @DmitryLukyanov we had a discussion in Slack where we determined that it IS safe to reference the `_heartbeatDelay` variable WITHOUT a `lock` even though we use a `lock` when setting it.

The conclusion is that at worst we would read the previous value rather than the current value of `_heartbeatDelay`.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+
+                // Wait for all heartbeats to complete
+                await Task.WhenAny(allHeartbeatsRecieved.Task, Task.Delay(1000));
+            }
+
+            allHeartbeatsRecieved.Task.Status.Should().Be(TaskStatus.RanToCompletion);
+            calledConcurrently.Should().Be(false);
+        }
+
+        [Fact(Timeout = 10000)]
```
DmitryLukyanov, line 123: This is quite foggy argument which looks like not recommended anymore: https://github.com/xunit/xunit/issues/217#issuecomment-65907484 
It looks like it still work for some cases (like here), but it doesn't work for sync code paths and for example a case like this:

      [Fact(Timeout = 3000)]
        public  Task Should_convert_from_InsertRequest_to_BsonDocument()
        {
            while (true) ;
        }

Apparently, it works here, so pointing here just to pay attention on this, but I think the whole test can be rewritten without this and without below Task.Yield

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            mockConnection.Setup(c => c.Open(It.IsAny<CancellationToken>())); // no action is required
+            mockConnection.Setup(c => c.OpenAsync(It.IsAny<CancellationToken>())).Returns(Task.FromResult(true)); // no action is required
+            mockConnection
+                .Setup(c => c.ReceiveMessageAsync(It.IsAny<int>(), It.IsAny<IMessageEncoderSelector>(), It.IsAny<MessageEncoderSettings>(), It.IsAny<CancellationToken>()))
+                .Returns(GetIsMasterResponse);
+
+            Task<ResponseMessage> GetIsMasterResponse()
+            {
+                var doc = s_primaries.Contains(serverId) ? primaryDoc : secondaryDoc;
```
DmitryLukyanov, line 311: `doc`: we don't use abbreviations anymore.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+                connectionModeSwitch: __connectionModeSwitch,
+                serverSelectionTimeout: TimeSpan.FromSeconds(30),
+                endPoints: new[] { __endPoint1 });
+
+            var allHeartbeatsRecieved = new TaskCompletionSource<bool>(false);
+            const int heartbeatsExpectedMinCount = 3;
+            int heartbeatsCount = 0, isInHeartbeat = 0;
+            var calledConcurrently = false;
+
+            void BlockHeartbeatRequested()
```
DmitryLukyanov, line 69: according to our agreement, we place local methods to the end of methods

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            var serverFactoryMock = new Mock<IClusterableServerFactory>();
+            serverFactoryMock
+                .Setup(f => f.CreateServer(It.IsAny<ClusterId>(), It.IsAny<IClusterClock>(), It.IsAny<EndPoint>()))
+                .Returns(serverMock.Object);
+
+            using (var cluster = new MultiServerCluster(clusterSettings, serverFactoryMock.Object, new EventCapturer()))
+            {
+                cluster._minHeartbeatInterval(TimeSpan.FromMilliseconds(10));
+                cluster._minHeartbeatInterval().Should().Be(TimeSpan.FromMilliseconds(10));
```
DmitryLukyanov, line 106: this is unnecessary, I don't think that we need to test our test code

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: this is not required. See ClusterTests suggestions

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
```diff
@@ -619,5 +619,8 @@ private void AddOrRemoveServers(ClusterDescription clusterDescription)
     internal static class ClusterReflector
     {
         public static InterlockedInt32 _state(this Cluster cluster) => (InterlockedInt32)Reflector.GetFieldValue(cluster, nameof(_state));
+
+        public static TimeSpan _minHeartbeatInterval(this Cluster cluster) => (TimeSpan)Reflector.GetFieldValue(cluster, nameof(_minHeartbeatInterval));
```
DmitryLukyanov, line 623: this is not required

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
```diff
@@ -619,5 +619,8 @@ private void AddOrRemoveServers(ClusterDescription clusterDescription)
...
+        public static void _minHeartbeatInterval(this Cluster cluster, TimeSpan timeSpan) => Reflector.SetFieldValue(cluster, nameof(_minHeartbeatInterval), timeSpan);
```
DmitryLukyanov, line 624: We still update just a static value. It's safe since we don't allow running tests in parallel: 

>public static void __minHeartbeatInterval(this Cluster cluster, TimeSpan timeSpan) => 
Reflector.SetStaticFieldValue(typeof(Cluster), nameof(__minHeartbeatInterval), timeSpan);

DmitryLukyanov (commented):
Initial review

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
rstam, line 53: `s_primaries` does not follow our naming conventions, use `__primaries` (i.e. two leading underscores for static fields).

rstam: Would a more specific name be appropriate? Perhaps `__stalledPrimaries` (not sure I fully understand how you are using this field).

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+        [Fact]
+        public async Task RapidHeartbeatTimerCallback_should_not_be_called_reentrantly()
+        {
+            var clusterSettings = new ClusterSettings(
+                connectionMode: __clusterConnectionMode,
+                connectionModeSwitch: __connectionModeSwitch,
+                serverSelectionTimeout: TimeSpan.FromSeconds(30),
+                endPoints: new[] { __endPoint1 });
+
+            var allHeartbeatsRecieved = new TaskCompletionSource<bool>(false);
```
rstam, line 64: Fix typo by renaming to `allHeartbeatsReceived`

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+            const int heartbeatsExpectedMinCount = 3;
+            int heartbeatsCount = 0, isInHeartbeat = 0;
+            var calledConcurrently = false;
```
rstam, line 67: Since we're testing reentrancy and not concurrency I recommend naming this variable `calledReentrantly`

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+            {
+                cluster._minHeartbeatInterval(TimeSpan.FromMilliseconds(10));
+                cluster._minHeartbeatInterval().Should().Be(TimeSpan.FromMilliseconds(10));
+
+                ForceClusterId(cluster, __clusterId);
+
+                cluster.Initialize();
+
+                // Trigger Cluster._rapidHeartbeatTimer
+                var selectServerTask = cluster.SelectServerAsync(CreateWritableServerAndEndPointSelector(__endPoint1), CancellationToken.None);
```
rstam, line 113: minor: replace `var selectServerTask` with discard variable `_` since we never use the value of `selectServerTask`?

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+            calledConcurrently.Should().Be(false);
+        }
+
+        [Fact(Timeout = 10000)]
+        public async Task Ensure_no_deadlock_after_primary_update()
+        {
+            // Force async execution, otherwise test timeout won't be respected
+            await Task.Yield();
+
+            var noLongerPrimaryStalled = new TaskCompletionSource<bool>();
```
rstam, line 129: The word order sounds off here. Perhaps `primaryNoLongerStalled` is more grammatical?

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: I missed the case when we can update this static value and then it will affect other tests since they will use the changed value insted the default one. In this case I see that this change is required

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
BorisDog: This change is not related to parallelism.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
BorisDog: You are right, this comment is not accurate. Should be: "This change is to enable easier and safer configuration in tests"

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
DmitryLukyanov: Discussed offline, as I understood, the only reason why this line was added here is to cover a case if the reflection behavior will be changed in the future. Right now, this validation can be only in two states:
* Always pass
* The error will be triggered in the previous reflection steps.
So it will never fail.

But even in the case when the reflection behavior may be changed, the only disadvantage we will have is failing in line 119. I think we should not make any changes with such unreliable assumptions.

My main consern, that we should not test our test code (at least with the reason that something in dotnet will be change on some day or with assumption that we wrote test incorrectly). 
We use settings values via reflection a lot and I don't think that we need to consider such validation anywhere including here.

rstam: I was surprised by this line also but decided to let it pass.

This line is NOT testing the driver at all. It's testing that line 105 worked.

I'd be fine with removing this line.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Referring to "InvalidatedBecause:NoLongerPrimary" event handler stalling.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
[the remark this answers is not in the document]

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -615,5 +626,11 @@ private static class State
             public const int Open = 1;
             public const int Disposed = 2;
         }
+
+        private static class RapidHeartbeatTimerCallbackState
+        {
+            public const int Idle = 0;
```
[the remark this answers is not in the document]

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
+﻿/* Copyright 2020-present MongoDB Inc.
```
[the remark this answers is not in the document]
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,329 @@
...
+
+                return new ConnectionDescription(
+                    mockConnection.Object.ConnectionId,
+                    new IsMasterResult(doc),
+                    new BuildInfoResult(new BsonDocument("version", serverVersion)));
+            }
+        }
+    }
+}
+;
```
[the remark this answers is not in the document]
BorisDog: Done.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: This code heavily relies on precise timings tuning.
In case of reflection failure, test troubleshooting would be disproportionately complicated. Therefore choosing "overtesting" approach in this case.

DmitryLukyanov: >This code heavily relies on precise timings tuning.

there is no real reasons to assume that something will be wrong with reflection. We always can say that settings some important test value can be failed via dotnet bug (regardless reflection or no), but we don't test it, so I still don't see reasons to check it. For now, I see a case when someone will investigate this not simple test and will see this asserting which can make his thinking that there is an idea behind this which in turns will confuse.

rstam: When testing a *unit* it is fair to assume that everything *outside* the unit works (whether that's reflection or anything else). If something *outside* the unit needs to be tested... then it should be tested somewhere else.

For example, in this test you *assumed* that mocking worked (as well as lots of other things).

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+            {
+                var isMasterDoc = _currentPrimaries.Contains(serverId) ? primaryDoc : secondaryDoc;
+
+                ResponseMessage result = MessageHelper.BuildReply(new RawBsonDocument(isMasterDoc.ToBson()));
+                return Task.FromResult(result);
+            }
+
+            ConnectionDescription GetConnectionDescription()
+            {
+                var isMasterDoc = _currentPrimaries.Contains(serverId) ? primaryDoc : secondaryDoc;
```
DmitryLukyanov, line 319: we don't use abbreviations anymore (`doc`)

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
rstam: I have no knowledge about this xUnit feature and/or its reliability.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+            }
+        }
+
+        [Fact(Timeout = 10000)]
+        public async Task Ensure_no_deadlock_after_primary_update()
+        {
+            // Force async execution, otherwise test timeout won't be respected
+            await Task.Yield();
+
+            var noLongerPrimaryEventStalled = new TaskCompletionSource<bool>();
```
rstam, line 129: I still don't understand this variable name.

Perhaps you could explain in English what this variable represents and that will either help me understand how you came up with this name, or it will help us collectively come up with a better name.

rstam (changes requested):

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+#pragma warning restore CS0618 // Type or member is obsolete
+        private readonly static ClusterId __clusterId = new ClusterId();
+        private readonly static bool? __directConnection = null;
+        private readonly static EndPoint __endPoint1 = new DnsEndPoint("localhost", 27017);
+        private readonly static EndPoint __endPoint2 = new DnsEndPoint("localhost", 27018);
+        private readonly static TimeSpan __heartbeatInterval = TimeSpan.FromMilliseconds(200);
+        private readonly static ServerId __serverId1 = new ServerId(__clusterId, __endPoint1);
+        private readonly static ServerId __serverId2 = new ServerId(__clusterId, __endPoint2);
+
+        private HashSet<ServerId> _currentPrimaries;
```
DmitryLukyanov, line 53: can this be moved into the local variable? It looks like yes

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
JamesKovacs: The comment explains the what, not the why. Robert's comment better explains the why.

JamesKovacs (changes requested):

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterTests.cs
BorisDog: Done

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
BorisDog: Done

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
rstam: @BorisDog says offline that the purpose of this Task is to record that we DID stall during processing the event, not that the event IS stalled.

BorisDog: Our lines length should be definitely capped.
In this case it's nearly same length as the line above, therefore I think it's appropriate usage here.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+        {
+            var clusterSettings = new ClusterSettings(
+                connectionMode: __clusterConnectionMode,
+                connectionModeSwitch: __connectionModeSwitch,
+                serverSelectionTimeout: TimeSpan.FromSeconds(30),
+                endPoints: new[] { __endPoint1 });
+
+            var allHeartbeatsReceived = new TaskCompletionSource<bool>(false);
+            const int heartbeatsExpectedMinCount = 3;
+            int heartbeatsCount = 0, isInHeartbeat = 0;
```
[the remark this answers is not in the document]
BorisDog: Our lines length should be definitely capped.
In this case it's nearly same length as the line above, therefore I think it's appropriate usage here.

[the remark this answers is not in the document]
rstam: In my opinion there is never a good reason to combine multiple declarations in one line.

And line length is definitely not one.

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
rstam: I think your comment was meant for somewhere else.

rstam: Feel free to improve this variable name or leave it as is...

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+        [Fact]
+        public async Task RapidHeartbeatTimerCallback_should_ignore_reentrant_calls()
+        {
+            var clusterSettings = new ClusterSettings(
+                connectionMode: __clusterConnectionMode,
+                connectionModeSwitch: __connectionModeSwitch,
+                serverSelectionTimeout: TimeSpan.FromSeconds(30),
+                endPoints: new[] { __endPoint1 });
+
+            var allHeartbeatsReceived = new TaskCompletionSource<bool>(false);
```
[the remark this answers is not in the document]
BorisDog: Done

tests/MongoDB.Driver.Core.Tests/Jira/CSharp3302Tests.cs
BorisDog: Done.

DmitryLukyanov (approved):
LGTM

JamesKovacs (approved):
LGTM

N.B. Deferring questions around testing strategy for later.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -75,6 +76,7 @@ internal abstract class Cluster : ICluster
         private readonly ICoreServerSessionPool _serverSessionPool;
         private readonly ClusterSettings _settings;
         private readonly InterlockedInt32 _state;
+        private readonly InterlockedInt32 _rapidHeartbeatTimerCallbackState;
```
rstam, line 79: nit: we keep fields in alphabetical order

no need to make any changes at this late stage as we're trying to get the release out

rstam (approved):
LGTM

PR #436, opened 2021-01-13
Increasing EG timeout from 30 to 45 minutes

BorisDog:


JamesKovacs (approved):
LGTM

evergreen/evergreen.yml
```diff
@@ -13,8 +13,8 @@ stepback: true
 command_type: system
 
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
-# 12 minutes is the longest we'll ever run
-exec_timeout_secs: 1800
+# 30 minutes is the longest we'll ever run
```
MikalaiMazurenka, line 16: Shouldn't it say 45 minutes?

MikalaiMazurenka (approved):
LGTM with minor note. I don't know the reason why 12 minutes were specified in the first place, so I don't have strong concerns about the new value either.

evergreen/evergreen.yml
BorisDog: We don't know actually. I think **_exec_timeout_secs_** should have some buffer over the time we think the test will take. So assuming for now they'll take ~30 mins.

MikalaiMazurenka: Sounds like it is the case

jyemin: Looking at the file history, this comment has almost always been wrong.  The first version said 12 minutes, but the exec_timeout_secs was 3600 (one hour).  I think it should either say 45 minutes, or be removed (since it's redundant with the actual code)

PR #441, opened 2021-01-26
CSHARP-2750: Upgrade C# language version to 9.

BorisDog:
CSHARP-2750:
- Upgrade to C# 9

Legacy projects where not updated.


BorisDog:
[EG](https://evergreen.mongodb.com/version/6010593a1e2d173581aa2c01) (internal link)

MongoDB.ruleset
```diff
@@ -21,7 +21,7 @@
     <Rule Id="CA1060" Action="Warning" />
     <Rule Id="CA1061" Action="None" />
     <Rule Id="CA1063" Action="None" />
-    <Rule Id="CA1065" Action="Warning" />
+    <Rule Id="CA1065" Action="Info" />
```
BorisDog, line 24: As described in [2750](https://jira.mongodb.org/browse/CSHARP-2750), CA1065 warnings are treated as errors.
Downgrading this warning to information.
This also might be solved by upgrading Microsoft.CodeAnalysis.FxCopAnalyzers version. But this can have other side effects. Also this package is deprecated. I think we should tackle this under different code analysis cleanup effort.

CSharpDriver.Bson.slnf
```diff
@@ -0,0 +1,10 @@
+{
```
DmitryLukyanov, line 1: do we really need this? I would say the best thing we can do here is to add `.slnf` into git ignore. And then if anyone will be interested in this (with any filter conditions), then he will be able to have their own version of this file. I don't feel that it's useful to have added filters under git. 

But I will be happy to hear other reviewers.

MongoDB.ruleset
DmitryLukyanov: FYI: It looks like this is a related article: https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-legacy-analysis-to-net-analyzers?view=vs-2019

But I'm ok with this change as is

CSharpDriver.Bson.slnf
rstam: I didn't know about this feature before, but when Boris taught me about it recently I found them really useful because they can be used to make the Test Explorer load more quickly.

I have no strong feelings whether these should or should not be in the repo.

However, I do feel they shouldn't be added as part of this PR. They have nothing to do with upgrading the language version to C# 8.

src/MongoDB.Bson/MongoDB.Bson.csproj
```diff
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard1.5;netstandard2.0;netstandard2.1;net452</TargetFrameworks>
     <TargetFrameworks Condition="'$(IsWindows)'!='true'">netstandard1.5;netstandard2.0;netstandard2.1</TargetFrameworks>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8</LangVersion>
```
DmitryLukyanov, line 9: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults
Looking at this:
>You should edit the project file and remove that node. Then, your project will use the compiler version recommended for your target framework.

Should we follow this recommendation?

CSharpDriver.Bson.slnf
DmitryLukyanov: I don't see why it should be just Bson project? Why not just Core? Or just Driver? (the last two will be more useful) So, not sure that it's a good idea to place it under git. However, I don't insist on it

CSharpDriver.Driver.slnf
```diff
@@ -0,0 +1,17 @@
...
+      "tests\\MongoDB.Bson.TestHelpers\\MongoDB.Bson.TestHelpers.csproj",
+      "tests\\MongoDB.Bson.Tests\\MongoDB.Bson.Tests.csproj",
+      "tests\\MongoDB.Driver.Core.TestHelpers\\MongoDB.Driver.Core.TestHelpers.csproj",
+      "tests\\MongoDB.Driver.Core.Tests\\MongoDB.Driver.Core.Tests.csproj",
+      "tests\\MongoDB.Driver.Examples\\MongoDB.Driver.Examples.csproj",
+      "tests\\MongoDB.Driver.TestHelpers\\MongoDB.Driver.TestHelpers.csproj",
+      "tests\\MongoDB.Driver.Tests\\MongoDB.Driver.Tests.csproj"
+    ]
+  }
+}
```
DmitryLukyanov, line 17: the last line in the file should be a blank line. Then you won't see this "red circle"

tests/AtlasConnectivity.Tests/AtlasConnectivity.Tests.csproj
```diff
@@ -1,12 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
DmitryLukyanov, line 1: I'm not sure about this line, we see this diff from time to time. If this is autoformatting change, then let's leave it, otherwise, this should be reverted

src/MongoDB.Bson/MongoDB.Bson.csproj
DmitryLukyanov: yeah, if this works in this way, then I agree that it's better to leave LangVersion

MongoDB.ruleset
BorisDog: Exactly

DmitryLukyanov (approved):
LGTM

rstam (changes requested):
Overall this looks fine. I have two questions:

1. I thought <LangVersion> was the "old" way to specify the language version. The "new" way is to let Visual Studio automatically pick the language version based on the targeted frameworks. I would like to see some references stating that it is still OK to use <LangVersion>

2. Not all .csproj files have been changed to specify C# 9. I'm assuming this is an oversight. Either specify C# 9 for the remaining projects that don't already specify C# 9 or explain why those projects were left out of this PR

BorisDog:
@rstam 

1. Yes, as stated [here ](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) we can change that value if we want to. The only side effect is that we will have only partial C# 9 features support.
2. Yes, missed 2 projects. Legacy projects where left out intentionally. As no active work is expected there we don't have any good reason to upgrade.

rstam:
> 1. Yes, as stated [here ](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) we can change that value if we want to. The only side effect is that we will have only partial C# 9 features support.

> 2. Yes, missed 2 projects. Legacy projects where left out intentionally. As no active work is expected there we don't have any good reason to upgrade.

1. Thanks for the link to the C# Language Versioning page. It does state "Choosing a language version newer than the default can cause hard to diagnose compile-time and runtime errors." so we have to stay on our toes.

2. OK. I wasn't sure whether you had chosen to not upgrade the language version from some projects or just missed them.

rstam (approved):
LGTM with the caveat that I'm relying on the analysis of others that it is OK to manually override the LangVersion and pick a version NEWER than the default.

BorisDog:
Merged.

PR #442, opened 2021-01-27
CSHARP-2009: Inserting correct discriminator on Upsert

BorisDog:
Few differences from the original PR :

1. Original PR maintains the update object on FilteredMongoCollectionBase.cs level, while I think it is very specific to 
OfTypeMongoCollection and should be handled on that level.
2. No need to create the update object if upsert is never called.
3. SetOnInsert is only called on upsert, and not for every update. Also seems that adding SetOnInsert only on "upsert" logic naturally belongs to OfTypeMongoCollection.

### DIFF
--- a/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
+++ b/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
@@ -340,6 +340,11 @@ namespace MongoDB.Driver
             return _filter & filter;
         }
 
+        protected virtual UpdateDefinition<TDocument> GetUpdateDefinition(UpdateDefinition<TDocument> updateDefinition, bool isUpsert)
+        {
+            return updateDefinition;
+        }
+
         private IEnumerable<WriteModel<TDocument>> CombineModelFilters(IEnumerable<WriteModel<TDocument>> models)
         {
             return models.Select<WriteModel<TDocument>, WriteModel<TDocument>>(x =>
@@ -372,7 +377,9 @@ namespace MongoDB.Driver
                         };
                     case WriteModelType.UpdateMany:
                         var updateManyModel = (UpdateManyModel<TDocument>)x;
-                        return new UpdateManyModel<TDocument>(CombineFilters(updateManyModel.Filter), updateManyModel.Update)
+                        return new UpdateManyModel<TDocument>(
+                            CombineFilters(updateManyModel.Filter),
+                            GetUpdateDefinition(updateManyModel.Update, updateManyModel.IsUpsert))
                         {
                             ArrayFilters = updateManyModel.ArrayFilters,
                             Collation = updateManyModel.Collation,
@@ -381,7 +388,9 @@ namespace MongoDB.Driver
                         };
                     case WriteModelType.UpdateOne:
                         var updateOneModel = (UpdateOneModel<TDocument>)x;
-                        return new UpdateOneModel<TDocument>(CombineFilters(updateOneModel.Filter), updateOneModel.Update)
+                        return new UpdateOneModel<TDocument>(
+                            CombineFilters(updateOneModel.Filter),
+                            GetUpdateDefinition(updateOneModel.Update, updateOneModel.IsUpsert))
                         {
                             ArrayFilters = updateOneModel.ArrayFilters,
                             Collation = updateOneModel.Collation,
--- a/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
+++ b/src/MongoDB.Driver/FilteredMongoCollectionBase.cs
@@ -340,7 +340,7 @@ namespace MongoDB.Driver
             return _filter & filter;
         }
 
-        protected virtual UpdateDefinition<TDocument> GetUpdateDefinition(UpdateDefinition<TDocument> updateDefinition, bool isUpsert)
+        protected virtual UpdateDefinition<TDocument> AdjustUpdateDefinition(UpdateDefinition<TDocument> updateDefinition, bool isUpsert)
         {
             return updateDefinition;
         }
@@ -379,7 +379,7 @@ namespace MongoDB.Driver
                         var updateManyModel = (UpdateManyModel<TDocument>)x;
                         return new UpdateManyModel<TDocument>(
                             CombineFilters(updateManyModel.Filter),
-                            GetUpdateDefinition(updateManyModel.Update, updateManyModel.IsUpsert))
+                            AdjustUpdateDefinition(updateManyModel.Update, updateManyModel.IsUpsert))
                         {
                             ArrayFilters = updateManyModel.ArrayFilters,
                             Collation = updateManyModel.Collation,
@@ -390,7 +390,7 @@ namespace MongoDB.Driver
                         var updateOneModel = (UpdateOneModel<TDocument>)x;
                         return new UpdateOneModel<TDocument>(
                             CombineFilters(updateOneModel.Filter),
-                            GetUpdateDefinition(updateOneModel.Update, updateOneModel.IsUpsert))
+                            AdjustUpdateDefinition(updateOneModel.Update, updateOneModel.IsUpsert))
                         {
                             ArrayFilters = updateOneModel.ArrayFilters,
                             Collation = updateOneModel.Collation,
--- a/src/MongoDB.Driver/OfTypeMongoCollection.cs
+++ b/src/MongoDB.Driver/OfTypeMongoCollection.cs
@@ -54,9 +54,9 @@ namespace MongoDB.Driver
             return new OfTypeMongoCollection<TRootDocument, TDerivedDocument>(_rootDocumentCollection, WrappedCollection.WithWriteConcern(writeConcern), Filter);
         }
 
-        protected override UpdateDefinition<TDerivedDocument> GetUpdateDefinition(UpdateDefinition<TDerivedDocument> updateDefinition, bool isUpsert)
+        protected override UpdateDefinition<TDerivedDocument> AdjustUpdateDefinition(UpdateDefinition<TDerivedDocument> updateDefinition, bool isUpsert)
         {
-            var result = base.GetUpdateDefinition(updateDefinition, isUpsert);
+            var result = base.AdjustUpdateDefinition(updateDefinition, isUpsert);
 
             if (isUpsert)
             {
@@ -64,8 +64,8 @@ namespace MongoDB.Driver
                 var discriminatorValue = discriminatorConvention.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
 
                 var builder = new UpdateDefinitionBuilder<TDerivedDocument>();
-                var derivedUpdate = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
-                result = builder.Combine(result, derivedUpdate);
+                var setOnInsertDiscriminator = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
+                result = builder.Combine(result, setOnInsertDiscriminator);
             }
 
             return result;
--- a/src/MongoDB.Driver/OfTypeMongoCollection.cs
+++ b/src/MongoDB.Driver/OfTypeMongoCollection.cs
@@ -13,7 +13,7 @@
 * limitations under the License.
 */
 
-using System;
+using MongoDB.Bson.Serialization;
 
 namespace MongoDB.Driver
 {
@@ -52,6 +52,23 @@ namespace MongoDB.Driver
         public override IMongoCollection<TDerivedDocument> WithWriteConcern(WriteConcern writeConcern)
         {
             return new OfTypeMongoCollection<TRootDocument, TDerivedDocument>(_rootDocumentCollection, WrappedCollection.WithWriteConcern(writeConcern), Filter);
+        }
+
+        protected override UpdateDefinition<TDerivedDocument> GetUpdateDefinition(UpdateDefinition<TDerivedDocument> updateDefinition, bool isUpsert)
+        {
+            var result = base.GetUpdateDefinition(updateDefinition, isUpsert);
+
+            if (isUpsert)
+            {
+                var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(typeof(TDerivedDocument));
+                var discriminatorValue = discriminatorConvention?.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
+
+                var builder = new UpdateDefinitionBuilder<TDerivedDocument>();
+                var derivedUpdate = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
+                result = builder.Combine(result, derivedUpdate);
+            }
+
+            return result;
         }
     }
 }
--- a/src/MongoDB.Driver/OfTypeMongoCollection.cs
+++ b/src/MongoDB.Driver/OfTypeMongoCollection.cs
@@ -61,7 +61,7 @@ namespace MongoDB.Driver
             if (isUpsert)
             {
                 var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(typeof(TDerivedDocument));
-                var discriminatorValue = discriminatorConvention?.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
+                var discriminatorValue = discriminatorConvention.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
 
                 var builder = new UpdateDefinitionBuilder<TDerivedDocument>();
                 var derivedUpdate = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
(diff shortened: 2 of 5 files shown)


JamesKovacs (changes requested):
Looks like you're on the right track here. I tested the following program before and after this change.

Prior to the change, we upsert a C document with type discriminator of `_t: "C"`, which then returns no results when we query for documents of type B. With the change, we correctly create the C document with `_t: ["A","B","C"]` and return that document when we query for documents of type B.

I noticed one minor code fix and I think we are good to merge.

```
using System;
using MongoDB.Bson;
using MongoDB.Bson.Serialization.Attributes;
using MongoDB.Driver;

namespace Csharp2009
{
    class Program
    {
        static void Main(string[] args)
        {
            var client = new MongoClient();
            var db = client.GetDatabase("csharp2009");
            var coll = db.GetCollection<A>("coll");
            var justTheCs = coll.OfType<C>();

            var filter = Builders<C>.Filter.Eq(x => x.Id, 42);
            var update = Builders<C>.Update.Set(x => x.Number, 99);
            var options = new UpdateOptions { IsUpsert = true };
            justTheCs.UpdateOne(filter, update, options);

            var allTheBs = coll.OfType<B>();
            foreach (var b in allTheBs.Find(Builders<B>.Filter.Empty).ToList()) {
                Console.WriteLine(b);
            }
        }
    }

    [BsonDiscriminator(RootClass = true)]
    class A {
        public int Id { get; set; }
    }
    class B : A {
        public int Number { get; set; }
    }
    class C : B {
    }
}
```

src/MongoDB.Driver/OfTypeMongoCollection.cs
```diff
@@ -53,5 +53,22 @@ public override IMongoCollection<TDerivedDocument> WithWriteConcern(WriteConcern
...
         }
+
+        protected override UpdateDefinition<TDerivedDocument> GetUpdateDefinition(UpdateDefinition<TDerivedDocument> updateDefinition, bool isUpsert)
+        {
+            var result = base.GetUpdateDefinition(updateDefinition, isUpsert);
+
+            if (isUpsert)
+            {
+                var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(typeof(TDerivedDocument));
+                var discriminatorValue = discriminatorConvention?.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
```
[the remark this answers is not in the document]
BorisDog: Done.
In PipelineStageDefinition.OfType LookupDiscriminatorConvention result is validated as well but I don't see how it can be null.

jyemin:
Added @rstam as second reviewer

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -396,47 +405,54 @@ public OfTypeMongoCollectionTests()
 
         [Theory]
         [ParameterAttributeData]
-        public void BulkWrite_with_UpdateOne(
-            [Values(false, true)] bool async)
+        public void BulkWrite_with_UpdateOne<TRoot, TDerived>(
+            [ClassValues(typeof(SerializedTypeInfos))]
+            (TRoot, TDerived, string Discriminator) typeInfo,
+            [Values(false, true)] bool isUpsert,
+            [Values(false, true)] bool async) where TDerived : TRoot
```
JamesKovacs, line 412: Same idea here. I think it would be more readable if we moved the update and upsert tests into separate theories.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -879,6 +910,36 @@ public class C : B
         {
             public int PropC;
         }
+
+        [BsonDiscriminator(RootClass = true)]
+        [BsonKnownTypes(typeof(B_Discriminated), typeof(C_Discriminated))]
+        public class A_Discriminated
+        {
+            public int PropA;
```
JamesKovacs, line 918: Nitpick... These are fields, not properties.

JamesKovacs (changes requested):
Overall looks good. I'm concerned with the complexity of the test methods. Consider whether update/upsert should be separated into their own test methods. Happy to discuss.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: We can feed rest for parameters via SerializedTypeInfos. @rstam what do you think?

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -347,47 +348,55 @@ public OfTypeMongoCollectionTests()
...
                 Collation = collation,
                 Hint = new BsonDocument("_id", 1),
-                IsUpsert = true
+                IsUpsert = isUpsert
             };
+
             var options = new BulkWriteOptions();
+            var updateString = isUpsert ?
+                $"{{$set: {{x: 1}}, $setOnInsert: {{{typeInfo.Discriminator}}}}}" :
+                "{$set: {x: 1}}";
```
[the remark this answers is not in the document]
BorisDog: Would one line look better here? cc @rstam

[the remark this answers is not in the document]
JamesKovacs: @BorisDog and I discussed via Slack. A single-line ternary is harder to read and changing this to an if-else is really long and verbose. Open to other suggestions though this concern would be addressed if we separated the update/upsert tests into separate test methods.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -347,47 +348,55 @@ public OfTypeMongoCollectionTests()
 
         [Theory]
         [ParameterAttributeData]
-        public void BulkWrite_with_UpdateMany(
-            [Values(false, true)] bool async)
+        public void BulkWrite_with_UpdateMany<TRoot, TDerived>(
+            [ClassValues(typeof(SerializedTypeInfos))]
+            (TRoot, TDerived, string Discriminator) typeInfo,
+            [Values(false, true)] bool isUpsert,
+            [Values(false, true)] bool async) where TDerived : TRoot
```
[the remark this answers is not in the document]
rstam: Let's be consistent on whether parameters are on one line or two.

Also, let's put the `where` constraint on a separate line so that it's more easily seen (this is one of our conventions).

In other words:

    [Theory]
    [ParameterAttributeData]
    public void BulkWrite_with_UpdateMany<TRoot, TDerived>(
        [ClassValues(typeof(UpdateTestCases))] (TRoot, TDerived, string Discriminator) testCase,
        [Values(false, true)] bool isUpsert,
            where TDerived : TRoot

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -347,47 +348,55 @@ public OfTypeMongoCollectionTests()
 
         [Theory]
         [ParameterAttributeData]
-        public void BulkWrite_with_UpdateMany(
-            [Values(false, true)] bool async)
+        public void BulkWrite_with_UpdateMany<TRoot, TDerived>(
+            [ClassValues(typeof(SerializedTypeInfos))]
```
rstam, line 352: I would like to rename `SerializedTypeInfos` to `UpdateTestCases` and rename the `typeInfo` parameter to `testCase`.

I think those names are more accurate.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -347,47 +348,55 @@ public OfTypeMongoCollectionTests()
...
+            (TRoot, TDerived, string Discriminator) typeInfo,
+            [Values(false, true)] bool isUpsert,
+            [Values(false, true)] bool async) where TDerived : TRoot
         {
-            var subject = CreateSubject();
+            var (derivedCollectionMock, ofTypeCollection) = CreateSubject<TRoot, TDerived>();
```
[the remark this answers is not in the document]
rstam: I'm fine with this type of assignment in general.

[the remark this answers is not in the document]
rstam: I would like to rename `derivedCollectionMock` to `mockDerivedCollection`

and to rename `ofTypeCollection` to `subject`

Just about every unit test should have a variable named `subject`.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -840,6 +856,21 @@ private IClientSessionHandle CreateSession(bool usingSession)
             return new OfTypeMongoCollection<A, B>(_rootCollection, _derivedCollection, _ofTypeFilter);
         }
 
+        private (Mock<IMongoCollection<TDerived>> Derived, OfTypeMongoCollection<TRoot, TDerived> OfType) CreateSubject<TRoot, TDerived>()
```
rstam, line 859: Rename `Derived` to `MockDerivedCollection` and rename `OfType` to `Subject`

    private (Mock<IMongoCollection<TDerived>> MockDerivedCollection, OfTypeMongoCollection<TRoot, TDerived> Subject) CreateSubject<TRoot, TDerived>()

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -840,6 +856,21 @@ private IClientSessionHandle CreateSession(bool usingSession)
...
+            where TDerived : TRoot
+        {
+            var mockRootCollectionMock = new Mock<IMongoCollection<TRoot>>();
```
rstam, line 862: Rename to `mockRootCollection`

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -840,6 +856,21 @@ private IClientSessionHandle CreateSession(bool usingSession)
...
+            mockRootCollectionMock.SetupGet(c => c.CollectionNamespace).Returns(CollectionNamespace.FromFullName("foo.bar"));
+            mockRootCollectionMock.SetupGet(c => c.Settings).Returns(new MongoCollectionSettings());
+
+            var mockDerivedCollectionMock = new Mock<IMongoCollection<TDerived>>();
```
rstam, line 866: Rename to `mockDerivedCollection`

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -840,6 +856,21 @@ private IClientSessionHandle CreateSession(bool usingSession)
...
+            mockDerivedCollectionMock.SetupGet(c => c.CollectionNamespace).Returns(CollectionNamespace.FromFullName("foo.bar"));
+            mockDerivedCollectionMock.SetupGet(c => c.Settings).Returns(new MongoCollectionSettings());
+
+            var ofTypeCollection = new OfTypeMongoCollection<TRoot, TDerived>(mockRootCollectionMock.Object, mockDerivedCollectionMock.Object, _ofTypeFilter);
```
rstam, line 870: Rename `ofTypeCollection` to `subject`

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -879,6 +910,36 @@ public class C : B
         {
             public int PropC;
         }
+
+        [BsonDiscriminator(RootClass = true)]
+        [BsonKnownTypes(typeof(B_Discriminated), typeof(C_Discriminated))]
+        public class A_Discriminated
```
rstam, line 916: Rename `A_Discriminated` to `AHierarchical` and same for the B and C versions.

The key difference between `A` and `AHierarchical` is NOT whether it's discriminated or not (in these tests BOTH are discriminated) but rather that `A` etc..  uses a scalar discriminator and `AHierchical` etc.. uses a hierarchical (i.e. array) discriminator.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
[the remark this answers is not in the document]
rstam: I'm willing to let this one slide. If you want to rewrite it you can.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
rstam: I think `SerializedTypeInfos` is fine as is, though I have asked for it to be renamed to `UpdateTestCases`.

rstam (changes requested):
I think these changes look really good. I am requesting some name changes that I hope you agree with.

While the two modified tests are more complex than we typically like to see a test be, I think they are fine as is.

src/MongoDB.Driver/FilteredMongoCollectionBase.cs
```diff
@@ -340,6 +340,11 @@ private FilterDefinition<TDocument> CombineFilters(FilterDefinition<TDocument> f
             return _filter & filter;
         }
 
+        protected virtual UpdateDefinition<TDocument> GetUpdateDefinition(UpdateDefinition<TDocument> updateDefinition, bool isUpsert)
```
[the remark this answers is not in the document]
BorisDog: Tnx, changed to AdjustUpdateDefinition.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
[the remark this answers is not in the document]
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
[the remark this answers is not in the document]
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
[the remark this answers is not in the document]
BorisDog: Left as is.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
BorisDog: Renamed

src/MongoDB.Driver/OfTypeMongoCollection.cs
```diff
@@ -53,5 +53,22 @@ public override IMongoCollection<TDerivedDocument> WithWriteConcern(WriteConcern
...
+        {
+            var result = base.GetUpdateDefinition(updateDefinition, isUpsert);
+
+            if (isUpsert)
+            {
+                var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(typeof(TDerivedDocument));
+                var discriminatorValue = discriminatorConvention.GetDiscriminator(typeof(TRootDocument), typeof(TDerivedDocument));
+
+                var builder = new UpdateDefinitionBuilder<TDerivedDocument>();
+                var derivedUpdate = builder.SetOnInsert(discriminatorConvention.ElementName, discriminatorValue);
```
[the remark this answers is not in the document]
BorisDog: Done

JamesKovacs (approved):
LGTM

BorisDog:
Merged.

PR #443, opened 2021-01-27
Renamed Tools directory to tools to fix issue with both Tools and tools directories being created on Linux

JamesKovacs:
`build.sh`/`build.ps1` from the Cake repo use `tools` as the name for the tools directory. Our `build.cake` file uses the name `Tools`. On NTFS (Windows) and APFS (OSX), `Tools` is created first and then references to `tools` simply use `Tools` because the file systems are case-aware but not case-sensitive. `ext4` and other Linux file systems are case-sensitive and this leads to both `tools` and `Tools` directories on Linux. Because only `/Tools` is in the `.gitignore` file, Linux shows a dirty working copy after a build because the build script has downloaded files into `/tools`.

This PR addresses the issue by standardizing on `tools`, which is the casing used by Cake itself.


.gitignore
```diff
@@ -65,4 +65,4 @@ src/packages
 # Other
 artifacts
 packages
-/Tools
+/tools
```
DmitryLukyanov, line 68: it's probably should be also changed here https://github.com/mongodb/mongo-csharp-driver/blob/master/build.cmd#L2

DmitryLukyanov: and here https://github.com/mongodb/mongo-csharp-driver/blob/master/build/build.fsx#L1

JamesKovacs: Those are orphaned build files related to our Fake build that is no longer used. See CSHARP-3246 (which we should probably just do).

DmitryLukyanov (approved):
I didn't try to build it via cake, but I assume James tried it, and it works. The changes themselves look correct. LGTM

PR #444, opened 2021-01-29
CSHARP-3281: Versioned MongoDB API for Drivers

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3281
Evergreen: https://evergreen.mongodb.com/version/603e5d1d2a60ed639b8e56b8
Specification: https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst


jyemin:
Added Dmitry as first reviewer, Robert as second.

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatTestRunner.cs
```diff
@@ -283,7 +305,11 @@ public class TestCaseFactory : JsonDrivenTestCaseFactory
             #endregion
 
             // protected properties
-            protected override string PathPrefix => "MongoDB.Driver.Tests.Specifications.unified_test_format.tests.valid_pass.";
+            protected override string[] PathPrefixes => new[]
```
MikalaiMazurenka, line 308: `versioned API` tests mostly reside in a different folder. I was deciding between adding a path to `unified tests` runner or creating an additional test runner for versioned API tests similar to `UnifiedTestFormatNegativeTestRunner` and decided to go with the first option, but that can be altered.

MikalaiMazurenka (commented):
Ready for review!

src/MongoDB.Driver.Core/Core/Servers/ServerFactory.cs
```diff
@@ -33,12 +33,43 @@ internal class ServerFactory : IClusterableServerFactory
...
+                  directConnection,
+                  settings,
+                  connectionPoolFactory,
+                  serverMonitoryFactory,
+                  eventSubscriber,
+                  null)
+        {
+        }
+
+        public ServerFactory(
```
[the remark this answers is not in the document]
DmitryLukyanov: I would say, the driver changes should be the first priority, how big changes in the tests should be a secondary question

src/MongoDB.Driver.Core/Core/WireProtocol/CommandWireProtocol.cs
```diff
@@ -97,6 +102,35 @@ internal class CommandWireProtocol<TCommandResult> : IWireProtocol<TCommandResul
...
+                  additionalOptions,
+                  postWriteAction,
+                  responseHandling,
+                  resultSerializer,
+                  messageEncoderSettings,
+                  null)
+        {
+        }
+
+        public CommandWireProtocol(
```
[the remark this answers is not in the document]
DmitryLukyanov: similar to the above, the number of changes in the tests is a secondary question

src/MongoDB.Driver.Core/Core/Authentication/IAuthenticator.cs
```diff
@@ -42,6 +41,15 @@ public interface IAuthenticator
...
         void Authenticate(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken);
 
+        /// <summary>
+        /// Authenticates the connection.
+        /// </summary>
+        /// <param name="connection">The connection.</param>
+        /// <param name="description">The connection description.</param>
+        /// <param name="serverApi">The server API.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        void Authenticate(IConnection connection, ConnectionDescription description, ServerApi serverApi, CancellationToken cancellationToken);
```
DmitryLukyanov, line 51: this is a BC. See this commit with a suggestion on how to avoid this: https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/0fe4d7e027cf03e34e191e2048cf98709a4b6cff

DmitryLukyanov (changes requested):
NOTE: I didn't look at tests yet

src/MongoDB.Driver.Core/ServerApi.cs
```diff
@@ -0,0 +1,124 @@
...
+        private readonly ServerApiVersion _version;
+
+        // constructors
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServerApi" /> class.
+        /// </summary>
+        /// <param name="version">The server API version.</param>
+        /// <param name="strict">The flag for strict server API version enforcement.</param>
+        /// <param name="deprecatationErrors">The flag for treating deprecated server APIs as errors.</param>
+        public ServerApi(ServerApiVersion version, bool? strict = null, bool? deprecatationErrors = null)
```
DmitryLukyanov, line 40: it's probably should be:

        public ServerApi(ServerApiVersion version, Optional<bool> strict = default, Optional<bool> deprecatationErrors = default)
        {
            _version = Ensure.IsNotNull(version, nameof(version));
            _strict = strict.WithDefault(false);
            _deprecatationErrors = deprecatationErrors.WithDefault(false);
        }

src/MongoDB.Driver.Legacy/MongoServerSettings.cs
```diff
@@ -592,6 +594,19 @@ public MongoServerAddress Server
             }
         }
 
+        /// <summary>
```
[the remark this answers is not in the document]
MikalaiMazurenka: Great catch!
Done.

src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -574,6 +576,19 @@ public string SdamLogFilename
             }
         }
 
+        /// <summary>
+        /// Gets or sets the server api.
```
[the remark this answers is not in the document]
MikalaiMazurenka: Nice find!
Done.

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -291,6 +291,19 @@ public static IEnumerable<string> GetModules()
             }
         }
 
+        public static BsonDocument GetServerParameters()
```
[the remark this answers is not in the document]
MikalaiMazurenka: Nope. The property shouldn't do any requests or complex things ([Choosing Between Properties and Methods](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)) or [Property Usage Guidelines: Properties vs. Methods](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/bzwdh01d(v=vs.71)#properties-vs-methods))

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -291,6 +291,19 @@ public static IEnumerable<string> GetModules()
...
+        {
+            using (var session = StartSession())
+            using (var binding = CreateReadBinding(session))
```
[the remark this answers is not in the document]
MikalaiMazurenka: There's no existing session to fork (and I see no point in forking just created session) and btw the method just above (`GetModules()`) uses the same exact code. Unless I'm missing something

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -722,6 +722,45 @@ public void Command_should_update_the_session_and_cluster_cluster_times()
                 server.ClusterClock.ClusterTime.Should().Be(actualClusterTime);
             }
         }
+
+        [SkippableFact]
+        public void Command_should_use_serverApi()
+        {
+            RequireServer.Check().VersionGreaterThanOrEqualTo("3.6").ClusterTypes(ClusterType.ReplicaSet, ClusterType.Sharded);
```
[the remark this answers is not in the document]
MikalaiMazurenka: Nice find!
It seems I just blindly copypasted the check from above method. My bad.
Done.

tests/MongoDB.Driver.Legacy.Tests/MongoServerSettingsTests.cs
```diff
@@ -112,6 +112,7 @@ public void TestClone()
 #pragma warning restore 618
             settings.SslSettings = new SslSettings { CheckCertificateRevocation = !url.TlsDisableCertificateRevocationCheck };
             settings.SdamLogFilename = "unimatrix-zero";
+            settings.ServerApi = new ServerApi(ServerApiVersion.V1, true, true);
```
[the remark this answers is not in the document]
MikalaiMazurenka: Good point!
Done.

MikalaiMazurenka (commented):
Ready for reivew!

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
```diff
@@ -122,5 +124,99 @@ public void Constructor_should_throw_an_ArgumentNullException_when_credential_is
             sentMessages[0].Should().Be("{opcode: \"query\", requestId: " + actualRequestId0 + ", database: \"source\", collection: \"$cmd\", batchSize: -1, slaveOk: true, query: {getnonce: 1}}");
             sentMessages[1].Should().Be("{opcode: \"query\", requestId: " + actualRequestId1 + ", database: \"source\", collection: \"$cmd\", batchSize: -1, slaveOk: true, query: {authenticate: 1, user: \"user\", nonce: \"2375531c32080ae8\", key: \"21742f26431831d5cfca035a08c5bdf6\"}}");
         }
+
+        [Theory]
+        [ParameterAttributeData]
+        public void Authenticate_should_send_serverApi_with_command_wire_protocol(
```
DmitryLukyanov, line 130: can query/command wire protocol be just a new Values argument (similar in other places)?

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
```diff
@@ -122,5 +124,99 @@ public void Constructor_should_throw_an_ArgumentNullException_when_credential_is
...
+            connection.EnqueueReplyMessage(getNonceReply);
+            connection.EnqueueReplyMessage(authenticateReply);
+            connection.Description = __descriptionQueryWireProtocol;
+
+            var expectedRequestId = RequestMessage.CurrentGlobalRequestId + 1;
+
+            Action act;
+            if (async)
+            {
+                act = () => subject.AuthenticateAsync(connection, __descriptionQueryWireProtocol, CancellationToken.None).GetAwaiter().GetResult();
```
DmitryLukyanov, line 199: I would prefer a new testing way here, but I will defer it to you

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -722,6 +723,46 @@ public void Command_should_update_the_session_and_cluster_cluster_times()
                 server.ClusterClock.ClusterTime.Should().Be(actualClusterTime);
             }
         }
+
+        [SkippableFact]
+        public void Command_should_use_serverApi()
```
DmitryLukyanov, line 728: async?

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
```diff
@@ -1013,6 +1019,21 @@ public void TestServer()
...
+            var settings = new MongoClientSettings();
+            Assert.Equal(null, settings.ServerApi);
+
+            var serverApi = new ServerApi(ServerApiVersion.V1, true, true);
+            settings.ServerApi = serverApi;
+            Assert.Equal(serverApi, settings.ServerApi);
+
+            settings.Freeze();
+            Assert.Equal(serverApi, settings.ServerApi);
+            Assert.Throws<InvalidOperationException>(() => { settings.ServerApi = serverApi; });
```
DmitryLukyanov, line 1034: let's avoid using an old testing format. I understand that probably this was copied from somewhere, but it's quite a small test to change it

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatTestRunner.cs
```diff
@@ -38,33 +38,51 @@ public sealed class UnifiedTestFormatTestRunner : IDisposable
...
+            BsonArray runOnRequirements,
+            string skipReason,
+            BsonArray operations,
+            BsonArray expectedEvents,
+            BsonArray outcome,
+            bool async)
         {
-            if (!schemaVersion.StartsWith("1.0"))
+            var schemaSemanticVersion = SemanticVersion.Parse(schemaVersion);
+            if (schemaSemanticVersion < new SemanticVersion(1, 0, 0) ||
```
DmitryLukyanov, line 70: should it be placed into RequireServer or similar?

src/MongoDB.Driver.Core/Core/Authentication/MongoDBCRAuthenticator.cs
```diff
@@ -113,7 +124,7 @@ public BsonDocument CustomizeInitialIsMasterCommand(BsonDocument isMasterCommand
         }
 
         // private methods
-        private CommandWireProtocol<BsonDocument> CreateAuthenticateProtocol(BsonDocument getNonceReply)
+        private CommandWireProtocol<BsonDocument> CreateAuthenticateProtocol(BsonDocument getNonceReply, ServerApi serverApi)
```
DmitryLukyanov, line 127: nit: technically this is not required since we have a local `_serverApi`, I would use it in the same way everywhere, check similar issues

DmitryLukyanov (commented):
Some comments and then it will LGTM

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -80,18 +81,47 @@ internal sealed class Server : IClusterableServer
...
+                 settings,
+                 endPoint,
+                 connectionPoolFactory,
+                 serverMonitorFactory,
+                 eventSubscriber,
+                 null)
+        {
+        }
+
+        public Server(
```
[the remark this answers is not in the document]
MikalaiMazurenka: Missed this one.
Done

src/MongoDB.Driver.Core/ServerApi.cs
```diff
@@ -0,0 +1,124 @@
...
+
+using System;
+using System.Text;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Shared;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Represents a server api.
```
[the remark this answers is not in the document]
MikalaiMazurenka: Missed one. Good catch!
Done

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: I've already tried that and it makes test less readable because of the number of `if` `else` clauses. So potentially yes, but I'd avoid that.

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: I stick to existing format in the file.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
MikalaiMazurenka: Great catch!
Done

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
MikalaiMazurenka: I merely use existing file format.

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatTestRunner.cs
MikalaiMazurenka: No. Since it is not an attribute of environment but of an implementation.

src/MongoDB.Driver.Core/Core/Authentication/GssapiAuthenticator.cs
```diff
@@ -111,7 +110,21 @@ public static string ServiceRealmPropertyName
...
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GssapiAuthenticator"/> class.
+        /// </summary>
+        /// <param name="credential">The credential.</param>
+        /// <param name="properties">The properties.</param>
+        /// <param name="serverApi">The server API.</param>
+        public GssapiAuthenticator(
```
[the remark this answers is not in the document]
MikalaiMazurenka: Done.
Updated all authenticators except for default.

MikalaiMazurenka (commented):
Ready for review!

src/MongoDB.Driver.Core/Core/Authentication/MongoDBCRAuthenticator.cs
DmitryLukyanov: I would say that it's better to use the same approach everywhere, for example, you don't use this mechanic for `_credential`. And even more, with your suggestion, you will need to pass all global variables through all call stack which doesn't look right and we definitely don't do this in other places. If you strongly feel that your approach is better, I will rely on you and @rstam here

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatTestRunner.cs
DmitryLukyanov: yeah, got it, thanks

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
DmitryLukyanov: it depends on which test you took for example, in this file, there are two approached. The general rule, use a new approach for new tests

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -722,6 +734,66 @@ public void Command_should_update_the_session_and_cluster_cluster_times()
...
+                            DatabaseNamespace.Admin,
+                            command,
+                            null, // payloads
+                            NoOpElementNameValidator.Instance,
+                            null, // additionalOptions
+                            null, // postWriteAction
+                            CommandResponseHandling.Return,
+                            BsonDocumentSerializer.Instance,
+                            new MessageEncoderSettings(),
+                            cancellationToken).GetAwaiter().GetResult();
```
DmitryLukyanov, line 789: expected formatting here:

                        channel.CommandAsync(
                           ...
                            cancellationToken)
                        .GetAwaiter()
                        .GetResult();

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
DmitryLukyanov: this file contains different formats

DmitryLukyanov: And actually, I missed that you mix them. Then, I think you should choose the latest one

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
DmitryLukyanov: Not sure that making a full copy of these test (except some input/assert arguments) makes these tests more clear. Consider this refactoring:

        [Theory]
        [ParameterAttributeData]
        public void Authenticate_should_send_serverApi(
            [Values(false, true)] bool useServerApi,
            [Values(false, true)] bool useQueryProtocol,
        {
            var dateTime = DateTime.UtcNow;
            var clientNonce = __randomByteGenerator.Generate(ClientNonceLength);
            var serverNonce = Combine(clientNonce, __randomByteGenerator.Generate(ClientNonceLength));
            var host = "sts.amazonaws.com";
            var credential = new UsernamePasswordCredential("$external", "permanentuser", "FAKEFAKEFAKEFAKEFAKEfakefakefakefakefake");
            var serverApi = useServerApi ? new ServerApi(ServerApiVersion.V1, true, true) : null;

            AwsSignatureVersion4.CreateAuthorizationRequest(
                dateTime,
                credential.Username,
                credential.Password,
                null,
                serverNonce,
                host,
                out var authHeader,
                out var timestamp);

            var mockClock = new Mock<IClock>();
            mockClock.Setup(x => x.UtcNow).Returns(dateTime);

            var mockRandomByteGenerator = new Mock<IRandomByteGenerator>();
            mockRandomByteGenerator.Setup(x => x.Generate(It.IsAny<int>())).Returns(clientNonce);

            var expectedClientFirstMessage = new BsonDocument
            {
                { "r", clientNonce },
                { "p", (int)'n' }
            };
            var expectedClientSecondMessage = new BsonDocument
            {
                { "a", authHeader },
                { "d", timestamp }
            };
            var serverFirstMessage = new BsonDocument
            {
                { "s", serverNonce },
                { "h", host }
            };

            var saslStartReplyString = $"{{ conversationId : 1, done : false, payload : BinData(0,\"{ToBase64(serverFirstMessage.ToBson())}\"), ok : 1 }}";
            var saslContinueReplyString = "{ conversationId : 1, done : true, payload : BinData(0,\"\"), ok : 1}";

            var subject = new MongoAWSAuthenticator(credential, null, mockRandomByteGenerator.Object, mockClock.Object, serverApi);

            var connection = new MockConnection(__serverId);
            connection.Description = ConfigureConnection();

            if (async)
            {
                subject.AuthenticateAsync(connection, connection.Description, CancellationToken.None).GetAwaiter().GetResult();
            }
            else
            {
                subject.Authenticate(connection, connection.Description, CancellationToken.None);
            }

            SpinWait.SpinUntil(() => connection.GetSentMessages().Count >= 2, TimeSpan.FromSeconds(5)).Should().BeTrue();

            var sentMessages = MessageHelper.TranslateMessagesToBsonDocuments(connection.GetSentMessages());
            sentMessages.Count.Should().Be(2);

            var actualRequestId0 = sentMessages[0]["requestId"].AsInt32;
            var actualRequestId1 = sentMessages[1]["requestId"].AsInt32;

            var expectedMessages = GetExpectedMessages();
            sentMessages[0].Should().Be(expectedMessages.ExpectedStartMessage);
            sentMessages[1].Should().Be(expectedMessages.ExpectedContinueMessage);

            ConnectionDescription ConfigureConnection()
            {
                if (useQueryProtocol)
                {
                    var saslStartReply = MessageHelper.BuildReply(RawBsonDocumentHelper.FromJson(saslStartReplyString));
                    var saslContinueReply = MessageHelper.BuildReply(RawBsonDocumentHelper.FromJson(saslContinueReplyString));
                    connection.EnqueueReplyMessage(saslStartReply);
                    connection.EnqueueReplyMessage(saslContinueReply);
                    return __descriptionQueryWireProtocol;
                }
                else
                {
                    var saslStartResponse = MessageHelper.BuildCommandResponse(RawBsonDocumentHelper.FromJson(saslStartReplyString));
                    var saslContinueResponse = MessageHelper.BuildCommandResponse(RawBsonDocumentHelper.FromJson(saslContinueReplyString));
                    connection.EnqueueCommandResponseMessage(saslStartResponse);
                    connection.EnqueueCommandResponseMessage(saslContinueResponse);
                    return __descriptionCommandWireProtocol;
                }
            }

            (string ExpectedStartMessage, string ExpectedContinueMessage) GetExpectedMessages()
            {
                var expectedServerApiString = useServerApi ? ", apiVersion : \"1\", apiStrict : true, apiDeprecationErrors : true" : "";
                string expectedFirstMessage;
                string expectedSecondMessage;
                if (useQueryProtocol)
                {
                    expectedFirstMessage = GetExpectedSaslStartQueryMessage(actualRequestId0, expectedClientFirstMessage, expectedServerApiString);
                    expectedSecondMessage = GetExpectedSaslContinueQueryMessage(actualRequestId1, expectedClientSecondMessage, expectedServerApiString);
                }
                else
                {
                    expectedFirstMessage = GetExpectedSaslStartCommandMessage(actualRequestId0, expectedClientFirstMessage, expectedServerApiString);
                    expectedSecondMessage = GetExpectedSaslContinueCommandMessage(actualRequestId1, expectedClientSecondMessage, expectedServerApiString);
                }
                return (expectedFirstMessage, expectedSecondMessage);
            }
        }

DmitryLukyanov: NOTE: I think all logic from here: `ConfigureConnection` can be moved into `MessageHelper` like:

    MessageHelper.BuildResponse<ReplyMessage>(...)
    MessageHelper.BuildResponse<CommandResponseMessage>(...)

and we can decide which message should be processed based on the T argument

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
```diff
@@ -122,5 +124,99 @@ public void Constructor_should_throw_an_ArgumentNullException_when_credential_is
...
+            if (async)
+            {
+                act = () => subject.AuthenticateAsync(connection, __descriptionCommandWireProtocol, CancellationToken.None).GetAwaiter().GetResult();
+            }
+            else
+            {
+                act = () => subject.Authenticate(connection, __descriptionCommandWireProtocol, CancellationToken.None);
+            }
+
+            act.ShouldNotThrow();
```
DmitryLukyanov, line 159: this is not necessary, since we can take (and do it in other places) the fact that the test hasn't failed as the flag that nothing was failed :)

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: Conditional asserts in unit tests are harder to maintain and understand. I would like to hear @rstam opinion on this before implementing these changes.

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: I removed it here and in similar places.
Done.

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: This file contains one format, which I am using.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
MikalaiMazurenka: Done

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
MikalaiMazurenka: All of the tests which assert a particular option of `MongoClientSettings` use `Assert.Equal` format which I am using. Mixing the formats will reduce readability.

MikalaiMazurenka (commented):
Ready for review!

src/MongoDB.Driver.Core/ServerApi.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// </summary>
+    public class ServerApi : IEquatable<ServerApi>
+    {
+        // fields
+        private readonly bool? _deprecatationErrors;
```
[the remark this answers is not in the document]
rstam: Fix typo in field name, should be:

    _deprecationErrors

Make sure it's fixed everywhere.

rstam (changes requested):
Reviewed it rapidly and looks good initially.

I will review once more a little more carefully once someone else has given it the first approval.

src/MongoDB.Driver.Core/ServerApi.cs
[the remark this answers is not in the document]
MikalaiMazurenka: Fixed this typo everywhere.
Done

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
MikalaiMazurenka: Switched new tests to new format.
Done.

tests/MongoDB.Driver.Tests/MongoClientSettingsTests.cs
MikalaiMazurenka: Switched new tests to new format.
Done.

MikalaiMazurenka (commented):
Ready for review!

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -722,6 +734,69 @@ public void Command_should_update_the_session_and_cluster_cluster_times()
...
+
+            using (var cluster = CoreTestConfiguration.CreateCluster(builder))
+            using (var session = cluster.StartSession())
+            {
+                var cancellationToken = CancellationToken.None;
+                var server = (Server)cluster.SelectServer(WritableServerSelector.Instance, cancellationToken);
+                using (var channel = server.GetChannel(cancellationToken))
+                {
+                    var command = BsonDocument.Parse("{ ping : 1 }");
+                    if (async)
```
DmitryLukyanov, line 759: should be inverted

DmitryLukyanov (approved):
Some typo fix and then LGTM
@rstam, there are a few comments where you're mentioned, it would be good to hear your opinion there

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
MikalaiMazurenka: Done

src/MongoDB.Driver.Core/Core/Authentication/DefaultAuthenticator.cs
```diff
@@ -43,14 +43,29 @@ public class DefaultAuthenticator : IAuthenticator
         /// </summary>
         /// <param name="credential">The credential.</param>
         public DefaultAuthenticator(UsernamePasswordCredential credential)
-            : this(credential, new DefaultRandomStringGenerator())
+            : this(credential, new DefaultRandomStringGenerator(), serverApi: null)
```
rstam, line 46: Why not remove this constructor?

Seems like defaulting `serverApi` to `null` makes it too easy to forget to call the constructor that takes an explicit `serverApi` value (which might be `null` of course).

Same for all other places where you default `serverApi: null`.

src/MongoDB.Driver.Core/ServerApi.cs
```diff
@@ -0,0 +1,124 @@
...
+        private readonly ServerApiVersion _version;
+
+        // constructors
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServerApi" /> class.
+        /// </summary>
+        /// <param name="version">The server API version.</param>
+        /// <param name="strict">The flag for strict server API version enforcement.</param>
+        /// <param name="deprecationErrors">The flag for treating deprecated server APIs as errors.</param>
+        public ServerApi(ServerApiVersion version, Optional<bool?> strict = default(Optional<bool?>), Optional<bool?> deprecationErrors = default(Optional<bool?>))
```
rstam, line 40: This can be shortened to:

    public ServerApi(ServerApiVersion version, Optional<bool?> strict = default, Optional<bool?> deprecationErrors = default)

rstam (changes requested):
Possible changes? Not sure. See comments.

rstam:
@DmitryLukyanov I did not review the tests. I'm assuming you did.

MikalaiMazurenka (commented):
Ready for review!
@rstam could you please check the comments where you mentioned to help with decisions.

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -400,6 +411,25 @@ bool IsSessionAcknowledged()
                     return true;
                 }
             }
+
+            void AddServerApiElementsIfNecessaryAndNotAlreadyAdded()
+            {
+                var commandName = _command.GetElement(0).Name;
+                if (commandName == "getMore")
```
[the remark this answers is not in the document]
rstam: I don't have a better solution either.

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoDBCRAuthenticatorTests.cs
rstam: Agreed that the simpler the test is the better.

I defer to you two what's simplest here.

src/MongoDB.Driver.Core/Core/Authentication/MongoDBCRAuthenticator.cs
rstam: > It is generally preferable to explicitly specify local parameters whenever possible which increases readability and reusability of code.

Maybe, but when I see this parameter it seems like the reason for adding it is because you want to pass in an arbitrary value. If you always pass in `_serverApi` then having it as a parameter here seems redundant (and even confusing because it implies the wrong thing).

src/MongoDB.Driver.Core/Core/Authentication/DefaultAuthenticator.cs
rstam: Yes of course I understand the goal of non-breaking.

But this isn't a class an application would ever call... and leaving the old constructor in place makes it easy to forget to pass the `serverApi` which would be a bug now.

I leave it up to you (and maybe @DmitryLukyanov could comment also) whether you want to prioritize backward compatibility (in a class an application wouldn't ever use) or safety.

DmitryLukyanov: Let's make old ctors obsolete?

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatTestRunner.cs
DmitryLukyanov: I checked how it looks on different tickets. I think we should create a new test unified runner for each ticket, like a new CrudUnifiedTestRunner and so on, that will use this test runner inside.

DmitryLukyanov: nit: Also, it looks like this factory (together with ClassData method) also should be moved from this class into a new one. And after all, you will have one common class(without mentions of particular tests) which you will reuse everywhere, and also two specific test runners for `unified` logic. I think this is out of scope for your PR, so see whether you want to work on it now or no

MikalaiMazurenka: Moved `versioned-api` tests into separate test runner. Further improvements require too many changes, so should be made in scope of a dedicated story.

MikalaiMazurenka (commented):
Ready for review!

rstam (approved):
LGTM

PR #446, opened 2021-01-29
CSHARP-3321: Sharing a MongoClient for metadata lookup can lead to deadlock in drivers using automatic encryption.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/60148e575623433e9c236559

### DIFF
--- a/src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
+++ b/src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
@@ -28,25 +28,60 @@ namespace MongoDB.Driver.Encryption
 {
     internal sealed class AutoEncryptionLibMongoCryptController : LibMongoCryptControllerBase, IBinaryDocumentFieldDecryptor, IBinaryCommandFieldEncryptor
     {
+        #region static
+        public static AutoEncryptionLibMongoCryptController Create(IMongoClient client, CryptClient cryptClient, AutoEncryptionOptions autoEncryptionOptions)
+        {
+            var lazyInternalClient = new Lazy<IMongoClient>(() => CreateInternalClient());
+            var keyVaultClient = autoEncryptionOptions.KeyVaultClient ?? lazyInternalClient.Value;
+            var metadataClient = autoEncryptionOptions.BypassAutoEncryption ? null : lazyInternalClient.Value;
+            var internalClient = lazyInternalClient.IsValueCreated ? lazyInternalClient.Value : null;
+
+            return new AutoEncryptionLibMongoCryptController(
+                internalClient,
+                keyVaultClient,
+                metadataClient,
+                cryptClient,
+                autoEncryptionOptions);
+
+            IMongoClient CreateInternalClient()
+            {
+                var internalClientSettings = client.Settings.Clone();
+                internalClientSettings.AutoEncryptionOptions = null;
+                internalClientSettings.MinConnectionPoolSize = 0;
+                return new MongoClient(internalClientSettings);
+            }
+        }
+        #endregion
+
         // private fields
-        private readonly IMongoClient _client;
+        private readonly IMongoClient _internalClient;
+        private readonly IMongoClient _metadataClient;
         private readonly IMongoClient _mongocryptdClient;
         private readonly MongocryptdFactory _mongocryptdFactory;
 
         // constructors
-        public AutoEncryptionLibMongoCryptController(
-            IMongoClient client,
+        private AutoEncryptionLibMongoCryptController(
+            IMongoClient internalClient,
+            IMongoClient keyVaultClient,
+            IMongoClient metadataClient,
             CryptClient cryptClient,
             AutoEncryptionOptions autoEncryptionOptions)
             : base(
                   Ensure.IsNotNull(cryptClient, nameof(cryptClient)),
-                  Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultClient ?? client,
+                  Ensure.IsNotNull(keyVaultClient, nameof(keyVaultClient)),
                   Ensure.IsNotNull(Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultNamespace, nameof(autoEncryptionOptions.KeyVaultNamespace)))
         {
-            _client = Ensure.IsNotNull(client, nameof(client)); // _client might not be fully constructed at this point, don't call any instance methods on it yet
+            _internalClient = internalClient; // can be null
+            _metadataClient = metadataClient; // can be null
             _mongocryptdFactory = new MongocryptdFactory(autoEncryptionOptions.ExtraOptions);
             _mongocryptdClient = _mongocryptdFactory.CreateMongocryptdClient();
         }
+
+        // internal properties
+        /// <summary>
+        /// this property is used by DisposableMongoClient.Dispose to unregister the internal cluster.
+        /// </summary>
+        internal IMongoClient InternalClient => _internalClient;
 
         // public methods
         public byte[] DecryptFields(byte[] encryptedDocumentBytes, CancellationToken cancellationToken)
@@ -153,7 +188,13 @@ namespace MongoDB.Driver.Encryption
         // private methods
         private void ProcessNeedCollectionInfoState(CryptContext context, string databaseName, CancellationToken cancellationToken)
         {
-            var database = _client.GetDatabase(databaseName);
+            if (_metadataClient == null)
+            {
+                // should not be reached
+                throw new InvalidOperationException("Metadata client is null.");
+            }
+
+            var database = _metadataClient.GetDatabase(databaseName);
             var filterBytes = context.GetOperation().ToArray();
             var filterDocument = new RawBsonDocument(filterBytes);
             var filter = new BsonDocumentFilterDefinition<BsonDocument>(filterDocument);
@@ -165,7 +206,13 @@ namespace MongoDB.Driver.Encryption
 
         private async Task ProcessNeedCollectionInfoStateAsync(CryptContext context, string databaseName, CancellationToken cancellationToken)
         {
-            var database = _client.GetDatabase(databaseName);
+            if (_metadataClient == null)
+            {
+                // should not be reached
+                throw new InvalidOperationException("Metadata client is null.");
+            }
+
+            var database = _metadataClient.GetDatabase(databaseName);
             var filterBytes = context.GetOperation().ToArray();
             var filterDocument = new RawBsonDocument(filterBytes);
             var filter = new BsonDocumentFilterDefinition<BsonDocument>(filterDocument);
--- a/tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
+++ b/tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
@@ -18,10 +18,12 @@ using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Bson;
 using MongoDB.Driver.Core.Clusters;
-using MongoDB.Driver.Core.Operations;
 
 namespace MongoDB.Driver.TestHelpers
 {
+    /// <summary>
+    /// This class is used in tests to close down the MongoClient cluster (and also the cluster for the internal MongoClient if there is one).
+    /// </summary>
     public class DisposableMongoClient : IMongoClient, IDisposable
     {
         private readonly IMongoClient wrapped;
@@ -237,6 +239,15 @@ namespace MongoDB.Driver.TestHelpers
         public void Dispose()
         {
             ClusterRegistry.Instance.UnregisterAndDisposeCluster(wrapped.Cluster);
+
+            if (wrapped is MongoClient mongoClient)
+            {
+                var internalClient = mongoClient.LibMongoCryptController?.InternalClient;
+                if (internalClient != null)
+                {
+                    ClusterRegistry.Instance.UnregisterAndDisposeCluster(internalClient.Cluster);
+                }
+            }
         }
     }
 }
--- a/tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
+++ b/tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
@@ -26,7 +26,6 @@ using MongoDB.Bson.TestHelpers.XunitExtensions;
 using MongoDB.Driver.Core;
 using MongoDB.Driver.Core.Bindings;
 using MongoDB.Driver.Core.Clusters;
-using MongoDB.Driver.Core.Configuration;
 using MongoDB.Driver.Core.Events;
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Core.Operations;
@@ -81,7 +80,7 @@ namespace MongoDB.Driver.Tests.Specifications.client_side_encryption.prose_tests
                 .SkipWhen(SupportedOperatingSystem.Linux, SupportedTargetFramework.NetStandard15)
                 .SkipWhen(SupportedOperatingSystem.MacOS, SupportedTargetFramework.NetStandard15);
 
-            var eventCapturer = new EventCapturer().Capture<CommandStartedEvent>(e => e.CommandName == "insert");
+            var eventCapturer = CreateEventCapturer(commandNameFilter: "insert");
             using (var client = ConfigureClient())
             using (var clientEncrypted = ConfigureClientEncrypted(kmsProviderFilter: "local", eventCapturer: eventCapturer))
             {
@@ -722,6 +721,213 @@ namespace MongoDB.Driver.Tests.Specifications.client_side_encryption.prose_tests
         }
 
         [SkippableTheory]
+        [MemberData(nameof(DeadlockTest_MemberData))]
+        public void DeadlockTest(
+            string _,
+            int maxPoolSize,
+            bool bypassAutoEncryption,
+            string keyVaultMongoClientKey,
+            int expectedNumberOfClients,
+            string[] clientEncryptedEventsExpectation,
+            string[] clientKeyVaultEventsExpectation,
+            bool async)
+        {
+            RequireServer.Check().Supports(Feature.ClientSideEncryption);
+            RequirePlatform
+                .Check()
+                .SkipWhen(SupportedOperatingSystem.Linux, SupportedTargetFramework.NetStandard15)
+                .SkipWhen(SupportedOperatingSystem.MacOS, SupportedTargetFramework.NetStandard15);
+
+            var clientKeyVaultEventCapturer = CreateEventCapturer();
+            using (var client_keyvault = CreateMongoClient(maxPoolSize: 1, writeConcern: WriteConcern.WMajority, readConcern: ReadConcern.Majority, eventCapturer: clientKeyVaultEventCapturer))
+            using (var client_test = ConfigureClient(clearCollections: true, writeConcern: WriteConcern.WMajority, readConcern: ReadConcern.Majority))
+            {
+                var dataKeysCollection = GetCollection(client_test, __keyVaultCollectionNamespace);
+                var externalKey = JsonFileReader.Instance.Documents["external.external-key.json"];
+                Insert(dataKeysCollection, async, externalKey);
+
+                var externalSchema = JsonFileReader.Instance.Documents["external.external-schema.json"];
+                CreateCollection(client_test, __collCollectionNamespace, new BsonDocument("$jsonSchema", externalSchema));
+
+                using (var client_encryption = ConfigureClientEncryption(client_test.Wrapped as MongoClient, kmsProviderFilter: "local"))
+                {
+                    var value = "string0";
+                    var encryptionOptions = new EncryptOptions(
+                        algorithm: EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic.ToString(),
+                        alternateKeyName: "local");
+                    var ciphertext = ExplicitEncrypt(client_encryption, encryptionOptions, value, async);
+
+                    var eventCapturer = CreateEventCapturer().Capture<ClusterOpeningEvent>();
+
+                    using (var client_encrypted = ConfigureClientEncrypted(
+                        kmsProviderFilter: "local",
+                        maxPoolSize: maxPoolSize,
+                        bypassAutoEncryption: bypassAutoEncryption,
+                        eventCapturer: eventCapturer,
+                        externalKeyVaultClient: GetKeyVaultMongoClientByKey()))
+                    {
+                        IMongoCollection<BsonDocument> collCollection;
+                        if (client_encrypted.Settings.AutoEncryptionOptions.BypassAutoEncryption)
+                        {
+                            collCollection = GetCollection(client_test, __collCollectionNamespace);
+                            Insert(
+                                collCollection,
+                                async,
+                                new BsonDocument
+                                {
+                                    { "_id", 0 },
+                                    { "encrypted", ciphertext }
+                                });
+                        }
+                        else
+                        {
+                            collCollection = GetCollection(client_encrypted, __collCollectionNamespace);
+                            Insert(
+                                collCollection,
+                                async,
+                                new BsonDocument
+                                {
+                                    { "_id", 0 },
+                                    { "encrypted", value }
+                                });
+                        }
+
+                        collCollection = GetCollection(client_encrypted, __collCollectionNamespace);
+                        var findResult = Find(collCollection, BsonDocument.Parse("{ _id : 0 }"), async).Single();
+                        findResult.Should().Be($"{{ _id : 0, encrypted : '{value}' }}");
+                        var events = eventCapturer.Events.ToList();
+                        AssertEvents(events.OfType<CommandStartedEvent>(), clientEncryptedEventsExpectation);
+                        if (clientKeyVaultEventsExpectation != null)
+                        {
+                            AssertEvents(clientKeyVaultEventCapturer.Events.OfType<CommandStartedEvent>(), clientKeyVaultEventsExpectation);
+                        }
+
+                        AssertNumberOfClients(events.OfType<ClusterOpeningEvent>());
+                    }
+                }
+
+                IMongoClient GetKeyVaultMongoClientByKey()
+                {
+                    switch (keyVaultMongoClientKey)
+                    {
+                        case "client_keyvault":
+                            return client_keyvault;
+                        default:
+                            return null;
+                    }
+                }
+            }
+
+            void AssertEvents(IEnumerable<CommandStartedEvent> events, string[] expectedEventsDetails)
+            {
+                for (int i = 0; i < expectedEventsDetails.Length; i++)
+                {
+                    var arguments = expectedEventsDetails[i].Split(';');
+                    (string CommandName, string Database) expectedEventDetails = (arguments[0], arguments[1]);
+                    var @event = events.ElementAt(i);
+                    @event.DatabaseNamespace.DatabaseName.Should().Be(expectedEventDetails.Database);
+                    @event.CommandName.Should().Be(expectedEventDetails.CommandName);
+                }
+                events.Count().Should().Be(expectedEventsDetails.Count());
+            }
+
+            void AssertNumberOfClients(IEnumerable<ClusterOpeningEvent> events)
+            {
+                events.Count().Should().Be(expectedNumberOfClients);
+            }
+        }
+
+        public static IEnumerable<object[]> DeadlockTest_MemberData()
+        {
+            var testCases = new List<object[]>();
+            testCases.AddRange(
+                CasesWithAsync(
+                    name: "case 1",
+                    maxPoolSize: 1,
+                    bypassAutoEncryption: false,
+                    keyVaultMongoClient: null,
+                    expectedNumberOfClients: 2,
+                    clientEncryptedEventsExpectation:
+                    new[]
+                    {
+                        "listCollections;db",
+                        "find;keyvault",
+                        "insert;db",
+                        "find;db"
+                    },
+                    clientKeyVaultEventsExpectation: null));
+            testCases.AddRange(
+                CasesWithAsync(
+                    name: "case 2",
+                    maxPoolSize: 1,
+                    bypassAutoEncryption: false,
+                    keyVaultMongoClient: "client_keyvault",
+                    expectedNumberOfClients: 2,
+                    clientEncryptedEventsExpectation:
+                    new[]
+                    {
+                        "listCollections;db",
+                        "insert;db",
+                        "find;db"
+                    },
+                    clientKeyVaultEventsExpectation:
+                    new[]
+                    {
+                        "find;keyvault"
+                    }));
+            testCases.AddRange(
+                CasesWithAsync(
+                    name: "case 3",
+                    maxPoolSize: 1,
+                    bypassAutoEncryption: true,
+                    keyVaultMongoClient: null,
+                    expectedNumberOfClients: 2,
+                    clientEncryptedEventsExpectation:
+                    new[]
+                    {
+                        "find;db",
+                        "find;keyvault"
+                    },
+                    clientKeyVaultEventsExpectation: null));
+            testCases.AddRange(
+                CasesWithAsync(
+                    name: "case 4",
+                    maxPoolSize: 1,
+                    bypassAutoEncryption: true,
+                    keyVaultMongoClient: "client_keyvault",
+                    expectedNumberOfClients: 1,
+                    clientEncryptedEventsExpectation:
+                    new[]
+                    {
+                        "find;db",
+                    },
+                    clientKeyVaultEventsExpectation:
+                    new[]
+                    {
+                        "find;keyvault"
+                    }));
+
+            // cases 5-8 use "MaxPoolSize: 0" which is not supported by the c# driver
+
+            return testCases;
+
+            IEnumerable<object[]> CasesWithAsync(
+                string name,
+                int maxPoolSize,
+                bool bypassAutoEncryption,
+                string keyVaultMongoClient,
+                int expectedNumberOfClients,
+                string[] clientEncryptedEventsExpectation,
+                string[] clientKeyVaultEventsExpectation)
+            {
+                foreach (var async in new[] { true, false })
+                {
+                    yield return new object[] { name, maxPoolSize, bypassAutoEncryption, keyVaultMongoClient, expectedNumberOfClients, clientEncryptedEventsExpectation, clientKeyVaultEventsExpectation, async };
+                }
+            }
+        }
+
+        [SkippableTheory]
         [ParameterAttributeData]
         public void ExternalKeyVaultTest(
             [Values(false, true)] bool withExternalKeyVault,
@@ -733,9 +939,17 @@ namespace MongoDB.Driver.Tests.Specifications.client_side_encryption.prose_tests
                 .SkipWhen(SupportedOperatingSystem.Linux, SupportedTargetFramework.NetStandard15)
                 .SkipWhen(SupportedOperatingSystem.MacOS, SupportedTargetFramework.NetStandard15);
 
+            IMongoClient externalKeyVaultClient = null;
+            if (withExternalKeyVault)
+            {
+                var externalKeyVaultClientSettings = DriverTestConfiguration.GetClientSettings().Clone();
+                externalKeyVaultClientSettings.Credential = MongoCredential.FromComponents(null, null, "fake-user", "fake-pwd");
+                externalKeyVaultClient = new MongoClient(externalKeyVaultClientSettings);
+            }
+
             var clientEncryptedSchema = new BsonDocument("db.coll", JsonFileReader.Instance.Documents["external.external-schema.json"]);
             using (var client = ConfigureClient())
-            using (var clientEncrypted = ConfigureClientEncrypted(clientEncryptedSchema, withExternalKeyVault))
+            using (var clientEncrypted = ConfigureClientEncrypted(clientEncryptedSchema, externalKeyVaultClient: externalKeyVaultClient))
             using (var clientEncryption = ConfigureClientEncryption(clientEncrypted.Wrapped as MongoClient))
             {
                 var datakeys = GetCollection(client, __keyVaultCollectionNamespace);
@@ -857,9 +1071,13 @@ namespace MongoDB.Driver.Tests.Specifications.client_side_encryption.prose_tests
         }
 
         // private methods
-        private DisposableMongoClient ConfigureClient(bool clearCollections = true)
+        private DisposableMongoClient ConfigureClient(
+            bool clearCollections = true,
+            int? maxPoolSize = null,
+            WriteConcern writeConcern = null,
+            ReadConcern readConcern = null)
         {
-            var client = CreateMongoClient();
+            var client = CreateMongoClient(maxPoolSize: maxPoolSize, writeConcern: writeConcern, readConcern: readConcern);
             if (clearCollections)
             {
                 var clientKeyVaultDatabase = client.GetDatabase(__keyVaultCollectionNamespace.DatabaseNamespace.DatabaseName);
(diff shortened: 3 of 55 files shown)


src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,24 +28,74 @@ namespace MongoDB.Driver.Encryption
...
     internal sealed class AutoEncryptionLibMongoCryptController : LibMongoCryptControllerBase, IBinaryDocumentFieldDecryptor, IBinaryCommandFieldEncryptor
     {
+        #region static
+        private static IMongoClient CreateInternalClient(IMongoClient client)
+        {
+            var internalClientMongoSettings = client.Settings.Clone();
+            internalClientMongoSettings.AutoEncryptionOptions = null;
+            internalClientMongoSettings.MinConnectionPoolSize = 0;
+            // this step is not mentioned in the spec. Unset this value to be able to reuse the created internal client instances
+            internalClientMongoSettings.ClusterConfigurator = null;
```
DmitryLukyanov, line 38: not sure that this line is required and safe. The current test design doesn't call this method, so in theory, we can remove this line (without affecting tets), but if anyone will configure ClusterConfigurator, he won't be able to reuse internalClient, if he creates more than one parent mongo client, however leaving this line as is can also be dangerous.

DmitryLukyanov: So, probably this should be removed, and maybe instead we can document this somewhere, but I would like to use another set of eyes here before.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,24 +28,74 @@ namespace MongoDB.Driver.Encryption
...
+        }
+
+        private static Lazy<IMongoClient> CreateLazyInternalClient(IMongoClient client, IMongoClient internalMongoClient)
+        {
+            return new Lazy<IMongoClient>(() =>
+            {
+                return
+                    // this client that provided externally is out of the spec and is used only internally,
+                    // mostly to be able to dispose the internal client in tests
+                    internalMongoClient ??
```
DmitryLukyanov, line 49: this client is added to have control over the used client here, in particular, to be able to dispose of the internal client after each test. Otherwise, after a big enough number of launched tests, the tests become really slow down and, finally, starts to fail

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -153,7 +203,7 @@ protected override async Task ProcessStateAsync(CryptContext context, string dat
         // private methods
         private void ProcessNeedCollectionInfoState(CryptContext context, string databaseName, CancellationToken cancellationToken)
         {
-            var database = _client.GetDatabase(databaseName);
+            var database = _metadataClient.GetDatabase(databaseName);
```
DmitryLukyanov, line 206: I will add isNotNull validation here in the next commit

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -154,7 +155,13 @@ protected override bool TryConfigureClientOption(MongoClientSettings settings, B
             switch (option.Name)
             {
                 case "autoEncryptOpts":
-                    settings.AutoEncryptionOptions = ConfigureAutoEncryptionOptions(option.Value.AsBsonDocument);
+                    {
+                        var autoEncryptionOptions = ConfigureAutoEncryptionOptions(option.Value.AsBsonDocument);
+                        var disposableMongoClient = CreateDisposableInternalClient(settings.ClusterConfigurator);
+                        AddDisposable(disposableMongoClient);
```
DmitryLukyanov, line 161: the used client will be disposed after each test

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -215,6 +222,18 @@ private AutoEncryptionOptions ConfigureAutoEncryptionOptions(BsonDocument autoEn
             return autoEncryptionOptions;
         }
 
+        private DisposableMongoClient CreateDisposableInternalClient(Action<ClusterBuilder> clusterBuilder)
+        {
+            var internalClientMongoSettings = DriverTestConfiguration.GetClientSettings().Clone();
+            // mimic the driver-side logic 
+            internalClientMongoSettings.AutoEncryptionOptions = null;
```
DmitryLukyanov, line 229: this logic can be moved into a separate static method, then we will be able to reuse it on the tests side

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +722,250 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+            }
+
+            Mock<DisposableMongoClient> CreateDisposableInternalClientMock(MongoClientSettings clientSettings)
+            {
+                var internalClientMongoSettings = clientSettings.Clone();
+                // mimic the driver-side logic 
+                internalClientMongoSettings.AutoEncryptionOptions = null;
+                internalClientMongoSettings.MinConnectionPoolSize = 0;
+
+                return new Mock<DisposableMongoClient>(new MongoClient(internalClientMongoSettings))
```
DmitryLukyanov, line 868: in the current test implementation, we create an internal client on the test side all the time regardless we will use this client in the driver or no. So, the way with events doesn't really work for us. But mocking helps with this :)

rstam (changes requested):
I was having a real hard time trying to convince myself these changes were correct and how to reconcile them with what the spec said.

Would you please consider an alternate approach prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3321-mongoclientapproach

The main points of this alternate approach are:

1. Let MongoClient create and own the internal client just like the spec says
2. Modify DisposableMongoClient to automatically clean up the cluster for the internal client (if any)

Note: the somewhat awkard use of an internal interface is because C# does not let you define internal members of an interface. The entire interface has to be either public or internal.

Let me know what you think. Thanks!

rstam:
While I have no concerns in principle about `MongoClient` creating and owning the internal client, I do appreciate your concerns about the `IMongoClientInternal` interface approach. It's a limitation of the C# language that we can't add internal methods to an interface, so there is no less awkward way to add internal methods to `IMongoClient`.

I would be OK I guess with your suggestion that `AutoEncryptionLibMongoCryptController` be the one to create and own the internal client. Since `MongoClient` has at most one instance of an `AutoEncryptionLibMongoCryptController`, letting the controller create and own the internal client is the same as letting `MongoClient` create and own the internal client just with one level of indirection.

I refactored my first `MongoClient` based proposal to move the creation and owning of the internal client to `AutoEncryptionLibMongoCryptController` and came up with:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3321-controllerapproach

Take a look and see what you think. With these changes it's much closer to what you have proposed, but with some differences that I do consider important:

* no changes to the `AutoEncryptionLibMongoCryptController` constructor parameters
* delay creation of the internal client until it is first needed
* don't pass in an internal client from tests , let the driver create the internal client as it normally would
* so therefore no changes to `AutoEncryptionOptions` (no `InternalClient` property)
* let `DisposableMongoClient` clean up the internal client also 

I haven't made any changes to the tests. I assume they would need to be refactored also.

jyemin:
> delay creation of the internal client until it is first needed

I'm not sure I agree that this is important.  If an application is configuring client-side encryption, they are almost certainly going to need the internal client, so is there benefit in delaying creation?

DmitryLukyanov:
NOTE: I'm almost finished refactoring my previous version with taking Robert's suggestions into account, I need just to check the prose tests

rstam:
>> delay creation of the internal client until it is first needed

> I'm not sure I agree that this is important. If an application is configuring client-side encryption, they are almost certainly going to need the internal client, so is there benefit in delaying creation?

You have a good point. The subtle issue is that the `AutoEncryptionLibMongoCryptController` is created from the `MongoClient` constructor. That means that the `client` passed in to the `AutoEncryptionLibMongoCryptController` is NOT yet fully constructed and initialized and cannot yet be used.

However, to create the internal client we only need to access the `Settings` of the `MongoClient`, and that should be safe as long as the `_settings` field of `MongoClient` is initialized before the `AutoEncryptionLibMongoCryptController` constructor is called.

So while delaying creating the internal client is 100% safe, creating it up front should also be safe if we are careful.

rstam:
The latest commit of my branch:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3321-controllerapproach

shows what it would look like if we don't delay creating the internal client.

DmitryLukyanov:
> 
> 
> While I have no concerns in principle about `MongoClient` creating and owning the internal client, I do appreciate your concerns about the `IMongoClientInternal` interface approach. It's a limitation of the C# language that we can't add internal methods to an interface, so there is no less awkward way to add internal methods to `IMongoClient`.
> 
> I would be OK I guess with your suggestion that `AutoEncryptionLibMongoCryptController` be the one to create and own the internal client. Since `MongoClient` has at most one instance of an `AutoEncryptionLibMongoCryptController`, letting the controller create and own the internal client is the same as letting `MongoClient` create and own the internal client just with one level of indirection.
> 
> I refactored my first `MongoClient` based proposal to move the creation and owning of the internal client to `AutoEncryptionLibMongoCryptController` and came up with:
> 
> https://github.com/rstam/mongo-csharp-driver/tree/csharp3321-controllerapproach
> 
> Take a look and see what you think. With these changes it's much closer to what you have proposed, but with some differences that I do consider important:
> 
>     * no changes to the `AutoEncryptionLibMongoCryptController` constructor parameters
> 
>     * delay creation of the internal client until it is first needed
> 
>     * don't pass in an internal client from tests , let the driver create the internal client as it normally would
> 
>     * so therefore no changes to `AutoEncryptionOptions` (no `InternalClient` property)
> 
>     * let `DisposableMongoClient` clean up the internal client also
> 
> 
> I haven't made any changes to the tests. I assume they would need to be refactored also.

> * no changes to the `AutoEncryptionLibMongoCryptController` constructor parameters

why? I think I can rewrite code without this, but it will take time and I don't see any benefits with this yet.

> delay creation of the internal client until it is first needed

this will always be done in this way with the exception of tests, but in the current version it's true for all case

>* don't pass in an internal client from tests , let the driver create the internal client as it normally would

done

>* so therefore no changes to `AutoEncryptionOptions` (no `InternalClient` property)

done

> * let `DisposableMongoClient` clean up the internal client also

It could be done as you suggested, but I would not do it. With your suggestion, we move the FLE specific details into the Dispose level which is just a wrapper for a client. The problem here is related to the fact that we have a "gap" in the disposing approaches, in particular, I mean that MongoClient and, in turns, FLE auto controller don't implement IDisposable interface, so we actually have no control over this. So since we cannot fix it, I would not create any specific workaround and instead fix it on the particular test level (when it matters)

>delay creation of the internal client until it is first needed

this is how it's defined in the spec: https://github.com/mongodb/specifications/commit/947be4cf82a8e44f1589d5ac19413d06ebc11e9d#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR311

DmitryLukyanov:
@rstam , please check the latest commit

rstam:
>> no changes to the AutoEncryptionLibMongoCryptController constructor parameters
> why? I think I can rewrite code without this, but it will take time and I don't see any benefits with this yet.

Why? Because the changes you made to how an `AutoEncryptionLibMongoCryptController` is created are unnecessarily complicated (all the static methods).

As for rewriting, did you look at my branch? It already has the proposed simpler changes.

DmitryLukyanov:
ready for review: https://evergreen.mongodb.com/version/601ab85d0305b905dce46f48

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
+            var keyVaultClient = autoEncryptionOptions.KeyVaultClient ?? lazyInternalClient.Value;
+            var metadataClient = autoEncryptionOptions.BypassAutoEncryption ? null : lazyInternalClient.Value;
+            var internalClient = lazyInternalClient.IsValueCreated ? lazyInternalClient.Value : null;
+
+            return new AutoEncryptionLibMongoCryptController(
+                keyVaultClient,
+                metadataClient,
+                cryptClient,
+                autoEncryptionOptions,
+                internalClient);
```
rstam, line 44: The proposed change was:

    return new AutoEncryptionLibMongoCryptController(
        internalClient,
        keyVaultClient,
        metadataClient,
        cryptClient,
        autoEncryptionOptions);

Why did you arbitrarily move the `internalClient` parameter to the end?

We could discuss the order of the three client parameters, but certainly we should keep them together.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
+            }
+        }
+        #endregion
+
         // private fields
-        private readonly IMongoClient _client;
+        private readonly IMongoClient _internalClient;
         private readonly IMongoClient _mongocryptdClient;
         private readonly MongocryptdFactory _mongocryptdFactory;
+        private readonly IMongoClient _metadataClient;
```
rstam, line 60: Keep fields in alphabetical order.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
             _mongocryptdClient = _mongocryptdFactory.CreateMongocryptdClient();
+            _internalClient = internalClient; // can be null
         }
 
+        // internal properties
+        /// <summary>
+        /// This property is not required and is added only to be able to dispose this internal client on the test side.
+        /// This is required only because our MongoClient, and in turns this controller, is not disposable yet,
+        /// so we don't have control over disposing on the driver side
+        /// </summary>
```
rstam, line 85: This comment is superfluous.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
+        internal IMongoClient InternalClient => _internalClient;
```
rstam, line 86: `public` would have been fine because the whole class is `internal`.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
+            IMongoClient keyVaultClient,
+            IMongoClient metadataClient,
             CryptClient cryptClient,
-            AutoEncryptionOptions autoEncryptionOptions)
+            AutoEncryptionOptions autoEncryptionOptions,
+            IMongoClient internalClient)
             : base(
                   Ensure.IsNotNull(cryptClient, nameof(cryptClient)),
-                  Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultClient ?? client,
+                  Ensure.IsNotNull(keyVaultClient, nameof(keyVaultClient)),
```
rstam, line 71: `keyVault` will never be null.

But you can keep this validation if you want to.

tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
```diff
@@ -237,6 +236,17 @@ public IMongoClient WithWriteConcern(WriteConcern writeConcern)
         public void Dispose()
         {
             ClusterRegistry.Instance.UnregisterAndDisposeCluster(wrapped.Cluster);
+
+            // the below is required because MongoClient doesn't implement IDisposable,
+            // so we need to add this workaround to force disposing of some internal resources
```
rstam, line 241: This comment is misleading. Please remove it.

To the extent that it is true it applies to the whole class and not just to the lines below.

rstam: `DisposableMongoClient` is a class used in `Tests` **only** to close down the `Cluster` associated with the `MongoClient`.

Before there was always exactly one `Cluster` owned by the `MongoClient`. Now there **might** be two.

rstam (changes requested):

rstam:
Note: I have not yet started reviewing the tests because we were still working through the implementation itself.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
rstam: No problem

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
rstam: The property may not otherwise be needed, but the `_internalClient` field definitely is (If `MongoClient` and this class was `IDisposable` the `Dispose` method would need to use the `_internalClient` field to `Dispose` the internal client).

rstam: One of my objections to your comment is the judgmental tone.

Let's simplify to:

    // this property is used by DisposableMongoClient.Dispose to unregister the internal cluster

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
rstam: If you insist. It is not common to use `internal` properties when the whole class is `internal`.

tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
rstam: I think you missed my main point: these comments do NOT apply to just the lines below.

Any comments you might add apply to the `DisposableMongoClient` class as a whole.

rstam: If you feel the need to add some sort of comment add the following to the class:

    // This class is used in tests to close down the MongoClient cluster (and also the cluster for the internal MongoClient if there is one)

but please remove the comments on lines 240-241

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,63 @@ namespace MongoDB.Driver.Encryption
...
                   Ensure.IsNotNull(Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultNamespace, nameof(autoEncryptionOptions.KeyVaultNamespace)))
         {
-            _client = Ensure.IsNotNull(client, nameof(client)); // _client might not be fully constructed at this point, don't call any instance methods on it yet
+            _metadataClient = metadataClient;
             _mongocryptdFactory = new MongocryptdFactory(autoEncryptionOptions.ExtraOptions);
             _mongocryptdClient = _mongocryptdFactory.CreateMongocryptdClient();
+            _internalClient = internalClient; // can be null
```
rstam, line 77: Move to between 73-74. It is good to use the parameters in the order they appear in the method signature.

rstam (changes requested):
Will review the tests when we are done working through the pending change requests.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: done

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: done

tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
DmitryLukyanov: np, done.

>I think you missed my main point: these comments do NOT apply to just the lines below.

I didn't miss this, I just said that the original intention is not clear from the code itself. So, I'm ok to add this comment to the whole class

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
rstam: It's usually not necessary to add this kind of validation in private methods.

If we are the only ones calling this method we can make sure that it is not called when `_metadataClient` is null.

You can leave this if you want... but let's not be tempted to start adding argument validation to all our private methods.

For public methods of course we have to validate arguments.

rstam (commented):
Resolved some comments.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: we don't have control over when this method will be called since it's up to the libmongocrypt level logic. And we definitely can have null in the metadataClient. So, it's about preventing a null reference exception in case if something will be changed on the C level. So, I would say in this particular case it's required.

DmitryLukyanov: but I agree that this should not be a common rule

rstam:
@MikalaiMazurenka I've reviewed the driver changes but not the tests. The driver changes currently look OK to me.

You can be first reviewer on the tests. I'll do a second review of the tests once you are done.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
rstam: We do know whether this method will be called or not. If `AutoEncryptionOptions.BypassAutoEncryption` is `true` then we know that this method is never called.

That's precisely why we set `_metadataClient` to `null` when we know it won't be needed.

DmitryLukyanov: >We do know whether this method will be called or not. If AutoEncryptionOptions.BypassAutoEncryption is true then we know that this method is never called.

if there is no bug in the C level, then yes, but we don't have control over this. It can be changed at any moment, we can have a bug in this logic even now, and we may not see this just because it happens only in some circumstances

DmitryLukyanov: I don't say that this is expected, but I do say that libmongocrypt is an external library for the driver and it's better to validate their "output"

rstam: `_metadataClient` is not output of the libmongocrypt library.

We create it ourselves (unless `AutoEncryptionOptions.BypassAutoEncryption` is `true`).

As I understand it the C level can't call us back unless we call it first (which we won't do if `BypassAutoEncryption` is `true`).

Anyway... if you feel better with this `Ensure` here I'm still OK with you leaving it.

rstam: Note: `Ensure` is used for validating arguments... and therefore throws an `ArgumentNullException` here. But `_metadataClient` is not an argument, it's part of the object state.

Perhaps you should manually check for null and throw an `InvalidOperationException` instead.

DmitryLukyanov: the C level can potentially call this method even in decrypt steps, so in that case, this value will be null

DmitryLukyanov: >Perhaps you should manually check for null and throw an InvalidOperationException instead.

yeah, I also thought about this, I will do it

rstam: > the C level can potentially call this method even in decrypt steps, so in that case, this value will be null

Are you sure?

If that's true then why are we setting `_metadataClient` to `null` when `BypassAutoEncryption` is `true`?

DmitryLukyanov: I meant that it may be if there is a bug on the C level (with "routing"). But it's not expected according to the spec

DmitryLukyanov: >>    Perhaps you should manually check for null and throw an InvalidOperationException instead.

>yeah, I also thought about this, I will do it

done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+                    var ciphertext = ExplicitEncrypt(client_encryption, encryptionOptions, value, async);
+
+                    var eventCapturer = CreateEventCapturer().Capture<ClusterOpeningEvent>();
+
+                    using (var client_encrypted = ConfigureClientEncrypted(
+                        kmsProviderFilter: "local",
+                        maxPoolSize: maxPoolSize,
+                        bypassAutoEncryption: bypassAutoEncryption,
+                        eventCapturer: eventCapturer,
+                        externalKeyVaultClient: GetKeyVaultMongoClientByKey()))
```
MikalaiMazurenka, line 767: Local method `GetKeyVaultMongoClientByKey` is redundant. Consider:
```suggestion
                        externalKeyVaultClient: keyVaultMongoClientKey == "client_keyvault" ? client_keyvault : null))
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+                for (int i = 0; i < expectedEventsDetails.Length; i++)
+                {
+                    var arguments = expectedEventsDetails[i].Split(';');
+                    (string CommandName, string Database) expectedEventDetails = (arguments[0], arguments[1]);
+                    var @event = events.ElementAt(i);
+                    @event.DatabaseNamespace.DatabaseName.Should().Be(expectedEventDetails.Database);
+                    @event.CommandName.Should().Be(expectedEventDetails.CommandName);
+                }
+                events.Count().Should().Be(expectedEventsDetails.Count());
+            }
```
MikalaiMazurenka, line 832: Alternative if you apply tuple suggestion below:
```suggestion
            void AssertEvents(IEnumerable<CommandStartedEvent> events, (string CommandName, string DatabaseName)[] expectedEventsDetails)
            {
                for (int i = 0; i < expectedEventsDetails.Length; i++)
                {
                    var expectedEventDetails = expectedEventsDetails[i];
                    var @event = events.ElementAt(i);
                    @event.DatabaseNamespace.DatabaseName.Should().Be(expectedEventDetails.DatabaseName);
                    @event.CommandName.Should().Be(expectedEventDetails.CommandName);
                }
                events.Count().Should().Be(expectedEventsDetails.Count());
            }
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+                    keyVaultMongoClient: null,
+                    expectedNumberOfClients: 2,
+                    clientEncryptedEventsExpectation:
+                    new[]
+                    {
+                        "listCollections;db",
+                        "find;keyvault",
+                        "insert;db",
+                        "find;db"
+                    },
```
MikalaiMazurenka, line 857: Alternative if you apply tuple suggestion below:
```suggestion
                    new[]
                    {
                        ("listCollections", "db"),
                        ("find", "keyvault"),
                        ("insert", "db"),
                        ("find", "db")
                    },
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+            }
+
+            IEnumerable<object[]> CaseWithAsync(
+                string name,
+                int maxPoolSize,
+                bool bypassAutoEncryption,
+                string keyVaultMongoClient,
+                int expectedNumberOfClients,
+                string[] clientEncryptedEventsExpectation,
+                string[] clientKeyVaultEventsExpectation)
```
MikalaiMazurenka, line 925: Instead of splitting strings you can use tuples here:
```suggestion
                (string, string)[] clientEncryptedEventsExpectation,
                (string, string)[] clientKeyVaultEventsExpectation)
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+        [SkippableTheory]
+        [MemberData(nameof(DeadlockTest_MemberData))]
+        public void DeadlockTest(
+            string _,
+            int maxPoolSize,
+            bool bypassAutoEncryption,
+            string keyVaultMongoClientKey,
+            int expectedNumberOfClients,
+            string[] clientEncryptedEventsExpectation,
+            string[] clientKeyVaultEventsExpectation,
```
MikalaiMazurenka, line 732: If you apply tuples suggestion below:
```suggestion
            (string, string)[] clientEncryptedEventsExpectation,
            (string, string)[] clientKeyVaultEventsExpectation,
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+
+            foreach (var asyncCases in testCases)
+            {
+                foreach (var asyncCase in asyncCases)
+                {
+                    yield return asyncCase;
+                }
+            }
+
+            IEnumerable<object[]> CaseWithAsync(
```
MikalaiMazurenka, line 918: It is preferable to avoid using `object`. It can be substituted with a nested class `DeadlockTestData`. As a benefit, you will be able to skip unset fields in this case.

tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
```diff
@@ -237,6 +239,15 @@ public IMongoClient WithWriteConcern(WriteConcern writeConcern)
...
             ClusterRegistry.Instance.UnregisterAndDisposeCluster(wrapped.Cluster);
+
+            if (wrapped is MongoClient mongoClient)
+            {
+                var internalClient = mongoClient.LibMongoCryptController?.InternalClient;
+                if (internalClient != null)
+                {
+                    ClusterRegistry.Instance.UnregisterAndDisposeCluster(internalClient.Cluster);
+                }
+            }
```
MikalaiMazurenka, line 250: I am deeply concerned about this code, but since it is a test solution and I don't have any better suggestions, I recommend to just keep this problem in mind for future major refactoring.

MikalaiMazurenka (changes requested):
Great encryption tests refactoring!
A have a few minor suggestions on the deadlock prose tests.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: it doesn't really work. In this case, we will have a problem with xunit and shown tests results. NOTE: your suggestion is how I did this originally

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: answered above

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: answered above

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: it's done by purpose, otherwise, there is a problem with xunit serialization

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: answered above

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+                                new BsonDocument
+                                {
+                                    { "_id", 0 },
+                                    { "encrypted", value }
+                                });
+                        }
+
+                        collCollection = GetCollection(client_encrypted, __collCollectionNamespace);
+                        var findResult = Find(collCollection, BsonDocument.Parse("{ _id : 0 }"), async).Single();
+                        findResult.Should().Be(BsonDocument.Parse($"{{ _id : 0, encrypted : '{value}' }}"));
```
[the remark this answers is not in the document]
DmitryLukyanov: will check it

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: mostly agreed, I did this because the spec test idea was to map a provided string key to the client, currently, we have only one test key, but I think it's a valid assumption that the number of keys can be bigger further, so I would not change this

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+            RequirePlatform
+                .Check()
+                .SkipWhen(SupportedOperatingSystem.Linux, SupportedTargetFramework.NetStandard15)
+                .SkipWhen(SupportedOperatingSystem.MacOS, SupportedTargetFramework.NetStandard15);
+
+            var clientKeyVaultEventCapturer = CreateEventCapturer();
+            using (var client_keyvault = CreateMongoClient(maxPoolSize: 1, writeConcern: WriteConcern.WMajority, readConcern: ReadConcern.Majority, eventCapturer: clientKeyVaultEventCapturer))
+            using (var client_test = ConfigureClient(clearCollections: true, writeConcern: WriteConcern.WMajority, readConcern: ReadConcern.Majority))
+            {
+                var dataKeys = GetCollection(client_test, __keyVaultCollectionNamespace);
```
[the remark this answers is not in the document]
DmitryLukyanov: will check

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -28,26 +28,61 @@ namespace MongoDB.Driver.Encryption
...
+                cryptClient,
+                autoEncryptionOptions);
+
+            IMongoClient CreateInternalClient()
+            {
+                var internalClientSettings = client.Settings.Clone();
+                internalClientSettings.AutoEncryptionOptions = null;
+                internalClientSettings.MinConnectionPoolSize = 0;
+                return new MongoClient(internalClientSettings);
+            }
```
MikalaiMazurenka, line 52: One more suggestion, which you can apply at your own discretion, because it makes a bit more checks but gets rid of `Lazy`:
```suggestion
            IMongoClient internalClient = null;
            if (autoEncryptionOptions.KeyVaultClient == null || autoEncryptionOptions.BypassAutoEncryption == false)
            {
                var internalClientSettings = client.Settings.Clone();
                internalClientSettings.AutoEncryptionOptions = null;
                internalClientSettings.MinConnectionPoolSize = 0;
                internalClient = new MongoClient(internalClientSettings);
            }
            var keyVaultClient = autoEncryptionOptions.KeyVaultClient ?? internalClient;
            var metadataClient = autoEncryptionOptions.BypassAutoEncryption ? null : internalClient;

            return new AutoEncryptionLibMongoCryptController(
                internalClient,
                keyVaultClient,
                metadataClient,
                cryptClient,
                autoEncryptionOptions);
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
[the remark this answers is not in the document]
DmitryLukyanov: this happens here: https://github.com/mongodb/mongo-csharp-driver/pull/446/files#diff-0537777d32b586e768ac1e74c10677b06f8933c41fba72c3f11a6f74862925e7R743 (see `clearConnection` argument)

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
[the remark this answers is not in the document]
DmitryLukyanov: good suggestion, done

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: I think a lazy approach exactly matches to what is required by the spec. Also, I think the code written in this way is a bit more clear, so I would leave this as is

MikalaiMazurenka (approved):
LGTM

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -153,7 +188,13 @@ protected override async Task ProcessStateAsync(CryptContext context, string dat
         // private methods
         private void ProcessNeedCollectionInfoState(CryptContext context, string databaseName, CancellationToken cancellationToken)
         {
-            var database = _client.GetDatabase(databaseName);
+            if (_metadataClient == null)
+            {
+                // should not be reached
+                throw new InvalidOperationException("Metadata client must be configured.");
```
rstam, line 194: I'm not sure "configured" is the right word here. That makes it sound like it's the application's fault for not configuring something correctly.

Don't know if you want to reword this or not. Since this should never happen we don't actually expect anyone to ever see this exception, other than readers of the code.

Consider:

    throw new InvalidOperationException("Metadata client is null.");

Or if you don't want to have to figure out a correct error message just remove the check and let a NullReferenceException be thrown if _metadataClient is null (which as we know will never happen).

rstam (commented):
I think a comment you added is not accurate, but maybe we can let it slide.

I have reviewed the driver changes but not the tests. Let me check with @jyemin whether he wants me to review the tests or whether he wants to add another reviewer to review the tests.

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: I would leave it. I changed the error message

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
rstam, line 745: Consider naming this variable `dataKeysCollection`.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+
+            void AssertNumberOfClients(IEnumerable<ClusterOpeningEvent> events)
+            {
+                events.Count().Should().Be(expectedNumberOfClients);
+            }
+        }
+
+        public static IEnumerable<object[]> DeadlockTest_MemberData()
+        {
+            var testCases = new[]
```
rstam, line 842: I found the data type of this variable confusing because of the double nesting (which you then have to undo in lines 910-916 before returning the actual test cases).

A variable named `testCases` should be be of type `IEnumerable<object[]>`.

You should refactor this to:

    var testCases = new List<object[]>();
    testCases.AddRange(
        CasesWithAsync(
            name: "case 1",
            maxPoolSize: 1,
            bypassAutoEncryption: false,
            keyVaultMongoClient: null,
            expectedNumberOfClients: 2,
            clientEncryptedEventsExpectation:
            new[]
            {
                "listCollections;db",
                "find;keyvault",
                "insert;db",
                "find;db"
            },
            clientKeyVaultEventsExpectation: null));
    testCases.AddRange(
        CasesWithAsync(
            name: "case 2",
            maxPoolSize: 1,
            bypassAutoEncryption: false,
            keyVaultMongoClient: "client_keyvault",
            expectedNumberOfClients: 2,
            clientEncryptedEventsExpectation:
            new[]
            {
                "listCollections;db",
                "insert;db",
                "find;db"
            },
            clientKeyVaultEventsExpectation:
            new[]
            {
                "find;keyvault"
            }));
    testCases.AddRange(
        CasesWithAsync(
            name: "case 3",
            maxPoolSize: 1,
            bypassAutoEncryption: true,
            keyVaultMongoClient: null,
            expectedNumberOfClients: 2,
            clientEncryptedEventsExpectation:
            new[]
            {
                "find;db",
                "find;keyvault"
            },
            clientKeyVaultEventsExpectation: null));
    testCases.AddRange(
        CasesWithAsync(
            name: "case 4",
            maxPoolSize: 1,
            bypassAutoEncryption: true,
            keyVaultMongoClient: "client_keyvault",
            expectedNumberOfClients: 1,
            clientEncryptedEventsExpectation:
            new[]
            {
                "find;db",
            },
            clientKeyVaultEventsExpectation:
            new[]
            {
                "find;keyvault"
            }));

    // cases 5-8 use "MaxPoolSize: 0" which is not supported by the c# driver

    return testCases;

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+                // cases 5-8 use "MaxPoolSize: 0" which is not supported by the c# driver
+            };
+
+            foreach (var asyncCases in testCases)
+            {
+                foreach (var asyncCase in asyncCases)
+                {
+                    yield return asyncCase;
+                }
+            }
```
rstam, line 916: Lines 910-916 can be simplified to:

    return testCases;

after applying the suggestion above.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -721,6 +720,217 @@ void TestCase(ClientEncryption testCaseClientEncription, BsonDocument masterKey,
...
+
+            IEnumerable<object[]> CaseWithAsync(
```
rstam, line 918: Rename to `CasesWithAsync` (plural) because this method returns *two* test cases.

rstam (changes requested):
Overall test changes look good. I do have some suggested changes.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: done

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/60257e130305b9597258fe4f

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
DmitryLukyanov: >I don't think the spec requires the use of the Lazy<T> class.

I meant not .NET lazy, but lazy approach in general when we postpone calling a method until it's actually required

DmitryLukyanov:
ready for review

rstam (approved):
LGTM

tests/MongoDB.Driver.TestHelpers/DisposableMongoClient.cs
MikalaiMazurenka: It's about both. Mostly the class `DisposableMongoClient` itself (I think this ideally should be done by test setup / teardown, but I don't see any good ways of doing it here). And the dispose logic (the new code) should ideally look simpler in my opinion, but again I don't have any suggestions, just the impression that it doesn't look simple enough.

PR #447, opened 2021-02-01
CSHARP-3361: Disable parallel NuGet package restore

BorisDog:


build.cake
```diff
@@ -52,7 +52,11 @@ Task("Release")
 Task("Restore")
     .Does(() =>
     {
-        DotNetCoreRestore(solutionFullPath);
+        var restoreSettings = new DotNetCoreRestoreSettings 
+        {
+            DisableParallel = true
+        };
+        DotNetCoreRestore(solutionFullPath, restoreSettings);
```
rstam, line 59: Would it be worth adding a comment before 55 along the lines of:

    // disable parallel restore to work around apparent bugs in restore

so anyone who looks at this in the future knows why we did this (we'd prefer parallel restore because presumably it's faster, if only it worked reliably).

DmitryLukyanov (approved):
LGTM

build.cake
BorisDog: Done tnx.

rstam (approved):
LGTM

BorisDog:
Merged.

PR #448, opened 2021-02-01
CSHARP-3369: Adding solution filters

BorisDog:


CSharpDriver.Driver.slnf
```diff
@@ -0,0 +1,12 @@
+{
+  "solution": {
+    "path": "CSharpDriver.sln",
+    "projects": [
+      "src\\MongoDB.Bson\\MongoDB.Bson.csproj",
+      "src\\MongoDB.Driver.Core\\MongoDB.Driver.Core.csproj",
+      "src\\MongoDB.Driver\\MongoDB.Driver.csproj",
+      "tests\\MongoDB.Driver.TestHelpers\\MongoDB.Driver.TestHelpers.csproj",
```
DmitryLukyanov, line 8: if I'm not mistaken, we also should add this line `MongoDB.Driver.Core.TestHelpers` here

CSharpDriver.Driver.Core.slnf
```diff
@@ -0,0 +1,11 @@
+{
+  "solution": {
+    "path": "CSharpDriver.sln",
+    "projects": [
+      "src\\MongoDB.Bson\\MongoDB.Bson.csproj",
+      "src\\MongoDB.Driver.Core\\MongoDB.Driver.Core.csproj",
```
rstam, line 6: Add:

    "tests\\MongoDB.Bson.TestHelpers\\MongoDB.Bson.TestHelpers.csproj"

CSharpDriver.Driver.slnf
rstam: Also: 

    "tests\\MongoDB.Bson.TestHelpers\\MongoDB.Bson.TestHelpers.csproj"

rstam (changes requested):

CSharpDriver.Driver.Core.slnf
```diff
@@ -0,0 +1,11 @@
...
+      "tests\\MongoDB.Driver.Core.TestHelpers\\MongoDB.Driver.Core.TestHelpers.csproj",
+      "tests\\MongoDB.Driver.Core.Tests\\MongoDB.Driver.Core.Tests.csproj"
+    ]
+  }
+}
```
DmitryLukyanov, line 11: the last line should be empty, then you won't see this red circle

CSharpDriver.Driver.slnf
```diff
@@ -0,0 +1,15 @@
+{
+  "solution": {
+    "path": "CSharpDriver.sln",
+    "projects": [
+      "src\\MongoDB.Bson\\MongoDB.Bson.csproj",
+      "src\\MongoDB.Driver.Core\\MongoDB.Driver.Core.csproj",
+      "src\\MongoDB.Driver.GridFS\\MongoDB.Driver.GridFS.csproj",
```
DmitryLukyanov, line 7: nit: this is probably unnecessary since this lib references Core and Driver

DmitryLukyanov (approved):
minor nit and then LGTM

CSharpDriver.Driver.slnf
rstam: I agree.

But also... I don't think we need to create yet another filter for the GridFS project. That should be used very rarely.

rstam (approved):
LGTM after you apply Dmitry's suggestion to remove GridFS

CSharpDriver.Driver.slnf
DmitryLukyanov: Yeah, I'm just saying that this library is useless in this list since the related tests are not added

rstam: And I'm agreeing with you... we should remove this project from this filter.

BorisDog: Needed in MongoDB.Driver.Tests

BorisDog:
Merged.

PR #452, opened 2021-02-05
CSHARP-3355: Add more spec tests for invalid relaxed UUIDs.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/601d956b3e8e86586c978855

### DIFF
--- a/src/MongoDB.Bson/IO/JsonReader.cs
+++ b/src/MongoDB.Bson/IO/JsonReader.cs
@@ -2089,15 +2089,8 @@ namespace MongoDB.Bson.IO
             }
             VerifyToken("}");
 
-            var hexString = uuidToken.StringValue.Replace("-", "");
-            var bytes = BsonUtils.ParseHexString(hexString);
-            if (bytes.Length != 16)
-            {
-                var message = string.Format("Invalid $uuid string: '{0}'.", hexString);
-                throw new FormatException(message);
-            }
-
-            return new BsonBinaryData(bytes, BsonBinarySubType.UuidStandard);
+            var guid = Guid.Parse(uuidToken.StringValue);
+            return new BsonBinaryData(guid, GuidRepresentation.Standard);
         }
 
         private BsonValue ParseUUIDConstructor(string uuidConstructorName)
--- a/tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
+++ b/tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
@@ -1,0 +1,434 @@
+#!/usr/bin/env perl
+use v5.10;
+use strict;
+use warnings;
+use utf8;
+use open qw/:std :utf8/;
+
+use Getopt::Long;
+use Pod::Usage;
+
+use if $^O eq 'MSWin32', 'Win32::Console::ANSI';
+use Term::ANSIColor;
+
+use constant {
+    NULL              => "\x00",
+    BSON_TYPE         => "C",
+    BSON_ENAME        => "Z*",
+    BSON_TYPE_NAME    => "CZ*",
+    BSON_DOUBLE       => "d",
+    BSON_STRING       => "l/A",
+    BSON_BOOLEAN      => "C",
+    BSON_REGEX        => "Z*Z*",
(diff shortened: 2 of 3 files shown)


tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
```diff
@@ -0,0 +1,434 @@
+#!/usr/bin/env perl
```
jyemin, line 1: I think this file can be deleted

jyemin (changes requested):

tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
DmitryLukyanov: as far as I know, we copy such files under git to make a diff with the spec folder easier. So, if we have such a file under git we can just fully copy all files from the spec folder into our tests without any analysis

src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -2089,14 +2090,25 @@ private BsonValue ParseUuidExtendedJson()
...
 
+            if (uuidTokenStringValue.Count(c => c == '-') != 4 ||
+                uuidTokenStringValue[8] != '-' ||
+                uuidTokenStringValue[13] != '-' ||
+                uuidTokenStringValue[18] != '-' ||
+                uuidTokenStringValue[23] != '-')
+            {
+                throw new FormatException($"Invalid $uuid hyphens format: '{uuidTokenStringValue}'.");
+            }
+
```
rstam, line 2111: I think you can replace lines 2093-2112 with:

    var guid = Guid.Parse(uuidToken.StringValue);
    var binaryData = new BsonBinaryData(guid, GuidRepresentation.Standard);

`Guid.Parse` already does all the validations we need.

tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
DmitryLukyanov: at least mostly :)

rstam (changes requested):

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/601d9af00305b9115d9a8acd

src/MongoDB.Bson/IO/JsonReader.cs
DmitryLukyanov: good point, done

tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
jyemin: If that's the policy, then so be it, but this is a weird one to copy.

jyemin:
Removed James since Robert picked it up.

tests/MongoDB.Bson.Tests/Specifications/bson-corpus/tests/bsonview
DmitryLukyanov: it's not a defined policy, but this is how we did previously

rstam (approved):
LGTM

src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -18,6 +18,7 @@
 using System.Globalization;
 using System.IO;
 using System.Text.RegularExpressions;
+using System.Linq;
```
JamesKovacs, line 21: Not used.

JamesKovacs (changes requested):
Changes look good. Verified that:
- BsonBinarySubType is still set correctly.
- Same exception type (FormatException) is thrown as before.
- Tests run successfully.

One minor change required. You have an extraneous `using` statement.

src/MongoDB.Bson/IO/JsonReader.cs
DmitryLukyanov: Done

JamesKovacs (approved):
LGTM

PR #457, opened 2021-02-16
CSHARP-3427 Can't make use of compression when application path contain space

jeremy-waguet:
Because of the use or Uri class (don't know why ...), application path containing spaces (e.g. ClickOnce deployed app running on a user account with spaces) produce the following exception when trying to connect to DB with compression enabled:

> System.IO.FileNotFoundException: Could not find library snappy64.dll. Checked C:\Users\User%20Name\source\repos\...\bin\Release\net5.0-windows\snappy64.dll;C:\Users\User%20Name\source\repos\...\bin\Release\net5.0-windows\..\..\runtimes\win\native\snappy64.dll;C:\Users\User%20Name\source\repos\...\bin\Release\net5.0-windows\runtimes\win\native\snappy64.dll.

This pull request fix it. For now I have to disable compression on the connection string (to keep using official nuget packages)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
+++ b/src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
@@ -40,10 +40,7 @@ namespace MongoDB.Driver.Core.NativeLibraryLoader
         public virtual string GetLibraryBasePath()
         {
             var assembly = GetLibraryBaseAssembly();
-            var codeBase = assembly.CodeBase;
-            var uri = new Uri(codeBase);
-            var absolutePath = uri.AbsolutePath;
-            return Path.GetDirectoryName(absolutePath);
+            return Path.GetDirectoryName(assembly.Location);
         }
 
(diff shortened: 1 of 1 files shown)


DmitryLukyanov:
Hey @jeremy-waguet, can you please also create a Jira ticket for this PR?

Also, just to confirm, that you see exactly the above exception with folders like: `net5.0-windows....` and `repos...`? And the real folder names look the same but without spaces: `net5.0-windows` and `repos`?

jeremy-waguet:
@DmitryNaumov 

Here is the link the the Jira issue : https://jira.mongodb.org/browse/CSHARP-3427

The issue is reproducible when a space is present in the path. It can be with an intermediate directory or the last directory component. For example. Renaming ‘net5.0-windows’ to ‘net5.0-windows 123’ produces the same issue

DmitryLukyanov:
Hey @jeremy-waguet , thanks for your PR. The way you used in this PR doesn't cover all cases that we consider in this logic, for example it doesn't correctly work if we configure shadow-copying. We chose a bit different solution and it will be available in a next release. Thanks

PR #463, opened 2021-02-18
CSHARP-3231: Push directly to nuget.org

JamesKovacs:
This requires testing during our next release. If it doesn't work as expected, we may have to revert this change until we can investigate further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,6 @@
 
 You can get the latest stable release from the [official Nuget.org feed](https://www.nuget.org/packages/MongoDB.Driver) or from our [github releases page](https://github.com/mongodb/mongo-csharp-driver/releases).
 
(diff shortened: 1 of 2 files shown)


DmitryLukyanov:
this line probably should be changed too: `If you'd like to work with the bleeding edge, you can use our custom feed. Some packages on this feed are pre-release and, while they've passed all our tests, are not yet ready for production.`: https://github.com/mongodb/mongo-csharp-driver/blob/master/README.md

JamesKovacs:
> this line probably should be changed too: `If you'd like to work with the bleeding edge, you can use our custom feed. Some packages on this feed are pre-release and, while they've passed all our tests, are not yet ready for production.`: https://github.com/mongodb/mongo-csharp-driver/blob/master/README.md

Good catch. I never even noticed that we had a link to our MyGet feed in our documentation.

DmitryLukyanov (approved):
LGTM

BorisDog (approved):

BorisDog:
LGTM

PR #467, opened 2021-02-26
CSHARP-3303: Fix Monitor_sleep_at_least_minHeartbeatFreqencyMS test on replica sets

BorisDog:
Select single server in Monitor_sleep_at_least_minHeartbeatFreqencyMS_between_checks test (test fails on connections strings containing multiple servers)

[EG](https://evergreen.mongodb.com/version/6039742061837d59d55e0d7a) (Internal link)


tests/MongoDB.Driver.Tests/Specifications/server-discovery-and-monitoring/ServerDiscoveryAndMonitoringProseTests.cs
```diff
@@ -96,6 +96,7 @@ public void Monitor_sleep_at_least_minHeartbeatFreqencyMS_between_checks()
                 }}");
 
             var settings = DriverTestConfiguration.GetClientSettings();
+            settings.Servers = new[] { settings.Servers.First() };
```
DmitryLukyanov, line 99: I forgot that we can have multi hosts connection string on EG, good point

DmitryLukyanov (approved):
LGTM

PR #471, opened 2021-03-02
Updated documentation to v2.12.0.

JamesKovacs:


rstam (approved):
LGTM

Updating these docs is always a mystery, but as far as I can tell everything you've changed looks OK.

PR #474, opened 2021-03-11
CSHARP-3304: Consider server load during server selection

BorisDog:
[EG](https://evergreen.mongodb.com/version/60664f99e3c33117576b4303) (internal link)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
@@ -410,16 +410,24 @@ namespace MongoDB.Driver.Core.Clusters
         private class SelectServerHelper : IDisposable
         {
             private readonly Cluster _cluster;
+            private readonly List<IClusterableServer> _connectedServers;
+            private readonly List<ServerDescription> _connectedServerDescriptions;
             private ClusterDescription _description;
             private Task _descriptionChangedTask;
             private bool _serverSelectionWaitQueueEntered;
             private readonly IServerSelector _selector;
+            private readonly OperationsCountServerSelector _operationCountServerSelector;
             private readonly Stopwatch _stopwatch;
             private readonly DateTime _timeoutAt;
 
             public SelectServerHelper(Cluster cluster, IServerSelector selector)
             {
                 _cluster = cluster;
+
+                _connectedServers = new List<IClusterableServer>(_cluster._description?.Servers?.Count ?? 1);
+                _connectedServerDescriptions = new List<ServerDescription>(_connectedServers.Count);
+                _operationCountServerSelector = new OperationsCountServerSelector(_connectedServers);
+
                 _selector = DecorateSelector(selector);
                 _stopwatch = Stopwatch.StartNew();
                 _timeoutAt = DateTime.UtcNow + _cluster.Settings.ServerSelectionTimeout;
@@ -489,36 +497,47 @@ namespace MongoDB.Driver.Core.Clusters
 
                 MongoIncompatibleDriverException.ThrowIfNotSupported(_description);
 
-                var connectedServers = _description.Servers.Where(s => s.State == ServerState.Connected);
-                var selectedServers = _selector.SelectServers(_description, connectedServers).ToList();
+                _connectedServers.Clear();
+                _connectedServerDescriptions.Clear();
 
-                while (selectedServers.Count > 0)
+                foreach (var description in _description.Servers)
                 {
-                    var server = selectedServers.Count == 1 ?
-                        selectedServers[0] :
-                        __randomServerSelector.SelectServers(_description, selectedServers).Single();
-
-                    IClusterableServer selectedServer;
-                    if (_cluster.TryGetServer(server.EndPoint, out selectedServer))
+                    if (description.State == ServerState.Connected &&
+                        _cluster.TryGetServer(description.EndPoint, out var server))
                     {
-                        _stopwatch.Stop();
-                        var selectedServerEventHandler = _cluster._selectedServerEventHandler;
-                        if (selectedServerEventHandler != null)
-                        {
-                            selectedServerEventHandler(new ClusterSelectedServerEvent(
-                                _description,
-                                _selector,
-                                server,
-                                _stopwatch.Elapsed,
-                                EventContext.OperationId));
-                        }
-                        return selectedServer;
+                        _connectedServers.Add(server);
+                        _connectedServerDescriptions.Add(description);
                     }
-
-                    selectedServers.Remove(server);
                 }
 
-                return null;
+                var selectedServersDescriptions = _selector
+                    .SelectServers(_description, _connectedServerDescriptions)
+                    .ToList();
+
+                IServer selectedServer = null;
+
+                if (selectedServersDescriptions.Count > 0)
+                {
+                    var selectedServerDescription = selectedServersDescriptions.Count == 1
+                        ? selectedServersDescriptions[0]
+                        : __randomServerSelector.SelectServers(_description, selectedServersDescriptions).Single();
+
+                    selectedServer = _connectedServers.FirstOrDefault(s => EndPointHelper.Equals(s.EndPoint, selectedServerDescription.EndPoint));
+                }
+
+                if (selectedServer != null)
+                {
+                    _stopwatch.Stop();
+
+                    _cluster._selectedServerEventHandler?.Invoke(new ClusterSelectedServerEvent(
+                        _description,
+                        _selector,
+                        selectedServer.Description,
+                        _stopwatch.Elapsed,
+                        EventContext.OperationId));
+                }
+
+                return selectedServer;
             }
 
             public void WaitingForDescriptionToChange()
@@ -554,6 +573,7 @@ namespace MongoDB.Driver.Core.Clusters
                 }
 
                 allSelectors.Add(_cluster._latencyLimitingServerSelector);
+                allSelectors.Add(_operationCountServerSelector);
 
                 return new CompositeServerSelector(allSelectors);
             }
--- a/src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
@@ -1,0 +1,65 @@
+/* Copyright 2021-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System.Collections.Generic;
+using System.Linq;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.Servers;
+
+namespace MongoDB.Driver.Core.Clusters.ServerSelectors
+{
+    internal sealed class OperationsCountServerSelector : IServerSelector
+    {
+        private readonly IEnumerable<IClusterableServer> _clusterableServers;
+
+        public OperationsCountServerSelector(IEnumerable<IClusterableServer> clusterableServers)
+        {
+            _clusterableServers = clusterableServers;
+        }
+
+        // methods
+        /// <inheritdoc/>
+        public IEnumerable<ServerDescription> SelectServers(ClusterDescription cluster, IEnumerable<ServerDescription> servers)
+        {
+            var list = servers.ToList();
+            switch (list.Count)
+            {
+                case 0:
+                case 1:
+                    return list;
+                default:
+                    {
+                        // Follow the "Power of Two Choices" approach
+                        // https://web.archive.org/web/20191212194243/https://www.nginx.com/blog/nginx-power-of-two-choices-load-balancing-algorithm/
+                        var index1 = ThreadStaticRandom.Next(list.Count);
+                        var index2 = (index1 + 1 + ThreadStaticRandom.Next(list.Count - 1)) % list.Count;
+
+                        var endpoint1 = list[index1].EndPoint;
+                        var endpoint2 = list[index2].EndPoint;
+                        var server1 = _clusterableServers.First(s => EndPointHelper.Equals(s.Description.EndPoint, endpoint1));
+                        var server2 = _clusterableServers.First(s => EndPointHelper.Equals(s.Description.EndPoint, endpoint2));
+
+                        var selectedServer = server1.OutstandingOperationsCount < server2.OutstandingOperationsCount ? server1 : server2;
+
+                        return new[] { selectedServer.Description };
+                    }
+            }
+        }
+
+        /// <inheritdoc/>
+        public override string ToString() =>
+            nameof(OperationsCountServerSelector);
+    }
+}
--- a/src/MongoDB.Driver.Core/Core/Servers/Server.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/Server.cs
@@ -70,6 +70,8 @@ namespace MongoDB.Driver.Core.Servers
         private readonly InterlockedInt32 _state;
         private readonly ServerApi _serverApi;
 
+        private int _outstandingOperationsCount;
+
         private readonly Action<ServerOpeningEvent> _openingEventHandler;
         private readonly Action<ServerOpenedEvent> _openedEventHandler;
         private readonly Action<ServerClosingEvent> _closingEventHandler;
@@ -115,6 +117,7 @@ namespace MongoDB.Driver.Core.Servers
             _baseDescription = new ServerDescription(_serverId, endPoint, reasonChanged: "ServerInitialDescription", heartbeatInterval: settings.HeartbeatInterval);
             _currentDescription = _baseDescription;
             _serverApi = serverApi;
+            _outstandingOperationsCount = 0;
 
             eventSubscriber.TryGetEventHandler(out _openingEventHandler);
             eventSubscriber.TryGetEventHandler(out _openedEventHandler);
@@ -133,6 +136,8 @@ namespace MongoDB.Driver.Core.Servers
         public ServerId ServerId => _serverId;
 
         internal IClusterClock ClusterClock => _clusterClock;
+
+        int IClusterableServer.OutstandingOperationsCount => Interlocked.CompareExchange(ref _outstandingOperationsCount, 0, 0);
 
         // methods
         public void Dispose()
@@ -160,24 +165,32 @@ namespace MongoDB.Driver.Core.Servers
         public IChannelHandle GetChannel(CancellationToken cancellationToken)
         {
             ThrowIfNotOpen();
+            IConnectionHandle connection = null;
 
-            var connection = _connectionPool.AcquireConnection(cancellationToken);
             try
             {
+                Interlocked.Increment(ref _outstandingOperationsCount);
+                connection = _connectionPool.AcquireConnection(cancellationToken);
+
                 // ignoring the user's cancellation token here because we don't
                 // want to throw this connection away simply because the user
                 // wanted to cancel their operation. It will be better for the
                 // collective to complete opening the connection than the throw
                 // it away.
-
                 connection.Open(CancellationToken.None); // This results in the initial isMaster being sent
                 return new ServerChannel(this, connection);
             }
             catch (Exception ex)
             {
-                HandleBeforeHandshakeCompletesException(connection, ex);
+                Interlocked.Decrement(ref _outstandingOperationsCount);
 
-                connection.Dispose();
+                if (connection != null)
+                {
+                    HandleBeforeHandshakeCompletesException(connection, ex);
+
+                    connection.Dispose();
+                }
+
                 throw;
             }
         }
@@ -185,10 +198,13 @@ namespace MongoDB.Driver.Core.Servers
         public async Task<IChannelHandle> GetChannelAsync(CancellationToken cancellationToken)
         {
             ThrowIfNotOpen();
+            IConnectionHandle connection = null;
 
-            var connection = await _connectionPool.AcquireConnectionAsync(cancellationToken).ConfigureAwait(false);
             try
             {
+                Interlocked.Increment(ref _outstandingOperationsCount);
+                connection = await _connectionPool.AcquireConnectionAsync(cancellationToken).ConfigureAwait(false);
+
                 // ignoring the user's cancellation token here because we don't
                 // want to throw this connection away simply because the user
                 // wanted to cancel their operation. It will be better for the
@@ -199,9 +215,15 @@ namespace MongoDB.Driver.Core.Servers
             }
             catch (Exception ex)
             {
-                HandleBeforeHandshakeCompletesException(connection, ex);
+                Interlocked.Decrement(ref _outstandingOperationsCount);
 
-                connection.Dispose();
+                if (connection != null)
+                {
+                    HandleBeforeHandshakeCompletesException(connection, ex);
+
+                    connection.Dispose();
+                }
+
                 throw;
             }
         }
@@ -591,14 +613,19 @@ namespace MongoDB.Driver.Core.Servers
         {
             // fields
             private readonly IConnectionHandle _connection;
-            private bool _disposed;
             private readonly Server _server;
 
+            private readonly InterlockedInt32 _state;
+            private readonly bool _decrementOperationsCount;
+
             // constructors
-            public ServerChannel(Server server, IConnectionHandle connection)
+            public ServerChannel(Server server, IConnectionHandle connection, bool decrementOperationsCount = true)
             {
                 _server = server;
                 _connection = connection;
+
+                _state = new InterlockedInt32(ChannelState.Initial);
+                _decrementOperationsCount = decrementOperationsCount;
             }
 
             // properties
@@ -814,10 +841,14 @@ namespace MongoDB.Driver.Core.Servers
 
             public void Dispose()
             {
-                if (!_disposed)
+                if (_state.TryChange(ChannelState.Initial, ChannelState.Disposed))
                 {
+                    if (_decrementOperationsCount)
+                    {
+                        Interlocked.Decrement(ref _server._outstandingOperationsCount);
+                    }
+
                     _connection.Dispose();
-                    _disposed = true;
                 }
             }
 
@@ -1261,7 +1292,8 @@ namespace MongoDB.Driver.Core.Servers
             public IChannelHandle Fork()
             {
                 ThrowIfDisposed();
-                return new ServerChannel(_server, _connection.Fork());
+
+                return new ServerChannel(_server, _connection.Fork(), false);
             }
 
             private ReadPreference GetEffectiveReadPreference(bool slaveOk, ReadPreference readPreference)
@@ -1319,10 +1351,17 @@ namespace MongoDB.Driver.Core.Servers
 
             private void ThrowIfDisposed()
             {
-                if (_disposed)
+                if (_state.Value == ChannelState.Disposed)
                 {
                     throw new ObjectDisposedException(GetType().Name);
                 }
+            }
+
+            // nested types
+            private static class ChannelState
+            {
+                public const int Initial = 0;
+                public const int Disposed = 1;
             }
         }
     }
--- a/tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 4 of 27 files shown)


src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -557,6 +561,25 @@ private IServerSelector DecorateSelector(IServerSelector selector)
 
                 return new CompositeServerSelector(allSelectors);
             }
+
+            private IClusterableServer SelectServerByOperationsCount(List<IClusterableServer> servers)
```
DmitryLukyanov, line 565: It looks like this logic still can be moved into server selector similar to:

      /// <summary>
    /// TODO
    /// </summary>
    public class OperationCountServerSelector : IServerSelector
    {
        #region static
        /// <summary>
        /// TODO
        /// </summary>
        /// <param name="servers"></param>
        /// <returns></returns>
        private static (EndPoint ServerAddress, int Count)[] SomePrepareStuff(IClusterableServer[] servers)
        {
            return servers.Select(s => (s.EndPoint, s.OutstandingOperationsCount)).ToArray();
        }

        /// <summary>
        /// TODO
        /// </summary>
        /// <param name="servers"></param>
        /// <returns></returns>
        public static IServerSelector Create(IClusterableServer[] servers)
        {
            var inputData = SomePrepareStuff(servers);
            return new OperationCountServerSelector(inputData);
        }
        #endregion

        private readonly (EndPoint ServerAddress, int Count)[] _serverAndPoolDescriptions;

        /// <summary>
        /// 
        /// </summary>
        /// <param name="serverAndPoolDescriptions">TODO</param>
        public OperationCountServerSelector((EndPoint ServerAddress, int Count)[] serverAndPoolDescriptions)
        {
            _serverAndPoolDescriptions = serverAndPoolDescriptions;
        }

        /// <summary>
        /// TODO
        /// </summary>
        /// <param name="cluster"></param>
        /// <param name="servers"></param>
        /// <returns></returns>
        public IEnumerable<ServerDescription> SelectServers(ClusterDescription cluster, IEnumerable<ServerDescription> servers)
        {
            if (_serverAndPoolDescriptions.Length != servers.Count())
            {
                // TODO: ?
            }

            return servers
                .Select(
                    s =>
                    {
                        var poolDescription = _serverAndPoolDescriptions.Single(d => EndPointHelper.Equals(d.ServerAddress, s.EndPoint));
                        return new
                        {
                            ServerDescription = s,
                            OperationsCount = poolDescription.Count
                        };
                    })
                .OrderBy(c=>c.OperationsCount)
                .Select(s=>s.ServerDescription);
        }
    }

DmitryLukyanov: where the missed information is provided through ctor. So we will need to create a new instance of this selector in each attempt

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -491,34 +491,38 @@ public IServer SelectServer()
 
                 var connectedServers = _description.Servers.Where(s => s.State == ServerState.Connected);
                 var selectedServers = _selector.SelectServers(_description, connectedServers).ToList();
+                
+                var clusterableServers = new List<IClusterableServer>();
 
-                while (selectedServers.Count > 0)
+                while (selectedServers.Count > 0 && clusterableServers.Count < 2)
```
DmitryLukyanov, line 497: can you elaborate on why you expect not bigger than 2 severs here?

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -491,34 +491,38 @@ public IServer SelectServer()
...
                 {
                     var server = selectedServers.Count == 1 ?
                         selectedServers[0] :
                         __randomServerSelector.SelectServers(_description, selectedServers).Single();
```
DmitryLukyanov, line 501: as far as I understood the ticket description, the target of this ticket is the replacement of this randomSelector

DmitryLukyanov: so, the natural way to reach this is by moving connectionPool description to the ServerDescription, let's try to discuss this tomorrow with @rstam  once again before giving up here

DmitryLukyanov (commented):
Initial review

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: Other options to consider (I don't say that they're better than the previous ones):
1. Add a method to the server description like: `GetConnectionDescriptionState`. This will call some delegate (?) which is filled on the server-side and return the required count
2. Add an event on the server-side like `ConnectionDescriptionChnaged` and handle this on the cluster side

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: Resolved. I missed this in the spec.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: Discussed offline, the current way is the simplest

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: The above selector should be updated to reflect the spec requirements about choosing only 2 random servers

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+
+namespace MongoDB.Driver.Specifications.server_selection
+{
+    public sealed class InWindowTestRunner
+    {
+        private static class Schema
+        {
+            public static readonly string description = nameof(description);
+            public static readonly string topology_description = nameof(topology_description);
+            public static readonly string mocked_topology_state = nameof(mocked_topology_state);
```
JamesKovacs, line 40: Why is this file using snake case for field names?

JamesKovacs (commented):
Overall looking good. I added comments with my questions/concerns.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -489,36 +500,47 @@ public IServer SelectServer()
...
+                    .SelectServers(_description, _connectedServersDescriptions)
+                    .ToList();
 
-                while (selectedServers.Count > 0)
+                IServer selectedServer = null;
+
+                if (selectedServers.Any())
                 {
-                    var server = selectedServers.Count == 1 ?
+                    var selectedServerDescription = selectedServers.Count == 1 ?
```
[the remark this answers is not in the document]
DmitryLukyanov: also, expected formatting here:

                    var selectedServerDescription = selectedServers.Count == 1
                        ? selectedServers[0]
                        : __randomServerSelector.SelectServers(_description, selectedServers).Single();

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
```diff
@@ -0,0 +1,65 @@
...
+*/
+
+using System.Collections.Generic;
+using System.Linq;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.Servers;
+
+namespace MongoDB.Driver.Core.Clusters.ServerSelectors
+{
+    internal sealed class OperationsCountServerSelector : IServerSelector
```
DmitryLukyanov, line 23: the rest of serverSelectors are public because they're expected to possibly be configured from the user side, but maybe in this particular case because of ctor this class is too internal for this, not sure. @JamesKovacs what do you think?

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        public static List<ServerDescription> BuildServerDescriptions(BsonArray serverDescriptions, ClusterId clusterId, TimeSpan heartbeatInterval) =>
+            serverDescriptions.Select(x => BuildServerDescription((BsonDocument)x, clusterId, heartbeatInterval)).ToList();
+
+        // private methods
+#pragma warning disable CS0618 // Type or member is obsolete
+        private static (ClusterType, ClusterConnectionMode) GetClusterType(string type)
+        {
+            if (type.StartsWith(Schema.TopologyDescription.ClusterType.ReplicaSet))
```
DmitryLukyanov, line 119: switch?

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -591,14 +613,19 @@ private sealed class ServerChannel : IChannelHandle
         {
             // fields
             private readonly IConnectionHandle _connection;
-            private bool _disposed;
             private readonly Server _server;
 
+            private readonly InterlockedInt32 _state;
```
DmitryLukyanov, line 618: it sounds like we don't use `Open` state which can be confusing, do we really need this change?

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -414,12 +414,23 @@ private class SelectServerHelper : IDisposable
             private Task _descriptionChangedTask;
             private bool _serverSelectionWaitQueueEntered;
             private readonly IServerSelector _selector;
+
+            private readonly List<IClusterableServer> _connectedServers;
+            private readonly List<ServerDescription> _connectedServersDescriptions;
```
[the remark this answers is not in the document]
DmitryLukyanov: Can we remove this and use only `_connectedServers`?  ServerDescription is a field in a server

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -489,36 +500,47 @@ public IServer SelectServer()
...
+                }
+
+                var selectedServers = _selector
+                    .SelectServers(_description, _connectedServersDescriptions)
+                    .ToList();
 
-                while (selectedServers.Count > 0)
+                IServer selectedServer = null;
+
+                if (selectedServers.Any())
```
DmitryLukyanov, line 522: as I remember `Any` calls GetEnumerator which then calls `Dispose`, this is why we try to avoid Linq methods if possible and use the Count property

DmitryLukyanov (commented):
Is this prose test https://github.com/mongodb/specifications/commit/07790d1978ebe6b8ff7f574c9ac5e1365dd16e26#diff-8b993a775e4afe3d3151060c3903f9a1abb803ab0a3e314a2dcd79bf3e6cb272R279 implemented?

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+        private class TestCaseFactory : JsonDrivenTestCaseFactory
+        {
+            protected override string[] PathPrefixes => new[]
+            {
+                "MongoDB.Driver.Core.Tests.Specifications.server_selection.tests.in_window.",
+            };
+
+            protected override IEnumerable<JsonDrivenTestCase> CreateTestCases(BsonDocument document)
+            {
+                var name = GetTestCaseName(document, document, 0);
```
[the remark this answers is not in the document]
BorisDog: Good suggestion, done.

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
BorisDog: Unfortunately schema values are not consts but readonly.

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
```diff
@@ -0,0 +1,241 @@
...
+        private static TagSet BuildTagSet(BsonDocument tagSet)
+        {
+            JsonDrivenHelper.EnsureAllFieldsAreValid(tagSet, Schema.TopologyDescription.ServerTags.AllFields);
+
+            return new TagSet(tagSet.Elements.Select(x => new Tag(x.Name, x.Value.ToString())));
+        }
+
+        private static ServerType GetServerType(string type)
+        {
+            if (type == Schema.TopologyDescription.ServerType.RSPrimary)
```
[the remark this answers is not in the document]
BorisDog: Unfortunately schema values are not consts but readonly.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -70,6 +70,8 @@ internal sealed class Server : IClusterableServer
         private readonly InterlockedInt32 _state;
         private readonly ServerApi _serverApi;
 
+        private int _outstandingOperationsCount;
```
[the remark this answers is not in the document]
DmitryLukyanov: I think this is exactly the same case as we have with the server, I don't believe that anyone uses our pool interface. My suggestion will simplify the server logic since now we won't need to calculate the pool-related information on the server-side. Since we already have this data in the pool. Also, It will make the server logic easier since we won't need to "spread" logic to calculate one field across the server.cs(in several different parts of the server.cs). If you and @JamesKovacs disagreed with this, I won't insist.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -641,6 +718,70 @@ internal void ShouldInvalidateServer_should_return_expected_result_for_MongoWrit
...
+                    .Setup(c => c.OpenAsync(It.IsAny<CancellationToken>()))
+                    .Throws(new MongoAuthenticationException(connectionId, "Invalid login."));
+            }
+
+            var mockConnectionPool = new Mock<IConnectionPool>();
+            if (exceptionOnConnectionAcquire)
+            {
+                mockConnectionPool
+                    .Setup(p => p.AcquireConnection(It.IsAny<CancellationToken>()))
+                    .Throws(new TimeoutException("Timeout"));
```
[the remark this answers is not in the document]
DmitryLukyanov: As far as I can see, SemaphoreSlim doesn't trigger a timeout exception (I missed something?)

>We don't really need to care about the exception details here.

 I think it's a bit more important. The test can be investigated to understand the underlying driver logic, so the test described behavior should not contradict what the test can do.

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
DmitryLukyanov: Maybe not exactly this line, but you use deserialization below: `BsonSerializer.Deserialize<OperationsCount>`. I just want to say that this contradicts to how other test runners look like and I don't really see the reason for this yet

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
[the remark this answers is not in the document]
DmitryLukyanov: For simplicity, I would say it's better to avoid handling "the same" entities twice :) Line 516 is the only place when we need this field so the target here can be reached via:

               var selectedServers = _selector
                    .SelectServers(_description, _connectedServers.Select(s=>s.Description))
                    .ToList();

DmitryLukyanov:
> 
> 
> Is this prose test [mongodb/specifications@07790d1#diff-8b993a775e4afe3d3151060c3903f9a1abb803ab0a3e314a2dcd79bf3e6cb272R279](https://github.com/mongodb/specifications/commit/07790d1978ebe6b8ff7f574c9ac5e1365dd16e26#diff-8b993a775e4afe3d3151060c3903f9a1abb803ab0a3e314a2dcd79bf3e6cb272R279) implemented?

@BorisDog  have you had a chance to look at this?

src/MongoDB.Driver.Core/Core/Servers/Server.cs
DmitryLukyanov: not sure what is the best way here (maybe two state classes with inheritance?), but at least let's add a comment that we're not going to use `Open` state

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -489,36 +497,47 @@ public IServer SelectServer()
 
                 MongoIncompatibleDriverException.ThrowIfNotSupported(_description);
 
-                var connectedServers = _description.Servers.Where(s => s.State == ServerState.Connected);
-                var selectedServers = _selector.SelectServers(_description, connectedServers).ToList();
+                _connectedServers.Clear();
+                _connectedServerDescriptions.Clear();
```
DmitryLukyanov, line 501: nit: these lines are not really needed since we call `SelectServer` only once, but maybe it worth having this also

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
BorisDog: Right, using `BsonSerializer.Deserialize<Outcome>` as well. I find it very convenient to define the schema and object model in one place, and to deserialize it directly.

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
BorisDog: nice, done. tnx!

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+            public int operation_count { get; set; }
+        }
+
+        [Theory]
+        [ClassData(typeof(TestCaseFactory))]
+        public void RunTestDefinition(JsonDrivenTestCase testCase)
+        {
+            var testDefinition = testCase.Test;
+
+            JsonDrivenHelper.EnsureAllFieldsAreValid(testDefinition, Schema.RootFields);
```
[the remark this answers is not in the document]
BorisDog: I find that defining the schema in single place and using predefined sets as `Schema.RootFields`  greatly increases readability, and simplifies future alterations if any.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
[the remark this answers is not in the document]
DmitryLukyanov: Thanks, @BorisDog for clarification, I missed place where we trigger timeout exception in the pool

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
DmitryLukyanov: got it

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
```diff
@@ -0,0 +1,160 @@
...
+            ReplicaSetWithPrimary,
+            ReplicaSetNoPrimary,
+            Sharded,
+            Single,
+            Unknown
+        }
+
+        private enum ServerTypeTest
+        {
+           RSPrimary = ServerType.ReplicaSetPrimary,
```
DmitryLukyanov, line 40: incorrect tabs

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        private const string _collectionName = "test";
+        private const string _databaseName = "test";
+
+        /// <summary>
+        /// Test that starting a new transaction on a pinned ClientSession unpins the
+        /// session and normal server selection is performed for the next operation.
+        /// </summary>
+        [SkippableTheory]
+        [ParameterAttributeData]
+        public void SelectServer_loadbalancing_prose_test([Values(false, true)] bool async)
```
DmitryLukyanov, line 61: My impression about this test:
1. We don't really understand what happens here since auth variants unexpectedly don't work
2. Also, all prose tests before were really straightforward in meaning that it was enough to use only high-level API, in your implementation you use methods like `GetChannel` that I would not expect here.

Since we're going to investigate this further, I would move the whole prose test to the new ticket and check them there. cc: @JamesKovacs

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            RequireServer.Check()
+                .Supports(Feature.ShardedTransactions, Feature.FailPointsBlockConnection)
+                .ClusterType(ClusterType.Sharded);
+            RequireMultipleShardRouters();
+
+            // temporary disable the test on Win Auth topologies, due to operations timings irregularities
+            if (CoreTestConfiguration.ConnectionString.Tls == true &&
```
DmitryLukyanov, line 69: this way probably should work also, but this is a condition which we use to determine auth variant: ` CoreTestConfiguration.ConnectionString.Username != null;`

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -338,7 +413,7 @@ public void Dispose_should_dispose_the_server()
             subject.Description.ReasonChanged.Should().Contain("ChannelException during handshake");
             mockConnectionPool.Verify(p => p.Clear(), Times.Once);
         }
-
+
```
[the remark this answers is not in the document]
BorisDog: Thanks, fixed.

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/ServerSelectionTestHelpers.cs
BorisDog: Good catch, fixed

tests/MongoDB.Driver.Tests/ClusterTests.cs
BorisDog: Thanks, done.

tests/MongoDB.Driver.Tests/ClusterTests.cs
BorisDog: This test works well on not Win Auth topologies, so I'd prefer to investigate the problematic topologies in scope of a new ticket, and have the test running on all other topologies.

DmitryLukyanov: mostlikely because we run auth tests only on windows yet :)

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly IEnumerable<IClusterableServer> _clusterableServers;
+
+        public OperationsCountServerSelector(IEnumerable<IClusterableServer> clusterableServers)
+        {
+            _clusterableServers = clusterableServers;
+        }
+
+        // methods
+        /// <inheritdoc/>
+        public IEnumerable<ServerDescription> SelectServers(ClusterDescription cluster, IEnumerable<ServerDescription> servers)
```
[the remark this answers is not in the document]
JamesKovacs: Ah yes. If it's part of the public interface, then we are kind of stuck. We have to maintain the public interface definition even if it is unlikely that external parties would have implemented their own.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -160,35 +165,46 @@ public void Dispose()
         public IChannelHandle GetChannel(CancellationToken cancellationToken)
         {
             ThrowIfNotOpen();
+            IConnectionHandle connection = null;
 
-            var connection = _connectionPool.AcquireConnection(cancellationToken);
             try
             {
+                Interlocked.Increment(ref _outstandingOperationsCount);
+                connection = _connectionPool.AcquireConnection(cancellationToken);
```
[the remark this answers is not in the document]
JamesKovacs: Gotcha. Missed the `throw;` at the end of the catch.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -160,35 +165,46 @@ public void Dispose()
...
         public async Task<IChannelHandle> GetChannelAsync(CancellationToken cancellationToken)
         {
             ThrowIfNotOpen();
+            IConnectionHandle connection = null;
 
-            var connection = await _connectionPool.AcquireConnectionAsync(cancellationToken).ConfigureAwait(false);
             try
             {
+                Interlocked.Increment(ref _outstandingOperationsCount);
+                connection = await _connectionPool.AcquireConnectionAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
JamesKovacs: Missed the rethrow.

tests/MongoDB.Driver.Tests/ClusterTests.cs
JamesKovacs: Agreed on both counts.

JamesKovacs (changes requested):
Some minor changes requested.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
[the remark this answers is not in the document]
DmitryLukyanov: ok, then this is resolved

tests/MongoDB.Driver.Tests/ClusterTests.cs
DmitryLukyanov: ok, i will approve this now and will make the more deep check in the next ticket

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -0,0 +1,200 @@
...
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.Servers;
+using MongoDB.Driver.Core.TestHelpers;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.Driver.TestHelpers;
+using Xunit;
+
+namespace MongoDB.Driver.Tests
+{
+    public class ClusterTests
```
DmitryLukyanov, line 39: call it ServerSelectionProseTests

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        public void SelectServer_loadbalancing_prose_test([Values(false, true)] bool async)
+        {
+            RequireServer.Check()
+                .Supports(Feature.ShardedTransactions, Feature.FailPointsBlockConnection)
+                .ClusterType(ClusterType.Sharded);
+            RequireMultipleShardRouters();
+
+            // temporary disable the test on Win Auth topologies, due to operations timings irregularities
+            if (CoreTestConfiguration.ConnectionString.Username != null &&
+                RequirePlatform.GetCurrentOperatingSystem() == SupportedOperatingSystem.Windows)
```
DmitryLukyanov, line 70: I think Windows is not necessary here, and we will see the same behavior on all auth variants when we run this, so I would say, it's enough to say just: `RequireServer.Check().Authentication(false)`

DmitryLukyanov: or: `RequirePlatform.Check().SkipWhen(() => CoreTestConfiguration.ConnectionString.Username != null, SupportedOperatingSystem.Windows);`, but the first one looks enough to me

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -0,0 +1,200 @@
...
+using MongoDB.Driver.Core.TestHelpers;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.Driver.TestHelpers;
+using Xunit;
+
+namespace MongoDB.Driver.Tests
+{
+    public class ClusterTests
+    {
+        private static readonly HashSet<string> __commandsToNotCapture = new HashSet<string>
```
DmitryLukyanov, line 41: we definitely should do something with these commands at some point, we keep copying them all the time

DmitryLukyanov (approved):
See my last comments, then LGTM

tests/MongoDB.Driver.Core.Tests/Specifications/server-selection/InWindowTestRunner.cs
JamesKovacs: Now I understand what you're doing. Basically creating strongly-typed string constants. Interesting technique. I'm OK with it.

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
BorisDog: Not sure about exposing it publicly. As it relies on the functionality of `IClusterableServer` which is intended to be internal and is public only due historic reasons. Further more, we will probably consider converting all pseudo-internal API to internal in v3, so it's better to avoid expanding pseudo-internal API when possible.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -134,6 +137,8 @@ internal sealed class Server : IClusterableServer
 
         internal IClusterClock ClusterClock => _clusterClock;
 
+        int IClusterableServer.OutstandingOperationsCount => Interlocked.CompareExchange(ref _outstandingOperationsCount, 0, 0);
```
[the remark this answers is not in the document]
BorisDog: Good catch. Leftovers from refactoring. Fixed.

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/OperationsCountServerSelector.cs
JamesKovacs: I'll mark this as resolved. If a user needs this public for custom configuration reasons, they can file a JIRA issue to request it. I think the risk is low and I prefer not expanding our API surface anymore than is necessary.

JamesKovacs (approved):
LGTM

PR #481, opened 2021-03-23
CSHARP-3448: Update FLE documentation to mention that we currently don't support arm architecture.

DmitryLukyanov:
Only docs changes.
I restored changes from CSHARP-3300: https://github.com/mongodb/mongo-csharp-driver/pull/450/files#diff-525914cfce0449efb9693ca2513ce901aa04b611f50f800f2e42748fb922ecdfL24 with changed wording.


JamesKovacs (approved):
LGTM

PR #483, opened 2021-03-30
CSHARP-3386: Drivers should not inspect error messages for 'not master' or 'node is recovering'

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3386
EG: https://evergreen.mongodb.com/version/606284d92a60ed5c62e7451f
Spec changes which should explain everything: https://github.com/mongodb/specifications/commit/6b885961d53afc1551eec3fa89ccaf5e9118081c


tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
```diff
@@ -448,6 +448,7 @@ public void A_description_changed_event_with_a_heartbeat_exception_should_clear_
 
         [Theory]
         [InlineData((ServerErrorCode)(-1), false)]
```
DmitryLukyanov, line 450: we probably should review whether we use -1 in the current version as it was expected before. Previously it meant unassigned, not it means a wrong code

DmitryLukyanov (approved):
Almost LGTM, before it, check whether we need to review -1 value in ServerTests again

rstam (approved):
LGTM

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerTests.cs
DmitryLukyanov: With your current changes, the original test idea was changed also. In this -1 test and the current implementation, you will be in this code path: https://github.com/mongodb/mongo-csharp-driver/blob/f9567acdabe666fafb5ced55ce679780fbae7dfb/src/MongoDB.Driver.Core/Core/Servers/Server.cs#L394, but the expected behavior is to avoid this `if-then` and heading right to the else branch. So, if we leave the current changes, we still need to add the case with `(ServerErrorCode)null`, I would also add a comment that the -1 means a different thing here than in other places

MikalaiMazurenka: Done

MikalaiMazurenka (commented):
Ready for review

DmitryLukyanov (approved):
LGTM

PR #489, opened 2021-04-06
Update releases.toml.

DmitryLukyanov:
Doing it explicitly to ensure that nothing missed in my first releasing

### DIFF
--- a/Docs/landing/data/releases.toml
+++ b/Docs/landing/data/releases.toml
@@ -1,7 +1,7 @@
-current = "2.12.0"
+current = "2.12.1"
 
 [[versions]]
-  version = "2.12.0"
+  version = "2.12.1"
   status = "current"
   docs = "./2.12/"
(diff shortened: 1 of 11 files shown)


Docs/landing/data/releases.toml
```diff
@@ -1,7 +1,7 @@
 current = "2.12.1"
```
JamesKovacs, line 1: Need to update current.

JamesKovacs (changes requested):
Missed one.

JamesKovacs:
I would recommend doing the PR on `master` and then cherrypicking to `v2.12.x`.

Docs/landing/data/releases.toml
DmitryLukyanov: done

DmitryLukyanov:
moved this code into different PR for the master

PR #492, opened 2021-04-06
CSHARP-3462: Unobserved task exception on connection failure raises its head again

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3462
EG: https://evergreen.mongodb.com/version/608081e657e85a60d6a8b5f8

Note: This is leftover from [CSHARP-3102: Memory leak related to heartbeat when connecting to Atlas clusters](https://github.com/mongodb/mongo-csharp-driver/commit/a2303983c22639a114727ec3696cb667e7817b8b).
And the sync path (`BinaryConnection.ReceiveMessage`) did not produce an unhandled exception (or it is not raised in the event for some reason), but I included both paths in the unit test.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
@@ -371,7 +371,6 @@ namespace MongoDB.Driver.Core.Connections
                         catch (Exception ex)
                         {
                             _dropbox.AddException(ex);
-                            throw;
                         }
 
                         if (messageTask.IsCompleted)
@@ -442,7 +441,6 @@ namespace MongoDB.Driver.Core.Connections
                         catch (Exception ex)
                         {
                             _dropbox.AddException(ex);
-                            throw;
                         }
 
                         if (messageTask.IsCompleted)
@@ -786,7 +784,6 @@ namespace MongoDB.Driver.Core.Connections
             {
                 foreach (var taskCompletionSource in _messages.Values)
                 {
-                    taskCompletionSource.Task.IgnoreExceptions();
                     taskCompletionSource.TrySetException(exception); // has no effect on already completed tasks
                 }
             }
--- a/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
@@ -786,6 +786,7 @@ namespace MongoDB.Driver.Core.Connections
             {
                 foreach (var taskCompletionSource in _messages.Values)
                 {
+                    taskCompletionSource.Task.IgnoreExceptions();
                     taskCompletionSource.TrySetException(exception); // has no effect on already completed tasks
                 }
             }
--- a/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
@@ -357,7 +357,7 @@ namespace MongoDB.Driver.Core.Connections
                     Task.WaitAny(messageTask, receiveLockRequest.Task);
                     if (messageTask.IsCompleted)
                     {
-                        return _dropbox.RemoveMessage(responseTo);
+                        return _dropbox.RemoveMessage(responseTo); // also propagates exception if any
                     }
 
                     receiveLockRequest.Task.GetAwaiter().GetResult(); // propagate exceptions
@@ -375,7 +375,7 @@ namespace MongoDB.Driver.Core.Connections
 
                         if (messageTask.IsCompleted)
                         {
-                            return _dropbox.RemoveMessage(responseTo);
+                            return _dropbox.RemoveMessage(responseTo); // also propagates exception if any
                         }
 
                         cancellationToken.ThrowIfCancellationRequested();
@@ -427,7 +427,7 @@ namespace MongoDB.Driver.Core.Connections
                     await Task.WhenAny(messageTask, receiveLockRequest.Task).ConfigureAwait(false);
                     if (messageTask.IsCompleted)
                     {
-                        return _dropbox.RemoveMessage(responseTo);
+                        return _dropbox.RemoveMessage(responseTo); // also propagates exception if any
                     }
 
                     receiveLockRequest.Task.GetAwaiter().GetResult(); // propagate exceptions
@@ -445,7 +445,7 @@ namespace MongoDB.Driver.Core.Connections
 
                         if (messageTask.IsCompleted)
                         {
-                            return _dropbox.RemoveMessage(responseTo);
+                            return _dropbox.RemoveMessage(responseTo); // also propagates exception if any
                         }
 
                         cancellationToken.ThrowIfCancellationRequested();
--- a/tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
@@ -470,6 +470,68 @@ namespace MongoDB.Driver.Core.Connections
 
         [Theory]
         [ParameterAttributeData]
+        public void ReceiveMessage_should_not_produce_unobserved_task_exceptions_on_fail(
+            [Values(false, true)] bool async)
+        {
+            var unobservedTaskExceptionRaised = false;
+            var mockStream = new Mock<Stream>();
+            EventHandler<UnobservedTaskExceptionEventArgs> eventHandler = (s, args) =>
+            {
+                if (args.Exception.InnerException is MongoConnectionException)
+                {
+                    unobservedTaskExceptionRaised = true;
+                }
+            };
+
+            try
+            {
+                TaskScheduler.UnobservedTaskException += eventHandler;
+                var encoderSelector = new ReplyMessageEncoderSelector<BsonDocument>(BsonDocumentSerializer.Instance);
+
+                _mockStreamFactory
+                    .Setup(f => f.CreateStream(_endPoint, CancellationToken.None))
+                    .Returns(mockStream.Object);
+
+                if (async)
+                {
+                    mockStream
+                        .Setup(s => s.ReadAsync(It.IsAny<byte[]>(), It.IsAny<int>(), It.IsAny<int>(), It.IsAny<CancellationToken>()))
+                        .Throws(new SocketException());
+                }
+                else
+                {
+                    mockStream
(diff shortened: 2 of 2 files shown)


src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -786,6 +786,7 @@ public void AddException(Exception exception)
             {
                 foreach (var taskCompletionSource in _messages.Values)
                 {
+                    taskCompletionSource.Task.IgnoreExceptions();
```
rstam, line 789: I'm concerned that this is the wrong place to do this.

We don't want to ignore ALL unobserved task exceptions.

What we actually want to do is "observe" them all so that there are no unobserved exceptions.

Where was the unobserved task exception happening? I suspect it was in the heartbeat. Let's change the heartbeat code instead to observe the exception.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
MikalaiMazurenka: > I'm concerned that this is the wrong place to do this.

I think this is the only place where this exception can be observed, because it happens on the `taskCompletionSource` task object.

> We don't want to ignore ALL unobserved task exceptions.

This exception will be still be thrown: [BinaryConnection:374](https://github.com/mongodb/mongo-csharp-driver/blob/bb5b38d7407b37453ce1aac82e8a67f5526b2f86/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L374)

> What we actually want to do is "observe" them all so that there are no unobserved exceptions.

This is exactly what `TaskExtensions.IgnoreExceptions` is doing: [TaskExtensions.IgnoreExceptions](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Misc/TaskExtensions.cs#L22)
By accessing the `.Exception` property of the task, it's exception is considered as observed.

> Where was the unobserved task exception happening? I suspect it was in the heartbeat. Let's change the heartbeat code instead to observe the exception.

I didn't manage to reproduce a case when it was not caused by heartbeat, and the heartbeat code in my opinion definitely requires refactoring, but I'm inclined to fix the issue at the lowest possible point, which is `BinaryConnection`. You can check the added unit test, which shows that the issue can be reproduced (with mocks) even without heartbeat involvement.

Additionally, we have a very similar logic in `BinaryConection.Open`: [BinaryConection:231](https://github.com/mongodb/mongo-csharp-driver/blob/bb5b38d7407b37453ce1aac82e8a67f5526b2f86/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L231). The exception is set on the `taskCompletionSource.Task`, but observed `_openTask.IgnoreExceptions();` and thrown.

MikalaiMazurenka (commented):
Please check my answers. If you still have concerns I will try to clarify things further and potentially make changes if required.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
rstam: I understand that by "ignoring" the exception we are technically "observing" it as far as .NET is concerned.

But I don't understand how "ignoring" the exception can be the right thing to do. Seems like all exceptions should be propagated to the caller and caught somewhere.

An unobserved exception means the exception was not caught, which would be a bug in our code.

Am I missing something?

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
MikalaiMazurenka: I didn't manage to unit test it, but I reproduced a case when we have two threads reading from the same connection and one sets an exception on `taskCompletionSource`. In this case when the exception is thrown on line [BinaryConnection:374](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L374) the other thread unlocks and attempts to remove a message from dropbox on line [BinaryConnection:360](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L360) which will result in propagating exception from `taskCompletionSource` ([BinaryConnection:810](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L810)).
So for each ignore call on `taskCompletionSource` there will be this very exception thrown to the caller, either via throw on line [BinaryConnection:374](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L374) or by accessing the result of completion source task on line [BinaryConnection:810](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L810) in multithreaded scenario.
To clarify, by calling [Dropbox.AddException](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L785) we add the exception to `taskCompletionSource` task and throw it synchronously ([BinaryConnection:374](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L374)), which sort of doubles the number of exceptions in this case (one thrown and one added to task), requiring us to observe the exception in the unobserved `taskCompletionSource` task.

MikalaiMazurenka (commented):
I think this explanation is a bit better

MikalaiMazurenka:
@DmitryLukyanov suggested another approach with explicit observation of message task. I've checked it and looks way better than the previous one. Instead of adding the exception to `taskCompletionSource.Task` and then additionally throwing the exception on line [BinaryConnection:374](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L374), now we only add the exception to `taskCompletionSource.Task` and then observe it's result on line [BinaryConnection:379](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L379).

Note: I've force pushed the branch to include the fix for latest test fails from master to see the tests results clearly.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -371,7 +371,6 @@ private IByteBuffer ReceiveBuffer(int responseTo, CancellationToken cancellation
                         catch (Exception ex)
                         {
                             _dropbox.AddException(ex);
-                            throw;
```
MikalaiMazurenka, line 374: The exception will be observed and thrown by calling `return _dropbox.RemoveMessage(responseTo);` on the line 378 on the right side, because the [BinaryConnection.Dropbox.RemoveMessage](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs#L810) observes the message task.

rstam (approved):
LGTM

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -371,7 +371,6 @@ private IByteBuffer ReceiveBuffer(int responseTo, CancellationToken cancellation
                         catch (Exception ex)
                         {
                             _dropbox.AddException(ex);
```
DmitryLukyanov, line 373: this is probably not the issue for us since as I remember we don't call ReceiveBuffer with more than 1 attempt anymore (am I right?), but if we would do it, we would be able to have a memory leak in line 378 since RemoveMessage return buffer that is not disposed. But in our case it's not an issue, we cannot call addMessage since mostlikely the error will be thrown in line 368, so the buffer even is not created yet

DmitryLukyanov (commented):
Some comments and then it will be LGTM

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
MikalaiMazurenka: Done

tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
```diff
@@ -468,6 +468,68 @@ public void Dispose_should_raise_the_correct_events()
             }
         }
 
+        [Theory]
+        [ParameterAttributeData]
+        public void ReceiveMessage_should_not_produce_unobserved_task_exceptions_on_fail(
```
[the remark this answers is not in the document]
MikalaiMazurenka: Done

tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
```diff
@@ -468,6 +468,68 @@ public void Dispose_should_raise_the_correct_events()
...
+            [Values(false, true)] bool async)
+        {
+            var unobservedTaskExceptionRaised = false;
+            var mockStream = new Mock<Stream>();
```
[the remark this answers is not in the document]
MikalaiMazurenka: I decided to leave it in finally, since this block is present anyway.

tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
```diff
@@ -468,6 +468,68 @@ public void Dispose_should_raise_the_correct_events()
...
+            EventHandler<UnobservedTaskExceptionEventArgs> eventHandler = (s, args) =>
+            {
+                if (args.Exception.InnerException is MongoConnectionException)
```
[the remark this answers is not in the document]
MikalaiMazurenka: As discussed offline, now handling all of the exceptions and marking them as observed

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
MikalaiMazurenka: Discussed offline. There could be theoretically a race condition when this happens, but practically this shouldn't be an issue.

MikalaiMazurenka (commented):
Ready for review!

DmitryLukyanov (approved):
LGTM

PR #494, opened 2021-04-07
CSHARP-3515: Sporadic tests failures: TestGuidRefId

rstam:
See comment in  https://jira.mongodb.org/browse/CSHARP-3515 for an explanation of why these changes will probably fix this sporadic test failure.

### DIFF
--- a/tests/MongoDB.Bson.TestHelpers/GuidMode.cs
+++ b/tests/MongoDB.Bson.TestHelpers/GuidMode.cs
@@ -14,7 +14,7 @@
 */
 
 using System.Collections.Generic;
-using System.Linq;
+using MongoDB.Bson.IO;
 using Xunit.Abstractions;
 
 namespace MongoDB.Bson.TestHelpers
@@ -36,24 +36,6 @@ namespace MongoDB.Bson.TestHelpers
         // public static properties
         public static IReadOnlyList<GuidMode> All => __all;
 
-        public static GuidMode Current
-        {
-            get
-            {
-#pragma warning disable 618
-                var currentMode = BsonDefaults.GuidRepresentationMode;
-                if (currentMode == GuidRepresentationMode.V2)
-                {
-                    return new GuidMode(currentMode, BsonDefaults.GuidRepresentation);
-                }
-                else
-                {
-                    return new GuidMode(currentMode);
-                }
-#pragma warning restore 618
-            }
-        }
-
         // public static methods
         public static void Set(GuidRepresentationMode guidRepresentationMode, GuidRepresentation guidRepresentation = GuidRepresentation.Unspecified)
         {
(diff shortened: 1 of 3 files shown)


rstam:
Evergreen patch:

https://evergreen.mongodb.com/version/606de2e5c9ec446a1c4b9cbf

tests/MongoDB.Bson.TestHelpers/GuidMode.cs
```diff
@@ -36,24 +36,6 @@ public class GuidMode : IXunitSerializable
         // public static properties
         public static IReadOnlyList<GuidMode> All => __all;
 
-        public static GuidMode Current
```
rstam, line 39: Changed to a method because it does more work than a property should.

tests/MongoDB.Bson.TestHelpers/GuidMode.cs
```diff
@@ -62,9 +44,17 @@ public static void Set(GuidRepresentationMode guidRepresentationMode, GuidRepres
         }
         #endregion
 
+        // private fields
+        private BsonBinaryReaderSettings _defaultBsonBinaryReaderSettings;
+        private BsonBinaryWriterSettings _defaultBsonBinaryWriterSettings;
+        private BsonDocumentReaderSettings _defaultBsonDocumentReaderSettings;
+        private BsonDocumentWriterSettings _defaultBsonDocumentWriterSettings;
+        private JsonReaderSettings _defaultJsonReaderSettings;
+        private JsonWriterSettings _defaultJsonWriterSettings;
```
rstam, line 53: Capture these settings as part of the global state related to Guid modes.

Note: these will be null after calling the `GuidMode()` constructor, but they will have the current global state when a `GuidMode` is returned from calling `CaptureCurrentSettings()`.

tests/MongoDB.Driver.Tests/MongoDBRefTests.cs
```diff
@@ -158,9 +159,14 @@ public void TestEqualsWithoutDatabase()
             Assert.Equal(a1.GetHashCode(), a2.GetHashCode());
         }
 
-        [Fact]
-        public void TestGuidRefId()
+        [Theory]
+        [ParameterAttributeData]
+        [ResetGuidModeAfterTest]
```
rstam, line 164: This ensures the global state is restored at the end of the test.

tests/MongoDB.Driver.Tests/MongoDBRefTests.cs
```diff
@@ -158,9 +159,14 @@ public void TestEqualsWithoutDatabase()
...
+        public void TestGuidRefId(
+            [ClassValues(typeof(GuidModeValues))] GuidMode mode)
```
rstam, line 166: We are now running this test method in all of the possible GuidMode values.

tests/MongoDB.Driver.Tests/MongoDBRefTests.cs
```diff
@@ -158,9 +159,14 @@ public void TestEqualsWithoutDatabase()
...
         {
+            mode.Set();
```
rstam, line 168: This sets the global state, including the values of the 6 reader/writer defaults.

Note: this results in the 6 global reader/writer defaults being set to null here because they are lazily initialized.

MikalaiMazurenka (approved):
Though I didn't find the source of the fail as well, I think it is connected to parallel test execution and thus should be fixed by this change.
LGTM!

PR #496, opened 2021-04-07
Updated docs with v2.12.2 changes.

DmitryLukyanov:
This link should work https://dmitrylukyanov.github.io/mongo-csharp-driver after a while


JamesKovacs (approved):
LGTM

PR #497, opened 2021-04-12
CSHARP-2936: Support [Serializable] when targeting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2936
EG: https://evergreen.mongodb.com/version/6074d8d23627e056aee4fc20

Note: there were several `[Serializable]` occurrences in Legacy, but I omitted them.

### DIFF
--- a/src/MongoDB.Driver.Core/MongoExecutionTimeoutException.cs
+++ b/src/MongoDB.Driver.Core/MongoExecutionTimeoutException.cs
@@ -14,10 +14,10 @@
 */
 
 using System;
-using MongoDB.Bson;
-#if NET452
+#if !NETSTANDARD1_5
 using System.Runtime.Serialization;
 #endif
+using MongoDB.Bson;
 using MongoDB.Driver.Core.Connections;
 
 namespace MongoDB.Driver
@@ -25,7 +25,7 @@ namespace MongoDB.Driver
     /// <summary>
     /// Represents a MongoDB execution timeout exception.
     /// </summary>
-#if NET452
+#if !NETSTANDARD1_5
     [Serializable]
 #endif
     public class MongoExecutionTimeoutException : MongoServerException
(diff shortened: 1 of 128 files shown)


src/MongoDB.Driver.Core/MongoExecutionTimeoutException.cs
```diff
@@ -79,7 +79,7 @@ public MongoExecutionTimeoutException(ConnectionId connectionId, string message,
             AddErrorLabelsFromCommandResult(this, result);
         }
 
-#if NET452
+#if !NETSTANDARD1_5
```
rstam, line 82: Add between lines 90-91:

    _result = (BsonDocument)info.GetValue("_result", typeof(BsonDocument));

Add between lines 112-113:

    #if !NETSTANDARD1_5
            /// <inheritdoc/>
            public override void GetObjectData(SerializationInfo info, StreamingContext context)
            {
                base.GetObjectData(info, context);
                info.AddValue("_result", _result);
            }
    #endif

rstam (changes requested):
One small change requested.

Good job and nice attention to detail!

rstam:
@jyemin is it OK to ignore Legacy in this PR?

jyemin:
Sure as long as we're tracking it somewhere

src/MongoDB.Driver.Core/MongoExecutionTimeoutException.cs
MikalaiMazurenka: Good catch! I've updated the unit test to contain the `_result` field as well.

MikalaiMazurenka (commented):
Thanks for appreciation!
Ready for review

rstam (approved):
LGTM

PR #499, opened 2021-04-14
CSHARP-3299: Enable Secure Tests on Linux.

DmitryLukyanov:
This PR for mac also :) 
EG: https://evergreen.mongodb.com/version/607727e81e2d176bc70b43a9


evergreen/add-certs-if-needed.sh
```diff
@@ -16,16 +17,26 @@ if [[ "$SSL" != "ssl" ]]; then
...
+  elif  [[ "$OS" =~ Ubuntu|ubuntu ]]; then
+    # makes the client.pem trusted
+    # note: .crt is the equivalent format as .pem, but we need to make this renaming because update-ca-certificates supports only .crt
+    sudo cp -f $1 /usr/local/share/ca-certificates/ca.crt
+    sudo update-ca-certificates
+  else
+    # mac OS, the same trick as for above ubuntu step
+    sudo cp -f $1 ~/ca.crt
+    sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/ca.crt
+    #sudo security delete-certificate -c "<name of existing certificate>"
```
DmitryLukyanov, line 34: it sounds like this is not required, a similar line is presented for windows in the guide, but we don't use it, so I will remove this line also. I temporarily left this here for now just in case

evergreen/add-certs-if-needed.sh
```diff
@@ -16,16 +17,26 @@ if [[ "$SSL" != "ssl" ]]; then
   exit 0
 fi
 
-if [[ "$OS" =~ Windows|windows ]]; then
-  certutil.exe -addstore "Root" ${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem
-
-  if [[ "$OCSP_TLS_SHOULD_SUCCEED" != "nil" && "$OCSP_ALGORITHM" != "nil" ]]; then
-    certutil.exe -addstore "Root" ${DRIVERS_TOOLS}/.evergreen/ocsp/${OCSP_ALGORITHM}/ca.pem
+function make_trusted() {
```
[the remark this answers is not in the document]
DmitryLukyanov: I'm open to suggestion/comments/critic here

evergreen/evergreen.yml
```diff
@@ -1263,6 +1274,16 @@ buildvariants:
      - name: test-netstandard20
      - name: test-netstandard21
 
+# macos-1014 does not support SSL until 3.2
+- matrix_name: "secure-tests-macOS"
+  matrix_spec: { version: ["3.2", "3.6", "4.0", "4.2", "4.4", "latest"], topology: "*", auth: "auth", ssl: "ssl", os: "macos-1014" }
+  display_name: "${version} ${topology} ${auth} ${ssl} ${os}"
+  tags: ["tests-variant"]
+  tasks:
+     # - name: test-netstandard15: auth tests fail on macos with test-netstandard15, so skip them
```
DmitryLukyanov, line 1283: with this line, we will see a lot of errors like this: https://evergreen.mongodb.com/task_log_raw/dot_net_driver_secure_tests_macOS__version~4.0_os~macos_1014_topology~sharded_cluster_auth~auth_ssl~ssl_test_netstandard15_patch_804de83e2388793b111beef8ffeb09b5cf697451_60771cb87742ae6ad4a3fdd5_21_04_14_16_48_03/0?type=T#L37944. I think we don't need to spend time on investigating this issue for netstandard1.5

evergreen/run-tests.sh
```diff
@@ -121,16 +121,16 @@ if [[ "$OS" =~ Windows|windows ]]; then
     powershell.exe '.\build.ps1 -target' $TARGET
   else
     powershell.exe \
-      '$env:MONGO_X509_CLIENT_CERTIFICATE_PATH="'${MONGO_X509_CLIENT_CERTIFICATE_PATH}'";'\
-      '$env:MONGO_X509_CLIENT_CERTIFICATE_PASSWORD="'${MONGO_X509_CLIENT_CERTIFICATE_PASSWORD}'";'\
+      '$env:MONGO_X509_CLIENT_CERTIFICATE_PATH="${MONGO_X509_CLIENT_CERTIFICATE_PATH}";'\
```
DmitryLukyanov, line 124: the value must be without `'`s

evergreen/add-certs-if-needed.sh
```diff
@@ -16,16 +17,26 @@ if [[ "$SSL" != "ssl" ]]; then
...
+    # note: .crt is the equivalent format as .pem, but we need to make this renaming because update-ca-certificates supports only .crt
+    sudo cp -f $1 /usr/local/share/ca-certificates/ca.crt
+    sudo update-ca-certificates
+  else
+    # mac OS, the same trick as for above ubuntu step
+    sudo cp -f $1 ~/ca.crt
+    sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/ca.crt
+    #sudo security delete-certificate -c "<name of existing certificate>"
   fi
-else
```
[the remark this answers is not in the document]
DmitryLukyanov: not sure why `keytool` way doesn't work for us, I saw somewhere that this way works only with java apps, but I'm not sure about details, anyway, the currently added way replaces these lines, so I think we can end up with the currently added approach

DmitryLukyanov:
EG: after rebasing

evergreen/add-certs-if-needed.sh
JamesKovacs: Unless we are changing test root CAs frequently (and I doubt we are), I agree that we don't need to worry about automating cleanup of previously used root CAs. On EG, this won't matter either since we start with a clean machine every time. It would only matter on our own machines when troubleshooting EG scripts.

JamesKovacs: Regarding the branching logic, we should detect OSX rather than falling through to it as the default. If we ever decide to run our test suite on Amazon Linux, RHEL, or any other OS, we will try to run the OSX commands on it and they likely won't work. I would include an `else` that errs with "unsupported OS" or something like that.

evergreen/add-certs-if-needed.sh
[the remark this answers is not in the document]
JamesKovacs: `keytool` is a Java utility that manages the Java KeyStore. This isn't the same keystore used by the OS and other language runtimes. So the `keytool` code was likely a holdover from porting the Java impl and wouldn't work for anything other than Java.

evergreen/evergreen.yml
```diff
@@ -259,14 +259,15 @@ functions:
...
-          SSL=${SSL} evergreen/add-certs-if-needed.sh
+          SSL=${SSL} OS=${OS} evergreen/add-certs-if-needed.sh
           AUTH=${AUTH} \
             SSL=${SSL} \
             MONGODB_URI="${MONGODB_URI}" \
             TOPOLOGY=${TOPOLOGY} \
             OS=${OS} \
             COMPRESSOR=${COMPRESSOR} \
             CLIENT_PEM=${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem \
+            CAFILE=${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem \
```
JamesKovacs, line 270: Why do we pass this in as `CAFILE` but then use `${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem` to locate the certs in `add-certs-if-needed.sh`? If we are going to pass in `CAFILE`, should we not also pass in the OCSP ca.pem too?

evergreen/evergreen.yml
JamesKovacs: .NET Core 1.1 uses OpenSSL on macOS, which may not be installed in the expected location depending on macOS version, especially on a clean build. Starting in .NET Core 2.0, .NET Core switched to using the Apple Security Framework on macOS.

https://github.com/dotnet/runtime/issues/17523
https://github.com/dotnet/announcements/issues/21

We can either `brew install openssl` on macOS to ensure that OpenSSL is installed for .NET Core 1.1 or we can skip the `netstandard1.5` tests on macOS. I would suggest the latter with a comment mentioning the missing OpenSSL dependency.

evergreen/run-tests.sh
JamesKovacs: Odd that it worked before. But OK.

JamesKovacs (changes requested):
Some minor feedback and changes requested. Overall looking good.

evergreen/evergreen.yml
DmitryLukyanov: you're right, Removed

evergreen/evergreen.yml
DmitryLukyanov: done

evergreen/add-certs-if-needed.sh
DmitryLukyanov: done

DmitryLukyanov (commented):
EG: https://evergreen.mongodb.com/version/6081aee2c9ec447a1a7257bd

tests/MongoDB.Driver.Tests/ClusterTests.cs
```diff
@@ -65,12 +65,8 @@ public void SelectServer_loadbalancing_prose_test([Values(false, true)] bool asy
                 .ClusterType(ClusterType.Sharded);
             RequireMultipleShardRouters();
 
-            // temporary disable the test on Win Auth topologies, due to operations timings irregularities
```
DmitryLukyanov, line 68: this test fails on auth envs, so only a win condition is not relevant anymore

JamesKovacs (approved):
LGTM

PR #500, opened 2021-04-16
CSHARP-2938: Refactor Stream subclasses when targeting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2938
EG: https://evergreen.mongodb.com/version/6079a9c5562343543bf6c0cf

Note: I wasn't adding new overrides, just adjusting existing.

### DIFF
--- a/src/MongoDB.Driver.GridFS/DelegatingStream.cs
+++ b/src/MongoDB.Driver.GridFS/DelegatingStream.cs
@@ -13,11 +13,10 @@
 * limitations under the License.
 */
 
+#if !NETSTANDARD1_5
 using System;
-using System.IO;
-#if NET452
-using System.Runtime.Remoting;
 #endif
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -42,7 +41,7 @@ namespace MongoDB.Driver.GridFS
             _wrappedStream = wrappedStream;
         }
 
-        // properties        
+        // properties
         /// <inheritdoc/>
         public override bool CanRead
         {
@@ -95,7 +94,7 @@ namespace MongoDB.Driver.GridFS
         }
 
         // methods
-#if NET452
+#if !NETSTANDARD1_5
         /// <inheritdoc/>
         public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
         {
@@ -103,7 +102,7 @@ namespace MongoDB.Driver.GridFS
         }
 #endif
 
-#if NET452
+#if !NETSTANDARD1_5
         /// <inheritdoc/>
         public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
         {
@@ -111,7 +110,7 @@ namespace MongoDB.Driver.GridFS
         }
 #endif
 
-#if NET452
+#if !NETSTANDARD1_5
         /// <inheritdoc/>
         public override void Close()
         {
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver.GridFS/DelegatingStream.cs
```diff
@@ -125,15 +124,7 @@ public override Task CopyToAsync(Stream destination, int bufferSize, Cancellatio
             return _wrappedStream.CopyToAsync(destination, bufferSize, cancellationToken);
         }
 
-#if NET452
-       /// <inheritdoc/>
-        public override ObjRef CreateObjRef(Type requestedType)
```
MikalaiMazurenka, line 130: Removed this, because it looks like an incorrect logic: we shouldn't be creating object reference of the underlying stream, but of current `DelegatingStream` instance.

src/MongoDB.Driver.GridFS/DelegatingStream.cs
```diff
@@ -191,26 +182,8 @@ public override int GetHashCode()
             return _wrappedStream.GetHashCode();
         }
 
-#if NET452
-       /// <inheritdoc/>
-        public override object InitializeLifetimeService()
```
MikalaiMazurenka, line 196: Same concern: we shouldn't be calling it on the underlying stream, but on current `DelegatingStream` instance.

src/MongoDB.Driver.GridFS/DelegatingStream.cs
```diff
@@ -191,26 +182,8 @@ public override int GetHashCode()
...
-        {
-            return _wrappedStream.InitializeLifetimeService();
-        }
-#endif
-
-#if NET452
-        /// <inheritdoc/>
-        [Obsolete("Not supported by DelegatingStream.")]
-        protected override void ObjectInvariant()
```
MikalaiMazurenka, line 205: This method is already obsolete in NET452, so I don't see reasons for obsoleting it in `DelegatingStream`.

tests/MongoDB.Driver.GridFS.Tests/DelegatingStreamTests.cs
```diff
@@ -0,0 +1,386 @@
+﻿/* Copyright 2021-present MongoDB Inc.
```
MikalaiMazurenka, line 1: Since I was making changes to `DelegatingStream` I thought it was a good idea to cover them with unit tests. And why not add the rest of the methods in this case.

MikalaiMazurenka (commented):
I have removed several NET452 methods. I know it's not the scope of this PR and the changes may be tricky, so I will revert them if necessary.

src/MongoDB.Driver.GridFS/DelegatingStream.cs
rstam: Agreed (note: this method is from MarshalByRefObject not Stream)

src/MongoDB.Driver.GridFS/DelegatingStream.cs
rstam: Agreed

src/MongoDB.Driver.GridFS/DelegatingStream.cs
rstam: That's fine.

tests/MongoDB.Driver.GridFS.Tests/DelegatingStreamTests.cs
rstam: Good idea. Thanks for doing this.

rstam (approved):
LGTM

PR #501, opened 2021-04-19
CSHARP-2940: Check uses of FormatterServices when targetting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2940
EG: https://evergreen.mongodb.com/version/607d8e9c57e85a62cdb149ed

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonClassMap.cs
+++ b/src/MongoDB.Bson/Serialization/BsonClassMap.cs
@@ -20,7 +20,9 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+#if !NETSTANDARD1_5
 using System.Runtime.Serialization;
+#endif
 using MongoDB.Bson.IO;
 using MongoDB.Bson.Serialization.Conventions;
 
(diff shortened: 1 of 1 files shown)


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -419,29 +421,11 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
         private static Type GetFormatterServicesType()
         {
-#if NET452
-            return typeof(FormatterServices);
-#else
-            // TODO: once we depend on newer versions of .NET Standard we should be able to do this without reflection
-
-            try
-            {
-                // new approach which works on .NET Core 3.0
```
rstam, line 429: I think removing this code is OK.

I assume the justification is that on .NET Core 3.0 we will now end up at line 428 (right side) instead?

rstam (approved):
LGTM

src/MongoDB.Bson/Serialization/BsonClassMap.cs
MikalaiMazurenka: Yes, exactly. The default target for netcore 3.0 is netstandard 2.1 (from the targets we build), where the `FormatterServices` type is already present.

PR #503, opened 2021-04-19
CSHARP-3380: Iterate over all connections on connection checkout 

BorisDog:
CSHARP-3380: Iterate over all connections on connection checkout in ConnectionPool

[EG](https://evergreen.mongodb.com/version/607dc1f2a4cf47318e9e3122) (internal link)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
+++ b/src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -177,6 +178,7 @@ namespace MongoDB.Driver.Core.ConnectionPools
             private readonly IConnection _connection;
             private readonly ExclusiveConnectionPool _connectionPool;
             private readonly int _generation;
+            private bool _disposed;
 
             public PooledConnection(ExclusiveConnectionPool connectionPool, IConnection connection)
             {
@@ -205,9 +207,14 @@ namespace MongoDB.Driver.Core.ConnectionPools
                 get { return _generation; }
             }
 
+            public bool IsDisposed
+            {
+                get { return _disposed; }
+            }
+
             public bool IsExpired
             {
-                get { return _generation < _connectionPool.Generation || _connection.IsExpired; }
+                get { return _disposed || _generation < _connectionPool.Generation || _connection.IsExpired; }
             }
 
             public ConnectionSettings Settings
@@ -217,7 +224,11 @@ namespace MongoDB.Driver.Core.ConnectionPools
 
             public void Dispose()
             {
-                _connection.Dispose();
+                if (!_disposed)
+                {
+                    _connection.Dispose();
+                    _disposed = true;
+                }
             }
 
             public void Open(CancellationToken cancellationToken)
@@ -468,43 +479,54 @@ namespace MongoDB.Driver.Core.ConnectionPools
 
             public void Prune()
             {
+                PooledConnection[] expiredConnections;
                 lock (_lock)
                 {
-                    for (int i = 0; i < _connections.Count; i++)
-                    {
-                        if (_connections[i].IsExpired)
-                        {
-                            RemoveConnection(_connections[i]);
-                            _connections.RemoveAt(i);
-                            break;
-                        }
-                    }
+                    expiredConnections = _connections.Where(c => c.IsExpired).ToArray();
+                }
 
-                    SignalOrReset();
+                foreach (var connection in expiredConnections)
+                {
+                    lock (_lock)
+                    {
+                        // At this point connection is always expired and might be disposed
+                        // If connection is already disposed the removal logic was already executed
+                        if (connection.IsDisposed)
+                        {
+                            continue;
+                        }
+
+                        RemoveConnection(connection);
+                        _connections.Remove(connection);
+                        SignalOrReset();
+                    }
                 }
             }
 
             public PooledConnection Acquire()
             {
+                PooledConnection result = null;
+
                 lock (_lock)
                 {
-                    if (_connections.Count > 0)
+                    while (_connections.Count > 0 && result == null)
                     {
-                        var connection = _connections[_connections.Count - 1];
-                        _connections.RemoveAt(_connections.Count - 1);
+                        var lastIndex = _connections.Count - 1;
+                        var connection = _connections[lastIndex];
+                        _connections.RemoveAt(lastIndex);
                         if (connection.IsExpired)
                         {
                             RemoveConnection(connection);
                         }
                         else
                         {
-                            return connection;
+                            result = connection;
                         }
                     }
 
                     SignalOrReset();
                 }
-                return null;
+                return result;
             }
 
             public void Return(PooledConnection connection)
--- a/tests/MongoDB.Bson.TestHelpers/XunitExtensions/RandomSeedAttribute.cs
+++ b/tests/MongoDB.Bson.TestHelpers/XunitExtensions/RandomSeedAttribute.cs
@@ -1,0 +1,32 @@
+﻿/* Copyright 2021-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Diagnostics;
+using MongoDB.Bson.TestHelpers.XunitExtensions;
+
+namespace MongoDB.Bson.TestHelpers
+{
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+    public class RandomSeedAttribute : Attribute, IValueGeneratorAttribute
+    {
+        public object[] GenerateValues() => new object[]
+        {
+            int.MaxValue,
+            int.MaxValue / 2,
+            Environment.TickCount ^ Process.GetCurrentProcess().Id
+        };
+    }
+}
--- a/tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
@@ -108,6 +108,93 @@ namespace MongoDB.Driver.Core.ConnectionPools
             act.ShouldThrow<ArgumentNullException>();
         }
 
+        [Fact]
+        public void AcquireConnection_should_iterate_over_all_dormant_connections()
+        {
+            const int connectionsCount = 10;
+
+            var settings = _settings.With(
+                minConnections: 0,
+                maxConnections: connectionsCount,
+                waitQueueTimeout: TimeSpan.FromMilliseconds(1),
+                maintenanceInterval: TimeSpan.FromMilliseconds(10000));
+
+            var connectionsCreated = new HashSet<ConnectionId>();
+            var connectionsExpired = new HashSet<ConnectionId>();
+            var connectionsDisposed = new HashSet<ConnectionId>();
+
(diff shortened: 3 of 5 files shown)


tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -510,128 +723,93 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
...
-            var actualTimeouts = 0;
-            var expectedTimeouts = maxAcquiringCount - maxConnecting;
+            _capturedEvents.WaitForOrThrowIfTimeout(events => events.Where(e => e is ConnectionCreatedEvent).Count() >= connectionsCount, TimeSpan.FromSeconds(10));
+            subject.DormantCount.Should().Be(connectionsCount);
 
-            ThreadingUtilities.ExecuteOnNewThreads(maxAcquiringCount + maxConnecting + 1, threadIndex =>
+            // expire some of the connections
+            _capturedEvents.Clear();
+            var random = new Random();
+            var connectionsToExpire = new List<ConnectionId>();
```
JamesKovacs, line 785: `connectionsToExpire` is unused.

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -510,128 +723,93 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
...
-            var allConnections = new List<IConnection>();
-            var actualTimeouts = 0;
-            var expectedTimeouts = maxAcquiringCount - maxConnecting;
+            _capturedEvents.WaitForOrThrowIfTimeout(events => events.Where(e => e is ConnectionCreatedEvent).Count() >= connectionsCount, TimeSpan.FromSeconds(10));
+            subject.DormantCount.Should().Be(connectionsCount);
 
-            ThreadingUtilities.ExecuteOnNewThreads(maxAcquiringCount + maxConnecting + 1, threadIndex =>
+            // expire some of the connections
+            _capturedEvents.Clear();
+            var random = new Random();
```
JamesKovacs, line 784: Using an unseeded PRNG can make tests difficult to debug. It would be better to succeed and fail in a set order. For example, always fail on odd-numbered connections or fail every third connection. If we need to test variations on that, we should write it as a parameterized `[Theory]`.

JamesKovacs (changes requested):
Overall looking good. The new connection pruning mechanism - where we prune, release, and prune some more - looks good.

There are some unused variables in tests, which need to be removed. The only major piece of feedback is the use of PRNG in one of the tests, which I don't like. The lack of determinism in tests can make them hard to debug in the face of failure. While that is true whenever there is threading in tests, there is no reason to make our lives even harder by introducing a PRNG. :)

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -510,128 +723,93 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
...
+            }
 
-            expectedTimeouts.Should().Be(expectedTimeouts);
-            subject.PendingCount.Should().Be(0);
+            // ensure removed events are received in subsequent order, meaning all expired connections where removed in same pass
+            _capturedEvents.WaitForOrThrowIfTimeout(events => events.Count() >= connectionsExpired.Count * 2, TimeSpan.FromSeconds(10));
+
+            _capturedEvents.Events
+                .Take(connectionsCount * 2)
+                .Where(e => e is ConnectionPoolRemovingConnectionEvent)
```
DmitryLukyanov, line 802: use `OfType` instead (and in other places)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -468,43 +476,51 @@ public void Clear()
...
             {
+                PooledConnection result = null;
+
                 lock (_lock)
                 {
-                    if (_connections.Count > 0)
+                    while (_connections.Count > 0 && result == null)
                     {
-                        var connection = _connections[_connections.Count - 1];
+                        var connection = _connections.Last();
```
DmitryLukyanov, line 509: nit: previous one works a bit faster due to avoiding this cast: https://github.com/microsoft/referencesource/blob/master/System.Core/System/Linq/Enumerable.cs#L1090

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -468,43 +476,51 @@ public void Clear()
...
                         _connections.RemoveAt(_connections.Count - 1);
                         if (connection.IsExpired)
                         {
                             RemoveConnection(connection);
                         }
                         else
                         {
-                            return connection;
+                            result = connection;
```
DmitryLukyanov, line 517: is this change really required? It sounds like the previous one also works

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -468,43 +476,51 @@ public void Clear()
 
             public void Prune()
             {
+                PooledConnection[] expiredConnections;
```
DmitryLukyanov, line 479: can you explain this change? I don't see significant difference yet

DmitryLukyanov (commented):
Can you please explain highlighted place?

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
BorisDog: Thanks, good idea!

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: done

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Discussed offline

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -177,6 +178,7 @@ private sealed class PooledConnection : IConnection
             private readonly IConnection _connection;
             private readonly ExclusiveConnectionPool _connectionPool;
             private readonly int _generation;
+            private bool _isDisposed;
```
DmitryLukyanov, line 181: in other places, we mostly name this variable `disposed` (without `is`).

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -392,6 +479,130 @@ public void Constructor_should_throw_when_eventSubscriber_is_null()
...
+            var allAcquiringCountEvent = new CountdownEvent(maxAcquiringCount + initalAcquiredCount);
+            var blockEstablishmentEvent = new ManualResetEventSlim(true);
+            var establishingCount = new CountdownEvent(maxConnecting + initalAcquiredCount);
+
+            var mockConnectionFactory = new Mock<IConnectionFactory>();
+            mockConnectionFactory
+                .Setup(c => c.CreateConnection(It.IsAny<ServerId>(), It.IsAny<EndPoint>()))
+                .Returns(() =>
+                {
+                    var connectionMock = new Mock<IConnection>();
```
DmitryLukyanov, line 508: note: we have `MockConnection` class that already mocks some logic

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
DmitryLukyanov: I'm not sure what exactly this random stuff can say for us? That there is a random value that causes issues? But what if not all of the random values are problematic? How can we work with the failed tests which we don't know why exactly they were failed? I agree with @JamesKovacs , we need to be more explicit in this case

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -205,6 +207,11 @@ public int Generation
                 get { return _generation; }
             }
 
+            public bool IsDisposed
```
DmitryLukyanov, line 210: I would also add a note here about the reason why it's not enough to have only `IsExpired`

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
JamesKovacs: As discussed over Zoom, using `Random` with an unknown seed value can make test failures difficult to reproduce. I suggest creating a `Randomizer` test class that can generate and output a seed value for the test run so that failures can be more easily investigated. I would also suggest considering something akin to the following:

```
[Theory]
[InlineData(null)]
[InlineData(12345)]
[InlineData(67890)]
public void TestRandomly(int? seed)
{
    var random = seed.HasValue ? new Random(seed.Value) : new Random();
    // remainder of test code
}
```

This would allow for two fixed random seeds - with known test outputs - as well as an actual random seed to fuzz the test.

If we see a failure in one of these tests, this would allow us to quickly spot the root cause. Imagine if we just used truly random seeds and a particular test passed on Windows and Linux, but failed on OSX. Is it a problem with OSX? Was it the random seed value used? We have no idea. Take the same scenario but using the above technique. The fixed seeds pass on all platforms but a random seed fails on OSX. That is a pretty good indication that it is due to the random seed selected rather than a problem on OSX.

JamesKovacs (changes requested):
Discussed the need for PRNG seeding and options for fuzzing over Zoom.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: >I think IsDisposed and IsExpired are quite self-explanatory.

my point was that looking just at `PooledConnection` class, it's not clear how `IsDisposed` field related to `IsExpired`. Should we also consider `IsDisposed` field in places when we currently use only `IsExpired`. It would be good if we say that the only reason why we have this field is to avoid some double removing in particular place

DmitryLukyanov (approved):
LGTM

tests/MongoDB.Bson.TestHelpers/XunitExtensions/RandomSeedAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Diagnostics;
+using MongoDB.Bson.TestHelpers.XunitExtensions;
+
+namespace MongoDB.Bson.TestHelpers
+{
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+    public class RandomSeedAttribute : Attribute, IValueGeneratorAttribute
+    {
+        public object[] GenerateValues() => new object[] { Environment.TickCount ^ Process.GetCurrentProcess().Id };
```
JamesKovacs, line 25: Although we are generating a random seed, we will know the value since it will get reported in the xUnit output. Is that correct?

If so, I like the idea for its simplicity and debuggability. I would still like to see two fixed seed values so that tests are reproducible. I would suggest:

```
public object[] GenerateValues() => new object[] {
        int.MaxValue / 2,
        int.MaxValue,
        Environment.TickCount ^ Process.GetCurrentProcess().Id
};
```

I tried `int.MinValue` and zero but those generate the same sequence of pseudo-randoms as `int.MaxValue`. Another possibility would be to use 0 and 42 as the fixed random seeds.

JamesKovacs (changes requested):
Looks good but I still think we need two fixed random seeds. See my inline comment. Happy to discuss more if you'd like.

JamesKovacs (approved):
LGTM. Thank you for the discussion and incorporating feedback.

PR #508, opened 2021-04-26
CSHARP-2955: Check MethodHelper GetMethodDefinition method when targetting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2955
EG: https://evergreen.mongodb.com/version/608728e63066157f0b79adb5

### DIFF
--- a/src/MongoDB.Driver/Linq/MethodHelper.cs
+++ b/src/MongoDB.Driver/Linq/MethodHelper.cs
@@ -89,12 +89,11 @@ namespace MongoDB.Driver.Linq
             }
 
             var declaringTypeDefinition = methodInfo.DeclaringType.GetGenericTypeDefinition();
-
 #if NETSTANDARD1_5
             var bindingFlags = BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public;
             var parameterTypes = methodInfo.GetParameters().Select(p => p.ParameterType).ToArray();
             return declaringTypeDefinition.GetTypeInfo().GetMethods(bindingFlags)
-                .Where(m => m.MetadataToken == methodInfo.MetadataToken)
+                .Where(m => m.Name == methodInfo.Name && m.GetParameters().Count() == parameterTypes.Length) // TODO: need better matching
                 .Single();
 #else
             return (MethodInfo)MethodBase.GetMethodFromHandle(methodInfo.MethodHandle, declaringTypeDefinition.TypeHandle);
--- a/src/MongoDB.Driver/Linq/MethodHelper.cs
+++ b/src/MongoDB.Driver/Linq/MethodHelper.cs
@@ -89,11 +89,12 @@ namespace MongoDB.Driver.Linq
             }
 
             var declaringTypeDefinition = methodInfo.DeclaringType.GetGenericTypeDefinition();
+
 #if NETSTANDARD1_5
             var bindingFlags = BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public;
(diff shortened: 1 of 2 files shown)


src/MongoDB.Driver/Linq/MethodHelper.cs
```diff
@@ -89,11 +89,12 @@ public static MethodInfo GetMethodDefinition(MethodInfo methodInfo)
             }
 
             var declaringTypeDefinition = methodInfo.DeclaringType.GetGenericTypeDefinition();
+
 #if NETSTANDARD1_5
             var bindingFlags = BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public;
             var parameterTypes = methodInfo.GetParameters().Select(p => p.ParameterType).ToArray();
             return declaringTypeDefinition.GetTypeInfo().GetMethods(bindingFlags)
-                .Where(m => m.Name == methodInfo.Name && m.GetParameters().Count() == parameterTypes.Length) // TODO: need better matching
+                .Where(m => m.MetadataToken == methodInfo.MetadataToken)
```
MikalaiMazurenka, line 97: This change is tricky, but hear me out:
[MemberInfo.MetadataToken](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.metadatatoken?view=netstandard-1.5) is unique per method per module (effectively per compiled assembly). Here we are searching for a method of a generic type with the same signature, but with generic parameter not specified. From the metadata perspective, both of these methods are the same and we are looking in the very same type, so in my opinion the usage of `MetadataToken` is fine in this case.
Though if you still think the usage of `MetadataToken` is not safe, I'm perfectly fine with reverting this back and just leaving the tests.

rstam: Before I study whether this change is correct can you explain why you even made it?

The point of the ticket was to verify whether the NETSTANDARD2_0 code path was correct.

rstam (changes requested):

src/MongoDB.Driver/Linq/MethodHelper.cs
MikalaiMazurenka: In this case I misinterpreted the task description. Since there was nothing to change for netstandard2.0 I updated the netstandard1.5 path, since I saw a TODO there

rstam: The task was to see whether the NETSTANDARD2_0 (and now NETSTANDARD2_1) path need to be changed to use new APIs available in 2.0 and 2.1 (and presumably common to .NET 4.5.2).

I would leave the NETSTANDARD1_5 code path alone.

You can argue if you want that the NETSTANDARD1_5 code path should be changed, but since we are so close to removing support for NETSTANDARD1_5 anyway I wouldn't bother.

If we do bother, we probably need a different JIRA ticket and a justification.

rstam (changes requested):

src/MongoDB.Driver/Linq/MethodHelper.cs
MikalaiMazurenka: Agreed. Dropped this change.

MikalaiMazurenka (commented):
I've reverted the netstandard 1.5 changes, but left this case documented in unit tests.
If tests are not necessary here, I will cancel this PR and close the issue as already implemented.

rstam (approved):
LGTM

I leave it up to you whether you want to keep the tests or not.

This is LINQ2 code and will go away when LINQ2 goes away.

MikalaiMazurenka:
Closing, because the tests are not necessary, considering they will go away with the new LINQ implementation.

PR #509, opened 2021-04-27
CSHARP-2956: Check IncrementalMDNet45 vs IncrementalMDNetStandard16 when targetting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2956
EG: https://evergreen.mongodb.com/version/608817ef3627e06f01825208

### DIFF
--- a/src/MongoDB.Shared/IncrementalMD5.cs
+++ b/src/MongoDB.Shared/IncrementalMD5.cs
@@ -25,7 +25,7 @@ namespace MongoDB.Shared
 #if NET452
             return new IncrementalMD5Net45();
 #else
-            return new IncrementalMD5NetStandard16();
+            return new IncrementalMD5NetStandard();
 #endif
         }
 
@@ -35,6 +35,7 @@ namespace MongoDB.Shared
     }
 
 #if NET452
+    // Required for .NET 4.5.2 only.
     internal class IncrementalMD5Net45 : IncrementalMD5
     {
         private static readonly byte[] __emptyByteArray = new byte[0];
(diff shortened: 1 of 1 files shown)


src/MongoDB.Shared/IncrementalMD5.cs
```diff
@@ -25,7 +25,7 @@ public static IncrementalMD5 Create()
 #if NET452
             return new IncrementalMD5Net45();
 #else
-            return new IncrementalMD5NetStandard16();
+            return new IncrementalMD5NetStandard();
```
rstam, line 28: So ALL versions of .NET standard have to use the workaround? Even 2.0 and 2.1?

rstam: I was expecting to see this:

    #if !NETSTANDARD1_5
                return new IncrementalMD5Net45();
    #else
                return new IncrementalMD5NetStandard16();
    #endif

MikalaiMazurenka: That depends on what to perceive as a workaround.
MD5 algorithm is generally considered outdated (current .NET 4.5.2 path: [MD5 -> Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.md5?view=netframework-4.5.2#remarks) and current .NET Standard path: [HashAlgorithmName.MD5 -> Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hashalgorithmname.md5?view=net-5.0#remarks)).
I didn't check it thoroughly, but my understanding is that GridFS requires MD5, so we can't change that, thus I would go with the most comfortable path. From my point of view it's the one which uses the least custom code, meaning [IncrementalHash](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.incrementalhash?view=net-5.0), which provides the necessary functionality out of the box. Additional point to `IncrementalHash` is that it allows to switch the algorithm simply by specifying a different algorithm name in the `CreateHash` method.
But we can use the code you included in the comment without any drawback, so I'm fine with either path.

rstam: Thanks for clarifying. 

I had assumed that the netstandard1.5 code path was the inferior one (because that's almost always the case given the limitations of netstandard1.5).

It does indeed look like `IncrementalHash` is the better version of MD5 and that in this case it is (oddly enough) the net452 path that has the workaround.

rstam (approved):
LGTM

PR #513, opened 2021-04-28
CSHARP-2945: Check uses of Marshal when targetting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2945
EG: https://evergreen.mongodb.com/version/6089b241c9ec443023b2395c

Note: [Marshal.SizeOf\<T\>()](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.sizeof?view=netframework-4.5.2#System_Runtime_InteropServices_Marshal_SizeOf__1) and [Marshal.PtrToStructure\<T\>()](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.ptrtostructure?view=netframework-4.5.2#System_Runtime_InteropServices_Marshal_PtrToStructure__1_System_IntPtr_) are available for .NET 4.5.2.
The reason why these directives were still present is likely because they were not updated when target changed from .NET 4.5 to .NET 4.5.2 along with introducing multitargeting (and these overloads are not available for .NET 4.5).

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Authentication/Sspi/SecurityBufferDescriptor.cs
+++ b/src/MongoDB.Driver.Core/Core/Authentication/Sspi/SecurityBufferDescriptor.cs
@@ -64,34 +64,18 @@ namespace MongoDB.Driver.Core.Authentication.Sspi
             NumBuffers = buffers.Length;
 
             //Allocate memory for SecBuffer Array....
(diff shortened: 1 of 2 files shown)


rstam (approved):
LGTM

PR #514, opened 2021-04-29
CSHARP-3626: Fix NuGet install fail

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3626
EG: https://evergreen.mongodb.com/version/608ae7623627e066eccb6a71

Note: maybe it makes sense to rename the commit to something like "Pin NuGet client version" so that later it could be gracefully reverted with commit "Unpin NuGet client version" once the build ticket is done?


DmitryLukyanov (approved):
LGTM

build.ps1
```diff
@@ -120,7 +120,7 @@ $ADDINS_DIR = Join-Path $TOOLS_DIR "Addins"
 $MODULES_DIR = Join-Path $TOOLS_DIR "Modules"
 $NUGET_EXE = Join-Path $TOOLS_DIR "nuget.exe"
 $CAKE_EXE = Join-Path $TOOLS_DIR "Cake/Cake.exe"
-$NUGET_URL = "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+$NUGET_URL = "https://dist.nuget.org/win-x86-commandline/v5.7.1/nuget.exe"
```
DmitryLukyanov, line 123: note: this is taken from here: https://github.com/cake-build/example/blob/master/build.ps1#L118 So probably it's better to revert this change when the issue will be resolved

JamesKovacs (approved):
I don’t like editing files provided by third-party dependencies like Cake, but I don’t think we have any other choice here. LGTM.

PR #517, opened 2021-05-04
CSHARP-2946: Check uses of DNS methods when targetting netstandard2.0

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-2946
EG: https://evergreen.mongodb.com/version/60917b575623430b33b5630f

Note: this PR just adds a sync version of `Dns.GetHostAddresses`, because it is available since netstandard 2.0. The swap in target paths (`#if NET452`) is made to avoid negative conditions (just as a general rule, nothing specific).


rstam (approved):
This change is really interesting. Nice work.

LGTM

PR #519, opened 2021-05-07
CSHARP-3306: Implement and Test Connection Pool Paused State

BorisDog:


src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -147,67 +144,54 @@ public int UsedCount
...
+        public async Task<IConnectionHandle> AcquireConnectionAsync(CancellationToken cancellationToken)
+        {
+            using var helper = new AcquireConnectionHelper(this);
+            return await helper.AcquireConnectionAsync(cancellationToken).ConfigureAwait(false);
         }
 
         public void Clear()
         {
-            ThrowIfNotOpen();
+            lock (_poolState)
```
DmitryLukyanov, line 179: is this part implemented (`lock on the TopologyDescription when clearing the pool`): https://github.com/mongodb/specifications/commit/beda39620fb39938a8d46eeaf63b75f807fe7c4e#diff-52a48d3b84e9e90e30fb87b3799c6302750f1ed2817b2abbead44e7a7bb6bb47R1214?

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -246,6 +249,10 @@ private void OnDescriptionChanged(object sender, ServerDescriptionChangedEventAr
             {
                 _connectionPool.Clear();
             }
+            else if (e.NewServerDescription.State == ServerState.Connected)
+            {
+                _connectionPool.SetReady();
```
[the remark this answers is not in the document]
DmitryLukyanov: and this one: https://github.com/mongodb/specifications/commit/beda39620fb39938a8d46eeaf63b75f807fe7c4e#diff-52a48d3b84e9e90e30fb87b3799c6302750f1ed2817b2abbead44e7a7bb6bb47R1394?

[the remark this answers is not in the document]
DmitryLukyanov: is this also needs be added: https://github.com/mongodb/specifications/commit/beda39620fb39938a8d46eeaf63b75f807fe7c4e#diff-52a48d3b84e9e90e30fb87b3799c6302750f1ed2817b2abbead44e7a7bb6bb47R1626?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -144,20 +244,70 @@ public void Finally()
...
             {
-                if (_enteredPool)
+                _pool._checkingOutConnectionEventHandler?
+                    .Invoke(new ConnectionPoolCheckingOutConnectionEvent(_pool._serverId, EventContext.OperationId));
+
+                _pool._poolState.ThrowIfDisposedOrNotReady();
+
+                // enter the wait-queue, deprecated feature
+                int freeSlots;
+                do
```
DmitryLukyanov, line 257: can you please point me out on the doc with description for this logic?

DmitryLukyanov (commented):
Initial review

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -29,6 +29,38 @@ public enum SemaphoreWaitResult
             Entered
         }
 
+        public sealed class SemaphoreSignalableAwaiter : IDisposable
```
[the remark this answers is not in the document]
DmitryLukyanov: I looked again here and now I see that this design is useful, but I would still want to see a different naming here

src/MongoDB.Driver.Core/Core/Servers/Server.cs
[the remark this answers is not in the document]
DmitryLukyanov: 4. related to the  #2, the ticket https://jira.mongodb.org/browse/DRIVERS-1773. But I  confirmed with Patric that expected way is to use IsDataBearing conditon here, so I would make a sync with the spec wording now. Also, adding condition on `e.OldServerDescription.Type == ServerType.Unknown` solves one of my other comments regarding the fact that we always set ready even if it had been set before. With this change we will set `ready` only in required situations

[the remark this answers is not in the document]
JamesKovacs: We need `IsDataBearing` in this condition. `ServerType.Arbiter` is never going to be data-bearing and will never be selected. As the condition stands, we would setup a connection pool for arbiters, which is incorrect.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,218 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
+
+                    if (_poolQueueWaitResult == SemaphoreSlimSignalable.SemaphoreWaitResult.Entered)
                     {
-                        connection = await connectionCreator.CreateOpenedOrReuseAsync(cancellationToken).ConfigureAwait(false);
+                        PooledConnection pooledConnection = null;
+                        var timeout = EnsureTimeout(stopwatch);
+
+                        using (var connectionCreator = new ConnectionCreator(_pool, timeout))
+                        {
+                            pooledConnection = connectionCreator.CreateOpenedOrReuse(cancellationToken);
```
DmitryLukyanov, line 187: why `pooledConnection` cannot be null anymore?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -144,20 +252,70 @@ public void Finally()
...
+            }
+
+            private Exception CreateException(Stopwatch stopwatch) =>
+                _poolQueueWaitResult switch
+                {
+                    SemaphoreSlimSignalable.SemaphoreWaitResult.Signaled =>
+                        MongoConnectionPoolPausedException.ForConnectionPool(_pool._endPoint),
+                    SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut =>
+                        new TimeoutException($"Timed out waiting for a connection after {stopwatch.ElapsedMilliseconds}ms."),
+                    _ => new ArgumentOutOfRangeException(nameof(_poolQueueWaitResult))
```
DmitryLukyanov, line 314: we throw `ArgumentOutOfRangeException` only in validating ctor arguments, it should be `InvalidOperationException`

DmitryLukyanov: also, is any other state expected here?

src/MongoDB.Driver.Core/Core/Servers/Server.cs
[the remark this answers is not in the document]
DmitryLukyanov: as it was discussed in the slack with Shane, it's better to not use this condition: `e.OldServerDescription.Type == ServerType.Unknown` just in case something wrong goes down in SDAM handling, with this condition we may never mark pool ready

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -29,6 +29,7 @@ public class ConnectionPoolSettings
         private readonly int _minConnections;
         private readonly int _waitQueueSize;
         private readonly TimeSpan _waitQueueTimeout;
+        private readonly int _maxConnecting;
```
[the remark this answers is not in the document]
DmitryLukyanov: as we discussed in slack, we usually don't place any internal field and constants into such options class, but I also was wrong about the above guide requirement, so I will rely on you and @JamesKovacs  here

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -338,32 +345,35 @@ private void HandleChannelException(IConnection connection, Exception ex)
...
             {
-                _connectionPool.Clear();
+                // non connection exception
                 return;
             }
 
-            if (ex is MongoConnectionException mongoConnectionException)
+            var (invalidateAndClear, cancelCheck) = ex switch
+            {
+                MongoAuthenticationException => (invalidateAndClear: true, cancelCheck: false),
```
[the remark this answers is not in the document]
DmitryLukyanov: thanks

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -144,20 +255,75 @@ public void Finally()
...
+            }
+
+            private Exception CreateException(Stopwatch stopwatch) =>
+                _poolQueueWaitResult switch
+                {
+                    SemaphoreSlimSignalable.SemaphoreWaitResult.Signaled =>
+                        MongoConnectionPoolPausedException.ForConnectionPool(_pool._endPoint),
+                    SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut =>
+                        new TimeoutException($"Timed out waiting for a connection after {stopwatch.ElapsedMilliseconds}ms."),
+                    _ => new InvalidOperationException($"Invalid {_poolQueueWaitResult}")
```
DmitryLukyanov, line 322: dot in the end of message

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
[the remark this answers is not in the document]
DmitryLukyanov: I will rely on you and @JamesKovacs  here

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,210 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
-                var handler = _pool._checkingOutConnectionEventHandler;
-                if (handler != null)
+                _state = State.Initial;
+            }
+
+            public State State => _state;
+            public bool IsDisposed => _state == State.Disposed;
+            public bool IsNotDisposed => _state != State.Disposed;
+
+            public bool TransitionState(State newState)
```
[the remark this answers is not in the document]
DmitryLukyanov: got it.
note: I agree that `TryX` method should not throw, but at least this follows how it's implemented currently: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Misc/InterlockedInt32.cs#L49.

also, the fact that we cannot imagine a correct name here rings a bit that we makes more steps than it could be inside this method, but I don't want to make it over-complicated, so will rely on you and @JamesKovacs

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,221 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
+
+                __transitions[(int)State.Uninitialized, (int)State.Disposed] = true;
+                __transitions[(int)State.Paused, (int)State.Disposed] = true;
+                __transitions[(int)State.Ready, (int)State.Disposed] = true;
+                __transitions[(int)State.Disposed, (int)State.Disposed] = true;
             }
 
-            // public methods
-            public void CheckingOutConnection()
+            public PoolState(string poolIdentifier)
```
DmitryLukyanov, line 65: it's always Endpoint, I think no reason to use string here especially if it requires changes in `MongoConnectionPoolPausedException`

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: I don't agree with this. There is no need to introduce any logic that it's not used. Unless we will have a real case when the identification is not endpoint, I don't think we need to make such change now.

I think a typed safe way is always better than `string`ed.

>Also we let pool decide how to identify itself

currently it will be always endpoint

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: nit: we usually add a comment to the switch branches that are note expected to be used (unless a driver error), like: `should not be reached`, this will show that this case/if branch is added just in case

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
DmitryLukyanov: I think the main point here is that it allows two actions, returning `true`/`false` as the method result (`Try` part) and throwing exception if some internal condition is wrong

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: We are not introducing new logic. We are preserving the old functionality and simplifying existing one by encapsulating the decision on what pool id is to pool itself. Additional side effect is that endpoint is serialized only once.

> currently it will be always endpoint
Not necessarily, we might want to add additional ID decoration later, for example for debugging purposes.

Note that this approach will also be useful/needed when we introduce logging.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: There is not really `Try` part in this method. Returning `true/false` is not well described by `Try`, as it indicates whether the transition was self-transition or not.
Added a comment explaining that, maybe in addition we can think of some naming that indicates the self-transition nuance.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Thanks, done.

tests/MongoDB.Driver.Tests/Specifications/Runner/MongoClientJsonDrivenTestRunnerBase.cs
```diff
@@ -321,17 +327,39 @@ protected virtual bool TryConfigureClientOption(MongoClientSettings settings, Bs
...
+                    settings.ConnectionMode = ConnectionMode.Automatic;
+                    settings._connectionModeSwitch( ConnectionModeSwitch.UseDirectConnection);
+#pragma warning restore CS0618 // Type or member is obsolete
+                    settings.DirectConnection = isDirectConnection;
+
+                    if (isDirectConnection)
+                    {
+                        settings.Servers = new MongoServerAddress[]
+                        {
+                            settings.Servers.First()
```
DmitryLukyanov, line 350: is this mentioned somewhere in the spec?

src/MongoDB.Driver.Core/Core/Bindings/SingleServerReadBinding.cs
```diff
@@ -86,6 +86,15 @@ public void Dispose()
 
         private IChannelSourceHandle GetChannelSourceHelper()
         {
+            // server might be in unknown state due to previous failed operation, allow description to be updated
+            // this is done instead of server selection
+            // TODO parameterize timeout and avoid busy wait, or offload waiting to server level
+            // Should be addressed by CSHARP-3556
+            if (_server.Description.State == ServerState.Disconnected)
```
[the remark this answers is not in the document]
DmitryLukyanov: is this condition about cases that can happen when we select server in `Unknown` state [here](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.GridFS/GridFSBucket.cs#L971)? If so, then it looks like the question for server selector..

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -152,6 +152,11 @@ public TCommandResult Execute(IConnection connection, CancellationToken cancella
                 TransactionHelper.UnpinServerIfNeededOnCommandException(_session, exception);
                 throw;
             }
+            finally
+            {
+                // temporary workaround until https://jira.mongodb.org/browse/DRIVERS-1655 is resolved
+                TransactionHelper.UnpinServerIfNeededOnAbortTransaction(_session, _command);
```
DmitryLukyanov, line 158: looking at this line in the driver ticket: `Server responds with ShutdownInProgress without TransientTransactionError labels`

should this happen only in catch block since we will appear there if `ShutdownInProgress` is triggered?

DmitryLukyanov: also, maybe it's better just to update this condition: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs#L558?

DmitryLukyanov: or to place this on transaction level similar to what we have for commit: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs#L277

tests/MongoDB.Driver.Core.TestHelpers/FailPoint.cs
```diff
@@ -187,11 +189,17 @@ private void ConfigureOff()
...
                 { "mode", "off" }
             };
-            Configure(command);
+            Configure(command, true);
         }
 
-        private void ExecuteCommand(BsonDocument command)
+        private void ExecuteCommand(BsonDocument command, bool waitForConnected)
         {
+            if (waitForConnected)
```
DmitryLukyanov, line 197: this looks like a hack.. even if it's valid change, 1 second might be not enough in some cases, and if so, it will be really hard to find this place, can you explain this change before we proceed?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,210 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
+            }
+
+            public void Dispose()
+            {
+                _cancellationTokenSource?.Cancel();
+                _cancellationTokenSource?.Dispose();
+            }
+        }
+
+        private sealed class AcquireConnectionHelper : IDisposable
```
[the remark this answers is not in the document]
DmitryLukyanov: regarding our offline discussion about cases that could not be covered by json tests, I would not want to check all json descriptions to see what is not covered. After quick look I think this logic is not covered: https://github.com/mongodb/mongo-csharp-driver/blob/3f246f1ba10d9745e4f5e1e7db31918616b55c6d/src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs#L302
I will leave this up to you whether you want to add this (or similar) test or no

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
DmitryLukyanov: ignore all the above, this case is fixed in one of my tickets. See this PR: https://github.com/mongodb/mongo-csharp-driver/pull/540

tests/MongoDB.Driver.Core.Tests/Specifications/connection-monitoring-and-pooling/ConnectionMonitoringAndPoolingTestRunner.cs
```diff
@@ -614,26 +634,88 @@ private void ResetConnectionId()
...
+                        eventCapturer.WaitForOrThrowIfTimeout(events => events.Any(e => e is ConnectionCreatedEvent), TimeSpan.FromMilliseconds(500));
+
+                        var connectionIdsToIgnore = new HashSet<int>(eventCapturer.Events
+                            .OfType<ConnectionCreatedEvent>()
+                            .Select(c => c.ConnectionId.LocalValue)
+                            .ToList());
+
+                        eventsFilter = o =>
+                        {
+                            if (o is ConnectionOpenedEvent ||
```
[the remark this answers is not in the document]
BorisDog: Nice!

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
BorisDog: Great! I'll revert if #540 completes beforehand.

src/MongoDB.Driver.Core/MongoConnectionPoolPausedException.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+        }
+
+#if !NETSTANDARD1_5
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MongoWaitQueueFullException"/> class.
+        /// </summary>
+        /// <param name="info">The SerializationInfo.</param>
+        /// <param name="context">The StreamingContext.</param>
+        protected MongoConnectionPoolPausedException(SerializationInfo info, StreamingContext context)
```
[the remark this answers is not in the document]
DmitryLukyanov: it looks like we have both ways, then I'm ok with any way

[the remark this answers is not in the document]
DmitryLukyanov: >also, as I understand, this ctor works in pair with GetObjectData method

this is missed

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenConfigureFailPointTest.cs
```diff
@@ -46,12 +46,22 @@ protected override async Task CallMethodAsync(CancellationToken cancellationToke
 
         protected virtual IServer GetFailPointServer()
         {
+            if (TestRunner.FailPointServer != null)
```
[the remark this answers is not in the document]
DmitryLukyanov: missed

tests/MongoDB.Driver.Tests/Specifications/Runner/MongoClientJsonDrivenTestRunnerBase.cs
DmitryLukyanov: missed

BorisDog: We can't have multiple servers with direct connection.

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenConfigureFailPointTest.cs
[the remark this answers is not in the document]
BorisDog: as `MongoClientJsonDrivenTestRunnerBase` configures failpoint, and later `configureFailPoint` operation shuts it off before test end, it has to be same server.
For example `minPoolSize-error.json`

src/MongoDB.Driver.Core/MongoConnectionPoolPausedException.cs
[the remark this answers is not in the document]
BorisDog: We don't have any special data to add in `GetObjectData`.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -338,32 +345,35 @@ private void HandleChannelException(IConnection connection, Exception ex)
 
         private void HandleBeforeHandshakeCompletesException(Exception ex)
         {
-            if (ex is MongoAuthenticationException)
+            if ((ex is not MongoConnectionException connectionException))
```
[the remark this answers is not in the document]
BorisDog: Tnx, done.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,221 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
+                try
                 {
-                    var timeSpentInWaitQueue = _stopwatch.Elapsed;
-                    using (var connectionCreator = new ConnectionCreator(_pool, _pool._settings.WaitQueueTimeout - timeSpentInWaitQueue))
+                    StartCheckingOut();
+
+                    var stopwatch = Stopwatch.StartNew();
+                    _poolQueueWaitResult = _pool._maxConnectionsQueue.WaitSignaled(_timeout, cancellationToken);
+
+                    if (_poolQueueWaitResult == SemaphoreSlimSignalable.SemaphoreWaitResult.Entered)
```
JamesKovacs, line 183: I'm pretty sure we have a race condition here. Imagine a thread enters the wait queue and nears its timeout. A second thread enters the wait queue around this time and gets signalled as the first thread times out. The first thread sets `_poolQueueWaitResult` to timed out. The second thread sets it to entered. When we hit the `if` block, they both succeed and checkout connections. Or if the ordering is reversed, both appear to time out even though the second thread just entered the wait queue and is nowhere near the timeout.

I suspect we should be storing `_poolQueueWaitResult` in a local variable rather than a member variable. Or is there some thread synchronization that I'm missing that avoids the above scenario?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -147,67 +153,43 @@ public int UsedCount
...
+            {
+                _poolState.ThrowIfNotInitialized();
 
-            _clearingEventHandler?.Invoke(new ConnectionPoolClearingEvent(_serverId, _settings));
+                if (_poolState.TransitionState(State.Paused))
```
JamesKovacs, line 180: Are you transitioning the pool to State.Paused? Are you checking if the pool is in State.Paused? I'm not understanding the intent.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -220,105 +202,109 @@ private PooledConnection CreateNewConnection()
...
         {
-            bool enteredPool = false;
-            try
+            using (var poolAwaiter = _maxConnectionsQueue.CreateAwaiter())
             {
-                // if it takes too long to enter the pool, then the pool is fully utilized
-                // and we don't want to mess with it.
-                enteredPool = await _poolQueue.WaitAsync(TimeSpan.FromMilliseconds(20), cancellationToken).ConfigureAwait(false);
-                if (!enteredPool)
+                var entered = await poolAwaiter.WaitSignaledAsync(TimeSpan.FromMilliseconds(20), cancellationToken).ConfigureAwait(false);
```
JamesKovacs, line 288: Why 20ms?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPoolFactory.cs
```diff
@@ -36,12 +36,12 @@ public ExclusiveConnectionPoolFactory(ConnectionPoolSettings settings, IConnecti
             _eventSubscriber = Ensure.IsNotNull(eventSubscriber, nameof(eventSubscriber));
         }
 
-        public IConnectionPool CreateConnectionPool(ServerId serverId, EndPoint endPoint)
+        public IConnectionPool CreateConnectionPool(ServerId serverId, EndPoint endPoint, IConnectionExceptionHandler connectionExceptionHandler)
```
JamesKovacs, line 39: Are we OK with adding parameters to public methods in Core?

src/MongoDB.Driver.Core/Core/ConnectionPools/IConnectionPool.cs
```diff
@@ -63,10 +59,15 @@ public interface IConnectionPool : IDisposable
...
         /// <summary>
-        /// Clears the connection pool.
+        /// Clears the connection pool and marks it as paused.
         /// </summary>
         void Clear();
 
+        /// <summary>
+        /// Marks the pool as ready.
+        /// </summary>
+        void SetReady();
```
JamesKovacs, line 69: What about adding a method to a public interface in Core? Is `IConnectionPool` exposed anywhere in our public API?

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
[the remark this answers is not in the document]
JamesKovacs: I'm OK with the `Awaiter` naming. Consider `SemaphoreSlimSignalableAwaiter`.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -135,6 +162,9 @@ public async Task<SemaphoreWaitResult> WaitSignaledAsync(TimeSpan timeout, Cance
             }
         }
 
+        public SemaphoreSignalableAwaiter CreateAwaiter() =>
```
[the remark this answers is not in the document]
JamesKovacs: `CreateAwaiter()` seems like the correct naming to me.

JamesKovacs (changes requested):
I haven't thoroughly vetted the tests as I want to discuss the other points raised first. I hope that you can prove me wrong about the apparent race condition noted below.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: Self-transition is transition from a state to itself, for example from Disposed to Disposed. It's allowed for some states, and not for others. It matters as `TransitionState` returns whether the transition was not self-transition.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,221 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
-                if (enteredPool)
+                var state = _state;
+                if (state == State.Paused)
                 {
-                    var timeSpentInWaitQueue = _stopwatch.Elapsed;
-                    using (var connectionCreator = new ConnectionCreator(_pool, _pool._settings.WaitQueueTimeout - timeSpentInWaitQueue))
-                    {
-                        connection = connectionCreator.CreateOpenedOrReuse(cancellationToken);
-                    }
+                    throw MongoConnectionPoolPausedException.ForConnectionPool(_poolIdentifier);
```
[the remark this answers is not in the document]
BorisDog: Not sure, that's how it was always used. I am happy to refactor if everybody is onboard.
cc @DmitryLukyanov

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
BorisDog: I guess that's this old heuristic for "almost immediate" allows to account for some unexpected execution delays. Did not find a reason to change that.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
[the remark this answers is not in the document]
BorisDog: Done.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Yes, `Pool` identifies itself via `Endpoint`. But there is no good reason for `PoolState` or other helpers to make the `Pool` identification decision on-behalf of the pool.
As `PoolState` has on other use for `Endpoint `beyond serializing it as pool identification, if would be cleaner just to receive the minimal information needed.
Also note that it's a bit cleaner for the PoolStateTests to create `PoolState` objects :)

src/MongoDB.Driver.Core/Core/ConnectionPools/IConnectionPool.cs
DmitryLukyanov: I think this is fine here (I don't see any way when someone can directly use this interface). I would also mention this in this ticket: https://jira.mongodb.org/browse/CSHARP-3679

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPoolFactory.cs
DmitryLukyanov: I think this is fine here (I don't see any way when someone can directly use this interface). I would also mention this in this ticket: https://jira.mongodb.org/browse/CSHARP-3679 (I think I also changed this interface in LB changes)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: this is old value, let's not change it at least in the current PR :)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
DmitryLukyanov: I'm ok with a factory method (but I also agreed with James that using just a ctor will provide exactly the same logic). But as I mentioned somewhere, I think we should not provide a string value, if we provide endpoint, it should a type safe value.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: I think if we will need to add a different identificator, we will need to add a different factory method or ctor, but I think it won't be needed :)

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
DmitryLukyanov: it's already pushed

tests/MongoDB.Driver.Tests/Specifications/Runner/MongoClientJsonDrivenTestRunnerBase.cs
DmitryLukyanov: I just surprised that this logic is not mentioned anywhere, I would expect so

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -32,110 +32,221 @@ namespace MongoDB.Driver.Core.ConnectionPools
...
-            public async Task<IConnectionHandle> EnteredPoolAsync(bool enteredPool, CancellationToken cancellationToken)
+            public void Start()
             {
-                _enteredPool = enteredPool;
-                PooledConnection connection = null;
+                _cancellationTokenSource?.Cancel();
+                _cancellationTokenSource = new CancellationTokenSource();
+
+                _maintenanceTask = _maintenanceTaskCreator(_cancellationTokenSource.Token);
+                _maintenanceTask.ConfigureAwait(false);
```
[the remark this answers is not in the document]
DmitryLukyanov: missed

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: missed

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: missed

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: I look at this quite simply: Does `PoolState` or any depended entity actually need the `Endpoint` object, are they expected to use it in any way in the future? So from design perspective any entity should receive no more information that it actually needs.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: In this case receiving string is type safe. The exception just needs a string and that's exactly what it gets. Not-type safe situations are when the object might not be of the expected type for any reason. It's not the case here.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: We just create and hold a reference to this task without inspecting the result. 
BTW we should aim to refactor this pattern for background tasks (that's used in other places) to regular threads.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Good question: `AcquireConnectionHelper` object is created per request, so effectively each member is isolated to the executing thread.

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
BorisDog: Great, will revert this.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
BorisDog: Transitioning the pool to Paused, and checking whether it was transitioned to Paused (it can be paused already).
Similar (but not identical) to the pattern we use in different places `if (_state.TryChange(State.Disposed)) {...}`

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
DmitryLukyanov: it's disputable, you can pass something like `abc` and it will be also "valid" and but not safe

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: if it doesn't need than we should not pass it, if we want to show this information anywhere inside (so we need this), then no reason to not help to class consumer about what this "identificator" is about, if it will be not only endpoint, then we will need a new ctor (or factory method)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: Why would not it be safe? If `ABCPool` decided to identify itself as `abc`, it's totally fine. Also note that this is exactly what is done in tests.
"Type safety" usually referrers to types mismatches, which can't happen here.

JamesKovacs (approved):
I see why we don't have a race condition in `AcquireConnectionHelper` now. Because each request creates its own instance, the `_poolQueueWaitResult` acts like a local variable. I had forgotten that `AcquireConnectionHelper` was instantiated on a per-request basis. Thanks for addressing/discussing my feedback. This PR now LGTM.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Not sure I understood the last comment completely. The only thing `PoolState` needs is some "hint" for exception message, thinking about that, we can even do without the pool identifier at all, as it's not consistent with other exceptions.
I am not sure why would be need new ctors or factories, as  `PoolState` should not be aware of this objects/serialization/formatting at all. Also if any other helper classes would like to add pool info to their exception (not sure why they don't), this info should formatted in single place consistently, not by helpers, and not by exceptions.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
BorisDog: Thanks, can't think how to improve the naming here. Let me know if there are any better ideas :)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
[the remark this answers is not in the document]
DmitryLukyanov: having ability to put any random string looks not the best solution to me, but it's minor, so I will rely on you here

DmitryLukyanov (approved):
LGTM

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -30,6 +30,8 @@ public class ConnectionPoolSettings
...
+        private readonly bool _isPausable;
```
BorisDog, line 34: @DmitryLukyanov `_isPausable` is used in LB mode.

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -107,6 +107,10 @@ protected override void InitializeSubClass()
 
             // propagate event to upper levels, this will be called only once
             TriggerServerDescriptionChanged(this, eventArgs);
+
+            // mark pool as ready, start the connection creation thread.
+            // note that the pool can not be paused in LB mode.
```
DmitryLukyanov, line 112: well it's not really true. The connectionPool is paused initially in all modes. And we transit the pool to the ready state here from the paused state. I don't say that it should be changed here (other than probably the comment wording), but it would be good to hear @JamesKovacs  opinion here also

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -113,32 +113,35 @@ protected override void Dispose(bool disposing)
...
             {
-                ConnectionPool.Clear();
+                // non connection exception
                 return;
             }
 
-            if (ex is MongoConnectionException mongoConnectionException)
+            var (invalidateAndClear, cancelCheck) = ex switch
+            {
+                MongoAuthenticationException => (invalidateAndClear: true, cancelCheck: false),
```
DmitryLukyanov, line 124: I forgot, does the spec require invalidating for auth exception now?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -286,66 +285,58 @@ public int GetGeneration(ObjectId? serviceId)
...
-                catch
+                while (!cancellationToken.IsCancellationRequested)
                 {
-                    // ignore exceptions
+                    try
+                    {
+                        await PrunePoolAsync(cancellationToken).ConfigureAwait(false);
+                        await EnsureMinSizeAsync(cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (MongoConnectionException exception)
```
DmitryLukyanov, line 299: does it make sense to rely on SDAM error handling and defer all exception to that point? Non `MongoConnectionException` will be ignored there anyway

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -727,7 +880,7 @@ public void Dispose()
             {
                 if (_connectingWaitStatus == SemaphoreSlimSignalable.SemaphoreWaitResult.Entered)
                 {
-                    _pool._connectingQueue.Release();
+                    _pool._maxConnectingQueue.Release();
                 }
 
                 if (_disposeConnection)
```
DmitryLukyanov, line 886: can we change this logic now?

DmitryLukyanov (changes requested):
It was a bit hard to review it since all the changes looked like a fully new changes. Some suggestions/questions are requested

tests/MongoDB.Driver.Core.Tests/Core/Helpers/MockConnection.cs
```diff
@@ -171,18 +175,27 @@ public List<RequestMessage> GetSentMessages()
...
         public Task OpenAsync(CancellationToken cancellationToken)
         {
+            _openingEventHandler?.Invoke(new ConnectionOpeningEvent(_connectionId, _connectionSettings, null));
+
             _openedAtUtc = DateTime.UtcNow;
             // _lastUsedAtUtc is set in the SendBufferAsync method in the BinaryConnection
             // which is one from methods called inside OpenAsync
             _lastUsedAtUtc = DateTime.UtcNow;
+
+            _openedEventHandler?.Invoke(new ConnectionOpenedEvent(_connectionId, _connectionSettings, TimeSpan.FromMilliseconds(100), null));
```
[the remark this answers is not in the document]
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -220,6 +224,20 @@ public void Invalidate(string reasonInvalidated, TopologyVersion responseTopolog
         // protected methods
         protected abstract void InitializeSubClass();
 
+        protected bool IsStandaloneTopology()
```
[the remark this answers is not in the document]
BorisDog: Done

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -212,10 +215,16 @@ private void SetDescription(ServerDescription newDescription)
 
         private void OnDescriptionChanged(object sender, ServerDescriptionChangedEventArgs e)
         {
-            if (e.NewServerDescription.HeartbeatException != null)
+            var newDescription = e.NewServerDescription;
+            if (newDescription.HeartbeatException != null)
             {
                 ConnectionPool.Clear();
             }
+            else if (newDescription.IsDataBearing ||
```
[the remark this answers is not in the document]
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
BorisDog: Yes: 

> elif isNetworkError(error) or (not error.completedHandshake and (isNetworkTimeout(error) or isAuthError(error)

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
BorisDog: Clarified this nuance. The pool can start in paused state, but can not be transitioned back to paused mode.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
BorisDog: I don't think so. SDAM error handling should be applied only in connectivity area, so I don't see why it should be piggybacked for other possible exceptions.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Thanks, removed that comment.  Errors during open are handled via SDAM.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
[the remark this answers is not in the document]
DmitryLukyanov: see ConnectionPoolSettingsTests class

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -150,72 +156,51 @@ public int UsedCount
...
-            _clearedEventHandler?.Invoke(new ConnectionPoolClearedEvent(_serverId, _settings));
+                    _clearedEventHandler?.Invoke(new ConnectionPoolClearedEvent(_serverId, _settings));
+                }
+            }
         }
 
         public void Clear(ObjectId serviceId)
         {
-            ThrowIfNotOpen();
+            lock (_poolState)
```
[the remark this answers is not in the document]
DmitryLukyanov: I would say this logic should be hidden inside this class, but ok

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -150,73 +156,55 @@ public int UsedCount
...
+            {
+                // access to _poolState should be synchronized
+                _poolState.ThrowIfNotReadyNonPausable();
+            }
 
+            // generation increment can happen outside lock, as _serviceStates is threadsafe
```
DmitryLukyanov, line 206: I don't see sense in this comment, other than it says that there is more work here to be done which is not true at least yet. I think this particular comment it's better to revert

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: >I don't think so. SDAM error handling should be applied only in connectivity area

I don't think that it's fully true statement. We catch all exceptions and it's up to SDAM how many of them we should handle. Yes, currently we handle only MongoConnectionException there but it's decision that we take on the SDAM level. In other words,  if at some moment we will start looking at more exceptions in SDAM, we won't need to change this logic. In other words (x2), we can simply delegate all/any exceptions handling in connection to the SDAM level

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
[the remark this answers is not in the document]
DmitryLukyanov: it looks like the spec won't be rephrased, then let's add a comment of why this wording contradicts to what the spec requires

tests/MongoDB.Bson.TestHelpers/JsonDrivenTests/JsonDrivenTest.cs
```diff
@@ -66,7 +66,8 @@ public virtual void Arrange(BsonDocument document)
                 SetArguments(document["arguments"].AsBsonDocument);
             }
 
-            if (document.Contains("error"))
+            if (document.TryGetElement("error", out var errorElement) &&
+                (!errorElement.Value.IsBoolean || errorElement.Value.AsBoolean))
```
DmitryLukyanov, line 70: can it be not `boolean`?

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -45,15 +47,35 @@ public class ConnectionPoolSettings
...
             Optional<int> minConnections = default(Optional<int>),
             Optional<int> waitQueueSize = default(Optional<int>),
-            Optional<TimeSpan> waitQueueTimeout = default(Optional<TimeSpan>))
+            Optional<TimeSpan> waitQueueTimeout = default(Optional<TimeSpan>)) : this(
+                maintenanceInterval: maintenanceInterval,
+                maxConnections: maxConnections,
+                minConnections: minConnections,
+                waitQueueSize: waitQueueSize,
+                waitQueueTimeout: waitQueueTimeout,
+                isPausable: true)
```
DmitryLukyanov, line 56: it's better to remove this, since you set only public fields here

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -45,15 +47,35 @@ public class ConnectionPoolSettings
             Optional<int> maxConnections = default(Optional<int>),
             Optional<int> minConnections = default(Optional<int>),
             Optional<int> waitQueueSize = default(Optional<int>),
-            Optional<TimeSpan> waitQueueTimeout = default(Optional<TimeSpan>))
+            Optional<TimeSpan> waitQueueTimeout = default(Optional<TimeSpan>)) : this(
```
DmitryLukyanov, line 50: expected formatting here:

    Optional<TimeSpan> waitQueueTimeout = default(Optional<TimeSpan>)) 
    : this(...

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -144,20 +270,76 @@ public void Finally()
...
+                do
                 {
-                    try
-                    {
-                        _pool._poolQueue.Release();
-                    }
-                    catch
+                    freeSlots = _pool._waitQueueFreeSlots;
+
+                    if (freeSlots == 0)
```
DmitryLukyanov, line 282: I still don't like this method, but I will rely on you here

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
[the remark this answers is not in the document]
BorisDog: Added tests there.

src/MongoDB.Driver.Core/MongoInternalDefaults.cs
```diff
@@ -21,5 +21,7 @@ public static class ConnectionPool
         {
             public const int MaxConnecting = 2;
         }
+
+        public const int SingleServerSelectionTimeoutMS = 1000;
```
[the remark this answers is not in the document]
DmitryLukyanov: not answered

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -849,6 +1220,227 @@ public void PrunePoolAsync_should_remove_all_expired_connections([RandomSeed]int
...
+        [Theory]
+        [ParameterAttributeData]
+        public void WaitQueue_should_be_cleared_on_pool_clear(
+            [Values(true, false)] bool isAsync,
+            [Values(1, 2, 5)] int blockedInQueueCount)
+        {
+            const int maxConnecting = 2;
+            var threadsCount = maxConnecting + blockedInQueueCount;
+            var waitQueueSize = threadsCount;
+            var settings = _settings.With(
```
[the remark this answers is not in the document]
DmitryLukyanov: missed

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: agreed, I missed that we can catch more exception here (unrelated to connection at all)

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: then you should rephrase it. I don't think `for now` should be left here.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
DmitryLukyanov: oh, then we probably specify both the internal options here, idk, I guess leave it as is, it's internal/private logic we can change it next time if we will find a better way

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
DmitryLukyanov: :(

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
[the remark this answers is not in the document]
BorisDog: Done.

DmitryLukyanov (approved):
LGTM

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -209,29 +208,35 @@ private void OnMonitorDescriptionChanged(object sender, ServerDescriptionChanged
 
         private void SetDescription(ServerDescription newDescription)
```
BorisDog, line 209: @DmitryLukyanov please review this change. It's a bug fix that was missed out in this PR.

DmitryLukyanov: can you please provide mode details?

BorisDog: @DmitryLukyanov Yes, setting server description has to happen after ready and before clear.

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -200,25 +198,44 @@ private void OnMonitorDescriptionChanged(object sender, ServerDescriptionChanged
...
         {
-            var oldDescription = Interlocked.CompareExchange(ref _currentDescription, value: newDescription, comparand: _currentDescription);
-            OnDescriptionChanged(sender: this, new ServerDescriptionChangedEventArgs(oldDescription, newDescription));
-        }
+            // Current assumption is SetDescription is always synchronized under _monitor.Lock.
+            // This synchronization technically can be violated by calling server.Invalidate not under _monitor.Lock.
+            // Therefore _currentDescription and ConnectionPool state can get out of sync.
+            var oldDescription = _currentDescription;
 
-        private void OnDescriptionChanged(object sender, ServerDescriptionChangedEventArgs e)
```
DmitryLukyanov, line 213: I would want to leave this method since it corresponds to the spec naming in pseudo code

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -200,25 +198,44 @@ private void OnMonitorDescriptionChanged(object sender, ServerDescriptionChanged
...
-        {
-            if (e.NewServerDescription.HeartbeatException != null)
+            if (newDescription.HeartbeatException != null || clearConnectionPool)
             {
+                // set new description before clearing the pool
+                _currentDescription = newDescription;
```
DmitryLukyanov, line 215: as It was discussed, this change has no practical effect. In theory I don't mind to leave this, but 2 points that bother me:
1. This makes me think that this is important for any logic we have which is not true. So I would want to not pay attention on something that has no real value.
2. This change requires removing `OnDescriptionChanged` method that is named and done similar to the spec pseudo code

So I would suggest revert this part. I assume that the solution for the failed tests that you mentioned before is related to the fact that now we don't make pool ready in case if the invalidating was triggered from the cluster level. Please correct me if I'm wrong

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -200,25 +198,44 @@ private void OnMonitorDescriptionChanged(object sender, ServerDescriptionChanged
             if (heartbeatReplyNotReceived ||
                 TopologyVersion.IsStalerThanOrEqualTo(currentDescription.TopologyVersion, e.NewServerDescription.TopologyVersion))
             {
-                SetDescription(e.NewServerDescription);
+                SetDescription(e.NewServerDescription, false);
             }
         }
 
-        private void SetDescription(ServerDescription newDescription)
+        private void SetDescription(ServerDescription newDescription, bool clearConnectionPool)
```
DmitryLukyanov, line 205: I think this name `clearConnectionPool` should be precise. Looking at this argument I feel like this argument is responsible for clearing the pool, but it's true only partially. What about: `forceClearConnectionPool`?

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -44,7 +43,7 @@ internal class DefaultServer : Server
         #endregion
 
         private readonly ServerDescription _baseDescription;
-        private ServerDescription _currentDescription;
+        private volatile ServerDescription _currentDescription;
```
DmitryLukyanov, line 46: I think the previous logic still works. If you want to proceed with this, we can discuss it

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -200,25 +198,44 @@ private void OnMonitorDescriptionChanged(object sender, ServerDescriptionChanged
             if (heartbeatReplyNotReceived ||
                 TopologyVersion.IsStalerThanOrEqualTo(currentDescription.TopologyVersion, e.NewServerDescription.TopologyVersion))
             {
-                SetDescription(e.NewServerDescription);
+                SetDescription(e.NewServerDescription, false);
```
DmitryLukyanov, line 201: let's add a explicit argument name

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
BorisDog: @JamesKovacs could you please review this change?

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
BorisDog: Discussed offline. Will be addressed in follow up ticket.

PR #521, opened 2021-05-07
Linq3 AST directory code review

rstam:


rstam:
In this code review let's limit ourselves to the AST directory.

A number of subsequent PRs will review other directories. The goal is to break the code review into more manageable chunks.

rstam:
Notes:

Because these classes are so closely tied to the server we prioritize names that mirror exactly what the server uses even if those names conflict with naming standards we would otherwise use in the driver.

Very few classes are intended to be instantiated directly. Instead all instances are intended to be created using static factory methods (as in the `AstExpression`, `AstFilter`, `AstProject` and `AstStage` classes). Not only are factory methods convenient, but they provide a place to centralize processing like constant folding and `$and`/`$or` flattening. Note that C# `Expression` classes are also created using factory methods so this is a familiar pattern.

All AST classes are intended to be immutable. If you see any cases of mutability be sure to note them.

Note that the `AstElemMatchFilterOperation` uses the magic field name `$elem` to represent the element being matched. This magic field name is removed when `Render` is called, but it allows us to use all existing `AstFilter` classes with `AstElemMatchFilterOperation`. We can discuss further via Zoom if there are questions.

An open question is when to `Render` single arguments with or without `[]` surrounding the arguments. The server documentation is a little inconsistent on this. Currently the AST code tries to mimic the server documentation. Alternatives might be to *always* use `[]` even for single arguments, or to *only* use `[]` for single arguments *iff* the value is an array (in which case `[]` is needed to disambiguate).

JamesKovacs (approved):
The following is a summary of review notes and discussions on Slack.

- `AstNodeType` has some unused enum values
  - `GtFilter`: There are no corresponding $lt, $lte, and $gte filters because those are all represented by `AstBinaryExpression` with an `AstBinaryOperator` as is $gt.
  - `IncludeFieldProjectionSpecification`
  - `ListLocalSessionsStage`
  - `ReplaceRootStage`
  - `SetFieldProjectSpecification`
  - Fixed
- `AstAndFilter` line 56: `!(filter is AstFieldOperationrFilter)` => `filter is not AstFieldOperationFilter`
  - Changed
- Tidy using statements
  - Done
- `AstStage.Group` methods should chain to one another rather than duplicating code
  - Done
- Stages and filters are only tested implicitly and many stages are completely untested
  - Much of this code is tested implicitly by the LINQ2 tests
  - Additional code coverage will be added as needed
- `AstBucketAutoStage`: Consider making `granularity` an enum
- `AstLookupStageUncorrelatedMatch` is never used (`AstLookupStage.cs`)
- `AstMergeStage` renders `one` instead of `on`
  - Fixed
- `AstReplaceRootStage.NodeType` is `AstNodeType.RedactStage`
- `AstReplaceWithStage.NodeType` is `AstNodeType.RedactStage`
  - No `AstNodeType.ReplaceWithStage`
- `AstSampleStage.NodeType` is `AstNodeType.RedactStage`
- `AstSkipStage.NodeType` is `AstNodeType.LimitStage`
- `AstListLocalSessionsStage` is `AstNodeType.ListSessionsStage`
  - `AstNodeType` issues have been fixed
- Missing `AstSortByCountStage`
  - Added
- C# 9 introduced `is not`
  - `if (!(filter is AstFieldOperationFilter fieldOperationFilter))`
    can be written as
  - `if (filter is not AstFieldOperationFilter fieldOperationFilter)`
- Flattening logic (e.g. `AstFilter.And` and other places) belongs in an optimizer, not in a static factory method
  - Discussed advantages of having simple micro-optimizations in the static factory methods and having the optimizer focus more on large-scale optimizations
- `AstFilterField.SubField`... Should path be `$CURRENT` or `$$CURRENT`?
  - Fields are written as `$field` where `field` is the path and rendered as `${path}`
  - Thus `$$CURRENT` has a path of `$CURRENT`, which looks odd but is correct. `${path}` evaluates to `$$CURRENT`
- `AstExpression.Let(var1, var2, var3, @in)` should use capacity 3
- `AstExpression.Add` only optimizes `Int32`, why not other constant numeric types?
- Consider adding constant folding for other operations like multiply, divide, subtract, etc.
  - Only enough constant folding was added to get the LINQ2 tests passing
  - Additional constant folding can be added in the future as needed
- `AstExpression.Not` could be flattened if `expr` is a constant expression
- Add guard clauses
  - `AstExpression.Divide`
  - `AstExpression.Log(arg, @base)` should have guard clause `Ensure.IsGreaterThanZero(@base)`
  - `AstExpression.Sqrt`
  - Design is to mostly delegate through to the server, except when required by C# AST manipulation code or obvious not null conditions
- `AstIndexOf<<>>Expression` are identical except for `$indexOf<<>>`, which is similar to `AstBinaryExpression` and similar
  - `$indexOfArray` (arrays) is conceptually different than `$indexOfBytes`/`$indexOfCP` (strings)
  - Therefore best to leave as separate classes
  - Not deemed worthwhile to combine `$indexOfBytes` and `$indexOfCP` into one class

rstam:
Closed as we are done looking at this directory.

PR #523, opened 2021-05-10
CSHARP-3534: Workarounds for killAllSessions in unified test runner.

DmitryLukyanov:
Spec: https://github.com/mongodb/specifications/commit/1b4ce24071065b9c00263939dbb7e9524147c068#diff-f718c246f51e05b3731a7b03b176a0d4b0c381a9261810742d2677823f963f98R2638

### DIFF
--- a/tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
+++ b/tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
@@ -324,12 +324,8 @@ namespace MongoDB.Driver.Tests.UnifiedTestOperations
                     // SERVER-54216
                     return true;
                 }
-                if (!Feature.KillAllSessions.IsSupported(serverVersion) && ex.Code == (int)ServerErrorCode.CommandNotFound)
-                {
-                    // if the command is executed on a pre-3.6 server
-                    // NOTE: this is not the case for the c# driver, but left it here just to make full sync with the spec
-                    return true;
-                }
+                // the spec also mentions CommandNotFound code, but it's not the case for c#
+                // since we don't call killAllSessions for servers less than 3.6
 
                 return false;
             }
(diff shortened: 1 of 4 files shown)


tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -305,21 +296,42 @@ private IUnifiedTestOperation CreateOperation(BsonDocument operation, UnifiedEnt
...
+                {
+                    // SERVER-38335
+                    return true;
+                }
+                if (ex.Code == (int)ServerErrorCode.Unauthorized)
+                {
+                    // SERVER-54216
+                    return true;
+                }
+                if (!Feature.KillAllSessions.IsSupported(serverVersion) && ex.Code == (int)ServerErrorCode.CommandNotFound)
```
DmitryLukyanov, line 327: I asked the spec commit's creator about how this case can happen, will update it if it will be required in the next commit

DmitryLukyanov: agreed, it should be removed but since the spec is not really clear with this, I'm going to add a comment that mention the reason of why we don't consider the 3rd code

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -305,21 +296,42 @@ private IUnifiedTestOperation CreateOperation(BsonDocument operation, UnifiedEnt
...
+                var adminDatabase = client.GetDatabase(DatabaseNamespace.Admin.DatabaseName);
 
                 try
                 {
                     adminDatabase.RunCommand<BsonDocument>(command);
                 }
-                catch (MongoCommandException exception) when (
-                    CoreTestConfiguration.ServerVersion < new SemanticVersion(4, 1, 9) &&
-                    exception.Code == (int)ServerErrorCode.Interrupted)
+                catch (MongoCommandException ex) when (ShouldIgnoreError(ex))
```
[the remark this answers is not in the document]
DmitryLukyanov: currently we have 2.5 conditions here, I think it's enough to move this into a new method

MikalaiMazurenka (approved):
LGTM

PR #526, opened 2021-05-12
CSHARP-3527: Consider accepting EncryptionAlgorithm enum in EncryptOptions.

DmitryLukyanov:
NOTE: I launched tests only locally, let me know if I need to make it possible to launch all tests on EG before a new libmongocrypt will be pushed

### DIFF
--- a/CSharpDriver.sln
+++ b/CSharpDriver.sln
@@ -48,11 +48,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		MongoDBTest.ruleset = MongoDBTest.ruleset
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SkippableTests", "tests\SkippableTests\SkippableTests.csproj", "{D198833A-6AC3-4327-8B02-5095455192D0}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkippableTests", "tests\SkippableTests\SkippableTests.csproj", "{D198833A-6AC3-4327-8B02-5095455192D0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MongoDB.Driver.TestConsoleApplication", "tests\MongoDB.Driver.TestConsoleApplication\MongoDB.Driver.TestConsoleApplication.csproj", "{2E5780D2-29A5-483C-9CA2-844F45A66D0C}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MongoDB.Driver.TestConsoleApplication", "tests\MongoDB.Driver.TestConsoleApplication\MongoDB.Driver.TestConsoleApplication.csproj", "{2E5780D2-29A5-483C-9CA2-844F45A66D0C}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AstrolabeWorkloadExecutor", "tests\AstrolabeWorkloadExecutor\AstrolabeWorkloadExecutor.csproj", "{B90F025F-89D3-436A-AD78-6AA304A6E240}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AstrolabeWorkloadExecutor", "tests\AstrolabeWorkloadExecutor\AstrolabeWorkloadExecutor.csproj", "{B90F025F-89D3-436A-AD78-6AA304A6E240}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MongoDB.Libmongocrypt", "..\libmongocrypt\bindings\cs\MongoDB.Libmongocrypt\MongoDB.Libmongocrypt.csproj", "{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -136,6 +138,10 @@ Global
 		{B90F025F-89D3-436A-AD78-6AA304A6E240}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{B90F025F-89D3-436A-AD78-6AA304A6E240}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{B90F025F-89D3-436A-AD78-6AA304A6E240}.Release|Any CPU.Build.0 = Release|Any CPU
+		{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -160,6 +166,7 @@ Global
 		{D198833A-6AC3-4327-8B02-5095455192D0} = {E472BDF5-61F1-461D-872B-9F53BB3ACA80}
 		{2E5780D2-29A5-483C-9CA2-844F45A66D0C} = {E472BDF5-61F1-461D-872B-9F53BB3ACA80}
 		{B90F025F-89D3-436A-AD78-6AA304A6E240} = {E472BDF5-61F1-461D-872B-9F53BB3ACA80}
+		{D95C7F2B-8A25-4B11-8C2F-79D39AEADE01} = {D2012971-32BB-4C5F-BFC4-30A9994AB152}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {24BEC44B-92B0-43AA-9B15-163459D0C098}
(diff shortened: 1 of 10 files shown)


CSharpDriver.sln
```diff
@@ -48,11 +48,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		MongoDBTest.ruleset = MongoDBTest.ruleset
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SkippableTests", "tests\SkippableTests\SkippableTests.csproj", "{D198833A-6AC3-4327-8B02-5095455192D0}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkippableTests", "tests\SkippableTests\SkippableTests.csproj", "{D198833A-6AC3-4327-8B02-5095455192D0}"
```
DmitryLukyanov, line 51: this will be reverted when the new nuget version will be available

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -149,6 +148,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\libmongocrypt\bindings\cs\MongoDB.Libmongocrypt\MongoDB.Libmongocrypt.csproj" />
```
DmitryLukyanov, line 151: this will be reverted when the new nuget version will be available

src/MongoDB.Driver/MongoDB.Driver.csproj
```diff
@@ -59,6 +58,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\libmongocrypt\bindings\cs\MongoDB.Libmongocrypt\MongoDB.Libmongocrypt.csproj" />
```
DmitryLukyanov, line 61: this will be reverted when the new nuget version will be available

tests/MongoDB.Driver.Examples/ExplicitEncryptionExamples.cs
```diff
@@ -20,7 +20,6 @@
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
 using MongoDB.Driver.Encryption;
-using MongoDB.Libmongocrypt;
```
DmitryLukyanov, line 23: we usually don't leave so simple changes, but this ns can confuse, so it's better to remove all usage of it (related to algorithm)

BorisDog (approved):

src/MongoDB.Driver/Encryption/EncryptOptions.cs
```diff
@@ -23,6 +23,16 @@ namespace MongoDB.Driver.Encryption
...
+        #region static
+        private static string ConvertEnumAlgorithmToString(EncryptionAlgorithm encryptionAlgorithm) =>
+            encryptionAlgorithm switch
+            {
+                EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic => "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic",
+                EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Random => "AEAD_AES_256_CBC_HMAC_SHA_512-Random",
+                _ => encryptionAlgorithm.ToString(),
+            };
+        #endregion
+
```
JamesKovacs, line 35: This should be part of `EncryptionAlgorithm`. See below.

src/MongoDB.Driver/Encryption/EncryptionAlgorithm.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Deterministic algorithm.
+        /// </summary>
+        AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic,
+        /// <summary>
+        /// Random algorithm.
+        /// </summary>
+        AEAD_AES_256_CBC_HMAC_SHA_512_Random
+    }
+}
```
JamesKovacs, line 32: I would recommend making this a smart enum so that we can encapsulate all the conversion logic within the class. Something like this:

```
public class EncryptionAlgorithm
{
    public static EncryptionAlgorithm AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic = new EncryptionAlgorithm("AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic");
    public static EncryptionAlgorithm AEAD_AES_256_CBC_HMAC_SHA_512_Random = new EncryptionAlgorithm("AEAD_AES_256_CBC_HMAC_SHA_512-Random");

    private readonly string _algorithm;

    private EncryptionAlgorithm(string algorithm)
    {
        _algorithm = algorithm;
    }

    public override string ToString() => _algorithm;
}
```

When using this smart enum class, it behaves just like an enum, but you can hang methods off it like `ToString()`:

```
Console.WriteLine(EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic);
Console.WriteLine(EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Random);
```

You can optionally add implicit casts to `string` so you can just use the smart enum wherever the string version is expected.

```
public static implicit operator string(EncryptionAlgorithm algorithm) => algorithm.ToString();
```

This would save you from having the overload (as you can just use the existing string parameter method), but makes the enum itself less discoverable. I think it's probably best not to add an implicit string conversion operator.

JamesKovacs (changes requested):
Comments inline.

src/MongoDB.Driver/Encryption/EncryptionAlgorithm.cs
DmitryLukyanov: this looks like more complicated than we need to here, @BorisDog what do you think? I will rely on both of you :)

BorisDog: I like the "smart" enums technique for more complicated cases. I think we can keep things simple here and just use regular enums.

JamesKovacs: I'd still prefer to keep all the `EncryptionAlgorithm`-related code together in a smart enum class, but this is a relatively simple case. I'm OK with leaving this as-is.

JamesKovacs (approved):
LGTM

DmitryLukyanov:
the failures I observed yesterday are gone away (The latest [EG](https://evergreen.mongodb.com/version/60f6150d3e8e86392bd473e0)). It looks like it's related to the scope of this ticket https://jira.mongodb.org/browse/DRIVERS-1541. So closing this PR.

damon-mcgraw-rack:
@DmitryLukyanov  this is probably not the best place for this but I believe that when this PR was merged and then MongoDb.libmongocrypt nuget package published that it created a binary incompatibility with the MongoDb.Driver nuget package. I was wondering when an update MongoDb.Driver package might be released which would fix the incompatibility.

The specific error I get is (please forgive the formatting)
`Error loading assemblies: Error loading MongoDB.Driver, Version=2.13.0.0, Culture=neutral, PublicKeyToken=null: System.TypeLoadException: Could not load type MongoDB.Libmongocrypt.EncryptionAlgorithm from assembly MongoDB.Libmongocrypt, Version=1.2.2.0, Culture=neutral, PublicKeyToken=null`

reverting to MongoDb.Libmongocrypt 1.2.1 resolves the issue but I'd rather not lock to that version if a new driver is coming soon.

DmitryLukyanov:
Hello @damon-mcgraw-rack thanks for your message, I don't see this error with any currently released driver versions. Can you please provide more details about your use case?

damon-mcgraw-rack:
@DmitryLukyanov Thanks for the quick response!
We have an IIS hosted web API targeting .Net Framework 4.7, the API loads various plugins through reflection depending on it's configuration and some of those plugins use the C# mongo driver. I'm seeing the error in our CI server, as well as our test environment when the plugin is loaded. 
We're using MongoDb.Driver 2.13 and MongoDb.LibMongoCrypt 1.2.2 when I see the error. With MongoDb.LibMongocrypt 1.2.1 the error goes away.

I suspected the namespace changed when the enum was moved and saw that this also updated the mongocrypt version used by the driver.
I'll try to provide a more specific example.

DmitryLukyanov:
yeah, more specific example will be helpful, `MongoDb.Driver 2.13` should look at `MongoDb.LibMongocrypt 1.2.1`, so I'm still now sure what exactly leads to your issue

damon-mcgraw-rack:
I think that's exactly it, I was using MongoDb.Driver 2.13 with MongoDb.LibMongoCrypt 1.2.2 as there isn't a newer release of MongoDb.Driver.

DmitryLukyanov:
by default, the `2.13 mongo driver` should work only with `MongoDb.LibMongocrypt 1.2.1`, so the question here is how the `2.13 driver` in your case sees `1.2.2 MongoDb.LibMongocrypt` ?

damon-mcgraw-rack:
We update the packages frequently and just pull the latest available, so once the 1.2.2 MongoDb.LibMongoCrypt was available it was automatically pulled in. I can lock the MongoDb.LibMongoCrypt version to 1.2.1 until a new driver is released if there are no current plans to release a new MongoDb.Driver package.

DmitryLukyanov:
>We update the packages frequently 

so you changed this [reference](https://github.com/mongodb/mongo-csharp-driver/blob/v2.13.0/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj#L129) manually/via reflection somehow?

damon-mcgraw-rack:
The nuget package is a packages.config file and our process just goes through and updates anything in that file.
It's similar to just doing `Get-Project -All | update-package` from the package manager console in VS. 
So yes once the package is include from the initial addition of MongoDb.Driver, it is updated independently.

DmitryLukyanov:
Hello @damon-mcgraw-rack , the driver 2.13.1 has been released. This release works with the latest libmongocrypt version, so it should solve your issue.

damon-mcgraw-rack:
@DmitryLukyanov I've confirmed that with the new packages my issue is resolved. Thank you for the quick response!

PR #527, opened 2021-05-12
CSHARP-3655: Removed deprecated URI option slaveOk=true

JamesKovacs:
The slaveOk URI option has been deprecated in favour of readPreference=secondaryPreferred since at least MongoDB 2.2. Many drivers (such as pymongo) have already removed it while others never implemented it because it predated the URI options spec.


rstam (approved):
LGTM

PR #529, opened 2021-05-13
CSHARP-3427: Can't make use of compression when application path contain space.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/609d8e0e61837d3a55f5a870

### DIFF
--- a/src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
+++ b/src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
@@ -26,24 +26,25 @@ namespace MongoDB.Driver.Core.NativeLibraryLoader
         public virtual bool IsX32ModeSupported => false;
 
         // public methods
+        public virtual string GetBaseAssemblyUri() => typeof(RelativeLibraryLocatorBase).GetTypeInfo().Assembly.CodeBase;
+
         public string GetLibraryAbsolutePath(OperatingSystemPlatform currentPlatform)
         {
             var relativePath = GetLibraryRelativePath(currentPlatform);
             return GetAbsolutePath(relativePath);
         }
 
-        public virtual Assembly GetLibraryBaseAssembly()
+        public virtual string GetLibraryBaseAssemblyPath()
         {
-            return typeof(RelativeLibraryLocatorBase).GetTypeInfo().Assembly;
+            var baseAssemblyPathUri = GetBaseAssemblyUri();
+            var uri = new Uri(baseAssemblyPathUri);
+            return Uri.UnescapeDataString(uri.AbsolutePath);
         }
 
         public virtual string GetLibraryBasePath()
         {
-            var assembly = GetLibraryBaseAssembly();
-            var codeBase = assembly.CodeBase;
-            var uri = new Uri(codeBase);
-            var absolutePath = uri.AbsolutePath;
-            return Path.GetDirectoryName(absolutePath);
+            var absoluteAssemblyPath = GetLibraryBaseAssemblyPath();
+            return Path.GetDirectoryName(absoluteAssemblyPath);
         }
 
         public abstract string GetLibraryRelativePath(OperatingSystemPlatform currentPlatform);
--- a/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
@@ -32,9 +32,10 @@ namespace MongoDB.Driver.Core.Tests.Core.NativeLibraryLoader
         [InlineData("&mongo$csharp@driver%", "&mongo$csharp@driver%")]
         public void GetLibraryBasePath_should_get_correct_paths(string rootTestFolder, string expectedRootTestFolder)
         {
-            string testAssemblyCodeBaseUri = null; // use a default one for null
+            string testAssemblyCodeBaseUri = null; // use assembly-based CodeBase for null
             if (rootTestFolder != null)
             {
+                // mock assembly-based CodeBase path
                 var assemblyPath = Path.Combine(
                     RequirePlatform.GetCurrentOperatingSystem() == SupportedOperatingSystem.Windows ? "C:/" : @"\\data",
                     rootTestFolder,
--- a/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
@@ -1,0 +1,96 @@
+﻿/* Copyright 2021-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.IO;
+using FluentAssertions;
+using MongoDB.Driver.Core.NativeLibraryLoader;
+using MongoDB.Driver.TestHelpers;
+using MongoDB.Shared;
+using Xunit;
+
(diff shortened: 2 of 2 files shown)


src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
```diff
@@ -26,24 +26,25 @@ internal abstract class RelativeLibraryLocatorBase : ILibraryLocator
         public virtual bool IsX32ModeSupported => false;
 
         // public methods
+        public virtual string GetBaseAssemblyUri() => typeof(RelativeLibraryLocatorBase).GetTypeInfo().Assembly.CodeBase;
```
DmitryLukyanov, line 29: the user's suggestion with assembly.Location doesn't work at least with cases when we configure `shadowCopy` in tests. See this [patch](https://evergreen.mongodb.com/version/609d37859ccd4e554a5c78b2) and .net framework variant.

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
```diff
@@ -26,24 +26,25 @@ internal abstract class RelativeLibraryLocatorBase : ILibraryLocator
...
             return GetAbsolutePath(relativePath);
         }
 
-        public virtual Assembly GetLibraryBaseAssembly()
+        public virtual string GetLibraryBaseAssemblyPath()
         {
-            return typeof(RelativeLibraryLocatorBase).GetTypeInfo().Assembly;
+            var baseAssemblyPathUri = GetBaseAssemblyUri();
+            var uri = new Uri(baseAssemblyPathUri);
+            return Uri.UnescapeDataString(uri.AbsolutePath);
```
JamesKovacs, line 41: We are dealing with file paths, which are not URIs. So we shouldn't be using `Uri` to determine the absolute path. We should be using the `Path` API, which is intended for file paths.

```
public virtual string GetLibraryBaseAssemblyPath()
{
    var baseAssemblyPathUri = GetBaseAssemblyUri();
    return Path.GetFullPath(baseAssemblyPathUri);
}
```

JamesKovacs (changes requested):
We should be using `Path.GetFullPath()` not `Uri.AbsolutePath`. (See below.)

As well `RelativeLibraryLocatorBase` is an internal class, but has many public and virtual methods that don't need to be public or virtual. Similarly `GetBaseAssemblyUri` doesn't return a `Uri` but returns a path as a string. We should rename this method to `GetBaseAssemblyPath` or similar.

Overall we should remove our dependence on `Uri` entirely in this class and should instead use the file API (e.g. Path/File/Directory) instead.

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
DmitryLukyanov: the `baseAssemblyPathUri` is actually `Uri` in string form (with `file://` prefix), not sure how `Path.GetFullPath` works with this value and whether Uri-like value is expected input value, but i will check it

JamesKovacs: It is strange how `Assembly.CodeBase` is marked obsolete and we are told to use `Assembly.Location` instead, but `Assembly.Location` says to use `Assembly.CodeBase` if you need the original and not shadow-copied path. So looks like we need to use `Assembly.CodeBase` even if it is marked obsolete.

Though I am curious why the shadow-copied assemblies can't load the unmanaged DLLs. I guess the unmanaged DLLs don't get shadow-copied along with their managed counterparts.

I did check and you are correct that `Assembly.CodeBase` is a URI as a string. Maybe `Uri.LocalPath` - which returns the local operating-system representation of a file name - will do the trick:

https://docs.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=net-5.0

```
var uriPath = typeof(RelativeLibraryLocatorBase).GetTypeInfo().Assembly.CodeBase;
var uri = new Uri(uriPath);
var path = uri.LocalPath;
```

MikalaiMazurenka: `Assembly.CodeBase` in fact contains a path without escaped characters and only differs in slashes and `file:///` prefix. And by using `Uri` you escape/unescape the characters, which isn't guaranteed to work properly. Consider the example where this wouldn't work and consider using the third option in the test I provided: https://gist.github.com/MikalaiMazurenka/b5fff9af75ebeb0c1cf645c6793519d4

tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.IO;
+using FluentAssertions;
+using MongoDB.Driver.Core.NativeLibraryLoader;
+using MongoDB.Driver.TestHelpers;
+using MongoDB.Shared;
+using Xunit;
+
+namespace MongoDB.Driver.Core.Tests.Core.NativeLibraryLoader
+{
+    public class NativeLibraryLoaderTests
```
MikalaiMazurenka, line 26: It was hard to figure out what's going on here. Consider providing additional comments.

MikalaiMazurenka (commented):
Great work! I'm ready to approve it as it, but I would like you to consider some suggestions first.

JamesKovacs (approved):
LGTM

DmitryLukyanov:
@MikalaiMazurenka found a new issue in this logic that is related to handling `#` sign. Uri has a special handling of this where it places all content after `#` (including `#`) into a special field called `Fragment`, so the solution can be just:
>return Uri.UnescapeDataString(uri.AbsolutePath + uri.Fragment)

where Fragment is a string empty if there is no `#`: https://docs.microsoft.com/en-us/dotnet/api/system.uri.fragment?view=net-5.0#remarks 
The good part of the original solution was that there were no changes in logic itself other than fixing in escaping logic, but if we want to fix the new issue also, we don't have a choice.

Regarding the suggestion with just manual `Replace` everything we don't need to, I would avoid it because:
* We have "strange" but at the same time built-in logic for Uri handling so I would prefer it.
* The manual changes here look like a platform-dependent or at least can be platform-dependent in future.

So, @MikalaiMazurenka , @JamesKovacs please take a look at this comment. 
If we decide to fix a new issue now, I created a temp branch with suggested solution [here](https://github.com/DmitryLukyanov/mongo-csharp-driver/tree/csharp3427_suggestion), also related [EG](https://evergreen.mongodb.com/version/60b11741d1fe0727fb1f79ef).

DmitryLukyanov:
note: the mentioned exception https://docs.microsoft.com/en-us/dotnet/api/system.uri.fragment?view=net-5.0#exceptions is the same as here https://docs.microsoft.com/en-us/dotnet/api/system.uri.absolutepath?view=net-5.0#exceptions (also checked it in the source)

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
DmitryLukyanov: >you escape/unescape the characters, which isn't guaranteed to work properly

I'm not sure what do you mean (other than handling of `#`), from my understanding escaping/unescaping should always work in both ways with the same results

tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
DmitryLukyanov: will do

JamesKovacs:
It feels like we are plugging holes rather than solving the actual problem. What we need to know is the location where we are loading files from irregardless of shadow copying. Since shadow copying is controlled by the AppDomain, what does `AppDomain.CurrentDomain.BaseDirectory` point to when shadow copying? Is it the original assembly location prior to shadow copying? And if that’s the case, does it work on all our supported TFMs?

DmitryLukyanov:
Discussed offline, the current next step is creating a new ticket that will consider additional (not space related) issues and possible refactoring of this logic

tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
DmitryLukyanov: done

DmitryLukyanov:
follow-up ticket: https://jira.mongodb.org/browse/CSHARP-3693

JamesKovacs (approved):
LGTM

MikalaiMazurenka (approved):
LGTM

PR #532, opened 2021-05-18
CSHARP-3664: Clarify whether maxPoolSize allows 0 option or no.

DmitryLukyanov:


rstam (changes requested):
There are still some loose ends. 

There are some places in `ClusterBuilderExtensions` that aren't mapping `0` to `int.MaxValue`

We should probably add validation to the setter of `MongoDefaults.MaxConnectionPoolSize` the same way we did to `MongoClientSettings`. (Do we need to do anything about `MongoServerSettings`?).

Should `MongoUrl` allow `0` (like `ConnectionString`) or not? I'm inclined to say the same as `ConnectionString` because `MongoUrl` is simply another class representing a connection string, but sadly that means there are now 2 places that allow `0` instead of just one.

Given the above loose ends I won't object if the team wants to give another think to how/where to interpret `0` as infinity.

DmitryLukyanov:
>(Do we need to do anything about MongoServerSettings?).

I would not touch legacy, we almost never change it in new ticket, however I don't mind if you and @JamesKovacs  want it :) 

>ClusterBuilderExtensions 

in this class we set values to `ConnectionPoolSettings` that has this validation, so I would not worry about this.

>MongoDefaults.MaxConnectionPoolSize

we usually don't validate values there, but I don't mind if you and @JamesKovacs  think differently

>MongoUrl 

I would not touch this class (and builder) and postpone this logic until clientSettings

DmitryLukyanov:
> in this class we set values to `ConnectionPoolSettings` that has this validation, so I would not worry about this.

actually I misunderstood you, yes, we need this logic there

src/MongoDB.Driver/MongoDefaults.cs
```diff
@@ -82,7 +83,16 @@ public static int ComputedWaitQueueSize
                 }
                 else
                 {
-                    return (int)(__waitQueueMultiple * __maxConnectionPoolSize);
+                    var computedWaitQueueSize = __waitQueueMultiple * __maxConnectionPoolSize;
```
DmitryLukyanov, line 86: I'm curious how other drivers solve this issue?

DmitryLukyanov: since currently we cannot have `__maxConnectionPoolSize=0`, this change (to fix `overflow`) is out of scope for this ticket, so I will probably revert it

rstam (changes requested):
As I mentioned in Slack I consider the only two valid alternatives here to be:

1. Allow 0 only in the two classes that represent a connection string: MongoUrl and ConnectionString
2. Allow 0 everywhere

Your current commit doesn't quite satisfy 1 because MongoUrl is not consistent with ConnectionString in how it handles 0. The two classes both represent a connection string and should handle 0 the same way.

Alternative 1 (based on your commits) is prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3664-zeroonconnectionstringonly

Alternative 2 (starting from scratch) is prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3664-zeroeverywhere

I have to say alternative 2 looks a lot cleaner.

DmitryLukyanov:
I disagree about the second approach, the points that I mentioned in the slack:
* in the pool we will need to keep in mind that if we need a maxconnections we should not look at ConnectionPoolSettings and instead we need to look at some workaround which is a separate variable in your case
* also, we will need to make some changes in calculating _waitQueueSize to handle 0  that now will be possible

and all the above without any real profit, we use instead of fixing of one line.

The first one already discussed a lot, I also disagree with this, but I can approve this if @JamesKovacs also agreed with this

rstam:
I guess the first step to resolving pending comments in this PR is to decide which of the following approaches we are going to take.

1. Allow 0 only in the two classes that represent a connection string: MongoUrl and ConnectionString
2. Allow 0 everywhere

Then we can work on resolving the pending code review comments.

@JamesKovacs can you provide your opinion on which of the two approaches would be better?

JamesKovacs:
I don't have a strong opinion on which of Robert's approaches to take. I'd be happy with either. Both handle the waitQueueSize calculation properly.

This PR currently does not handle waitQueueSize correctly and throws an exception when you try to execute `var client = new MonogClient("mongodb://localhost:27017?maxPoolSize=0");`. The reason is that the result of this calculation is `int.MinSize`:
```
var maxPoolSize = int.MaxValue;
var waitQueueMultiple = 5.0d;
var waitQueueSize = (int)(waitQueueMultiple * maxPoolSize);
Console.WriteLine(waitQueueSize);  // -2147483648 aka int.MinSize
```

rstam (changes requested):
Overall looks good.

I would like to centralize the `GetComputedWaitQueueSize` and `GetEffectivemaxConnections` logic by centralizing in one single place at the lowest level possible and then all higher levels can share the logic.

See the comments. Suggested changes are prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3664-movehelpermethodslower

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -49,7 +50,9 @@ public class ConnectionPoolSettings
             _maintenanceInterval = Ensure.IsInfiniteOrGreaterThanOrEqualToZero(maintenanceInterval.WithDefault(TimeSpan.FromMinutes(1)), "maintenanceInterval");
             _maxConnections = Ensure.IsGreaterThanZero(maxConnections.WithDefault(100), "maxConnections");
             _minConnections = Ensure.IsGreaterThanOrEqualToZero(minConnections.WithDefault(0), "minConnections");
-            _waitQueueSize = Ensure.IsGreaterThanOrEqualToZero(waitQueueSize.WithDefault(_maxConnections * 5), "waitQueueSize");
+#pragma warning disable CS0618 // Type or member is obsolete
+            _waitQueueSize = Ensure.IsGreaterThanOrEqualToZero(waitQueueSize.WithDefault(ConnectionStringConversions.GetComputedWaitQueueSize(_maxConnections, 5)), "waitQueueSize");
```
DmitryLukyanov, line 54: I think this logic should not be here, because:
1. We make all required calculations before.
2. Anyway it's deprecated and not used.

Left it as is since it's not important

src/MongoDB.Driver/MongoUrl.cs
```diff
@@ -225,7 +226,8 @@ public int ComputedWaitQueueSize
                 }
                 else
                 {
-                    return (int)(_waitQueueMultiple * _maxConnectionPoolSize);
+                    var effectiveMaxConnections = ConnectionStringConversions.GetEffectiveMaxConnections(_maxConnectionPoolSize);
```
DmitryLukyanov, line 229: this should use the value provided from the builder level, but left as is to avoid changes on the previous logic, anyway it's not used

src/MongoDB.Driver.Legacy/MongoServerSettings.cs
```diff
@@ -887,7 +887,9 @@ public static MongoServerSettings FromUrl(MongoUrl url)
             serverSettings.IPv6 = url.IPv6;
             serverSettings.MaxConnectionIdleTime = url.MaxConnectionIdleTime;
             serverSettings.MaxConnectionLifeTime = url.MaxConnectionLifeTime;
-            serverSettings.MaxConnectionPoolSize = url.MaxConnectionPoolSize;
+#pragma warning disable CS0436 // Type conflicts with imported type
```
rstam, line 890: I would rather this #pragma not be here

The problem here is that we have TWO copies of the same `ConnectionStringConversions` type visible:

1. In THIS assembly because we added a link to it in the `Shared` folder in `MongoDB.Driver.Legacy.csproj`
2. Another in the `MongoDB.Driver` assembly because we can see its internals (due to `InternalsVisibleTo`)

Since we can see the existing copy in `MongoDB.Driver` why don't we just use it instead of creating a name conflict by having two copies.

rstam (changes requested):
Looks good except for the type name conflict which I think we can avoid.

src/MongoDB.Driver.Legacy/MongoServerSettings.cs
DmitryLukyanov: oh, I missed this, I thought that we configured it only for tests (I probably needed spending more time on this change :( ). I will change it.

DmitryLukyanov: done

rstam: > I thought that we configured it only for tests 

Yes. `[InternalsVisibleTo]` should only be used for tests.

I don't know why `MongoDB.Driver.Legacy` has to see `MongoDb.Driver` internals. Some hack I suppose. Which we probably allowed since for years now we've been hoping to get rid of `MongoDB.Driver.Legacy`

rstam (approved):
LGTM

JamesKovacs (approved):
LGTM

PR #534, opened 2021-05-20
CSHARP-3660: Updated server error codes and internal/private variables/fields/methods to use modern terminology.

JamesKovacs:
This PR encompasses the straightforward renames that don't affect the public API. The intent is that this PR should be relatively straightforward to review as there should be no externally visible changes.

There are still many instances of deprecated terminology in the source code, but those will require implementing new public methods/classes and deprecating existing ones. This more complex work will be done in a separate PR.


jyemin (approved):
I'm good with all the API changes in core (mostly parameter name changes).

LGTM

rstam (approved):
LGTM to me also

PR #535, opened 2021-05-24
CSHARP-3683: Stabilize server selection prose test

BorisDog:


DmitryLukyanov:
what's the goal of this change? Just to ensure that non auth variants work well?

DmitryLukyanov (approved):
LGTM

I sounds like we didn't add any solution for the underlying issue here.. but I can confirm that in my tests the issue appears on non auth env much rarely.

PR #538, opened 2021-05-31
CSHARP-3694: Sporadic tests failures: ReceiveMessage_should_not_produce_unobserved_task_exceptions_on_fail

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3694
EG:

1. https://evergreen.mongodb.com/version/60b159ac3e8e860bb9483486
2. https://evergreen.mongodb.com/version/60b15f2b850e61283ccb1f42
3. https://evergreen.mongodb.com/version/60b15f360305b9589f2969c4
4. https://evergreen.mongodb.com/version/60b15f4256234370cdbbfb99

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
@@ -477,8 +477,10 @@ namespace MongoDB.Driver.Core.Connections
             var mockStream = new Mock<Stream>();
             EventHandler<UnobservedTaskExceptionEventArgs> eventHandler = (s, args) =>
             {
-                unobservedTaskExceptionRaised = true;
-                args.SetObserved();
+                if (args.Exception.InnerException is MongoConnectionException)
+                {
+                    unobservedTaskExceptionRaised = true;
+                }
             };
 
             try


tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
```diff
@@ -477,8 +477,10 @@ public void Dispose_should_raise_the_correct_events()
             var mockStream = new Mock<Stream>();
             EventHandler<UnobservedTaskExceptionEventArgs> eventHandler = (s, args) =>
             {
-                unobservedTaskExceptionRaised = true;
-                args.SetObserved();
+                if (args.Exception.InnerException is MongoConnectionException)
```
DmitryLukyanov, line 480: as far as I understand, it's still better to have this part: `args.SetObserved();`

DmitryLukyanov: and as I mentioned previously, I would highlight that we know that there are more exceptions here that we don't know reasons for and that validation on `MongoConnectionException` is a quick fix to make tests green

MikalaiMazurenka: > as far as I understand, it's still better to have this part: args.SetObserved();

This call is redundant on targets above .NET Framework 4.0: [ThrowUnobservedTaskExceptions/Remarks](https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/throwunobservedtaskexceptions-element?redirectedfrom=MSDN#remarks)

> validation on MongoConnectionException is a quick fix to make tests green

This is not a "quick fix". This is a change for test to validate the exact path of the test without validating all of the driver's behavior. The other exception which was caught here has a stack trace originating from Cluster code ([EG log](https://evergreen.mongodb.com/task_log_raw/dot_net_driver_secure_tests__version~3.2_os~windows_64_topology~sharded_cluster_auth~auth_ssl~ssl_test_net452_patch_cf62d156f72a517c4bb61046ddb23104f58ece0c_60a6e3d60ae6066cd0ca76dc_21_05_20_22_33_58/0?type=T#L1479)), which does not get called in this test.

DmitryLukyanov: I would disagree with this, if a test unexpectedly fails, it's not a good way just to ignore it and say this error is not we expect to see here so we will not count it and even won't mention it anywhere and hoping when someone else will work in this area he will need to make all this research again. 

However, I missed that we don't use a cluster in BinaryConnection test, then this error saved on appdomain level (or similar) before your test. It's still a question and should be investigated at some time, but probably it's related to your test only indirectly. So, I will rely on @JamesKovacs  whether it makes sense to mention this behavior here or no

MikalaiMazurenka: > I would disagree with this, if a test unexpectedly fails, it's not a good way just to ignore it and say this error is not we expect to see here so we will not count it and even won't mention it anywhere and hoping when someone else will work in this area he will need to make all this research again.

The purpose of this PR (and the PR which added the test) is not exploratory testing, so your concern is invalid.

MikalaiMazurenka: I have created the ticket to ensure the discovered issue isn't lost: [Unobserved task exception originating from Cluster](https://jira.mongodb.org/browse/CSHARP-3695)

DmitryLukyanov (approved):
LGTM

tests/MongoDB.Driver.Core.Tests/Core/Connections/BinaryConnectionTests.cs
JamesKovacs: The goal of this PR is to resolve the sporadic failures in this test. The fact that another test is leaking exception state into this test isn't good, but is not a problem with this test. I like @MikalaiMazurenka's solution of getting this PR approved and moving the investigation of other tests leaking exceptions into CSHARP-3695. This is a deeper investigation and reconsideration of how we approach task execution in the driver.

JamesKovacs (approved):
LGTM

PR #539, opened 2021-05-31
LINQ3: PR to review ExpressionToFilterTranslators

rstam:


JamesKovacs (changes requested):
Overall LGTM. A few minor corrections (and things to think about) before we move onto the next section...

ExpressionToFilterTranslator.Translate uses exceptions for flow control
- not sure if there is a better way
- line 38 consider `catch (ExpressionNotSupportedException) when (exprOk)`
- additional catch block around `TranslateUsingAggregationOperators` so we can throw an exception indicating that we were unable to translate using filters or $expr?

- LINQ3 placeholders shouldn't look like MQL variables - $$CURRENT, $elem, etc.
  - Choose something unique such as <<CURRENT>> or <<ELEM>> to make it clear that these placeholders don't translate into MQL variables

- ContainsMethodToFilterTranslator
  - private Translate calls TypeImplementsIEnumerable again
  - uses magic $elem placeholder

- StringExpressionToRegexFilterTranslator
  - private static fields can be made readonly
  - extra space between bool and CanTranslate
  - TranslateComparisonExpression only handles constant on the right
    - Add comment noting that ComparisonExpressionToFilterTranslator.Translate flips the operands if constant is on the left, which is why we only need to consider constant on right
  - line 571 s/verifing/verifying/

- EqualsMethodToFilterTranslator
  - line 31 should be logical and not bitwise

- AllWithContainsInPredicateMethodToFilterTranslator
  - fix comment on line 40

- AnyWithContainsInPredicateMethodToFilterTranslator
  - fix comments on lines 40 and 48

- ModuloComparisonExpressionToFilterTranslator
  - only accepts int and long, but any numeric type is acceptable

- FirstMethodToFilterFieldTranslator
  - unused using

- GetItemMethodToFilterFieldTranslator
  - unused usings

- MethodCallExpressionToFilterFieldTranslator
  - First() but no Last()
  - Repro: `GroupBy_where_select_anonymous_type_with_duplicate_accumulators_syntax` test but with `where g.Last().B == "Balloon"` instead of `First()`

rstam:
Done with review

PR #541, opened 2021-06-01
CSHARP-3427: Fix test.

DmitryLukyanov:
Evergreen: https://evergreen.mongodb.com/version/60b6a2611e2d173aca1bb5ae

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
@@ -26,10 +26,14 @@ namespace MongoDB.Driver.Core.Tests.Core.NativeLibraryLoader
     public class NativeLibraryLoaderTests
     {
         [Theory]
-        [InlineData(null, "mongo-csharp-driver")] // the default assembly-based logic
         [InlineData("mongo-csharp-driver", "mongo-csharp-driver")]
         [InlineData("mongo csharp driver", "mongo csharp driver")]
         [InlineData("&mongo$csharp@driver%", "&mongo$csharp@driver%")]
+
+        // the default assembly-based logic. Ideally expectedRootTestFolder should be mongo-csharp-driver,
+        // but since it's not mocked logic, it limits us where we can run our tests from. Avoid it by
+        // making a test assertation less strict
+        [InlineData(null, "")]
         public void GetLibraryBasePath_should_get_correct_paths(string rootTestFolder, string expectedRootTestFolder)
         {
             string testAssemblyCodeBaseUri = null; // use assembly-based CodeBase for null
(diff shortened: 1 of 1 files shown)


MikalaiMazurenka (commented):
I'd recommend to change the name of the commit to be more specific.
My take is "Relax assertion of native library loader test"

DmitryLukyanov:
d

> 
> 
> I'd recommend to change the name of the commit to be more specific.
> My take is "Relax assertion of native library loader test"

done

tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
```diff
@@ -26,10 +26,14 @@ namespace MongoDB.Driver.Core.Tests.Core.NativeLibraryLoader
...
         [Theory]
-        [InlineData(null, "mongo-csharp-driver")] // the default assembly-based logic
         [InlineData("mongo-csharp-driver", "mongo-csharp-driver")]
         [InlineData("mongo csharp driver", "mongo csharp driver")]
         [InlineData("&mongo$csharp@driver%", "&mongo$csharp@driver%")]
+
+        // the default assembly-based logic. Ideally expectedRootTestFolder should be mongo-csharp-driver,
+        // but since it's not mocked logic, it limits us where we can run our tests from. Avoid it by
+        // making a test assertation less strict
+        [InlineData(null, "")]
```
MikalaiMazurenka, line 36: Consider using the suggested refactoring: https://github.com/MikalaiMazurenka/mongo-csharp-driver/commit/bd94d56be40f2e154c48f85b33795cc68592b7cf

MikalaiMazurenka (commented):
I think these two cases are better to be split

tests/MongoDB.Driver.Core.Tests/Core/NativeLibraryLoader/NativeLibraryLoaderTests.cs
DmitryLukyanov: the validation and asserting steps here are the same, the only difference is input data, the ideal solution here is `ClassData` but it looks like overkill for this case, a simple `if` solves everything we need to. So I would leave this as is

MikalaiMazurenka: Alright, then could you remove the extra blank line and consider changing the 3 line comment to
```suggestion
        [InlineData(null, "")] // This case validates only the ending of the path with GetBaseAssemblyUri being used
```

DmitryLukyanov: I wanted to explain why this change is necessary since it's kind of not obvious exception, you comment doesn't provide this info, however I don't mind about changing messages, it will be good to hear @JamesKovacs  opinion here also

JamesKovacs (approved):
LGTM

MikalaiMazurenka (approved):
LGTM

PR #545, opened 2021-06-04
CSHARP-3703: Fix NullReferenceException when Id property has no getter.

JamesKovacs:
While having an `Id` property without a getter is an uncommon design choice, we actually throw a `NullReferenceException` in `NamedIdMemberConvention` rather than simply not returning the getter-less `Id` member. This prevents the use of other attributes or conventions to identify another property as the `_id`. For example the following classes will throw when attempting to serialize them:

```
[BsonNoId]
class Class1 {
    public int Id { set {} }
}

class Class2 {
    [BsonIgnore]
    public int Id { set {} }
}

class Class2 {
    [BsonIgnore]
    public int Id { set {} }

    [BsonId]
    public int AnotherId { get; set; }
}
```

The added null check in `NamedIdMemberConvention` allows all of the above classes to be mapped.

### DIFF
--- a/tests/MongoDB.Bson.Tests/Serialization/Conventions/IdMemberConventionsTests.cs
+++ b/tests/MongoDB.Bson.Tests/Serialization/Conventions/IdMemberConventionsTests.cs
@@ -44,6 +44,14 @@ namespace MongoDB.Bson.Tests.Serialization.Conventions
             public ObjectId _id { get; set; }
         }
 
+        private class TestClassNoIdGetter
+        {
+            public Object Id
+            {
+                set { }
+            }
+        }
+
         [Fact]
         public void TestNamedIdMemberConventionWithTestClassA()
         {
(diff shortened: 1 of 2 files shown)


tests/MongoDB.Bson.Tests/Serialization/Conventions/IdMemberConventionsTests.cs
```diff
@@ -82,5 +90,14 @@ public void TestNamedIdMemberConventionWithTestClassD()
             Assert.NotNull(classMap.IdMemberMap);
             Assert.Equal("_id", classMap.IdMemberMap.MemberName);
         }
+
+        [Fact]
+        public void TestNamedIdMemberConventionWillNotReturnIdWithoutGetter()
+        {
+            var convention = new NamedIdMemberConvention("Id", "id", "_id");
```
rstam, line 97: Maybe name variable `subject` like we usually do?

tests/MongoDB.Bson.Tests/Serialization/Conventions/IdMemberConventionsTests.cs
```diff
@@ -82,5 +90,14 @@ public void TestNamedIdMemberConventionWithTestClassD()
...
+            var classMap = new BsonClassMap<TestClassNoIdGetter>();
+            convention.Apply(classMap);
```
rstam, line 99: It's not required, specially in really short tests, but generally we put one blank line between the Arrange and Act sections and another between the Act and Assert sections.

rstam (approved):
LGTM with minor optional changes requested.

tests/MongoDB.Bson.Tests/Serialization/Conventions/IdMemberConventionsTests.cs
JamesKovacs: I was matching the conventions of the other tests in this file for both the naming and layout. I didn't want to reformat the entire test file when adding a single new test.

PR #562, opened 2021-07-01
CSHARP-3676: SRV support for load balancer mode + SDAM.

DmitryLukyanov:
EG:  https://evergreen.mongodb.com/version/60de54362fbabe3344a05f95

Steps to work with LB (nginx) on windows:

1. Download nginx: http://nginx.org/en/download.html
2. Update `nginx-1.21.0\conf\nginx.conf` on:

			#user  nobody;
			worker_processes  1;
			#error_log  logs/error.log;
			#error_log  logs/error.log  notice;
			#error_log  logs/error.log  info;
			# pid        nginx/logs/nginx.pid;
			events {
				worker_connections  1024;
			}
			stream {
				server {
				   listen 17017;
				   proxy_pass single_server_backend;
				}
				server {
				   listen 17018;
				   proxy_pass multi_server_backend;
				}
				upstream multi_server_backend {
					server localhost:27017;
					server localhost:27018;
				}
				upstream single_server_backend {
					server localhost:27017;
				}
			}

pay attention on ports 17017 and 17018 that represents single mongos and two mongos accordingly.
So, to use them a driver should consume the following connection strings: mongodb://localhost17017?loadBalanced=true or mongodb://localhost17018?loadBalanced=true
3.
cd c:\
unzip nginx-1.21.0.zip
cd nginx-1.21.0
start nginx
4. Verify that nginx has been successfully launched:

		C:\nginx-1.21.0>tasklist /fi "imagename eq nginx.exe"
		Image Name                     PID Session Name        Session#    Mem Usage
		========================= ======== ================ =========== ============
		nginx.exe                     1680 Console                    1     14,188 K
		nginx.exe                    14920 Console                    1     14,356 K

5. Run server with 2 mongos:
mlaunch init --sharded 1 --replicaset --mongos 2 --setParameter enableTestCommands=1
where mongos will be launched on 27017 and 27018 ports.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Bindings/CoreServerSessionPool.cs
+++ b/src/MongoDB.Driver.Core/Core/Bindings/CoreServerSessionPool.cs
@@ -95,15 +95,21 @@ namespace MongoDB.Driver
         private bool IsAboutToExpire(ICoreServerSession session)
         {
             var logicalSessionTimeout = _cluster.Description.LogicalSessionTimeout;
-            if (!session.LastUsedAt.HasValue || !logicalSessionTimeout.HasValue)
+            var clusterType = _cluster.Description.Type;
+
+            if (clusterType == ClusterType.LoadBalanced)
             {
-                return true;
+                return false;  // sessions never expire in load balancing mode
             }
-            else
+            else if (session.LastUsedAt.HasValue && logicalSessionTimeout.HasValue)
             {
                 var expiresAt = session.LastUsedAt.Value + logicalSessionTimeout.Value;
                 var timeRemaining = expiresAt - DateTime.UtcNow;
                 return timeRemaining < TimeSpan.FromMinutes(1);
+            }
+            else
+            {
+                return true;
             }
         }
 
--- a/src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
+++ b/src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
@@ -1,0 +1,333 @@
+﻿/* Copyright 2021-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Bindings;
+using MongoDB.Driver.Core.Clusters.ServerSelectors;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.Events;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.Servers;
+using MongoDB.Libmongocrypt;
+
+namespace MongoDB.Driver.Core.Clusters
+{
+    /// <summary>
+    /// Represents the cluster that uses loadbalancing mode.
+    /// </summary>
+    internal class LoadBalancedCluster : ICluster, IDnsMonitoringCluster
+    {
+        private readonly IClusterClock _clusterClock;
+        private readonly ClusterId _clusterId;
+        private readonly ClusterType _clusterType = ClusterType.LoadBalanced;
+        private CryptClient _cryptClient = null;
+        private ClusterDescription _description;
+        private readonly IDnsMonitorFactory _dnsMonitorFactory;
+        private Thread _dnsMonitorThread;
+        private readonly CancellationTokenSource _dnsMonitorCancellationTokenSource;
+        private readonly IEventSubscriber _eventSubscriber;
+        private IClusterableServer _server;
+        private readonly IClusterableServerFactory _serverFactory;
+        private readonly TaskCompletionSource<bool> _serverReadyTaskCompletionSource;
+        private readonly ICoreServerSessionPool _serverSessionPool;
+        private readonly ClusterSettings _settings;
+        private readonly InterlockedInt32 _state;
+
+        private readonly Action<ClusterClosingEvent> _closingEventHandler;
+        private readonly Action<ClusterClosedEvent> _closedEventHandler;
+        private readonly Action<ClusterOpeningEvent> _openingEventHandler;
+        private readonly Action<ClusterOpenedEvent> _openedEventHandler;
+        private readonly Action<ClusterDescriptionChangedEvent> _descriptionChangedEventHandler;
+
+        public LoadBalancedCluster(
+            ClusterSettings settings,
+            IClusterableServerFactory serverFactory,
+            IEventSubscriber eventSubscriber)
+            : this(
+                  settings,
+                  serverFactory,
+                  eventSubscriber,
+                  dnsMonitorFactory: new DnsMonitorFactory(new EventAggregator())) // should not trigger any events
+        {
+        }
+
+        public LoadBalancedCluster(
+            ClusterSettings settings,
+            IClusterableServerFactory serverFactory,
+            IEventSubscriber eventSubscriber,
+            IDnsMonitorFactory dnsMonitorFactory)
+        {
+#pragma warning disable CS0618 // Type or member is obsolete
+            Ensure.That(settings.ConnectionModeSwitch != ConnectionModeSwitch.UseConnectionMode, $"{nameof(ConnectionModeSwitch.UseConnectionMode)} must not be used for a {nameof(LoadBalancedCluster)}.");
+            if (settings.ConnectionModeSwitch == ConnectionModeSwitch.UseDirectConnection)
+            {
+                Ensure.That(!settings.DirectConnection.GetValueOrDefault(), $"DirectConnection mode is not supported for {nameof(LoadBalancedCluster)}.");
+            }
+#pragma warning restore CS0618 // Type or member is obsolete
+            Ensure.That(settings.LoadBalanced, $"Only Load balanced mode is supported for a {nameof(LoadBalancedCluster)}.");
+
+            Ensure.IsEqualTo(settings.EndPoints.Count, 1, nameof(settings.EndPoints.Count));
+            Ensure.IsNull(settings.ReplicaSetName, nameof(settings.ReplicaSetName));
+
+            _clusterClock = new ClusterClock();
+            _clusterId = new ClusterId();
+
+            _dnsMonitorCancellationTokenSource = new CancellationTokenSource();
+            _dnsMonitorFactory = Ensure.IsNotNull(dnsMonitorFactory, nameof(dnsMonitorFactory));
+            _settings = Ensure.IsNotNull(settings, nameof(settings));
+
+            _serverFactory = Ensure.IsNotNull(serverFactory, nameof(serverFactory));
+            _serverReadyTaskCompletionSource = new TaskCompletionSource<bool>();
+
+            _serverSessionPool = new CoreServerSessionPool(this);
+
+            _state = new InterlockedInt32(State.Initial);
+
+            _description = ClusterDescription.CreateInitial(
+                _clusterId,
+#pragma warning disable CS0618 // Type or member is obsolete
+                ClusterConnectionMode.Automatic,
+                ConnectionModeSwitch.UseConnectionMode,
+#pragma warning restore CS0618 // Type or member is obsolete
+                null);
+
+            _eventSubscriber = eventSubscriber;
+            eventSubscriber.TryGetEventHandler(out _closingEventHandler);
+            eventSubscriber.TryGetEventHandler(out _closedEventHandler);
+            eventSubscriber.TryGetEventHandler(out _openingEventHandler);
+            eventSubscriber.TryGetEventHandler(out _openedEventHandler);
+            eventSubscriber.TryGetEventHandler(out _descriptionChangedEventHandler);
+        }
+
+        public ClusterId ClusterId => _clusterId;
+        public CryptClient CryptClient => _cryptClient;
+
+        public ClusterDescription Description => _description;
+
+        public ClusterSettings Settings => _settings;
+
+        public event EventHandler<ClusterDescriptionChangedEventArgs> DescriptionChanged;
+
+        // public methods
+        public ICoreServerSession AcquireServerSession()
+        {
+            ThrowIfDisposed();
+            return _serverSessionPool.AcquireSession();
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (_state.TryChange(State.Disposed))
+            {
+                if (disposing)
+                {
+                    _dnsMonitorCancellationTokenSource.Cancel();
+                    _dnsMonitorCancellationTokenSource.Dispose();
+
+                    _closingEventHandler?.Invoke(new ClusterClosingEvent(ClusterId));
+                    var stopwatch = Stopwatch.StartNew();
+                    if (_server != null)
+                    {
+                        _server.DescriptionChanged -= ServerDescriptionChangedHandler;
+                        _server.Dispose();
+                    }
+                    _closedEventHandler?.Invoke(new ClusterClosedEvent(ClusterId, stopwatch.Elapsed));
+                }
+            }
+        }
+
+        public void Initialize()
+        {
+            ThrowIfDisposed();
+
+            if (_state.TryChange(State.Initial, State.Open))
+            {
+                var stopwatch = Stopwatch.StartNew();
+                _openingEventHandler?.Invoke(new ClusterOpeningEvent(ClusterId, Settings));
+
+                if (_settings.KmsProviders != null || _settings.SchemaMap != null)
+                {
+                    _cryptClient = CryptClientCreator.CreateCryptClient(_settings.KmsProviders, _settings.SchemaMap);
+                }
+
+                var endPoint = _settings.EndPoints.Single();
+                if (_settings.Scheme != ConnectionStringScheme.MongoDBPlusSrv)
+                {
+                    _server = CreateInitializedServer(endPoint);
+                }
+                else
+                {
+                    // _server will be created after srv resolving
+                    var dnsEndPoint = (DnsEndPoint)endPoint;
+                    var lookupDomainName = dnsEndPoint.Host;
+                    var monitor = _dnsMonitorFactory.CreateDnsMonitor(this, lookupDomainName, _dnsMonitorCancellationTokenSource.Token);
+                    _dnsMonitorThread = monitor.Start();
+                }
+
+                _openedEventHandler?.Invoke(new ClusterOpenedEvent(ClusterId, Settings, stopwatch.Elapsed));
+            }
+        }
+
+        public IServer SelectServer(IServerSelector _, CancellationToken cancellationToken)
+        {
+            ThrowIfDisposed();
+
+            var timeoutTask = Task.Delay(_settings.ServerSelectionTimeout, cancellationToken);
+            var index = Task.WaitAny(_serverReadyTaskCompletionSource.Task, timeoutTask);
+            if (index != 0)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                throw CreateTimeoutException(_description); // _description will contain dnsException
+            }
+
+            return
+                _server ??
+                throw new InvalidOperationException("The server must be created before usage."); // should not be reached
+        }
+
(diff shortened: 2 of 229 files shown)


src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -389,5 +390,16 @@ private void ThrowIfNotOpen()
...
             }
         }
+
+        public void Clear(ObjectId serviceId)
+        {
+            ThrowIfNotOpen();
+
+            _clearingEventHandler?.Invoke(new ConnectionPoolClearingEvent(_serverId, _settings));
+
+            // TODO: temporary reverted
```
DmitryLukyanov, line 400: this will be added in the scope of CMAP changes

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -309,7 +311,9 @@ private async Task OpenHelperAsync(CancellationToken cancellationToken)
                 helper.OpeningConnection();
                 _stream = await _streamFactory.CreateStreamAsync(_endPoint, cancellationToken).ConfigureAwait(false);
                 helper.InitializingConnection();
-                _description = await _connectionInitializer.InitializeConnectionAsync(this, cancellationToken).ConfigureAwait(false);
+                _description = await _connectionInitializer.HandshakeAsync(this, cancellationToken).ConfigureAwait(false);
```
DmitryLukyanov, line 314: we need to know serviceId even though the connection fails, so I split this logic into 2 parts: handshake itself and authentication (+getLastError)

src/MongoDB.Driver.Core/Core/Connections/IsMasterResult.cs
```diff
@@ -319,10 +319,30 @@ public ServerType ServerType
                     return ServerType.ShardRouter;
                 }
 
+                if (ServiceId != null)
+                {
+                    return ServerType.LoadBalanced; // TODO: change when Service Id will be supported by server
```
DmitryLukyanov, line 324: this should be updated according to the discussion we had on today meeting with Jeff (this should be filled only for LB)

src/MongoDB.Driver.Core/Core/Servers/DefaultServer.cs
```diff
@@ -0,0 +1,281 @@
...
+            if (__invalidatingExceptions.Contains(exception.GetType()))
+            {
+                invalidatingResponseTopologyVersion = null;
+                return true;
+            }
+
+            var exceptionsToCheck = new[]
+            {
+                exception as MongoCommandException,
+                (exception as MongoWriteConcernException)?.MappedWriteConcernResultException
```
DmitryLukyanov, line 246: this line corresponds to these lines in the master: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Servers/Server.cs#L526-L542. In 2 words, we just make the same logic as in ExceptionMapper only because the actual exception can be hidden inside `MongoWriteConcernException.writeConcernError`, I extract it into `MappedWriteConcernResultException` and as a result no reason to analyze exception codes on this level anymore.

However, to save time I'm ok to move this logic into a different ticket

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            // no-opt
+        }
+
+        protected override void HandleBeforeHandshakeCompletesException(Exception ex)
+        {
+            // drivers MUST NOT perform SDAM error handling for any errors that occur before the MongoDB Handshake
+
+            if (ex is MongoAuthenticationException mongoAuthenticationException &&
+                mongoAuthenticationException.ServiceId.HasValue) // this value will be always filled for MongoAuthenticationException, adding this condition just in case
```
DmitryLukyanov, line 70: `mongoAuthenticationException.ServiceId.HasValue` this value should always present for LB mode, maybe it makes sense to trigger exception if for some reason it's null?

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -320,7 +320,7 @@ public void RequestHeartbeat_should_force_another_heartbeat()
             capturedEvents.Any().Should().BeFalse();
         }
 
-        [Fact]
+        [Fact(Skip = "Will check this test")]
```
DmitryLukyanov, line 323: this is failing, I will investigate it now

src/MongoDB.Driver.Core/Core/Bindings/CoreServerSessionPool.cs
```diff
@@ -95,16 +95,22 @@ public void ReleaseSession(ICoreServerSession session)
         private bool IsAboutToExpire(ICoreServerSession session)
         {
             var logicalSessionTimeout = _cluster.Description.LogicalSessionTimeout;
-            if (!session.LastUsedAt.HasValue || !logicalSessionTimeout.HasValue)
+            var clusterType = _cluster.Description.Type;
```
DmitryLukyanov, line 98: related to sessions expiring

src/MongoDB.Driver.Core/Core/Clusters/MultiServerCluster.cs
```diff
@@ -66,23 +66,15 @@ internal sealed class MultiServerCluster : Cluster, IDnsMonitoringCluster
 #pragma warning disable CS0618 // Type or member is obsolete
             if (settings.ConnectionModeSwitch == ConnectionModeSwitch.UseDirectConnection)
             {
-                if (settings.DirectConnection.GetValueOrDefault())
-                {
-                    throw new ArgumentException("DirectConnection is not supported for a MultiServerCluster.");
-                }
+                Ensure.That(!settings.DirectConnection.GetValueOrDefault(), "DirectConnection is not supported for a MultiServerCluster.");
```
rstam, line 69: Will never throw because base class constructor would have already thrown.

src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
```diff
@@ -0,0 +1,329 @@
...
+                  eventSubscriber,
+                  dnsMonitorFactory: new DnsMonitorFactory(new EventAggregator())) // should not trigger any events
+        {
+        }
+
+        public LoadBalancedCluster(
+            ClusterSettings settings,
+            IClusterableServerFactory serverFactory,
+            IEventSubscriber eventSubscriber,
+            IDnsMonitorFactory dnsMonitorFactory)
```
rstam, line 76: I thought it was a bug that this class was NOT calling the base class constructor.

And then I discovered to my total surprise that `LoadBalancedCluster`  is not a subclass of `Cluster`.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -283,18 +283,27 @@ public Task OpenAsync(CancellationToken cancellationToken)
...
 
                 helper.OpenedConnection();
             }
             catch (Exception ex)
             {
+                if (ShouldSetHandshakeDescription(_description, handshakeDescription))
+                {
+                    // if we have successful handshake description, we should propogate it to upper levels
+                    _description = handshakeDescription;
+                }
```
rstam, line 306: I think lines 302-306 can be more simply written as:

                _description ??= handshakeDescription;

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -304,18 +313,27 @@ private void OpenHelper(CancellationToken cancellationToken)
...
```
rstam, line 336: Same comment as above.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -709,6 +727,9 @@ private bool AnyMessageNeedsToBeCompressed(IEnumerable<RequestMessage> messages)
             compressedMessageEncoder.WriteMessage(compressedMessage);
         }
 
+        private bool ShouldSetHandshakeDescription(ConnectionDescription connectionDescription, ConnectionDescription handshakeDescription) =>
+            connectionDescription == null && handshakeDescription != null;
+
```
rstam, line 732: This method is probably not necessary. See comments above.

src/MongoDB.Driver.Core/Core/Connections/IConnectionInitializer.cs
```diff
@@ -13,16 +13,17 @@
...
 using System.Threading.Tasks;
-using MongoDB.Driver.Core.Servers;
 
 namespace MongoDB.Driver.Core.Connections
 {
     internal interface IConnectionInitializer
     {
-        ConnectionDescription InitializeConnection(IConnection connection, CancellationToken cancellationToken);
-        Task<ConnectionDescription> InitializeConnectionAsync(IConnection connection, CancellationToken cancellationToken);
+        ConnectionDescription Handshake(IConnection connection, CancellationToken cancellationToken);
```
rstam, line 23: Method names should be verbs and `Handshake` is a noun. Perhaps `Initialize` or  `SendHello`?

src/MongoDB.Driver.Core/Core/Connections/IConnectionInitializer.cs
```diff
@@ -13,16 +13,17 @@
...
+        ConnectionDescription ConnectionAuthentication(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken);
```
rstam, line 24: Method names should be verbs and `ConnectionAuthentication` is a noun. How about more simply `Authenticate`?

src/MongoDB.Driver.Core/Core/Connections/IConnectionInitializer.cs
```diff
@@ -13,16 +13,17 @@
...
+
+        Task<ConnectionDescription> HandshakeAsync(IConnection connection, CancellationToken cancellationToken);
+        Task<ConnectionDescription> ConnectionAuthenticationAsync(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken);
```
rstam, line 27: Alphabetical order?

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -0,0 +1,122 @@
...
+                  serverApi)
+        {
+            _baseDescription = _currentDescription = new ServerDescription(ServerId, endPoint, reasonChanged: "ServerInitialDescription");
+        }
+
+        public override ServerDescription Description => Interlocked.CompareExchange(ref _currentDescription, value: null, comparand: null);
+
+        protected override void Dispose(bool disposing)
+        {
+            // no-opt
```
rstam, line 62: Change `no-opt` to `no-op` everywhere

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -211,8 +192,7 @@ public void Initialize()
 
                 var stopwatch = Stopwatch.StartNew();
                 _connectionPool.Initialize();
-                _monitor.DescriptionChanged += OnMonitorDescriptionChanged;
-                _monitor.Initialize();
+                Initializing();
```
rstam, line 195: This method name is confusing... what's the difference between `Initialize` and `Initializing`. They are basically synonyms.

How about `InitializeSubClass`?

src/MongoDB.Driver.Core/Core/Servers/Server.cs
```diff
@@ -261,311 +240,53 @@ private void OnDescriptionChanged(object sender, ServerDescriptionChangedEventAr
...
+                return ex is OperationCanceledException;
             }
+
+            Exception GetEffectiveException(Exception ex) =>
+                ex is AggregateException aggregateException && aggregateException.InnerExceptions.Count == 1
+                    ? aggregateException.InnerException
+                    : ex;
         }
 
+        private bool IsOpened() => _state.Value == State.Open;
```
rstam, line 288: Rename to `IsOpen`

The `ed` ending is not needed.

src/MongoDB.Driver.Core/Core/Servers/ServerFactory.cs
```diff
@@ -63,20 +63,32 @@ internal class ServerFactory : IClusterableServerFactory
...
+                        clusterClock,
+                        _clusterConnectionMode,
+                        _connectionModeSwitch,
+                        _directConnection,
+                        _settings,
+                        endPoint,
+                        _connectionPoolFactory,
+                        _serverMonitorFactory,
+                        _eventSubscriber,
+                        _serverApi),
```
rstam, line 91: Remove redundant comma at end of line

src/MongoDB.Driver.Core/MongoWriteConcernException.cs
```diff
@@ -43,9 +43,29 @@ private static void AddErrorLabelsFromWriteConcernResult(MongoWriteConcernExcept
                 }
             }
         }
+
+        private static bool TryMapWriteConcernResultToException(ConnectionId connectionId, WriteConcernResult writeConcernResult, out Exception mappedException)
+        {
+            mappedException = null;
```
rstam, line 49: Move this assignment to immediately before `return false` line 63.

Not setting and out parameter to `null` at the top of the method allows the compiler to give us an error message if we forget to assign a value to an out parameter before returning `true`.

rstam (changes requested):
Overall looks good with minor changes requested.

The only file I had a lot of trouble verifying is `DefaultServer.cs`. I expected it to have only minor differences from the original `Server.cs` but it has major differences. I think they all look OK but my confidence level is low.

src/MongoDB.Driver.Core/Core/Clusters/SingleServerCluster.cs
```diff
@@ -49,6 +49,8 @@ internal SingleServerCluster(ClusterSettings settings, IClusterableServerFactory
             Ensure.IsEqualTo(settings.EndPoints.Count, 1, "settings.EndPoints.Count");
             _replicaSetName = settings.ReplicaSetName;  // can be null
 
+            Ensure.That(!settings.LoadBalanced, $"Load balanced mode is not supported for a {nameof(SingleServerCluster)}.");
```
DmitryLukyanov: also removed

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Servers/Server.cs
DmitryLukyanov: the second one is already in progress :), `InitializeSubClass` sounds good to me, done

src/MongoDB.Driver.Core/Core/Servers/Server.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Servers/ServerFactory.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/MongoWriteConcernException.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -304,18 +310,24 @@ private void OpenHelper(CancellationToken cancellationToken)
...
+                // we always need to have access to description after handshake regardless furher errors
+                _description = await _connectionInitializer.AuthenticateAsync(this, handshakeDescription, cancellationToken).ConfigureAwait(false);
                 _sendCompressorType = ChooseSendCompressorTypeIfAny(_description);
 
                 helper.OpenedConnection();
             }
             catch (Exception ex)
             {
+                // if we have successful handshake description, we should propogate it to upper levels
+                _description = handshakeDescription;
```
rstam, line 330: This is slightly different that what I had suggested:

    _description ??= handshakeDescription;

They might be the same thing, but only if you guarantee that no further exceptions are thrown after the assignment to `_description` in the `try` block.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -283,18 +283,24 @@ public Task OpenAsync(CancellationToken cancellationToken)
...
+
+            ConnectionDescription handshakeDescription = null;
             try
             {
                 helper.OpeningConnection();
                 _stream = _streamFactory.CreateStream(_endPoint, cancellationToken);
                 helper.InitializingConnection();
-                _description = _connectionInitializer.InitializeConnection(this, cancellationToken);
+                handshakeDescription = _connectionInitializer.SendHello(this, cancellationToken);
+                // we always need to have access to description after handshake regardless furher errors
```
rstam, line 294: Can we just remove this comment? It doesn't make sense to me.

If you insist on keeping it we will have to work on rewording it.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -283,18 +283,24 @@ public Task OpenAsync(CancellationToken cancellationToken)
...
+                _description = _connectionInitializer.Authenticate(this, handshakeDescription, cancellationToken);
                 _sendCompressorType = ChooseSendCompressorTypeIfAny(_description);
 
                 helper.OpenedConnection();
             }
             catch (Exception ex)
             {
+                // if we have successful handshake description, we should propogate it to upper levels
```
rstam, line 302: Can we just remove this comment? It doesn't make sense to me.

If you insist on keeping it we will have to work on rewording it.

src/MongoDB.Driver.Core/Core/Connections/IConnectionInitializer.cs
```diff
@@ -13,16 +13,16 @@
...
 namespace MongoDB.Driver.Core.Connections
 {
     internal interface IConnectionInitializer
     {
-        ConnectionDescription InitializeConnection(IConnection connection, CancellationToken cancellationToken);
-        Task<ConnectionDescription> InitializeConnectionAsync(IConnection connection, CancellationToken cancellationToken);
+        ConnectionDescription Authenticate(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken);
+        Task<ConnectionDescription> AuthenticateAsync(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken);
+        ConnectionDescription SendHello(IConnection connection, CancellationToken cancellationToken);
+        Task<ConnectionDescription> SendHelloAsync(IConnection connection, CancellationToken cancellationToken);
```
rstam, line 26: Note: I am unfamiliar with the use of "handshake" as a verb. I don't think it's very common. But using `SendHello` avoids the question of whether "handshake" can be used as a verb or not.

For example, this dictionary defines it as a noun ony:

https://www.dictionary.com/browse/handshake

rstam (changes requested):
A few minor questions remain. Almost ready to approve.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
DmitryLukyanov: my mistake, fixed

DmitryLukyanov (commented):
ready for review

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
rstam: Pending

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -304,18 +309,23 @@ private void OpenHelper(CancellationToken cancellationToken)
...
+
+            ConnectionDescription handshakeDescription = null;
             try
             {
                 helper.OpeningConnection();
                 _stream = await _streamFactory.CreateStreamAsync(_endPoint, cancellationToken).ConfigureAwait(false);
                 helper.InitializingConnection();
-                _description = await _connectionInitializer.InitializeConnectionAsync(this, cancellationToken).ConfigureAwait(false);
+                handshakeDescription = await _connectionInitializer.SendHelloAsync(this, cancellationToken).ConfigureAwait(false);
+                // we always need to have access to description after handshake regardless furher errors
```
rstam, line 320: Can we just remove this comment? It doesn't make sense to me.

If you insist on keeping it we will have to work on rewording it.

rstam (changes requested):
There are still some pending comments.

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
DmitryLukyanov: I thought you meant only comments in the catch block, to save time I'm ok to remove this also, done

rstam (approved):
LGTM

PR #563, opened 2021-07-06
CSHARP-3714: Implement missing test from Versioned API specification

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3714
EG: https://evergreen.mongodb.com/version/60e63dec2a60ed546b45f97e

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilderExtensions.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilderExtensions.cs
@@ -235,6 +235,10 @@ namespace MongoDB.Driver.Core.Configuration
             {
                 builder = builder.ConfigureCluster(s => s.With(serverSelectionTimeout: connectionString.ServerSelectionTimeout.Value));
             }
+            if (serverApi != null)
+            {
+                builder = builder.ConfigureCluster(s => s.With(serverApi: serverApi));
+            }
 
             return builder;
         }
--- a/tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
+++ b/tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
@@ -108,7 +108,7 @@ namespace MongoDB.Driver
             }
 
             builder = builder
-                .ConfigureWithConnectionString(__connectionString.Value)
+                .ConfigureWithConnectionString(__connectionString.Value, GetServerApi())
                 .ConfigureCluster(c => c.With(serverSelectionTimeout: TimeSpan.FromMilliseconds(int.Parse(serverSelectionTimeoutString))));
 
             if (__connectionString.Value.Tls.HasValue &&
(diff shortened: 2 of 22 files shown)


src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilderExtensions.cs
```diff
@@ -235,6 +235,10 @@ public static ClusterBuilder ConfigureWithConnectionString(this ClusterBuilder b
             {
                 builder = builder.ConfigureCluster(s => s.With(serverSelectionTimeout: connectionString.ServerSelectionTimeout.Value));
             }
+            if (serverApi != null)
```
MikalaiMazurenka, line 238: This is the only driver code change. I noticed that this method applies `serverApi` only to auth mechanisms, but should apply it to cluster builder as well.

tests/MongoDB.Driver.Core.Tests/Core/Operations/BulkMixedWriteOperationTests.cs
```diff
@@ -1453,7 +1453,34 @@ public void WriteConcern_should_work()
         [InlineData(new[] { 10000000, 10000000, 10000000, 10000000, 7999888 }, new[] { 4, 1 })]
         public void Execute_with_multiple_deletes_should_split_batches_as_expected_when_using_write_commands_via_opmessage(int[] requestSizes, int[] expectedBatchCounts)
         {
-            RequireServer.Check().Supports(Feature.CommandMessage);
+            RequireServer.Check().Supports(Feature.CommandMessage).RequireApiVersion(false);
```
MikalaiMazurenka, line 1456: The batch splitting depends on the current position in the buffer on the start of batch size calculation and `apiVersion` field wasn't initially expected here, so when present it shift the buffer position by 18, which I took into account by splitting tests with `serverApi` and without `serverApi`.

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilderExtensions.cs
jyemin: What is the effect of not having this?  What doesn't work?

MikalaiMazurenka: Without this change if user want to configure `ClusterBuilder` with `serverApi` and connection string they would write:
```
var builder = new ClusterBuilder().ConfigureWithConnectionString(connectionString, serverApi);
```
But only authentication mechanism will be configured with `serverApi` and the cluster itself will have `serverApi` set to `null`. So the complete configuration will need one more call:
```
var builder = new ClusterBuilder().ConfigureWithConnectionString(connectionString, serverApi).ConfigureCluster(c => c.With(serverApi: serverApi));
```
Which I stumbled upon myself, thinking that I implemented `ConfigureWithConnectionString` to configure `serverApi` for all aspects of cluster.
So while it's debatable whether this is a bug, without this change the `ConfigureWithConnectionString` method behavior is at least unintuitive.

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -293,6 +293,31 @@ private static BuildInfoResult RunBuildInfo()
...
 
+        public static bool RequireApiVersion
+        {
+            get
+            {
+                return Environment.GetEnvironmentVariable("MONGODB_API_VERSION") != null;
+            }
+        }
+
+        public static ServerApi GetServerApi()
```
JamesKovacs, line 304: This should be a lazy static property like `Cluster`, `ConnectionString`, etc.

JamesKovacs (changes requested):
Overall LGTM. Minor changes requested.

tests/MongoDB.Driver.Core.Tests/Core/Operations/BulkMixedWriteOperationTests.cs
MikalaiMazurenka: Great point!
Done

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
MikalaiMazurenka: I generally think this class should be reimplemented. I do not think that overcomplicating the code with excessive usage of statics is worth the tiny performance savings. But agree that this PR is not the point to start major test code changes.
Done

MikalaiMazurenka (commented):
Ready for review!

JamesKovacs (approved):
LGTM

PR #564, opened 2021-07-06
CSHARP-3673: Update connection pinning logic for LB mode.

DmitryLukyanov:


src/MongoDB.Driver.Core/Core/Authentication/SaslAuthenticator.cs
```diff
@@ -103,7 +103,7 @@ public void Authenticate(IConnection connection, ConnectionDescription descripti
                         var protocol = CreateCommandProtocol(command);
                         result = protocol.Execute(connection, cancellationToken);
                     }
-                    catch (MongoCommandException ex)
+                    catch (MongoException ex)
```
DmitryLukyanov, line 106: in the current test suite we need to wrap MongoConnectionException into MongoAuthException. Maybe it makes sense to wrap all errors here?

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -357,6 +359,7 @@ public void Dispose()
                     }
                 }
 
+                _currentTransaction?.UnpinAll();
```
DmitryLukyanov, line 362: previously we didn't need this because we didn't need to dispose the pinned server

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -236,12 +236,12 @@ public void Open(CancellationToken cancellationToken)
                 try
                 {
                     _connection.Open(cancellationToken);
+                    SetEffectiveGenerationIfRequired(_connection.Description);
```
DmitryLukyanov, line 239: we need to update the current connection generation based on connectionsState and serviceId from the helloResponse

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -251,40 +251,88 @@ public async Task OpenAsync(CancellationToken cancellationToken)
...
+                }
             }
 
-            public Task<ResponseMessage> ReceiveMessageAsync(int responseTo, IMessageEncoderSelector encoderSelector, MessageEncoderSettings messageEncoderSettings, CancellationToken cancellationToken)
+            public async Task<ResponseMessage> ReceiveMessageAsync(int responseTo, IMessageEncoderSelector encoderSelector, MessageEncoderSettings messageEncoderSettings, CancellationToken cancellationToken)
             {
-                return _connection.ReceiveMessageAsync(responseTo, encoderSelector, messageEncoderSettings, cancellationToken);
+                try
+                {
+                    return await _connection.ReceiveMessageAsync(responseTo, encoderSelector, messageEncoderSettings, cancellationToken).ConfigureAwait(false);
```
DmitryLukyanov, line 281: `await` needs to catch exception n this method

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -260,6 +273,19 @@ public void Dispose()
             }
         }
 
+        public int GetConnectionPoolGenerationForConnection(ConnectionDescription description)
```
DmitryLukyanov, line 276: I would suggest deprecate Clear method and Generation property and instead use this method and a new method like `ClearPoolForAffectedConnections` (or similar). This will allow us handle serviceId only on the pool level. But since Generation and Clear are still described in the spec, I'm ok to move this discussion to later (or maybe we can just don't care about it)

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
```diff
@@ -414,8 +416,11 @@ private Task ExecuteKillCursorsProtocolAsync(IChannelHandle channel, Cancellatio
         /// <inheritdoc/>
         public void Dispose()
         {
-            Dispose(true);
-            GC.SuppressFinalize(this);
+            if (!TryTerminationPendingMode())
```
DmitryLukyanov, line 419: if getMore is inProgress, we need to defer closing until getMore will be finished

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
```diff
@@ -441,13 +446,20 @@ protected virtual void Dispose(bool disposing)
...
+            if (!TryTerminationPendingMode() && !_closed)
             {
                 try
                 {
                     if (_cursorId != 0)
                     {
-                        KillCursors(cancellationToken);
+                        try
+                        {
+                            KillCursors(cancellationToken);
```
DmitryLukyanov, line 457: the spec says that we "MAY" protect killCursors against exception, but the current tests require this behavior

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
```diff
@@ -543,13 +572,16 @@ private void KillCursors(CancellationToken cancellationToken)
             using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(10)))
             using (var channel = _channelSource.GetChannel(cancellationTokenSource.Token))
             {
-                if (Feature.KillCursorsCommand.IsSupported(channel.ConnectionDescription.ServerVersion))
-                {
-                    ExecuteKillCursorsCommand(channel, cancellationToken);
-                }
-                else
+                if (!channel.Connection.IsExpired)
```
DmitryLukyanov, line 575: some tests require that if a connection isExpired, we should not try to run killCursors

tests/MongoDB.Driver.Core.Tests/LoadBalancingIntergationTests.cs
```diff
@@ -0,0 +1,672 @@
...
+                    }
+                    AssertCommand(eventCapturer, "abortTransaction", onlySingleCommand: false);
+                    AssertCheckInOnlyEvents(eventCapturer);
+                    AssertSessionReferenceCount(session, 0);
+                    AssertChannelReferenceCount(readBindingsBundle.RetryableReadContext.Channel, 0);
+                }
+            }
+        }
+
+        [SkippableTheory(Skip = "Investigate")]
```
DmitryLukyanov, line 347: this tests fails (for attempts > 1), because each new attempt increase referenceCounter and don't dispose enough of them, then when we got finishing like abortTransaction, the referenceCounter is bigger than expected so, actual resources like channel is not disposed

tests/MongoDB.Driver.Tests/Specifications/load-balancers/LoadBalancersTestRunner.cs
```diff
@@ -0,0 +1,88 @@
...
+    [Trait("Category", "SupportLoadBalancing")]
+    public sealed class LoadBalancersTestRunner
+    {
+        [SkippableTheory]
+        [ClassData(typeof(TestCaseFactory))]
+        public void Run(JsonDrivenTestCase testCase)
+        {
+            if (testCase.Name.Contains("wait-queue-timeouts"))
+            {
+                throw new SkipException("Not implemented yet.");
```
DmitryLukyanov, line 37: we decided to temporary disable these tests

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedErrorMatcher.cs
```diff
@@ -79,7 +79,11 @@ private void AssertErrorCodeName(Exception actualException, string expectedError
 
         private void AssertErrorContains(Exception actualException, string expectedSubstring)
         {
-            actualException.Message.Should().ContainEquivalentOf(expectedSubstring);
+            actualException
```
DmitryLukyanov, line 82: this change is related to this exception: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Connections/ConnectionInitializer.cs#L122. That will be wrapped in the next step into `MongoConnectionException`. Maybe it makes sense to consider all innerExceptions and not only the first one

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -140,7 +140,7 @@ public void AbortTransaction(CancellationToken cancellationToken = default(Cance
                 catch (Exception exception) when (ShouldRetryEndTransactionException(exception))
                 {
                     // unpin if retryable error
-                    _currentTransaction.PinnedServer = null;
+                    _currentTransaction.UnpinAll();
```
DmitryLukyanov, line 143: almost copy of mongos behavior for transactions

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -380,6 +383,7 @@ public void StartTransaction(TransactionOptions transactionOptions = null)
                 throw new InvalidOperationException("Transactions do not support unacknowledged write concerns.");
             }
 
+            _currentTransaction?.UnpinAll(); // unpin data if any when a new transaction is started
```
DmitryLukyanov, line 386: we didn't need this before, because we didn't need to dispose a pinned server

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <param name="readPreference">The read preference.</param>
+        /// <returns>An effective read binging.</returns>
+        public static IReadBindingHandle CreateEffectiveReadBinding(ICluster cluster, ICoreSessionHandle session, ReadPreference readPreference)
+        {
+            IReadBinding readBinding;
+            if (session.IsInTransaction &&
+                IsConnectionPinned(session.CurrentTransaction) &&
+                session.CurrentTransaction.State != CoreTransactionState.Starting)
+            {
+                readBinding = new ChannelReadWriteBinding(
```
DmitryLukyanov, line 44: use a pinned connection/server if possible

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            return new ReadWriteBindingHandle(readWriteBinding);
+        }
+
+        internal static IChannelSourceHandle CreateEffectiveGetMoreChannelSource(IChannelSourceHandle channelSource, long cursorId)
+        {
+            IChannelSource effectiveChannelSource;
+            if (IsInLoadBalancedMode(channelSource.ServerDescription) && cursorId != 0)
+            {
+                var getMoreChannel = channelSource.GetChannel(CancellationToken.None); // no need for cancellation token since we already have channel in the source
```
DmitryLukyanov, line 98: `channelSource.GetChannel(CancellationToken.None)` effectively means `channel.Fork()` for this case

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,174 @@
...
+                var forkedSession = session.Fork();
+
+                var pinnedChannelSource = new ChannelSourceHandle(
+                    new ChannelChannelSource(
+                        server,
+                        forkedChannel, 
+                        forkedSession));
+
+                if (session.IsInTransaction)
+                {
```
DmitryLukyanov, line 134: Since we fork the channel here it won't be disposed after cursor will be disposed

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -251,40 +251,88 @@ public async Task OpenAsync(CancellationToken cancellationToken)
...
             public void SetReadTimeout(TimeSpan timeout)
             {
                 _connection.SetReadTimeout(timeout);
             }
+
+            // private methods
+            private void EnrichExceptionDetails(MongoConnectionException ex)
+            {
+                // should be refactored in CSHARP-3720
+                ex.Generation = _generation;
```
DmitryLukyanov, line 325: _generation is already up-to-date in the previous step

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,174 @@
...
+                session.CurrentTransaction.State != CoreTransactionState.Starting)
+            {
+                readBinding = new ChannelReadWriteBinding(
+                    session.CurrentTransaction.PinnedServer,
+                    session.CurrentTransaction.PinnedChannel,
+                    session);
+            }
+            else
+            {
+                if (IsInLoadBalancedMode(cluster.Description) && IsConnectionPinned(session.CurrentTransaction))
```
DmitryLukyanov, line 51: this logic is mostly coping from here: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Clusters/IClusterExtensions.cs#L27

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+                var forkedChannel = channel.Fork();
+                var forkedSession = session.Fork();
+
+                var pinnedChannelSource = new ChannelSourceHandle(
+                    new ChannelChannelSource(
+                        server,
+                        forkedChannel,
+                        forkedSession));
+
+                if (session.IsInTransaction && !IsConnectionPinned(session.CurrentTransaction))
```
DmitryLukyanov, line 133: `!IsConnectionPinned(session.CurrentTransaction)` - a new logic, if we already pinned connection, no need to pin it again, it also allowed to remove a bit "hacky" code in the EndTransactionOperation.

src/MongoDB.Driver.Core/Core/Operations/RetryableReadContext.cs
```diff
@@ -14,15 +14,29 @@
...
 {
+    internal static class RetryableReadContextExtensions
+    {
+        public static void PinConnectionIfRequired(this RetryableReadContext context)
+        {
+            if (ChannelPinningHelper.TryCreatePinnedChannelSourceAndPinChannel(
+                context.ChannelSource,
+                context.Channel,
+                context.Binding.Session,
+                out var pinnedChannel))
```
rstam, line 32: I don't think an `out` combined with a `Tuple` is a good idea. I suggest:

    out var pinnedChannelSource,
    out var pinnedChannel))

src/MongoDB.Driver.Core/Core/Operations/RetryableReadContext.cs
```diff
@@ -14,15 +14,29 @@
...
+            {
+                context.ReplaceChannelSource(pinnedChannel.PinnedChannelSource);
+                context.ReplaceChannel(pinnedChannel.Channel);
```
rstam, line 35: Combined with the previous comment this can read better as:

    context.ReplaceChannelSource(pinnedChannelSource);
    context.ReplaceChannel(pinnedChannel);

src/MongoDB.Driver.Core/Core/Operations/RetryableWriteContext.cs
```diff
@@ -23,6 +23,23 @@
...
+    internal static class RetryableWriteContextExtensions
+    {
+        public static void PinConnectionIfRequired(this RetryableWriteContext context)
+        {
+            if (context.Binding.Session.IsInTransaction &&
+                ChannelPinningHelper.TryCreatePinnedChannelSourceAndPinChannel(
+                context.ChannelSource,
+                context.Channel,
+                context.Binding.Session,
+                out var pinnedChannel))
```
rstam, line 35: I don't think an `out` combined with a `Tuple` is a good idea. I suggest:

    out var pinnedChannelSource,
    out var pinnedChannel))

src/MongoDB.Driver.Core/Core/Operations/RetryableWriteContext.cs
```diff
@@ -23,6 +23,23 @@
...
+            {
+                context.ReplaceChannelSource(pinnedChannel.PinnedChannelSource);
+                context.ReplaceChannel(pinnedChannel.Channel);
```
rstam, line 38: Combined with the previous comment this can read better as:

    context.ReplaceChannelSource(pinnedChannelSource);
    context.ReplaceChannel(pinnedChannel);

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -80,7 +80,7 @@ protected override void HandleAfterHandshakeCompletesException(IConnection conne
             {
                 if (ex is MongoConnectionException mongoConnectionException &&
                     mongoConnectionException.Generation.HasValue &&
-                    mongoConnectionException.Generation.Value != ConnectionPool.Generation)
+                    mongoConnectionException.Generation.Value != ConnectionPool.GetConnectionPoolGenerationForConnection(connection.Description))
```
rstam, line 83: mongoConnectionException.Generation.Value != ConnectionPool.GetGeneration(connection.Description?.ServiceId))

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -481,25 +485,27 @@ private void EnsureTransactionsAreSupported()
...
-                }
-                else
-                {
-                    Feature.Transactions.ThrowIfNotSupported(connectedDataBearingServer.Version);
+                    case ServerType.Standalone:
+                        throw new NotSupportedException("Standalone servers do not support transactions.");
+                    case ServerType.ShardRouter:
+                        Feature.ShardedTransactions.ThrowIfNotSupported(connectedDataBearingServer.Version);
+                        break;
+                    case ServerType.LoadBalanced:
```
rstam, line 495: I don't know if it's too late too change this but it should have been:

    ServerType.LoadBalancer

ending with "er" not "ed"

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+    public static class ChannelPinningHelper
+    {
+        /// <summary>
+        /// Create effective read binding handle.
+        /// </summary>
+        /// <param name="cluster">The cluster,</param>
+        /// <param name="session">The session.</param>
+        /// <param name="readPreference">The read preference.</param>
+        /// <returns>An effective read binging.</returns>
+        public static IReadBindingHandle CreateEffectiveReadBinding(ICluster cluster, ICoreSessionHandle session, ReadPreference readPreference)
```
rstam, line 37: I think `CreateReadBinding` is a good enough name. Not sure "effective" is necessary.

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            IReadBinding readBinding;
+            if (session.IsInTransaction &&
+                IsConnectionPinned(session.CurrentTransaction) &&
+                session.CurrentTransaction.State != CoreTransactionState.Starting)
+            {
+                readBinding = new ChannelReadWriteBinding(
+                    session.CurrentTransaction.PinnedServer,
+                    session.CurrentTransaction.NewPinnedChannelHandleIfConfigured(),
```
rstam, line 46: session.CurrentTransaction.PinnedChannel.Fork(),

Just fork the channel. No need for a `NewPinnedChannelHandleIfConfigured` method

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+            return new ReadBindingHandle(readBinding);
+        }
+
+        /// <summary>
+        /// Create effective readwrite binding handle.
+        /// </summary>
+        /// <param name="cluster">The cluster.</param>
+        /// <param name="session">The session.</param>
+        /// <returns>An effective read write binging.</returns>
+        public static IReadWriteBindingHandle CreateEffectiveReadWriteBinding(ICluster cluster, ICoreSessionHandle session)
```
rstam, line 68: I think `CreateReadWriteBinding` is a good enough name. Not sure "effective" is necessary.

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            IReadWriteBinding readWriteBinding;
+            if (session.IsInTransaction &&
+                IsConnectionPinned(session.CurrentTransaction) &&
+                session.CurrentTransaction.State != CoreTransactionState.Starting)
+            {
+                readWriteBinding = new ChannelReadWriteBinding(
+                    session.CurrentTransaction.PinnedServer,
+                    session.CurrentTransaction.NewPinnedChannelHandleIfConfigured(),
```
rstam, line 77: session.CurrentTransaction.PinnedChannel.Fork(),

Just fork the channel. No need for a NewPinnedChannelHandleIfConfigured method

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+                    // unpin if the next operation is not under transaction
+                    session.CurrentTransaction.UnpinAll();
+                }
+                readWriteBinding = new WritableServerBinding(cluster, session);
+            }
+
+            return new ReadWriteBindingHandle(readWriteBinding);
+        }
+
+        internal static IChannelSourceHandle CreateEffectiveGetMoreChannelSource(IChannelSourceHandle channelSource, long cursorId)
```
rstam, line 93: I think `CreateGetMoreChannelSource` is a good enough name. Not sure "effective" is necessary.

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+            else
+            {
+                effectiveChannelSource = new ServerChannelSource(channelSource.Server, channelSource.Session.Fork());
+            }
+
+            return new ChannelSourceHandle(effectiveChannelSource);
+        }
+
+        internal static bool TryCreatePinnedChannelSourceAndPinChannel(
```
rstam, line 114: I would suggest `PinChannelSourceAndChannelIfRequired` as the name.

Normally `TryXyz` methods return false when they FAIL, not when an action is not required.

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+            IChannelSourceHandle channelSource,
+            IChannelHandle channel,
+            ICoreSessionHandle session,
+            out (IChannelSourceHandle PinnedChannelSource, IChannelHandle Channel) pinnedChannel)
```
rstam, line 118: An `out` with `Tuple` doesn't really enhance usability. Just use two `out` vars:

    out IChannelSourceHandle pinnedChannelSource,
    out IChannelHandle pinnedChannel)

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            pinnedChannel = default;
+            if (IsInLoadBalancedMode(channel.ConnectionDescription))
+            {
+                var server = channelSource.Server;
+                var forkedChannel = channel.Fork();
+                var forkedSession = session.Fork();
```
rstam, line 125: Implementation can be simplified to:

   if (IsInLoadBalancedMode(channel.ConnectionDescription))
   {
       var server = channelSource.Server;

       pinnedChannelSource = new ChannelSourceHandle(
           new ChannelChannelSource(
               server,
               channel.Fork(),
               session.Fork()));

       if (session.IsInTransaction && !IsConnectionPinned(session.CurrentTransaction))
       {
           session.CurrentTransaction.PinConnection(channel.Fork());
           session.CurrentTransaction.PinnedServer = server;
       }

       pinnedChannel = channel.Fork();

       return true;
   }

   pinnedChannelSource = null;
   pinnedChannel = null;
   return false;

src/MongoDB.Driver.Core/Core/ConnectionPools/IConnectionPool.cs
```diff
@@ -73,6 +69,13 @@ public interface IConnectionPool : IDisposable
         /// </summary>
         void Clear(ObjectId serviceId);
 
+        /// <summary>
+        /// Gets the connection pool generation for the connection.
+        /// </summary>
+        /// <param name="description">The connection pool description.</param>
+        /// <returns>A connection pool generation for the connection.</returns>
+        int GetConnectionPoolGenerationForConnection(ConnectionDescription description);
+
```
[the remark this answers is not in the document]
rstam: The argument should be a `serviceId` instead of a connection description.

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
rstam: I did not review this file because you said you were going to revert these changes.

src/MongoDB.Driver.Core/Core/Operations/RetryableReadContext.cs
```diff
@@ -14,15 +14,29 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Driver.Core.Bindings;
 using MongoDB.Driver.Core.Misc;
 
 namespace MongoDB.Driver.Core.Operations
 {
+    internal static class RetryableReadContextExtensions
+    {
+        public static void PinConnectionIfRequired(this RetryableReadContext context)
```
rstam, line 26: As commented elsewhere I think it might be better if Create did the pinning automatically if required.

rstam (changes requested):
Some/most of the changes proposed are prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp3673_v2-0708

Use my branch as a reference only, it is not intended to be fully fleshed out.

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
DmitryLukyanov: will change this

DmitryLukyanov (commented):
First replies

src/MongoDB.Driver.Core/Core/Authentication/SaslAuthenticator.cs
rstam: I leave it up to you and @jyemin whether a comment is necessary explaining why a connection error would be considered an authentication error

src/MongoDB.Driver.Core/Core/Events/CommandFailedEvent.cs
```diff
@@ -43,13 +44,29 @@ public struct CommandFailedEvent
         /// <param name="connectionId">The connection identifier.</param>
         /// <param name="duration">The duration.</param>
         public CommandFailedEvent(string commandName, Exception exception, long? operationId, int requestId, ConnectionId connectionId, TimeSpan duration)
```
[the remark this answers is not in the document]
rstam: The problem is it is too easy to "forget" to pass the serviceId.

Why not pass `null` in the few instances where we don't have it?

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
```diff
@@ -268,6 +268,7 @@ public IAsyncCursor<TResult> Execute(IReadBinding binding, CancellationToken can
 
             using (var context = RetryableReadContext.Create(binding, _retryRequested, cancellationToken))
             {
+                context.PinConnectionIfRequired();
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

originally I thought that transactions can be added not for all operations on the driver side. So I missed some places where it should be added. Also, I added pinning connection for RetryableReads only for operations that returns cursors based on this:
>When the driver is in load balancing mode and executing any cursor-initiating command, the driver MUST NOT check the connection back into the pool unless the command fails or the server returns a cursor ID of 0 (i.e. all documents are returned in a single batch)

However, it looks harmless to add this logic for all operations in LB. But I will think about it one more time tomorrow.

src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
```diff
@@ -177,7 +177,8 @@ public void Initialize()
                 var endPoint = _settings.EndPoints.Single();
                 if (_settings.Scheme != ConnectionStringScheme.MongoDBPlusSrv)
                 {
-                    _server = CreateInitializedServer(endPoint);
+                    _server = _serverFactory.CreateServer(_clusterType, _clusterId, _clusterClock, endPoint);
```
DmitryLukyanov, line 180: `CreateInitializedServer` was waiting until server would trigger description change, but even after that the _server variable itself could still be unassigned

src/MongoDB.Driver.Core/Core/ChannelPinningHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+                        channel.Fork(),
+                        session.Fork()));
+
+                if (session.IsInTransaction && !IsChannelPinned(session.CurrentTransaction))
+                {
+                    session.CurrentTransaction.PinChannel(channel.Fork());
+                    session.CurrentTransaction.PinnedServer = server;
+                }
+
+                pinnedChannel = channel.Fork();
```
DmitryLukyanov, line 136: done, I think I know why this change pinnedChannel = channel.`Fork()`; didn't cause any issue in my code, but I want to confirm it, so will check it tomorrow.

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
rstam: Actually the reason I suggested the change was because I saw places in the spec that were using `LoadBalancer` instead of `LoadBalanced`.

But searching more thoroughly I find that the spec is inconsistent and uses BOTH variations.

But it seems to use `LoadBalanced` more frequently, which is unfortunate because `er` seems more correct and is already used in the `ShardRouter` `ServerType` (i.e. not `ShardRouted`).

So I guess leave as is...

src/MongoDB.Driver.Core/Core/Events/CommandFailedEvent.cs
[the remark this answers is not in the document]
rstam: I think you are missing my point that by leaving this constructor in the code we are making it easy to "forget" to pass the `serviceId` in situations where it is required.

But in any case you may leave as is.

src/MongoDB.Driver.Core/Core/Operations/RetryableReadContext.cs
```diff
@@ -44,6 +42,17 @@ public static RetryableReadContext Create(IReadBinding binding, bool retryReques
...
+
+                if (ChannelPinningHelper.PinChannelSourceAndChannelIfRequired(
+                    context.ChannelSource,
+                    context.Channel,
+                    context.Binding.Session,
+                    out var pinnedChannelSource,
+                    out var pinnedChannel))
+                {
+                    context.ReplaceChannelSource(pinnedChannelSource);
+                    context.ReplaceChannel(pinnedChannel);
```
rstam, line 54: Here in the `Create` method is a good place to handle pinning. I like it.

But can't we just create the binding pinned or not? Do we have to create it "wrong" and then sometimes "replace" the channel source and channel?

rstam (commented):
Latest changes look good.

I added some minor comments.

src/MongoDB.Driver.Core/Core/Operations/RetryableReadContext.cs
DmitryLukyanov: we will need to call the currently presented steps in Initialize anyway since they include selecting Server and acquiring connection and we will need to dispose original handles anyway in further pinning logic. So we can move these replacement into Initialize method. But it will be the same steps that will make exactly the same logic. I would probably won't change it

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -227,7 +227,7 @@ public static CollectionNamespace GetCollectionNamespaceForTestMethod(string cla
             return new CollectionNamespace(__databaseNamespace.Value, collectionName);
         }
 
-        private static ConnectionString GetConnectionString()
+        public static ConnectionString CreateConnectionString()
```
rstam, line 230: This name change is unnecessary, and arguably harmful.

We are "getting" the connection string from an environment variable (with suitable defaults if the environment variable isn't set).

We're not really "creating" anything. Imagine we were returning a `string` instead of wrapping it in a `ConnectionString` and it's easier to see that "get" is more appropriate than "create".

rstam (changes requested):

rstam (approved):
LGTM

Pending comments can be deferred until later

DmitryLukyanov:
Squashed/rebased/merged.

PR #566, opened 2021-07-08
CSHARP-3396: Expose the Reason an Operation Fails Document Validation

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3396
EG: https://evergreen.mongodb.com/version/60e8a2a20305b9034bab85a2
Specification prose test: https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst#2-writeerrordetails-exposes-writeerrorserrinfo

Note: the first prose test was in unexpected location and namespace, so I moved it without making any changes there.


tests/MongoDB.Driver.Tests/Specifications/crud/prose-tests/CrudProseTests.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                var database = client.GetDatabase(DriverTestConfiguration.DatabaseNamespace.DatabaseName);
+                database.CreateCollection(collectionName, collectionOptions);
+                var collection = database.GetCollection<BsonDocument>(collectionName);
+
+                exception = Record.Exception(() => collection.InsertOne(new BsonDocument("x", 1)));
+            }
+
+            // Assert WriteError
+            var bulkWriteException = exception.InnerException.Should().BeOfType<MongoBulkWriteException<BsonDocument>>().Subject;
```
jyemin, line 102: What is the type of `exception`?  I don't really know what `InnerException` is...

jyemin, line 102: Does the exception have some sort of equivalent to Java's `toString()` method?  Or a standard Message property of type string that is printed as part of the exception stack trace?  If so, can we assert something about that message, and ensure that the write error details are serialized as a string into it?

MikalaiMazurenka: Yes, it has some custom logic around the `.ToString()` method, but it returns only the general info:
```
MongoDB.Driver.MongoWriteException: A write operation resulted in an error.\r\n  Document failed validation ---> MongoDB.Driver.MongoBulkWriteException`1[MongoDB.Bson.BsonDocument]: A bulk write operation resulted in one or more errors.\r\n  Document failed validation\r\n   at MongoDB.Driver.MongoCollectionImpl`1.BulkWrite(IClientSessionHandle session, IEnumerable`1 requests, BulkWriteOptions options, CancellationToken cancellationToken) in C:\\Projects\\MongoDB\\mongo-csharp-driver\\src\\MongoDB.Driver\\MongoCollectionImpl.cs:line 230\r\n   at 
...
```
Strictly speaking, `.ToString()` is default here and it just returns the `Message`, but the forming of the `Message` is a bit more custom, it includes the `WriteError`'s `Message` property, which in this case is `"Document failed validation"`, so the `code` and `details` fields won't make it into the exception message which the user would see.

So I see what is your concern here. The user won't see the exception details unless they access them explicitly, which is a normal practice, but requires some knowledge of driver exceptions and some custom error handling code.

**TL;DR:** Great catch! I will check if the code should be changed to include the `details` into resulting exception message.

MikalaiMazurenka (commented):
Answered the questions

jyemin (commented):
You re-requested review before making any further changes.  So just re-reviewing to get back into the right state.

tests/MongoDB.Driver.Tests/Specifications/crud/prose-tests/CrudProseTests.cs
jyemin: Cool,  for the record I think both the outer and inner exception should contain info about write error(s), including the details.

JamesKovacs: It would be worthwhile asserting that outer exception is a `MongoWriteException` as it expresses the intent that this operation results in a `MongoWriteException`.

Additionally you should `exception.Should().NotBeNull();` as you do on line 71. Otherwise line 102 will throw a NRE if we don't get a failure back as expected.

tests/MongoDB.Driver.Tests/Specifications/crud/prose-tests/CrudProseTests.cs
```diff
@@ -0,0 +1,142 @@
...
+        private DisposableMongoClient CreateDisposableClient(EventCapturer eventCapturer)
+        {
+            return DriverTestConfiguration.CreateDisposableClient((MongoClientSettings settings) =>
+            {
+                settings.HeartbeatInterval = TimeSpan.FromMilliseconds(5);
+                settings.ClusterConfigurator = c => c.Subscribe(eventCapturer);
+            });
+        }
+    }
+}
```
JamesKovacs, line 142: `ConfigureFailPoint` and `CreateDisposableClient` would be very useful in one of the `TestHelpers` projects so we can easily re-use them.

JamesKovacs (changes requested):
I don't see any changes to production code, only test code. I thought the goal of CSHARP-3396 was to make write errors more easily parsable and visible to developers. As far as I can see, the errors are still buried in inner exceptions inside stack traces.

tests/MongoDB.Driver.Tests/Specifications/crud/prose-tests/CrudProseTests.cs
MikalaiMazurenka: Done

MikalaiMazurenka: Done

MikalaiMazurenka:
Ready for review

tests/MongoDB.Driver.Tests/Specifications/crud/prose-tests/CrudProseTests.cs
jyemin: Consider this resolved by me.  I'll leave it to James to actually resolve.

jyemin (approved):
LGTM!

JamesKovacs (approved):
LGTM

PR #569, opened 2021-07-12
CSHARP-3721: Disable writes on snapshot sessions

BorisDog:


src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -351,6 +351,13 @@ private Type0CommandMessageSection<BsonDocument> CreateType0Section(ConnectionDe
                 }
             }
 
+
+            var snapshotReadConcernDocument = ReadConcernHelper.GetReadConcernForSnapshotSesssion(_session, connectionDescription);
+            if (snapshotReadConcernDocument != null)
+            {
+                extraElements.Add(new BsonElement("readConcern", snapshotReadConcernDocument));
+            }
+
```
rstam, line 360: I'm not fond of returning `null` as a magic value to indicate that it's *not* a snapshot read after all.

Consider something like this instead:

    if (ReadConcernHelper.IsSnapshotRead(_session, connectionDescription))
    {
        extraElements.Add(new BsonElement("readConcern", ReadConcernHelper.GetReadConcernForSnapshotSesssion(_session, connectionDescription)));
    }

rstam (changes requested):
I want to review again once there is a stable starting point.

src/MongoDB.Driver.Core/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
BorisDog: The intention is to stay consistent with other `ReadConcernHelper` methods and their usages. For example all the necessary validations like `session.IsCausallyConsistent` is performed by `ReadConcernHelper.GetReadConcernForCommand` and null return value indicates the absence of readconcern value.

rstam: I understand how it works. I'm just not fond of returning `null` as a magic value.

But it can stay as is.

rstam (approved):
LGTM

PR #570, opened 2021-07-12
CSHARP-3700: Support 'let' option for aggregate command

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3700
EG: https://evergreen.mongodb.com/version/60ed9f9b562343354c74fc64

Note: Public API `Feature` class was changed, which could be considered as a breaking change. The reason for the change is that existing `AggregateLet` feature was referring only to the `$lookup` stage of the aggregate pipeline, while the new feature applies to the whole `aggregate` operation.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
+++ b/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
@@ -36,7 +36,8 @@ namespace MongoDB.Driver.Core.Misc
         private static readonly Feature __aggregateFunction = new Feature("AggregateFunction", new SemanticVersion(4, 3, 4));
         private static readonly Feature __aggregateGraphLookupStage = new Feature("AggregateGraphLookupStage", new SemanticVersion(3, 4, 0, "rc0"));
         private static readonly Feature __aggregateHint = new Feature("AggregateHint", new SemanticVersion(3, 6, 0, "rc0"));
-        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(3, 6, 0));
+        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(5, 0, 0, ""));
+        private static readonly Feature __aggregateLookupLet = new Feature("AggregateLookupLet", new SemanticVersion(3, 6, 0));
         private static readonly Feature __aggregateMerge = new Feature("AggregateMerge", new SemanticVersion(4, 2, 0));
         private static readonly Feature __aggregateOut = new Feature("AggregateOut", new SemanticVersion(2, 6, 0));
         private static readonly Feature __aggregateOutToDifferentDatabase = new Feature("AggregateOutToDifferentDatabase", new SemanticVersion(4, 3, 0));
@@ -174,6 +175,11 @@ namespace MongoDB.Driver.Core.Misc
         /// Gets the aggregate let feature.
         /// </summary>
         public static Feature AggregateLet => __aggregateLet;
+
+        /// <summary>
+        /// Gets the aggregate lookup stage let feature.
+        /// </summary>
+        public static Feature AggregateLookupLet => __aggregateLookupLet;
 
         /// <summary>
         /// Gets the aggregate merge feature.
--- a/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
+++ b/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
@@ -36,8 +36,8 @@ namespace MongoDB.Driver.Core.Misc
         private static readonly Feature __aggregateFunction = new Feature("AggregateFunction", new SemanticVersion(4, 3, 4));
         private static readonly Feature __aggregateGraphLookupStage = new Feature("AggregateGraphLookupStage", new SemanticVersion(3, 4, 0, "rc0"));
         private static readonly Feature __aggregateHint = new Feature("AggregateHint", new SemanticVersion(3, 6, 0, "rc0"));
-        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(5, 0, 0, ""));
-        private static readonly Feature __aggregateLookupLet = new Feature("AggregateLookupLet", new SemanticVersion(3, 6, 0));
+        private static readonly Feature __aggregateOptionsLet = new Feature("AggregateOptionsLet", new SemanticVersion(5, 0, 0, ""));
+        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(3, 6, 0));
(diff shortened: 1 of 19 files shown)


src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -36,7 +36,8 @@ public class Feature
         private static readonly Feature __aggregateFunction = new Feature("AggregateFunction", new SemanticVersion(4, 3, 4));
         private static readonly Feature __aggregateGraphLookupStage = new Feature("AggregateGraphLookupStage", new SemanticVersion(3, 4, 0, "rc0"));
         private static readonly Feature __aggregateHint = new Feature("AggregateHint", new SemanticVersion(3, 6, 0, "rc0"));
-        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(3, 6, 0));
+        private static readonly Feature __aggregateLet = new Feature("AggregateLet", new SemanticVersion(5, 0, 0, ""));
+        private static readonly Feature __aggregateLookupLet = new Feature("AggregateLookupLet", new SemanticVersion(3, 6, 0));
```
JamesKovacs, line 40: Rather than renaming the existing one, we should consider `Feature.AggregateOptionsLet` for the new feature. What do you think?

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs
```diff
@@ -798,6 +837,25 @@ public void CreateCommand_should_throw_when_ReadConcern_is_set_but_not_supported
...
+            var subject = new AggregateOperation<BsonDocument>(_collectionNamespace, __pipeline, __resultSerializer, _messageEncoderSettings)
+            {
+                Let = new BsonDocument("x", "y")
+            };
+
+            var cursor = ExecuteOperation(subject, async);
+            var result = ReadCursorToEnd(cursor, async);
+
+            result.Should().NotBeNull();
+        }
```
JamesKovacs, line 857: The query is not using `$$x`. So we don't know that the `let` variable was set correctly. Note that for `$$x` must be used in the context of an `$expr`.

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateToCollectionOperationTests.cs
```diff
@@ -718,6 +755,25 @@ public void CreateCommand_should_return_expected_result_when_MaxTime_is_set(long
...
+            var subject = new AggregateToCollectionOperation(_collectionNamespace, __pipeline, _messageEncoderSettings)
+            {
+                Let = new BsonDocument("x", "y")
+            };
+
+            ExecuteOperation(subject, async);
+            var result = ReadAllFromCollection(new CollectionNamespace(_databaseNamespace, "awesome"), async);
+
+            result.Should().NotBeNull();
+        }
```
JamesKovacs, line 775: The `let` variables aren't being used in the query and we don't know that they're set correctly.

JamesKovacs (changes requested):
Overall looks good. One minor suggestion and some test improvements requested.

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
rstam: Definitely let's not rename an existing feature name.

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs
rstam: Or better yet refactor the test to use the let variable in a `$project` instead of in a `$match`.

rstam (changes requested):

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs
MikalaiMazurenka: Added tests for each stage.
Done.

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateToCollectionOperationTests.cs
MikalaiMazurenka: Done.

MikalaiMazurenka (commented):
Ready for review!

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -36,6 +36,7 @@ public class Feature
         private static readonly Feature __aggregateFunction = new Feature("AggregateFunction", new SemanticVersion(4, 3, 4));
         private static readonly Feature __aggregateGraphLookupStage = new Feature("AggregateGraphLookupStage", new SemanticVersion(3, 4, 0, "rc0"));
         private static readonly Feature __aggregateHint = new Feature("AggregateHint", new SemanticVersion(3, 6, 0, "rc0"));
+        private static readonly Feature __aggregateOptionsLet = new Feature("AggregateOptionsLet", new SemanticVersion(5, 0, 0, ""));
```
rstam, line 39: Maintain alphabetical order?

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -173,6 +174,11 @@ public class Feature
         /// <summary>
         /// Gets the aggregate let feature.
         /// </summary>
+        public static Feature AggregateOptionsLet => __aggregateOptionsLet;
```
rstam, line 177: Maintain alphabetical order?

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs
```diff
@@ -460,6 +472,33 @@ public void CreateCommand_should_throw_when_Collation_is_set()
...
+            {
+                Let = let
+            };
+
+            var connectionDescription = OperationTestHelper.CreateConnectionDescription(Feature.AggregateOptionsLet.FirstSupportedVersion);
+            var session = OperationTestHelper.CreateSession();
+
+            var result = subject.CreateCommand(connectionDescription, session);
+
+            var expectedResult = new BsonDocument
```
rstam, line 492: This is a dangerous form of testing, where you essentially copy/paste essential parts of the code under test into the test. 

If the code you copy/paste has a bug the test is going to pass in spite of the bug.

In the interest of getting the release out no change needed. Just keep in mind for the future.

rstam (changes requested):
The changes I am requesting are minor and could be skipped if James and Jeff are OK with the current version as-is.

JamesKovacs (approved):
LGTM

PR #572, opened 2021-07-12
Revert hello-timeout blockTimeMS back to 2000ms.

BorisDog:
Due to current minimal socket timeout of 500ms, server blocking failpoint duration in  is increased to 2000ms to eliminate `hello-timeout` test failures.


DmitryLukyanov (approved):
LGTM

PR #573, opened 2021-07-13
Update current version to 2.14 in reference docs.

rstam:


rstam:
Wrong version number. Starting over.

PR #577, opened 2021-07-13
Add missing await to unified list test operations.

DmitryLukyanov:


tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
```diff
@@ -59,11 +59,10 @@ public async Task<OperationResult> ExecuteAsync(CancellationToken cancellationTo
         {
             try
             {
-                using var cursor = _session == null
-                    ? await _database.ListCollectionsAsync(_options, cancellationToken)
-                    : await _database.ListCollectionsAsync(_session, _options, cancellationToken);
-
-                _ = cursor.ToListAsync(cancellationToken);
+                var cursor = _session == null
+                    ? await _database.ListCollectionsAsync(_options, cancellationToken).ConfigureAwait(false)
```
BorisDog, line 63: No need for `ConfigureAwait`

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
```diff
@@ -41,7 +41,7 @@ public OperationResult Execute(CancellationToken cancellationToken)
         {
             try
             {
-                using var cursor = _session == null
+                var cursor = _session == null
```
BorisDog, line 44: I don't think that we can skip disposing entities as a pattern.
If this is a temporary patch, we need a ticket for fixing it.

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
rstam: We seem to be inconsistent on whether to use `ConfigureAwait` in test projects or not.

MikalaiMazurenka:
Please change the PR name and commit name to something more specific, like `Add missing await to unified list operations`

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
BorisDog: AFAIK technically we don't have a reason to use `ConfigureAwait` in tests.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
DmitryLukyanov: This will make investigating of test results harder since we will need to consider one more event. And sometimes it can be painful during debugging. From the other hand this looks like overkill to me here, but let's see what the spec author will say

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
DmitryLukyanov: I think we did this sometimes, but at least in most of the cases it's not added. So, reverted

BorisDog (approved):

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListCollectionsOperation.cs
DmitryLukyanov: The issue is related to missed `await`. I forgot that we call `killCursors` only for `cursorId !=0`, so when we missed await, we can call `cursor.Dispose` before actual iterating. And then we will see the mentioned behavior. Adding await solves it.

rstam (approved):
LGTM

PR #578, opened 2021-07-13
CSHARP-3698: Calling cursor.next() and cursor.close() concurrently in Load Balanced mode uses the same connection simultaneously.

DmitryLukyanov:
This PR does **not** affect the current release.


rstam (approved):
LGTM

PR #581, opened 2021-07-20
CSHARP-3726: Skip failing test on 5.1.0-latest only.

rstam:


tests/MongoDB.Driver.Tests/Linq/MongoQueryableTests.cs
```diff
@@ -555,6 +555,11 @@ public void GroupJoin_syntax_with_select_many()
         public void GroupJoin_syntax_with_select_many_and_DefaultIfEmpty()
         {
             RequireServer.Check().VersionGreaterThanOrEqualTo("3.2.0");
+            var actualVersion = CoreTestConfiguration.ServerVersion;
+            if (actualVersion > SemanticVersion.Parse("5.0.0-") && actualVersion < SemanticVersion.Parse("5.0.0"))
```
rstam, line 559: In reality we know that version `5.0.0-alpha0-1698-gf4169a7` is *after* `5.0.0` but it "looks" like it's before (because it looks like a pre-release of `5.0.0`), hence the somewhat odd comparison.

JamesKovacs (approved):
LGTM

PR #582, opened 2021-07-20
CSHARP-2661: Show Test Results in Evergreen

jyemin:
CSHARP-2661

Supersedes https://github.com/durran/mongo-csharp-driver/pull/7

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -2,6 +2,7 @@
 #addin nuget:?package=Cake.Git&version=1.0.1
 #addin nuget:?package=Cake.Incubator&version=6.0.0
 #tool dotnet:?package=GitVersion.Tool&version=5.6.9
+#tool nuget:?package=JunitXml.TestLogger&version=3.0.98
 
 using System;
 using System.Text.RegularExpressions;
@@ -24,6 +25,7 @@ var docsDirectory = solutionDirectory.Combine("Docs");
 var docsApiDirectory = docsDirectory.Combine("Api");
 var srcDirectory = solutionDirectory.Combine("src");
 var testsDirectory = solutionDirectory.Combine("tests");
+var outputDirectory = solutionDirectory.Combine("build");
 var toolsDirectory = solutionDirectory.Combine("tools");
 var toolsHugoDirectory = toolsDirectory.Combine("Hugo");
 
@@ -51,7 +53,7 @@ Task("Restore")
     .Does(() =>
     {
         // disable parallel restore to work around apparent bugs in restore
-        var restoreSettings = new DotNetCoreRestoreSettings 
+        var restoreSettings = new DotNetCoreRestoreSettings
         {
             DisableParallel = true
         };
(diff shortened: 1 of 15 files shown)


build.cake
```diff
@@ -2,6 +2,7 @@
 #addin nuget:?package=Cake.Git&version=1.0.1
 #addin nuget:?package=Cake.Incubator&version=6.0.0
 #tool dotnet:?package=GitVersion.Tool&version=5.6.9
+#tool nuget:?package=JunitXml.TestLogger&version=3.0.98
```
jyemin, line 5: There was a merge conflict on this line.

build.cake
```diff
@@ -24,6 +25,7 @@ var docsDirectory = solutionDirectory.Combine("Docs");
 var docsApiDirectory = docsDirectory.Combine("Api");
 var srcDirectory = solutionDirectory.Combine("src");
 var testsDirectory = solutionDirectory.Combine("tests");
+var outputDirectory = solutionDirectory.Combine("build");
```
jyemin, line 28: There was a merge conflict on this line.

build.cake
jyemin, line 5: Updated to latest version, per James' recommendation

JamesKovacs: That looks kosher to me now.

build.cake
```diff
@@ -148,11 +150,14 @@ Task("Test")
             Console.WriteLine($"MONGO_X509_CLIENT_CERTIFICATE_PASSWORD={mongoX509ClientCertificatePassword}");
         }
 
+        var testResultsFile = outputDirectory.Combine("test-results").Combine($"TEST-{target.ToLowerInvariant()}-{DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()}.xml");
+        var logger = $"junit;LogFilePath={testResultsFile};FailureBodyFormat=Verbose";
```
DmitryLukyanov, line 154: it would be good to add comments about why we use anything java-related like junit.

tests/AstrolabeWorkloadExecutor/AstrolabeWorkloadExecutor.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0" />
+    <PackageReference Include="JunitXml.TestLogger" Version="2.1.81" />
```
JamesKovacs, line 28: The cake script is using `JunitXml.TestLogger `3.0.98 but the test assemblies are using 2.1.81. They should all use the same version. Same for all the PackageReferences below too.

JamesKovacs (changes requested):
Fix the NuGet package versions referenced and we are good to go.

build.cake
jyemin: What would you like it to say?  I'm not sure it's necessary as you can get the context from looking at the Jira ticket.

tests/AstrolabeWorkloadExecutor/AstrolabeWorkloadExecutor.csproj
jyemin: Done

build.cake
DmitryLukyanov: jira ticket(as I see) doesn't say why we use a java-related library instead of c# one. For example why it's not enough to use a way that was used by @BorisDog  in his logging [PR](https://github.com/BorisDog/mongo-csharp-driver/pull/9/files#diff-c6eace5a70f0c2d920dfa97e988d4cad3d0131dd2b029bb2646c96222e866cf8R156)? Is it impossible? If so, adding a note will handle such question later

jyemin: Can you draft the note that you want to see?  I'm not sure what it should say.

DmitryLukyanov: EG can currently parse only junit output?

jyemin: Done

JamesKovacs: We are just outputting test results in Junit format because that's a de facto standard. As far as I understand it, EG has better Junit format support than Xunit format support. If we add a comment, then something along the lines of:

```
// Evergreen provides better integration with the Junit test results format.
```

JamesKovacs (approved):
LGTM

build.cake
jyemin: @JamesKovacs looking at the patch build, it looks like this version bump might have broken things.

DmitryLukyanov (approved):
LGTM

PR #583, opened 2021-07-20
CSHARP-3387: Serverless testing

MikalaiMazurenka:
JIRA: https://jira.mongodb.org/browse/CSHARP-3387
EG (serverless only): https://evergreen.mongodb.com/version/610156453e8e8611432d8f5d
EG (full run): https://evergreen.mongodb.com/version/6101567a5623432f0eaea2cf

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -337,6 +337,11 @@ Task("TestGssapi")
         );
     });
 
+Task("TestGssapiNet452").IsDependentOn("TestGssapi");
+Task("TestGssapiNetStandard15").IsDependentOn("TestGssapi");
+Task("TestGssapiNetStandard20").IsDependentOn("TestGssapi");
+Task("TestGssapiNetStandard21").IsDependentOn("TestGssapi");
+
 Task("TestServerless")
     .IsDependentOn("Build")
     .DoesForEach(
@@ -348,28 +353,27 @@ Task("TestServerless")
                 NoBuild = true,
                 NoRestore = true,
                 Configuration = configuration,
-                ArgumentCustomization = args => args.Append("-- RunConfiguration.TargetPlatform=x64")
-                Filter = "Category=\"Serverless\"",
+                ArgumentCustomization = args => args.Append("-- RunConfiguration.TargetPlatform=x64"),
+                Filter = "Category=\"Serverless\""
             };
             switch (target.ToLowerInvariant())
             {
-                case "serverlessnet452": settings.Framework = "net452"; break;
-                case "serverlessnetstandard15": settings.Framework = "netcoreapp1.1"; break;
-                case "serverlessnetstandard20": settings.Framework = "netcoreapp2.1"; break;
-                case "serverlessnetstandard21": settings.Framework = "netcoreapp3.0"; break;
+                case "testserverlessnet452": settings.Framework = "net452"; break;
+                case "testserverlessnetstandard15": settings.Framework = "netcoreapp1.1"; break;
+                case "testserverlessnetstandard20": settings.Framework = "netcoreapp2.1"; break;
+                case "testserverlessnetstandard21": settings.Framework = "netcoreapp3.0"; break;
                 default: throw new ArgumentException($"Unexpected target: \"{target}\".");
             }
             DotNetCoreTest(
                 testProject.FullPath,
                 settings
             );
-        })
-    .DeferOnError();
+        });
 
-Task("TestGssapiNet452").IsDependentOn("TestGssapi");
-Task("TestGssapiNetStandard15").IsDependentOn("TestGssapi");
-Task("TestGssapiNetStandard20").IsDependentOn("TestGssapi");
-Task("TestGssapiNetStandard21").IsDependentOn("TestGssapi");
+Task("TestServerlessNet452").IsDependentOn("TestServerless");
+Task("TestServerlessNetStandard15").IsDependentOn("TestServerless");
+Task("TestServerlessNetStandard20").IsDependentOn("TestServerless");
+Task("TestServerlessNetStandard21").IsDependentOn("TestServerless");
 
 Task("TestLoadBalanced")
     .IsDependentOn("Build")
--- a/build.cake
+++ b/build.cake
@@ -313,7 +313,7 @@ Task("TestGssapi")
     .DoesForEach(
         GetFiles("./**/MongoDB.Driver.Tests.csproj"),
         testProject =>
-	{
+    {
         var settings = new DotNetCoreTestSettings
         {
             NoBuild = true,
@@ -329,11 +329,42 @@ Task("TestGssapi")
             case "testgssapinetstandard20": settings.Framework = "netcoreapp2.1"; break;
             case "testgssapinetstandard21": settings.Framework = "netcoreapp3.0"; break;
         }
-		DotNetCoreTest(
-			testProject.FullPath,
+        DotNetCoreTest(
+            testProject.FullPath,
             settings
-		);
-	});
+        );
+    });
+
+Task("TestServerless")
+    .IsDependentOn("Build")
+    .DoesForEach(
+        GetFiles("./**/MongoDB.Driver.Tests.csproj"),
+        testProject =>
+        {
+            var targetFramework = Environment.GetEnvironmentVariable("FRAMEWORK");
+            var framework = "";
+            switch (targetFramework.ToLowerInvariant())
+            {
+                case "net452": framework = "net452"; break;
+                case "netstandard15": framework = "netcoreapp1.1"; break;
+                case "netstandard20": framework = "netcoreapp2.1"; break;
+                case "netstandard21": framework = "netcoreapp3.0"; break;
+                default: throw new ArgumentException($"Unexpected framework: \"{framework}\"");
+            }
+            DotNetCoreTest(
+                testProject.FullPath,
+                new DotNetCoreTestSettings
+                {
+                    NoBuild = true,
+                    NoRestore = true,
+                    Configuration = configuration,
+                    ArgumentCustomization = args => args.Append("-- RunConfiguration.TargetPlatform=x64"),
+                    Filter = "Category=\"Serverless\"",
+                    Framework = framework
+                }
+            );
+        })
+    .DeferOnError();
 
 Task("TestGssapiNet452").IsDependentOn("TestGssapi");
 Task("TestGssapiNetStandard15").IsDependentOn("TestGssapi");
--- a/build.cake
+++ b/build.cake
@@ -161,6 +161,7 @@ Task("Test")
             case "testnetstandard15": settings.Framework = "netcoreapp1.1"; break;
             case "testnetstandard20": settings.Framework = "netcoreapp2.1"; break;
             case "testnetstandard21": settings.Framework = "netcoreapp3.0"; break;
+            default: throw new ArgumentException($"Unexpected target: \"{target}\".");
         }
         DotNetCoreTest(
             testProject.FullPath,
@@ -328,6 +329,7 @@ Task("TestGssapi")
             case "testgssapinetstandard15": settings.Framework = "netcoreapp1.1"; break;
             case "testgssapinetstandard20": settings.Framework = "netcoreapp2.1"; break;
             case "testgssapinetstandard21": settings.Framework = "netcoreapp3.0"; break;
+            default: throw new ArgumentException($"Unexpected target: \"{target}\".");
         }
         DotNetCoreTest(
             testProject.FullPath,
@@ -341,27 +343,25 @@ Task("TestServerless")
         GetFiles("./**/MongoDB.Driver.Tests.csproj"),
         testProject =>
         {
-            var targetFramework = Environment.GetEnvironmentVariable("FRAMEWORK");
-            var framework = "";
-            switch (targetFramework.ToLowerInvariant())
+            var settings = new DotNetCoreTestSettings
             {
-                case "net452": framework = "net452"; break;
-                case "netstandard15": framework = "netcoreapp1.1"; break;
-                case "netstandard20": framework = "netcoreapp2.1"; break;
-                case "netstandard21": framework = "netcoreapp3.0"; break;
-                default: throw new ArgumentException($"Unexpected framework: \"{framework}\"");
+                NoBuild = true,
+                NoRestore = true,
+                Configuration = configuration,
+                ArgumentCustomization = args => args.Append("-- RunConfiguration.TargetPlatform=x64")
+                Filter = "Category=\"Serverless\"",
+            };
+            switch (target.ToLowerInvariant())
+            {
+                case "serverlessnet452": settings.Framework = "net452"; break;
+                case "serverlessnetstandard15": settings.Framework = "netcoreapp1.1"; break;
+                case "serverlessnetstandard20": settings.Framework = "netcoreapp2.1"; break;
+                case "serverlessnetstandard21": settings.Framework = "netcoreapp3.0"; break;
+                default: throw new ArgumentException($"Unexpected target: \"{target}\".");
             }
             DotNetCoreTest(
                 testProject.FullPath,
-                new DotNetCoreTestSettings
-                {
-                    NoBuild = true,
-                    NoRestore = true,
-                    Configuration = configuration,
-                    ArgumentCustomization = args => args.Append("-- RunConfiguration.TargetPlatform=x64"),
-                    Filter = "Category=\"Serverless\"",
-                    Framework = framework
-                }
+                settings
             );
         })
     .DeferOnError();
--- a/evergreen/evergreen.yml
+++ b/evergreen/evergreen.yml
@@ -599,7 +599,6 @@ functions:
       params:
         working_dir: mongo-csharp-driver
         script: |
-          # set +o xtrace
           MONGODB_URI="${MONGODB_URI}" \
           MONGODB_SRV_URI="${MONGODB_SRV_URI}" \
           SERVERLESS_ATLAS_USER="${SERVERLESS_ATLAS_USER}" \
@@ -609,6 +608,34 @@ functions:
           COMPRESSOR=${COMPRESSOR} \
           FRAMEWORK=${FRAMEWORK} \
             evergreen/run-serverless-tests.sh
+
+  create-serverless-instance:
+    - command: shell.exec
+      params:
+        #silent: true
+        script: |
+          ${PREPARE_SHELL}
+          set +o xtrace
+          SERVERLESS_DRIVERS_GROUP=${SERVERLESS_DRIVERS_GROUP} \
+          SERVERLESS_API_PUBLIC_KEY=${SERVERLESS_API_PUBLIC_KEY} \
+          SERVERLESS_API_PRIVATE_KEY=${SERVERLESS_API_PRIVATE_KEY} \
+            bash ${DRIVERS_TOOLS}/.evergreen/serverless/create-instance.sh
+    - command: expansions.update
+      params:
+        file: serverless-expansion.yml
+
+  delete-serverless-instance:
+    - command: shell.exec
+      params:
+        #silent: true
+        script: |
+          ${PREPARE_SHELL}
+          set +o xtrace
+          SERVERLESS_DRIVERS_GROUP=${SERVERLESS_DRIVERS_GROUP} \
+          SERVERLESS_API_PUBLIC_KEY=${SERVERLESS_API_PUBLIC_KEY} \
+          SERVERLESS_API_PRIVATE_KEY=${SERVERLESS_API_PRIVATE_KEY} \
+          SERVERLESS_INSTANCE_NAME=${SERVERLESS_INSTANCE_NAME} \
+            bash ${DRIVERS_TOOLS}/.evergreen/serverless/delete-instance.sh
 
   publish-snapshot:
     - command: shell.exec
@@ -675,7 +702,6 @@ functions:
           ${PREPARE_SHELL}
           echo '{"results": [{ "status": "FAIL", "test_file": "Build", "log_raw": "No test-results.json found was created" } ]}' > ${PROJECT_DIRECTORY}/test-results.json
 
-# Note: Task groups do not run "pre" or "post" tasks, thus relevant tasks should be copied to "setup_group" section of the task group
 pre:
   - func: fetch-source
   - func: prepare-resources
@@ -687,6 +713,7 @@ pre:
 post:
   # Removed, causing timeouts
   # - func: upload-working-dir
+  - func: delete-serverless-instance
   - func: upload-mo-artifacts
   - func: upload-test-results
   - func: cleanup
@@ -863,10 +890,33 @@ tasks:
       commands:
         - func: publish-snapshot
 
-    - name: serverless-tests
-      tags: ["serverless"]
+    - name: serverless-tests-net452
       commands:
+        - func: create-serverless-instance
         - func: run-serverless-tests
+          vars:
+            FRAMEWORK: net452
+
+    - name: serverless-tests-netstandard15
+      commands:
+        - func: create-serverless-instance
+        - func: run-serverless-tests
+          vars:
+            FRAMEWORK: netstandard15
+
+    - name: serverless-tests-netstandard20
+      commands:
+        - func: create-serverless-instance
+        - func: run-serverless-tests
+          vars:
+            FRAMEWORK: netstandard20
+
+    - name: serverless-tests-netstandard21
+      commands:
+        - func: create-serverless-instance
+        - func: run-serverless-tests
+          vars:
+            FRAMEWORK: netstandard21
 
     - name: test-ocsp-rsa-valid-cert-server-staples-ca-responder
       tags: ["ocsp"]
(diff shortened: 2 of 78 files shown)


tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsTestRunner.cs
```diff
@@ -416,14 +417,38 @@ private void VerifyCollectionData(IEnumerable<BsonDocument> expectedDocuments)
         // nested types
         public class TestCaseFactory : JsonDrivenTestCaseFactory
         {
+            #region static
+            private static readonly string[] __serverlessIgnoredTestNames =
+            {
+                // https://jira.mongodb.org/browse/CLOUDP-92618
```
jyemin, line 423: Make sure not to commit with this link in the source code, as that project is private.  Maybe instead add a CSHARP ticket to re-enable these, and link the CSHARP ticket to the CLOUDP one.

jyemin (changes requested):
Just a quick scan, found one thing.

evergreen/evergreen.yml
```diff
@@ -1456,6 +1542,24 @@ buildvariants:
     - name: "test-load-balancer-netstandard20"
     - name: "test-load-balancer-netstandard21"
 
+- matrix_name: serverless-tests-windows
+  matrix_spec: { auth: "auth", ssl: "ssl", compressor: "zlib", os: "windows-64", framework: ["net452", "netstandard15", "netstandard20", "netstandard21"] }
+  display_name: "Serverless ${compressor} ${auth} ${ssl} ${os} ${framework}"
+  tasks:
+    - name: serverless-task-group
```
DmitryLukyanov, line 1549: unless I don't know some context, see how framework is handled in other places for example in LB (few lines above)

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsTestRunner.cs
MikalaiMazurenka: Done

evergreen/evergreen.yml
MikalaiMazurenka: task groups cannot be arranged the same way as other tasks. So in order to test multiple frameworks in task group they should be defined as axis.

MikalaiMazurenka (commented):
Ready for review

MikalaiMazurenka:
Some evergreen tasks failed with "system failure". That's because I've used all the available test instances, meaning nothing wrong with the tests or the code.

evergreen/evergreen.yml
DmitryLukyanov: can you explain why `task groups` is required? It looks like java doesn't use it: https://github.com/mongodb/mongo-java-driver/pull/763/files#diff-bcd2c099e589f1ef43957739e25b7ac58b6a847e92cc1e5cc9d9c91acb2fe9f9R252

DmitryLukyanov: My main point here, that it would be good to not introduce a new way to configure framework (if possible) since we already have one

evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+else
+  for var in TMP TEMP NUGET_PACKAGES NUGET_HTTP_CACHE_PATH APPDATA; do
+    export $var=/data/tmp;
+  done
+fi
+
+MONGODB_URI_SPLIT=(${MONGODB_URI//,/ })
+export MONGODB_URI_WITH_MULTIPLE_MONGOSES="${MONGODB_URI:0:10}${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${MONGODB_URI:10}/?tls=true&authSource=admin&compressors=$COMPRESSOR"
+export MONGODB_URI="${MONGODB_URI_SPLIT[0]:0:10}${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${MONGODB_URI_SPLIT[0]:10}/?tls=true&authSource=admin&compressors=$COMPRESSOR"
+export SERVERLESS=true
```
DmitryLukyanov, line 57: nit: I would call it: `SERVERLESS_ENABLED` or something similar

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
```diff
@@ -127,6 +127,18 @@ public RequireServer RunOn(BsonArray requirements)
...
+        public RequireServer Serverless(bool require = true)
+        {
+            var isServerless = CoreTestConfiguration.Serverless;
+
+            if (isServerless == require)
+            {
+                return this;
+            }
+
+            throw new SkipException("Test skipped because serverless is " + (require ? "required" : "not required"));
```
DmitryLukyanov, line 139: `dot` in the end of exception message

tests/MongoDB.Driver.Tests/Specifications/transactions/tests/README.rst
```diff
@@ -0,0 +1,663 @@
...
+
+recoveryToken
+^^^^^^^^^^^^^
+
+A ``recoveryToken`` value of ``42`` in a command-started event is a
+placeholder for an arbitrary recovery token. Drivers MUST assert that the
+actual command includes a "recoveryToken" field and SHOULD assert that field
+is a BSON document.
+
+Mongos Pinning Prose Tests
```
DmitryLukyanov, line 525: do we have such tests?

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatValidFailTestRunner.cs
```diff
@@ -22,6 +22,7 @@
 
 namespace MongoDB.Driver.Tests.Specifications.unified_test_format
 {
+    [Trait("Category", "unified-test-format")]
```
DmitryLukyanov, line 25: does anything set this?

build.cake
```diff
@@ -329,11 +329,42 @@ Task("TestGssapi")
...
+        );
+    });
+
+Task("TestServerless")
+    .IsDependentOn("Build")
+    .DoesForEach(
+        GetFiles("./**/MongoDB.Driver.Tests.csproj"),
+        testProject =>
+        {
+            var targetFramework = Environment.GetEnvironmentVariable("FRAMEWORK");
```
DmitryLukyanov, line 344: it's better to pass/handle this option in the same way as we do in other places in this file: https://github.com/mongodb/mongo-csharp-driver/blob/1c3488e2bdf2e05fc2c5d21ccaec6f5e83a4e981/build.cake#L325

build.cake
```diff
@@ -329,11 +329,42 @@ Task("TestGssapi")
...
+            var framework = "";
+            switch (targetFramework.ToLowerInvariant())
+            {
+                case "net452": framework = "net452"; break;
+                case "netstandard15": framework = "netcoreapp1.1"; break;
+                case "netstandard20": framework = "netcoreapp2.1"; break;
+                case "netstandard21": framework = "netcoreapp3.0"; break;
+                default: throw new ArgumentException($"Unexpected framework: \"{framework}\"");
```
DmitryLukyanov, line 352: we don't have such case option in other places here. I think it's better to choose the single approach maybe even with creating a cake method for this.

DmitryLukyanov: also, a `dot` in the end of error message

evergreen/evergreen.yml
```diff
@@ -13,8 +13,8 @@ stepback: true
 command_type: system
 
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
-# 12 minutes is the longest we'll ever run
-exec_timeout_secs: 1800
+# 42 minutes is the average for serverless tests (the longest of all, because of the heavy set up task)
```
DmitryLukyanov, line 16: this can be painful since it can hide timeout exception in some case. One of theory we have that at least partially the timeouts we have are related to performance. But I'm not sure whether we can implement it differently now

evergreen/evergreen.yml
```diff
@@ -1289,6 +1314,67 @@ axes:
...
+        params:
+          silent: true
+          script: |
+            ${PREPARE_SHELL}
+            set +o xtrace
+            SERVERLESS_DRIVERS_GROUP=${SERVERLESS_DRIVERS_GROUP} \
+            SERVERLESS_API_PUBLIC_KEY=${SERVERLESS_API_PUBLIC_KEY} \
+            SERVERLESS_API_PRIVATE_KEY=${SERVERLESS_API_PRIVATE_KEY} \
+            SERVERLESS_INSTANCE_NAME=${SERVERLESS_INSTANCE_NAME} \
+              bash ${DRIVERS_TOOLS}/.evergreen/serverless/delete-instance.sh
```
DmitryLukyanov, line 1374: it looks like this step never happens: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/serverless/delete-instance.sh#L26 at least I don't see it here: https://evergreen.mongodb.com/task_log_raw/dot_net_driver_serverless_tests_ubuntu__os~ubuntu_1804_auth~auth_ssl~ssl_compressor~zlib_framework~netstandard21_serverless_tests_patch_c04e50a455112fd71df86991024ff66bec1fab97_60f7ff932fbabe15604a8a3d_21_07_21_11_06_22/0?type=T

DmitryLukyanov (commented):
I don't know the serverless specific deeply. So, I'm rely on the fact that tests passed.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedListIndexesOperation.cs
```diff
@@ -63,7 +63,7 @@ public async Task<OperationResult> ExecuteAsync(CancellationToken cancellationTo
                     ? await _collection.Indexes.ListAsync(_listIndexesOptions, cancellationToken)
                     : await _collection.Indexes.ListAsync(_session, _listIndexesOptions, cancellationToken);
 
-                _ = cursor.ToListAsync(cancellationToken);
+                _ = await cursor.ToListAsync(cancellationToken);
```
BorisDog, line 66: +1

evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+
+# Don't trace since the URI contains a password that shouldn't show up in the logs
+set -o errexit # Exit the script with error if any of the commands fail
+
+# Supported/used environment variables:
+#   OS                                  Operating system, must be set
```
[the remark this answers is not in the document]
MikalaiMazurenka: Removed the `MONGODB_SRV_URI`.

evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+    export $var=z:\\data\\tmp
+  done
+else
+  for var in TMP TEMP NUGET_PACKAGES NUGET_HTTP_CACHE_PATH APPDATA; do
+    export $var=/data/tmp;
+  done
+fi
+
+MONGODB_URI_SPLIT=(${MONGODB_URI//,/ })
+export MONGODB_URI_WITH_MULTIPLE_MONGOSES="${MONGODB_URI:0:10}${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${MONGODB_URI:10}/?tls=true&authSource=admin&compressors=$COMPRESSOR"
```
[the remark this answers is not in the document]
MikalaiMazurenka: The script is using `SERVERLESS_ATLAS_USER` from environment variables.
1. These lines' purpose is to create a complete `MONGODB_URI` to use in the test run via already existing `CoreTestConfiguration` mechanism.
2. Safe unless the the mongodb uri structure itself changes or `create-instance` starts producing entirely different format. It looks scary because it's bash. This manipulation could be done with python, but I'm afraid that won't help, `build.cake` is also an option, but a less preferable one, so I'd stick to current implementation.
3. It's almost the same code but with injected credentials.

evergreen/run-serverless-tests.sh
MikalaiMazurenka: That's a specification requirement.

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
MikalaiMazurenka: Done

tests/MongoDB.Driver.Tests/Specifications/unified-test-format/UnifiedTestFormatValidFailTestRunner.cs
MikalaiMazurenka: Great catch! A leftover from removed category implementation.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -316,28 +316,16 @@ private void KillOpenTransactions(IMongoClient client)
...
-            {
-                if (serverVersion < new SemanticVersion(4, 1, 9) && ex.Code == (int)ServerErrorCode.Interrupted)
-                {
+                catch (MongoCommandException ex) when (
                     // SERVER-38335
-                    return true;
-                }
-                if (ex.Code == (int)ServerErrorCode.Unauthorized)
-                {
+                    serverVersion < new SemanticVersion(4, 1, 9) && ex.Code == (int)ServerErrorCode.Interrupted ||
```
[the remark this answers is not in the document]
MikalaiMazurenka: This implementation is better as it takes less space with the same effect and readability. As for the comment, it's redundant, since it's not applied to C# driver (as mentioned in the comment itself).

MikalaiMazurenka (commented):
Ready for review

evergreen/evergreen.yml
MikalaiMazurenka: > Is is possible to override exec_timeout_secs just for serverless tests?
For example looks like it's done here or here

Missed your comment. Thanks for this example! Looks like it's possible, so I'll try to incorporate this in the next commit.

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -299,6 +305,13 @@ private static ServerApi GetServerApi()
             return new ServerApi(ServerApiVersion.V1);
         }
 
+        private static bool GetServerless()
+        {
+            var serverless = Environment.GetEnvironmentVariable("SERVERLESS");
+
+            return serverless != null;
```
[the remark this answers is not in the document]
MikalaiMazurenka: Sounds like a valid option. I'll need some consideration and will likely incorporate this in the next commit.

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
```diff
@@ -300,6 +301,18 @@ private void SkipTestIfNeeded(BsonDocument definition, BsonDocument test)
                 RequireServer.Check().VersionLessThanOrEqualTo(maxServerVersion.AsString);
             }
 
+            if (definition.TryGetValue("serverless", out var serverless))
+            {
+                if (serverless == "forbid")
```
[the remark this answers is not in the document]
MikalaiMazurenka: Great catch! The specification does mention these values. I'll add them in the next commit.

MikalaiMazurenka (commented):
Acknowledging missed comments!

evergreen/evergreen.yml
DmitryLukyanov: is it mentioned somewhere that it should be `silent`ed? It will hide possible errors that we can see here. Also, at least the patch I checked in java didn't hide this logs

evergreen/run-serverless-tests.sh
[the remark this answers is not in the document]
DmitryLukyanov: I would consider a java [way](https://github.com/mongodb/mongo-java-driver/pull/763/files#diff-c1a39fdd9eb0e51d578e3b88ec364894b2c6dbef3bf40630ebdc6b89c44f830aR24)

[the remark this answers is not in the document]
DmitryLukyanov: also, maybe I'm wrong, but it looks like java uses only single host connection string. Does the spec say anything about it?

evergreen/evergreen.yml
```diff
@@ -632,13 +649,15 @@ functions:
...
       params:
         script: |
-          ${PREPARE_SHELL}
-          for i in $(find ${DRIVERS_TOOLS}/.evergreen ${PROJECT_DIRECTORY}/evergreen -name \*.sh); do
-            cat $i | tr -d '\r' > $i.new
-            mv $i.new $i
-          done
-          # Copy client certificate because symlinks do not work on Windows.
-          cp ${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem ${MONGO_ORCHESTRATION_HOME}/lib/client.pem
+          if [ "Windows_NT" = "$OS" ]; then
```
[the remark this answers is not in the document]
DmitryLukyanov: if it's a windows only fix, then it should be run only there. I think it wasn't added before, because we launched our tests only on windows long time. So, when we added linux/macos support, we just didn't notice this case. It's still a question why it didn't fail all this time, but it's minor

evergreen/evergreen.yml
```diff
@@ -843,6 +890,34 @@ tasks:
       commands:
         - func: publish-snapshot
 
+    - name: serverless-tests-net452
```
DmitryLukyanov, line 893: in other places it's called: `test-#feature#-netstandard21`

MikalaiMazurenka (commented):
Ready for review!

evergreen/evergreen.yml
MikalaiMazurenka: We always hide all the sensitive output in the logs. I'll check why Java doesn't do that (mind you: their PR is not merged yet)

jyemin: It doesn't matter in Java because our whole Evergreen project is private.  I think .NET's is too.

MikalaiMazurenka: Yep, they all are private. Then there's no serious reason to keep it silenced.
My opinion is that as far as I'm concerned till this point nothing sensitive has being written into the logs and if we want to make evergreen project public some day, no extra work would be required. Also, if something happens, the build can be rerun with the silence flag lifted, so it is only useful in terms of saving time and catching volatile errors in the scripts.
@jyemin what do you think? Should the restriction be lifted for the advantage of seeing the exact error?

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
```diff
@@ -257,69 +257,9 @@ private bool CanRunOn(ICluster cluster, BsonDocument requirement)
         {
             foreach (var item in requirement)
             {
-                switch (item.Name)
+                if (!IsRequirementSatisfied(item.Name, item.Value))
```
rstam, line 260: Would this work?

    return requirement.All(item => IsRequirementSatisifed(item.Name, item.value);

or if you change the parameter type of `IsRequirementSatisifed` to `BsonElement`:

    return requirement.All(IsRequirementSatisfied);

Might also be better to rename `requiirement` to `requirements` (plural).

rstam (changes requested):

evergreen/evergreen.yml
```diff
@@ -13,8 +13,7 @@ stepback: true
 command_type: system
 
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
-# 12 minutes is the longest we'll ever run
-exec_timeout_secs: 1800
+exec_timeout_secs: 1800 # 30 minutes
```
DmitryLukyanov, line 16: let's revert the original one since nothing really changed here

evergreen/run-serverless-tests.sh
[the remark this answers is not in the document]
DmitryLukyanov: now, this looks better, thanks

evergreen/run-serverless-tests.sh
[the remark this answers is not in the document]
DmitryLukyanov: > but there is also $OS environment variable which is used in the script but not passed directly

It's passed in other [places](https://github.com/mongodb/mongo-csharp-driver/blob/master/evergreen/evergreen.yml#L301). So even thought there is way(probably) to configure it globally, I would say it's always better to pass it explicitly to see these details without deep investigation.

>(the summary says Supported/used environment variables: which is not the same as passed

Yes, we're not consistent in this behavior (I suppose it was missed during code reviews), but the original intention of this section was to explicitly say which variables are passed and which of them are used as output. I think we should probably define this behavior better, but at the very least let's: 1. mention everything that it's mentioned on the evergreen config side (like SERVERLESS_ATLAS_USER/PASSWORD`) and 2. add `${OS}` as input parameter on evergreen side.

tests/MongoDB.Driver.Tests/Specifications/transactions/tests/README.rst
DmitryLukyanov: yes, we have them in `PinnedShardRouterTests`

evergreen/evergreen.yml
DmitryLukyanov: I think we pass sensitive data a lot for example here `set-temp-fle-aws-creds.sh` and the only option we have in this case is `set +o xtrace # Disable tracing.` for a particular file. But I think the easiest and probably safest value is what we use here: https://github.com/mongodb/mongo-csharp-driver/blob/master/evergreen/evergreen.yml#L283-L290. This way doesn't print this data to the output.

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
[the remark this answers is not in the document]
DmitryLukyanov: I think we need this logic here (I didn't check, but I think the idea is clear):

            if (definition.TryGetElement("serverless", out var serverless))
            {
                RequireServer.Check().RunOn(new BsonArray(new[] { serverless }));
            }

evergreen/evergreen.yml
DmitryLukyanov: most of them. I think a few different ones were missed during code review. No reason to not make unified for new ones

evergreen/evergreen.yml
```diff
@@ -667,6 +713,7 @@ pre:
 post:
   # Removed, causing timeouts
   # - func: upload-working-dir
+  - func: delete-serverless-instance
```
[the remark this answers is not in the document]
DmitryLukyanov: well, this behavior is surprised. The fix that I meant and that I used in LB PR doesn't actually work as expected. I filed a ticket to fix it later. Currently I'm ok with the current way

evergreen/run-serverless-tests.sh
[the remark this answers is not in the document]
MikalaiMazurenka: Done

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
MikalaiMazurenka: Great point!
Done

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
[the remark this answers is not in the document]
MikalaiMazurenka: Great point!
Done

MikalaiMazurenka (commented):
Ready for review!

evergreen/evergreen.yml
DmitryLukyanov: clear naming is this `test-#feature#-netstandard21` . we missed a few cases during previous code reviews, but it's just mistake it doesn't say that we should give up with this until fix that errors

evergreen/evergreen.yml
DmitryLukyanov: it says different, 12 minutes is the longest estimated/validated time for a run (which still might be true if we put aside timeout issues, idk the right value).

DmitryLukyanov: however I don't rely insist here. It's minor anyway. The only painful part that it will be shown in the history where we didn't change anything here. I will rely on other reviewers here

evergreen/evergreen.yml
DmitryLukyanov: missed

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
```diff
@@ -241,72 +252,65 @@ public RequireServer VersionLessThanOrEqualTo(string version)
         }
 
         // private methods
-        private bool CanRunOn(ICluster cluster, BsonDocument requirement)
+        private bool CanRunOn(BsonDocument requirements)
+        {
+            return requirements.All(IsRequirementSatisfied);
+        }
+
+        private bool IsAuthenticated() => CoreTestConfiguration.ConnectionString.Username != null;
```
DmitryLukyanov, line 260: methods order

evergreen/evergreen.yml
MikalaiMazurenka: I think `feature`-tests-`target` is clear naming.
So let's put this aside, since anyway I'm not going to rename every task in scope of this PR so even if I rename the tasks I've added, the issue will still persist.

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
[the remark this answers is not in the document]
MikalaiMazurenka: Unfortunately this didn't work. Reverted to the initial logic.

evergreen/evergreen.yml
DmitryLukyanov: It's not about clear naming or no, it's about the used format in this file. I will rely on @BorisDog and you here

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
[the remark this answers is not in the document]
DmitryLukyanov: I wasn't sure that `BsonArray` will parse `BsonElement`, then it should be:

            if (definition.TryGetElement("serverless", out var serverless))
            {
                RequireServer.Check().RunOn(new BsonArray(new[] { new BsonDocument(serverless.Name, serverless.Value)}));
            }

pay attention on the `TryGetElement` method

MikalaiMazurenka:
Dmitry's suggestion didn't work, the patch build fails:
https://evergreen.mongodb.com/version/60fb1548d6d80a4987269c18
I had to revert the passed environment variables change.

DmitryLukyanov:
> 
> 
> Dmitry's suggestion didn't work, the patch build fails:
> https://evergreen.mongodb.com/version/60fb1548d6d80a4987269c18
> I had to revert the passed environment variables change.

as I wrote, I wasn't sure in original implementation, since you said that it doesn't work, I needed to apply a small fix

tests/MongoDB.Driver.Tests/Specifications/crud/CrudTestRunner.cs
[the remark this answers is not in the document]
MikalaiMazurenka: As I already wrote, `crud-v1` tests will likely become obsolete along with `CrudTestRunner` before any updates are made to the specification, thus I don't see the benefits of updating this code unless necessary.
Current implementation is sufficient and is unlikely to require further changes.

MikalaiMazurenka (commented):
Ready for review

evergreen/evergreen.yml
MikalaiMazurenka: Removed silent, but left the tracing disabled.

evergreen/evergreen.yml
DmitryLukyanov: the matrix style has more options and they're not as obvious as the discussed one. I agreed that it should be standardize as well, but it's not the scope of this ticket. You should revert it.

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/RequireServer.cs
```diff
@@ -241,72 +252,65 @@ public RequireServer VersionLessThanOrEqualTo(string version)
         }
 
         // private methods
-        private bool CanRunOn(ICluster cluster, BsonDocument requirement)
+        private bool CanRunOn(BsonDocument requirements)
```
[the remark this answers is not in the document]
BorisDog: I would encourage us using expression for short methods as well. On a second thought, maybe we don't need `CanRunOn` method at all anymore. 
Up to you.

BorisDog (approved):
LGTM

DmitryLukyanov (approved):
LGTM
we can fix all pending question in next tickets

PR #584, opened 2021-07-21
CSHARP-3659: Use hello command for monitoring (including RTT) if supported.

JamesKovacs:
We look at the previous hello/legacy hello response to determine whether hello or legacy hello should be used for the next heartbeat or RTT command. The only slight complexity is that we re-use the previous command/protocol if we are streaming heartbeats from the server.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/IsMasterResult.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/IsMasterResult.cs
@@ -451,6 +451,17 @@ namespace MongoDB.Driver.Core.Connections
             get { return _wrapped; }
         }
 
+        /// <summary>
+        /// Gets whether the server support the hello command.
+        /// </summary>
+        /// <value>
+        /// True if helloOk:true was included in the response; false otherwise.
+        /// </value>
+        public bool HelloOk
+        {
+            get { return _wrapped.GetValue("helloOk", false).ToBoolean(); }
+        }
+
         // methods
         /// <inheritdoc/>
         public bool Equals(IsMasterResult other)
--- a/src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
@@ -98,10 +98,8 @@ namespace MongoDB.Driver.Core.Servers
         {
             // generate initial server description
             var newDescription = _baseDescription
-                .With(
-                    type: ServerType.LoadBalanced,
-                    reasonChanged: "Initialized",
-                    state: ServerState.Connected);
+                .With(reasonChanged: "Initialized",
+                    state: ServerState.Connected, type: ServerType.LoadBalanced);
             var oldDescription = Interlocked.CompareExchange(ref _currentDescription, value: newDescription, comparand: _currentDescription);
             var eventArgs = new ServerDescriptionChangedEventArgs(oldDescription, newDescription);
 
--- a/src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
@@ -98,8 +98,10 @@ namespace MongoDB.Driver.Core.Servers
         {
             // generate initial server description
             var newDescription = _baseDescription
-                .With(reasonChanged: "Initialized",
-                    state: ServerState.Connected, type: ServerType.LoadBalanced);
+                .With(
+                    type: ServerType.LoadBalanced,
+                    reasonChanged: "Initialized",
+                    state: ServerState.Connected);
             var oldDescription = Interlocked.CompareExchange(ref _currentDescription, value: newDescription, comparand: _currentDescription);
             var eventArgs = new ServerDescriptionChangedEventArgs(oldDescription, newDescription);
 
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
@@ -476,6 +476,7 @@ namespace MongoDB.Driver.Core.Servers
                 EndPointHelper.Equals(_endPoint, other._endPoint) &&
                 Equals(_heartbeatException, other._heartbeatException) &&
                 _heartbeatInterval == other._heartbeatInterval &&
+                _helloOk == other._helloOk &&
                 _lastHeartbeatTimestamp == other.LastHeartbeatTimestamp &&
                 _lastUpdateTimestamp == other._lastUpdateTimestamp &&
                 _lastWriteTimestamp == other._lastWriteTimestamp &&
@@ -505,6 +506,7 @@ namespace MongoDB.Driver.Core.Servers
                 .Hash(_endPoint)
                 .Hash(_heartbeatException)
                 .Hash(_heartbeatInterval)
+                .Hash(_helloOk)
                 .Hash(_lastHeartbeatTimestamp)
                 .Hash(_lastUpdateTimestamp)
                 .Hash(_lastWriteTimestamp)
@@ -666,6 +668,7 @@ namespace MongoDB.Driver.Core.Servers
                 electionId: _electionId,
                 heartbeatException: heartbeatException,
                 heartbeatInterval: _heartbeatInterval,
+                helloOk: _helloOk,
                 lastHeartbeatTimestamp: DateTime.UtcNow,
                 lastUpdateTimestamp: DateTime.UtcNow,
                 lastWriteTimestamp: _lastWriteTimestamp,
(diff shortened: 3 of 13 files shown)


src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -321,7 +320,14 @@ private async Task HeartbeatAsync(CancellationToken cancellationToken)
                     }
                     else
                     {
-                        helloProtocol = helloProtocol ?? InitializeHelloProtocol(connection);
+                        // If MoreToCome is true, that means we are streaming hello or legacy hello results and must
+                        // continue using the existing helloProtocol object.
+                        // Otherwise helloProtocol has either not been initialized or we may need to switch between
+                        // heartbeat commands based on the last heartbeat response.
+                        if (helloProtocol is not {MoreToCome: true})
```
DmitryLukyanov, line 327: This new syntax looks interesting, but it looks like it's just overkill here:
```
if (!helloProtocol.MoreToCome)
```
Maybe I'm missing something, but instead of simple bool check, we're writing much more letters:). Let me know what do you think about it

DmitryLukyanov (approved):
2 minor comments and then LGTM, but before it, can you explain how we can switch between hello formats in a fly?

src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
```diff
@@ -54,6 +54,7 @@ private static bool Equals(Exception x, Exception y)
         private readonly EndPoint _endPoint;
         private readonly Exception _heartbeatException;
         private readonly TimeSpan _heartbeatInterval;
+        private readonly bool _helloOk;
```
DmitryLukyanov, line 57: let's add tests for this

tests/MongoDB.Driver.Core.Tests/Core/Connections/IsMasterResultTests.cs
```diff
@@ -293,5 +293,30 @@ public void GetReplicaSetConfig_should_return_correct_info_when_the_server_is_a_
...
             config.Version.Should().Be(20);
         }
+
+        [Fact]
+        public void HelloOk_should_return_true_when_response_contains_helloOk_true()
+        {
+            var doc = new BsonDocument
+            {
+                { "ok", 1 },
+                { "helloOk", true }
```
DmitryLukyanov, line 303: what if the value missed?

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
JamesKovacs: The check is actually more complex because `helloProtocol` can be null. It is equivalent to:

```
if ( !(helloProtocol?.MoreToCome ?? false) )
```

I started with the above and Rider suggested simplifications that landed me on the new syntax, which seemed easier to read.

src/MongoDB.Driver.Core/Core/Connections/IsMasterResult.cs
```diff
@@ -451,6 +451,17 @@ public BsonDocument Wrapped
             get { return _wrapped; }
         }
 
+        /// <summary>
+        /// Gets whether the server support the hello command.
+        /// </summary>
+        /// <value>
+        /// True if helloOk:true was included in the response; false otherwise.
```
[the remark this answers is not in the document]
JamesKovacs: We use this elsewhere such as ClientSessionOptions.cs line 45, BsonTrie.cs line 81, 83, IByteBuffer.cs line 39, ...

src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -98,10 +98,8 @@ protected override void InitializeSubClass()
         {
             // generate initial server description
             var newDescription = _baseDescription
-                .With(
-                    type: ServerType.LoadBalanced,
-                    reasonChanged: "Initialized",
-                    state: ServerState.Connected);
+                .With(reasonChanged: "Initialized",
```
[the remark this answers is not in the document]
JamesKovacs: The reformatting was introduced by the refactoring to add the `helloOk` parameter and I thought I had reverted all of them. Missed this one. Fixed.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
DmitryLukyanov: interesting, then i'm ok with this, maybe only add `space` between `{` and `}`?

src/MongoDB.Driver.Core/Core/Connections/IsMasterResult.cs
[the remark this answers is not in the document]
DmitryLukyanov: Ok, i will rely on you here :)

src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
JamesKovacs: Good point. Added unit tests and found a bug. Fixed.

tests/MongoDB.Driver.Core.Tests/Core/Connections/IsMasterResultTests.cs
JamesKovacs: The next test on line 311 covers the case where `helloOk: true` is not present in the response. `helloOk: false` will never appear in the response, which is why we don't have a test for it.

DmitryLukyanov: got it

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerDescriptionTests.cs
```diff
@@ -50,6 +50,7 @@ public void Constructor_with_serverId_and_endPoint_only_should_return_disconnect
             subject.CanonicalEndPoint.Should().BeNull();
             subject.ElectionId.Should().BeNull();
             subject.EndPoint.Should().Be(__endPoint);
+            subject.HelloOk.Should().BeFalse();
```
MikalaiMazurenka, line 53: This test: `Equals_should_return_false_when_any_field_is_not_equal` should be updated as well.

src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
```diff
@@ -220,6 +224,17 @@ public TimeSpan HeartbeatInterval
             get { return _heartbeatInterval; }
         }
 
+        /// <summary>
+        /// Whether the server supports the hello command.
+        /// </summary>
+        /// <value>
+        /// True if the server responded with helloOk:true; false otherwise.
+        /// </value>
+        public bool HelloOk
```
MikalaiMazurenka, line 233: Regarding `.ToString()` method. It seems fine that `helloOk` is not there, but I would consider adding it if there are any advantages to it which I don't see atm.

MikalaiMazurenka (commented):
Looks great! My only major concern is about the readability of certain `if` statement.

src/MongoDB.Driver.Core/Core/Servers/ServerDescription.cs
JamesKovacs: I purposely didn't add `helloOk` to `ToString()` as I didn't see the value in exposing it either.

tests/MongoDB.Driver.Core.Tests/Core/Connections/HelloHelperTests.cs
```diff
@@ -21,23 +21,22 @@
...
+        [Theory]
+        [InlineData(true, false, "{ hello : 1, helloOk : true }")]
+        [InlineData(false, false, "{ isMaster : 1, helloOk : true }")]
+        [InlineData(false, true, "{ hello : 1, helloOk : true }")]
+        [InlineData(true, true, "{ hello : 1, helloOk : true }")]
+        public void CreateCommand_should_return_correct_hello_command(bool useServerApiVersion, bool helloOk, string expectedResult)
         {
-            var command = HelloHelper.CreateCommand(null);
-            command.Should().Be($"{{ isMaster : 1 }}");
+            ServerApi serverApi = useServerApiVersion? new ServerApi(ServerApiVersion.V1) : null;
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerDescriptionTests.cs
JamesKovacs: Fixed and found multiple test bugs in the process - which are now also fixed. Thanks!

MikalaiMazurenka (approved):
LGTM!

PR #585, opened 2021-07-23
CSHARP-3661: Remove oppressive language from the documentation.

JamesKovacs:
The only remaining occurrences of oppressive language in our docs are in the following places:

- the API docs (which will be handled in CSHARP-3660)
- third-party dependencies (jquery.js, underscore.js, etc.)
- explanation that `slaveOk=true` must be replaced with `readPreference` in What's New?


jyemin (approved):
LGTM

PR #587, opened 2021-07-27
CSHARP-3733: Don't modify the value of environment variables in test code.

DmitryLukyanov:
This change looks correct in any case. But it makes test code a bit more complicated than it was all the time before. Another option, can be just disabling LB for local test. But I think we will need to do the same change in future anyway, so probably it's better to add this now

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/LoadBalancingIntergationTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/LoadBalancingIntergationTests.cs
@@ -23,9 +23,11 @@ using MongoDB.Bson.TestHelpers;
 using MongoDB.Bson.TestHelpers.XunitExtensions;
 using MongoDB.Driver.Core.Bindings;
 using MongoDB.Driver.Core.Clusters;
+using MongoDB.Driver.Core.Configuration;
 using MongoDB.Driver.Core.Connections;
 using MongoDB.Driver.Core.Events;
 using MongoDB.Driver.Core.Operations;
+using MongoDB.Driver.Core.TestHelpers;
 using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
 using MongoDB.Driver.TestHelpers;
 using Xunit;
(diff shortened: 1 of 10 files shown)


tests/MongoDB.Driver.Core.Tests/LoadBalancingIntergationTests.cs
```diff
@@ -734,11 +736,9 @@ private void SkipIfNotLoadBalancingMode()
             RequirePlatform.Check().SkipWhen(SupportedOperatingSystem.MacOS);
             // Make sure that LB is started. "nginx" is a LB we use for windows testing
             RequireEnvironment.Check().ProcessStarted("nginx");
-            Environment.SetEnvironmentVariable("MONGODB_URI", "mongodb://localhost:17017?loadBalanced=true");
-            Environment.SetEnvironmentVariable("MONGODB_URI_WITH_MULTIPLE_MONGOSES", "mongodb://localhost:17018?loadBalanced=true");
             RequireServer
-                .Check()
-                .LoadBalancing(enabled: true, ignorePreviousSetup: true)
+                .ConfigureAndCheck(new LoadBalancedCoreEnvironmentConfiguration())
```
rstam, line 740: I don't have time to review more thoroughly but I was expecting the current

    RequireServer.Check()... // checks requirements against default test server (i.e. cluster)

to be enhanced with a new overload like:

    RequireServer.Check(cluster)... // checks requirements against a specified cluster

The sheer volume of changes in this PR makes me think there has to be a better way.

rstam (changes requested):

tests/MongoDB.Driver.Core.Tests/LoadBalancingIntergationTests.cs
DmitryLukyanov: technically, I have almost the same change but which is a bit more complicated since the same logic should be applied for unified test runner.
Current logic:
1. I moved all cluster creating logic from CoreTestConfiguration into a new static helper `ClusterBuilderHelper`.
2. `ClusterTestWrapper` takes an ICluster as ctor argument. Then we can use the provided cluster to call helper logic from CoreTestConfiguration
3. `CoreEnvironmentConfiguration` - a class that is responsible for working with env variables. This is the main target for LB test runner that allows us working with different env variables values
4. `TestClientsProvider` a factory for clients based on `CoreEnvironmentConfiguration`

we can discuss it later

DmitryLukyanov: Another option that I just found, we can create a new `DEBUG-like` configuration value, for example "LOCAL" or "UNSAFE" (or whatever) which can be specified only intentionally. I think It will remove any problems here

rstam (changes requested):
I've spent some more time attempting to review this PR this morning. I don't think this approach is viable.

I think you should try a new approach. 

The key new requirement is that we now want to have more than one `CoreTestConfiguration`. A default one (like now), and a new one for testing against a loadbalancer. And maybe more in the future.

The way to do that should be to change the `CoreTestConfiguration` class from a `static` class to a non-static class that we can have multiple instances of.

A static property called `CoreTestConfiguration.Default` can be used to point to the default instance of the `CoreTestConfiguration` which would correspond to the current static one.

A similar approach could be used for `DriverTestConfiguration`.

And `RequireServer.Check()` should have a new `RequireServer.Check(cluster)` overload that you can use when running tests against the non-default cluster.

DmitryLukyanov:
discussed offline, for now we didn't find good enough and at the same time a simple enough solution, so moving this PR to hold and related ticket to Open

JamesKovacs:
Closing this PR based on previous comments.

PR #588, opened 2021-07-29
CSHARP-3767: Skip tests that depend on the legacy wire protocol.

DmitryLukyanov:


tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
```diff
@@ -264,7 +272,7 @@ public void TestAggregateMaxTime()
         [SkippableFact]
         public void TestAggregateWriteConcern()
         {
-            RequireServer.Check().Supports(Feature.CommandsThatWriteAcceptWriteConcern).ClusterType(ClusterType.ReplicaSet);
+            RequireServer.Check().Supports(Feature.CommandsThatWriteAcceptWriteConcern, Feature.LegacyWireProtocol).ClusterType(ClusterType.ReplicaSet);
```
jyemin, line 275: What does this test have to do with the legacy wire protocol?

tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
```diff
@@ -1488,9 +1556,11 @@ public void TestFindOneAsWithMaxTime()
             }
         }
 
-        [Fact]
+        [SkippableFact]
         public void TestFindOneAsWithSkipAndSortyBy()
         {
+            RequireServer.Check().Supports(Feature.LegacyWireProtocol);
```
jyemin, line 1562: Huh... Does the Legacy driver not use FindOperation? Or is is the Insert that's failing?

JamesKovacs (commented):
How did you identify which tests to disable? Was it simply based on build failures? If there are tests that are using legacy wire protocol for their setup (e.g. OP_INSERT), but are testing something else, we should modify the setup to use OP_MSG for setup. We should only be disabling tests on 5.1+ that are *testing* legacy wire protocol, not tests that happen to be using legacy wire protocol as a side effect.

tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
```diff
@@ -3175,17 +3296,25 @@ public void TestSortAndLimit()
...
         public void TestGetStatsNoPadding()
         {
-            RequireServer.Check().VersionGreaterThanOrEqualTo("3.0").ClusterTypes(ClusterType.Standalone, ClusterType.ReplicaSet).StorageEngine("mmapv1");
+            RequireServer
+                .Check()
+                .VersionGreaterThanOrEqualTo("3.0")
+                .ClusterTypes(ClusterType.Standalone, ClusterType.ReplicaSet)
+                .StorageEngine("mmapv1")
+                .Supports(Feature.LegacyWireProtocol);
+
```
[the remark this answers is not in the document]
jyemin: Maybe we should remove, but let's do it in scope of a different ticket.

tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
DmitryLukyanov: it's interesting, I saw this error in original patch https://evergreen.mongodb.com/task_log_raw/dot_net_driver_unsecure_tests__version~latest_os~windows_64_topology~standalone_auth~noauth_ssl~nossl_test_net452_patch_44a8a50402fe554d86ecf4cefd9fc5610a85542a_6102d26e0ae6060253c430f3_21_07_29_16_08_34/0?type=T#L2198 but don't see it anymore. I need to check why it doesn't fail anymore

tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
DmitryLukyanov: similar to above, I need to check, but legacy uses FindCommand. So in this case it should not use the legacy wire protocol

DmitryLukyanov:
> 
> How did you identify which tests to disable? Was it simply based on build failures? If there are tests that are using legacy wire protocol for their setup (e.g. OP_INSERT), but are testing something else, we should modify the setup to use OP_MSG for setup. We should only be disabling tests on 5.1+ that are _testing_ legacy wire protocol, not tests that happen to be using legacy wire protocol as a side effect.

I will recheck it, I disabled all tests that failed on latest.

tests/MongoDB.Driver.Legacy.Tests/MongoCollectionTests.cs
DmitryLukyanov: it looks like most of the failed tests were affected by previous failed tests. I'm not sure how exactly it happens, but when I disabled only tests that use `unacknowledged  writes` or `$explain` (that use legacy wire protocol) all tests passed. I think it's enough for this ticket

tests/MongoDB.Driver.Legacy.Tests/Linq/ExplainTests.cs
```diff
@@ -41,9 +42,11 @@ public ExplainTests()
             _collection = LegacyTestConfiguration.Collection;
         }
 
-        [Fact]
+        [SkippableFact]
         public void TestExplainFromLinqQueryEqualsExplainFromCursor()
         {
+            RequireServer.Check().Supports(Feature.LegacyWireProtocol);
```
jyemin, line 48: Any idea what this test fails?  It doesn't seem to use the legacy protocol.

tests/MongoDB.Driver.Tests/Linq/MongoQueryableTests.cs
```diff
@@ -556,11 +556,8 @@ public void GroupJoin_syntax_with_select_many_and_DefaultIfEmpty()
         {
             RequireServer.Check().VersionGreaterThanOrEqualTo("3.2.0");
             // remove the skip exception when SERVER-58680 is fixed
-            var actualVersion = CoreTestConfiguration.ServerVersion;
```
jyemin, line 559: Seems like this should be in a separate commit, at the least.

tests/MongoDB.Driver.Legacy.Tests/Linq/ExplainTests.cs
DmitryLukyanov: if there is $explain, then we use old protocol

tests/MongoDB.Driver.Tests/Linq/MongoQueryableTests.cs
DmitryLukyanov: i don't mind

jyemin: ok, please do that before merging.

jyemin (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/MongoQueryableTests.cs
DmitryLukyanov: done

PR #589, opened 2021-08-02
CSHARP-3765: Clean-up a stop step for LB and Serverless tests.

DmitryLukyanov:
I didn't make any changes for LB since when I make the same change for stopping LB, the whole scripts fails with this error https://evergreen.mongodb.com/task_log_raw/dot_net_driver_load_balancer_tests_secure__version~latest_os~ubuntu_1804_topology~sharded_cluster_auth~auth_ssl~ssl_test_load_balancer_netstandard21_patch_a563175b58fffa0f0ca754c735caa4e32234aaad_6105f384d1fe071f0cb0c6d9_21_08_01_01_06_30/0?type=T#L1614. It looks like stopping logic should be done in the same scope as the main steps. Yes, if something will fail, the stopping logic won't be called. I asked @durran how we can make deal with this, since in the node he applied the same change

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -9,7 +9,7 @@ using System.Text.RegularExpressions;
 using System.Linq;
 using Cake.Common.Tools.DotNetCore.DotNetCoreVerbosity;
 
-var defaultTarget = "Default";
+const string defaultTarget = "Default";
 var target = Argument("target", defaultTarget);
 var configuration = Argument("configuration", "Release");
 
--- a/build.cake
+++ b/build.cake
@@ -9,7 +9,8 @@ using System.Text.RegularExpressions;
 using System.Linq;
 using Cake.Common.Tools.DotNetCore.DotNetCoreVerbosity;
 
-var target = Argument("target", "Default");
+var defaultTarget = "Default";
+var target = Argument("target", defaultTarget);
 var configuration = Argument("configuration", "Release");
 
 var gitVersion = GitVersion();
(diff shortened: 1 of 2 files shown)


BorisDog (approved):

build.cake
```diff
@@ -562,3 +563,11 @@ Task("DumpGitVersion")
     });
 
 RunTarget(target);
+
+void ThrowIfNotDefaultTarget(string @value)
```
DmitryLukyanov, line 567: we need to be able to run all tests if we don't specify any target as argument. But if we specify something, it should be parsed

evergreen/evergreen.yml
```diff
@@ -762,27 +767,27 @@ tasks:
         - func: bootstrap-mongo-orchestration
         - func: run-load-balancer
         - func: run-load-balancer-tests
-        - func: stop-load-balancer
           vars:
             FRAMEWORK: netstandard15
+        - func: stop-load-balancer
```
DmitryLukyanov, line 772: previously, the FRAMEWORK value was passed only into `stop-load-balancer` step

DmitryLukyanov: Also, it looks like this step is not 100% required, since the LB process is killed anyway when the current task is finished (I checked it with `ps -ax | grep haprox`), but I'm still not 100% sure in the details how it happens, but I would say it's harmless to have this step to ensure at least sometimes (if previous steps is not failed), we don't force killing LB process and run a special script for it

DmitryLukyanov: maybe it makes sense to add a comment about it?

build.cake
```diff
@@ -9,7 +9,8 @@ using System.Text.RegularExpressions;
 using System.Linq;
 using Cake.Common.Tools.DotNetCore.DotNetCoreVerbosity;
 
-var target = Argument("target", "Default");
+var defaultTarget = "Default";
```
BorisDog, line 12: nit: const?

DmitryLukyanov: done

PR #596, opened 2021-08-04
CSHARP-3781: Zstandard EG variant should set "zstd" as compression setup.

DmitryLukyanov:


jyemin (changes requested):
Can we limit this to just fixing the evergreen configuration?  I don't like mixing testing configuration fixes with improvements to the production code, in a single PR (let alone a single commit).

DmitryLukyanov:
ok, I will move the rest of the code into a different ticket

DmitryLukyanov:
Test EG: https://evergreen.mongodb.com/version/610b0362a4cf4727698858bd that will show NotImplemented errors, because I temporary changed the ZstandardCompressor to fail in Decompress. This will prove that this step was actually called

jyemin (approved):

PR #598, opened 2021-08-05
CSHARP-3784: Create a test to validate url options provided by EG.

DmitryLukyanov:


tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -154,6 +156,22 @@ public void Resolve_with_resolveHosts_should_return_expected_result(string conne
...
+        [Fact]
+        public void ServerConfiguration_should_have_only_expected_arguments()
+        {
+            Assert(CoreTestConfiguration.ConnectionString);
+            Assert(CoreTestConfiguration.ConnectionStringWithMultipleShardRouters);
+
+            void Assert(ConnectionString connectionString)
+            {
+                var nameValueCollection = connectionString._unknownOptions();
+                var collection = from key in nameValueCollection.Cast<string>()
```
BorisDog, line 168: Would `connectionString._unknownOptions().Any().ShouldBeFalse()` be sufficient?

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -1236,4 +1254,9 @@ public void When_calling_resolve_on_a_native_connection_string()
             resolved.Should().BeSameAs(subject);
         }
     }
+
+    public static class ConnectionStringReflector
+    {
+        public static NameValueCollection _unknownOptions(this ConnectionString connectionString) => (NameValueCollection)Reflector.GetFieldValue(connectionString, nameof(_unknownOptions));
```
BorisDog, line 1260: Can we use `connectionString.AllUnknownOptionNames` ?

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
DmitryLukyanov: yes, but I think it would be helpful to see what options exactly are missed in the test configuration

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
DmitryLukyanov: similar to the above, in this case we won't see what value is incorrect, only the key. I think it's better to see all details without "debugging"

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionStringTests.cs
BorisDog: Got it.

BorisDog (approved):
LGTM

PR #600, opened 2021-08-11
CSHARP-2914: Support Zstandard on Linux and MacOS.

DmitryLukyanov:
Steps to generate zstd libraries:
1. clone the git repo: https://github.com/facebook/zstd
2. `git submodule update --init` (not sure that this is required)
3. call `make` from the root folder
4. Check this path: `root_folder/lib/libzstd.so|.dylib`


src/MongoDB.Driver.Core/Core/Compression/Snappy/Snappy32NativeMethods.cs
```diff
@@ -108,12 +108,12 @@ private class SnappyLocator : RelativeLibraryLocatorBase
         {
             public override bool IsX32ModeSupported => true;
 
-            public override string GetLibraryRelativePath(OperatingSystemPlatform currentPlatform)
+            public override string GetLibraryName(OperatingSystemPlatform currentPlatform)
```
DmitryLukyanov, line 111: minor refactoring to avoid specifying the same relative path all the time

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/DarwinLibraryLoader.cs
```diff
@@ -54,11 +54,13 @@ public IntPtr GetFunctionPointer(string name)
         // nested types
         private static class NativeMethods
         {
-            [DllImport("libdl", CharSet = CharSet.Unicode)]
+#pragma warning disable CA2101 // Specify marshaling for P/Invoke string arguments
```
DmitryLukyanov, line 57: The base commit has changed logic comparing it with libmongocrypt bindings, I think originally I changed it to avoid CA2101 warning, but I didn't have a way to check that it's valid change. Now when I finally tried, I noticed that the change is invalid. So reverted it to the original one

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
```diff
@@ -26,40 +26,50 @@ internal abstract class RelativeLibraryLocatorBase : ILibraryLocator
...
+            return GetAbsolutePath(relativePath, libraryName);
         }
 
-        public virtual string GetLibraryBaseAssemblyPath()
+        public virtual string GetLibraryDirectoryRelativePath(OperatingSystemPlatform currentPlatform)
         {
-            var baseAssemblyPathUri = GetBaseAssemblyUri();
-            var uri = new Uri(baseAssemblyPathUri);
-            return Uri.UnescapeDataString(uri.AbsolutePath);
+            var rid = GetCurrentPlatformRuntimeIdentifier(currentPlatform);
```
DmitryLukyanov, line 48: this logic is always the same, so moved it to the helper level

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -58,7 +58,9 @@
 
   <!--Compression-->
   <PropertyGroup>
-    <CompressionRuntimesPath>runtimes/win/native</CompressionRuntimesPath>
+    <CompressionWinRuntimesPath>runtimes/win/native</CompressionWinRuntimesPath>
```
DmitryLukyanov, line 61: I will add more comments about logic in this and target files in the next commit

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -94,28 +96,58 @@
...
-  </ItemGroup>
-
   <PropertyGroup>
-    <ZstdBinaries>Core/Compression/Zstandard/lib/win</ZstdBinaries>
+    <ZstdWinBinaries>Core/Compression/Zstandard/lib/win</ZstdWinBinaries>
+    <ZstdLinuxBinaries>Core/Compression/Zstandard/lib/linux</ZstdLinuxBinaries>
+    <ZstdOsxBinaries>Core/Compression/Zstandard/lib/osx</ZstdOsxBinaries>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(IsWindows)' == true ">
```
DmitryLukyanov, line 105: the validation on OS was added here by mistake, this condition is compile time only, so since we build our nugets on windows, it will always give `true`

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -20,20 +20,27 @@
...
         private static string __testMessage = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz";
-#if WINDOWS
-        [Theory]
+
+        [SkippableTheory]
         [InlineData(CompressorType.Snappy)]
         [InlineData(CompressorType.ZStandard)]
         public void Compressor_should_read_the_previously_written_message(CompressorType compressorType)
         {
+            RequirePlatform
```
DmitryLukyanov, line 39: will remove this validation when snappy will be supported on non windows OSs

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/RelativeLibraryLocatorBase.cs
```diff
@@ -26,40 +26,50 @@ internal abstract class RelativeLibraryLocatorBase : ILibraryLocator
         public virtual bool IsX32ModeSupported => false;
 
         // public methods
+        public string GetBaseAssemblyDirectory()
```
DmitryLukyanov, line 29: minor refactoring + renaming, but the only real changes are:
1. Moved relative path logic on this helper level
2. Now we expect that the main driver will configure only the library name

DmitryLukyanov:
The pending question, how to configure whether it should be generated x32 or x64 library, will try to check the facebook guide/repo to find this info

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -74,11 +76,11 @@
   <ItemGroup Condition=" '$(IsWindows)' == true ">
```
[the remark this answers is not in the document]
DmitryLukyanov: i thought i removed all these conditions, so this was missed, will remove it in the next commit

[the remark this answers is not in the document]
JamesKovacs: Sounds good. 👍

build.cake
```diff
@@ -561,4 +564,212 @@ Task("DumpGitVersion")
         Information(gitVersion.Dump());
     });
 
+Task("PreparePackagingTests")
```
DmitryLukyanov, line 567: this is the scope of https://jira.mongodb.org/browse/CSHARP-2933, but I think this is useful to be add here

build.cake
```diff
@@ -561,4 +564,212 @@ Task("DumpGitVersion")
...
+        {
+            throw new Exception($"Package {mongoDbDriverPackageName} must be presented and unique.");
+        }
+        var mongoDriverPackageVersion = packagesList.Single(p => p.Name == mongoDbDriverPackageName).Version;
+        Information($"Package version {mongoDriverPackageVersion}");
+
+        var testDetails = new List<(string Moniker, string CsprojType, string Version, string Bitness, string Command)>
+        {
+            /* NonSDK will be added later
+            { ("v4.5.2", "NonSdk", mongoDriverPackageVersion) },
```
DmitryLukyanov, line 637: I didn't find a good way to create non-sdk app and consume the nuget. I will create a separate ticket for this

build.cake
```diff
@@ -561,4 +564,212 @@ Task("DumpGitVersion")
...
+        CleanDirectory(monikerTestFolder);
+
+        var csprojFileName = $"{monikerTestFolder.GetDirectoryName()}.csproj"; 
+        var csprojFullPath = monikerTestFolder.CombineWithFilePath(csprojFileName);
+
+        switch (command)
+        {
+            case "xunit":
+                {
+                    if (moniker == "net452" || moniker == "net472")
```
DmitryLukyanov, line 673: when we run tests for .net framework and if we don't consume the nuget directly. Then .net framework cannot resolve native libraries correctly. I saw this behavior in other nugets, so it can be not our issue, I will create a ticket to investigate it further (I have no good results for now)

DmitryLukyanov: >if we don't consume the nuget directly.

I mean if we consume MongoDB.Driver.Core nuget, then everything works fine, if we consume it indirectly (via MongoDB.Driver) then it doesn't work

build.cake
```diff
@@ -561,4 +564,212 @@ Task("DumpGitVersion")
...
+                            Framework = moniker,
+                            Configuration = configuration,
+                            ArgumentCustomization = args => args.Append($"-- RunConfiguration.TargetPlatform={testDetails.Bitness}")
+                        }
+                    );
+                } 
+                break;
+            case "console":
+                {
+                    Information("Creating console project..");
```
DmitryLukyanov, line 719: added this case, because console and xunit have a bit different nuget consuming behavior for .net framework

tests/MongoDB.Driver.Tests/EncryptionTests.cs
```diff
@@ -184,12 +184,6 @@ public void Mongocryptd_should_be_initialized_only_for_auto_encryption([Values(f
             }
         }
 
-        [Fact]
-        public void Mongocrypt_native_library_should_have_expected_version()
```
DmitryLukyanov, line 188: moved this test into a different file

tests/MongoDB.Driver.Tests/Packaging/PackagingTests.cs
```diff
@@ -0,0 +1,72 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+// this directive is configured/used only in cake packaging tests
+#if CONSOLE_TEST
```
DmitryLukyanov, line 17: this directive is set only in packaging tests and allows us to run this test as console application :)

build.cake
JamesKovacs: Minor nit. An ellipsis is three dots (...) not two.

evergreen/evergreen.yml
```diff
@@ -1436,6 +1436,24 @@ buildvariants:
...
      - name: test-netstandard21
 
+- matrix_name: "tests-zstandard-compression-linux"
+  matrix_spec: { compressor : "zstandard", auth: "noauth", ssl: "nossl", version: ["3.6", "4.0", "4.2", "4.4", "5.0", "latest"], topology: "standalone", os: "ubuntu-1804" }
+  display_name: "${version} ${compressor} ${topology} ${auth} ${ssl} ${os} "
+  tags: ["tests-variant"]
+  tasks:
+     - name: test-netstandard15
+     - name: test-netstandard20
+     - name: test-netstandard21
```
JamesKovacs, line 1446: CSHARP-2714 updates net452->net472 and removes netstandard15. Shouldn't we merge those changes in first rather than adding new targets that will be removed soon?

src/MongoDB.Driver.Core/MongoDB.Driver.Core.targets
```diff
@@ -2,26 +2,43 @@
...
+        <Link>%(FileName)%(Extension)</Link>
+      </Content>
+    </ItemGroup>
+    <!--macos-->
+    <ItemGroup>
+      <Content Include="$(MSBuildThisFileDirectory)../$(CompressionOsxRuntimesPath)/libzstd.dylib">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <Link>%(FileName)%(Extension)</Link>
+      </Content>
+    </ItemGroup>
```
JamesKovacs, line 43: Rather than special-casing snappy/zstd in all their various combinations of OS and bitness, why don't we have a more general directive that packages everything in compression runtimes into the nuget package? We would just need to match the layout in the repo to the layout in the nuget package - which we already do.

JamesKovacs (approved):
LGTM. Mentioned some minor suggestions that you can consider.

build.cake
```diff
@@ -561,4 +564,212 @@ Task("DumpGitVersion")
...
+            /* NonSDK will be added later
+            { ("v4.5.2", "NonSdk", mongoDriverPackageVersion) },
+            { ("v4.7.2", "NonSdk", mongoDriverPackageVersion) }, */
+
+            { ("net452", "SDK", mongoDriverPackageVersion, "x64", Command : "xunit") }, // TODO: add x32
+            { ("net472", "SDK", mongoDriverPackageVersion, "x64", Command: "xunit") },
+            { ("netcoreapp21", "SDK", mongoDriverPackageVersion, "x64", Command: "xunit") },
+            { ("netcoreapp30", "SDK", mongoDriverPackageVersion, "x64", Command: "xunit") },
+            { ("net50", "SDK", mongoDriverPackageVersion, "x64", Command: "xunit") }
+        };
```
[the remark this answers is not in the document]
DmitryLukyanov: yep, I removed netstadard1.5, but forgot about net452, will do it

evergreen/evergreen.yml
DmitryLukyanov: it's not pushed yet, so this change will be added later

src/MongoDB.Driver.Core/MongoDB.Driver.Core.targets
DmitryLukyanov: sorry, not sure that I understood it, can you please describe it a bit more?

JamesKovacs: Discussed via Slack.

evergreen/evergreen.yml
JamesKovacs: ACK.

PR #603, opened 2021-08-16
CSHARP-3799: Fix flacky AcquireConnection_should_iterate_over_all_dormant_connections test

BorisDog:
AcquireConnection_should_iterate_over_all_dormant_connections timings tuning

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
@@ -119,7 +119,7 @@ namespace MongoDB.Driver.Core.ConnectionPools
             var settings = _settings.With(
                 minConnections: 0,
                 maxConnections: connectionsCount,
-                waitQueueTimeout: TimeSpan.FromMilliseconds(1),
+                waitQueueTimeout: TimeSpan.FromMilliseconds(20),
                 maintenanceInterval: TimeSpan.FromMilliseconds(10000));
 
             var connectionsCreated = new HashSet<ConnectionId>();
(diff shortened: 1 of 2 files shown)


tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -119,7 +119,7 @@ public void AcquireConnection_should_iterate_over_all_dormant_connections()
             var settings = _settings.With(
                 minConnections: 0,
                 maxConnections: connectionsCount,
-                waitQueueTimeout: TimeSpan.FromMilliseconds(1),
+                waitQueueTimeout: TimeSpan.FromMilliseconds(20),
```
DmitryLukyanov, line 122: I'm not sure about the test idea, but the change looks safe. How often this test fails in master? Are we sure that it fixes it reliably?

DmitryLukyanov (approved):
One question and then LGTM

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
BorisDog: Approximately I observe it once in full EG run lately. Not sure 100% re. reliability, but 20ms has a big buffer over the worst case values I've seen in EG failures.

DmitryLukyanov: maybe try to run like 5 full patches to ensure that it's actually fixed and then we will good to go with this I guess

BorisDog: No signs of it in 3 full runs.

PR #604, opened 2021-08-16
CSHARP-3642: Fix race condition in the pool-checkout-no-idle test

BorisDog:


src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -66,7 +66,7 @@ public class ConnectionPoolSettings
             Optional<bool> isPausable = default(Optional<bool>),
             Optional<int> maxConnecting = default(Optional<int>))
         {
-            _maintenanceInterval = Ensure.IsInfiniteOrGreaterThanOrEqualToZero(maintenanceInterval.WithDefault(TimeSpan.FromMinutes(1)), nameof(maintenanceInterval));
+            _maintenanceInterval = maintenanceInterval.WithDefault(TimeSpan.FromMinutes(1));
```
BorisDog, line 69: We need to be able to disable background thread. In spec, that would be marked any negative value for background thread interval. Unfortunately we already support Infinite timeout, so the options are to use any other negative values, or a specific one.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -293,6 +293,12 @@ public int GetGeneration(ObjectId? serviceId)
         // private methods
         private async Task MaintainSizeAsync(CancellationToken cancellationToken)
         {
+            if (_settings.MaintenanceInterval != Timeout.InfiniteTimeSpan &&
```
DmitryLukyanov, line 296: I didn't find discussed approach that allows us to not run anything. So, since this option is only test related, I think we need to use reflection to disable this rather than introduce any special logic for this

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
DmitryLukyanov: Having two validations rules to initialize the same value is definitely wrong here. I think we should use reflection to achieve required behavior.

DmitryLukyanov (changes requested):

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: it's also possible to introduce a new option like `EnableMaintenance`, but I think that reflection is enough and the best way here

BorisDog: I prefer to use the same parameter that is specified in spec for disabling the thread. And I think internal is more suitable option over reflection.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
DmitryLukyanov: The purpose is to follow the same timespan agreements everywhere and do not introduce two different validation logic for the same field

BorisDog: I suggest to follow the spec "A negative value: never begin a Run.". But without exposing it publicly and not introducing BC by continuing supporting `Timestamp.Infinite`.

DmitryLukyanov: we fully follow to the spec only on the connection string level(when we parse connection string option) where we still work with `int` values, and use a C# agreements for Timespan on higher levels (where we work with c# object). Also, as it's mentioned here https://github.com/mongodb/mongo-csharp-driver/pull/604/files#diff-25f3f5fd3adcf23fba2d5d70a3122707faa25bba7155323c24e76fc9aeabbf14R41, this option is only test-related, so in any case no reason to extend it on any levels other than test

DmitryLukyanov: not fully the same, but similar logic can be found here https://github.com/mongodb/mongo-csharp-driver/commit/9c34b84089bfe124b6106986c96500197dd916d1#diff-f61be6b500f0cfc0bf6a15772c7b4ac352fe74be4c75f347563c1498e764dfccR20

BorisDog: Great, that's why we are keeping this option internal only. While there is no reason to depart from `backgroundThreadIntervalMS` definition in spec. 
I think this aligns well with our architecture, as we have single parameter to control the thread behaviour.

DmitryLukyanov: the reason is consistency and keeping simple things simple :)

BorisDog: That's exactly the intention. Extending the existing parameter, which also aligns well with spec.

DmitryLukyanov: I disagreed. 
>simple things simple
You introduce two different ways to work with the same field, it's not simple. The complexity here is introduced for test reason and I think should be limited to that level only.  

>consistency
your solution contradicts to consistency since we follow to the C# agreements to work with Timespans on any level higher than connection string

BorisDog: > You introduce two different ways to work with the same field
This is not accurate, adding addition option to the same field, while keeping the others as is. So extending existing behaviour, not adding a "new way".

> C# agreements 
We can't do much in this area, as `Timespan.Zero` and `Timespan.Infinite` are already part of public API and have different meaning. Negative value as specified in spec is perfectly suitable for the current purpose.

DmitryLukyanov: >So extending existing behaviour, not adding a "new way".

it's wording question, above I wrote `Having two validations rules`.

>We can't do much in this area,

it's not true, we have a lot of options. The best I see is to use reflection and don't change the main code at all (other than possible minor refactoring). It's also possible to introduce a new internal enabling field (but I think it's not necessary).

The key point, that how we handle such options are already defined and keeping introduce any new rules in this area will break consistency.

JamesKovacs: `Timeout.InfiniteTimeSpan` exists but `TimeSpan.Infinite` does not. Some APIs (like `System.Threading`) interpret a `TimeSpan` of -1 ms to be infinite, but this is convention within that API. `TimeSpan` itself doesn't have a concept of infinitely (which was a poor original design choice IMHO). Even the .NET Framework design team has argued about this and closed it as "won't fix":

https://github.com/dotnet/runtime/issues/30723

My preference is to keep special values from specs in our parsing code and not let them leak into our own internal code. So if the spec says "A negative value: never begin a Run." for `backgroundThreadIntervalMS`, then I would suggest an internal boolean representing whether the maintenance thread should run and a `TimeSpan` to represent that interval. Then we can have the special case logic in the parsing code to set the boolean and then straightforward `TimeSpan.FromMilliseconds(input)` for setting the interval.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
JamesKovacs: I would argue against reflection. It makes refactoring much more difficult than it needs to be. I ran into multiple places during the `hello` work where I renamed a method only to find broken tests because reflection was still looking for the old name. Yes, our tests caught this - which is good - but it still made the refactoring more time-consuming because I had to go back and fix a bunch of tests that were unwittingly broken during a simple rename refactoring.

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -1111,6 +1111,17 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
             }
         }
 
+        [Fact]
+        public async Task MaintainSizeAsync_should_not_run_with_negative_maintenanceInterval()
+        {
+            var settings = _settings.WithInternal(maintenanceInterval: TimeSpan.FromSeconds(-1));
```
JamesKovacs, line 1117: I don't like scattering special constants through our tests. I would suggest `Interval.Infinite` or `TimeSpanConstants.Infinite` (e.g. a class of our creation), which returns `Timeout.InfiniteTimeSpan`.

Note that `Timeout.InfiniteTimeSpan` is actually -1 ms, not -1 s. If we are going to use -1 to represent an infinite `TimeSpan`, let's be consistent on the units used elsewhere in the .NET Framework.

Another option since this is test-only is `_settings.WithInfiniteInterval()` or `_settings.DisableMaintenance()`, which would throw if `maintenanceInterval` is set. Make the intention clear rather than a side-effect of a special value.

tests/MongoDB.Driver.Core.Tests/Specifications/connection-monitoring-and-pooling/ConnectionMonitoringAndPoolingTestRunner.cs
```diff
@@ -577,6 +577,11 @@ private string MapExpectedEventNameToDriverTypeName(string expectedEventName)
                 {
                     switch (poolOption.Name)
                     {
+                        case "backgroundThreadIntervalMS":
+                            var intervalMS = poolOption.Value.ToInt32();
+                            var interval = TimeSpan.FromMilliseconds(intervalMS < 0 ? -1000 : intervalMS);
+                            connectionPoolSettings = connectionPoolSettings.WithInternal(maintenanceInterval: interval);
+                            break;
```
JamesKovacs, line 584: This is where I'd prefer to see the special case logic around -1 reside. Something like:

```
var intervalMS = poolOption.Value.ToInt32();
if (intervalMS == -1)
{
    connectionPoolSettings.DisableMaintenance();
}
else
{
    connectionPoolSettings.WithInterval(maintenanceInterval: TimeSpan.FromMilliseconds(intervalMS));
}
```

JamesKovacs (changes requested):
Adding my thoughts to the ongoing discussion...

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: it won't be really difficult, but I don't really insist, a new internal option also works for me

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -1111,6 +1108,36 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
...
+        public void Maintenance_should_not_run_with_infinite_maintenanceInterval()
+        {
+            var settings = _settings.With(maintenanceInterval: Timeout.InfiniteTimeSpan);
+
+            using var subject = CreateSubject(settings);
+
+            subject.Initialize();
+            subject.SetReady();
+
+            var isTaskNull = subject._maintenanceHelper()._maintenanceTask() == null;
```
BorisDog, line 1137: For some reason `subject._maintenanceHelper()._maintenanceTask().Should().BeNull()` hangs XUnit thread.

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -298,7 +298,7 @@ public void Constructor_should_throw_when_exceptionHandler_is_null()
             [Values(1, 3, 5)] int attempts,
             [Values(false, true)] bool async)
         {
-            var subjectSettings = new ConnectionPoolSettings(minConnections: 0);
+            var subjectSettings = new ConnectionPoolSettings(minConnections: 0, maintenanceInterval: TimeSpan.FromDays(1));
```
BorisDog, line 301: We want to have single run of background thread for stability.

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
JamesKovacs: I don't like that we need to reach into the internal state of the SUT - especially multiple levels - to wait on tasks or check internal state. But I can't think of a clean way to do this without significantly refactoring `ExclusiveConnectionPool`. We could add test-only internal methods to the class itself, but that is equally messy.

JamesKovacs (approved):
LGTM. I've got reservations about reaching into the internal state of the SUT for testing purposes, but I can't think of a way to avoid it without refactoring `ExclusiveConnectionPool` extensively.

JamesKovacs:
I like `IsRunning()` much better. Cleaner and you're not rooting around inside the SUT's internal state. 👍

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -1111,6 +1109,35 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
             }
         }
 
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(1000)]
```
DmitryLukyanov, line 1115: nit: consider `ValuesAttribute`

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -1177,7 +1204,7 @@ public void MaintainSizeAsync_should_not_try_new_attempt_after_failing_without_d
                     {
                         using var connection = AcquireConnection(subject, isAsync);
                     }
-                    
+
```
DmitryLukyanov, line 1207: empty line

tests/MongoDB.Driver.Core.Tests/Core/ConnectionPools/ExclusiveConnectionPoolTests.cs
```diff
@@ -1260,8 +1287,7 @@ public void PrunePoolAsync_should_remove_all_expired_connections([RandomSeed] in
             subject.Initialize();
             subject.SetReady();
 
-            _capturedEvents.WaitForOrThrowIfTimeout(events => events.Where(e => e is ConnectionCreatedEvent).Count() >= connectionsCount, TimeSpan.FromSeconds(10));
-            subject.DormantCount.Should().Be(connectionsCount);
+            SpinWait.SpinUntil(() => subject.DormantCount == connectionsCount, TimeSpan.FromSeconds(1));
```
DmitryLukyanov, line 1290: can you explain this?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -161,30 +161,40 @@ public void ThrowIfNotInitialized()
...
             {
+                _interval = interval;
                 _maintenanceTaskCreator = Ensure.IsNotNull(maintenanceTaskCreator, nameof(maintenanceTaskCreator));
             }
 
+            public bool IsRunning => _maintenanceTask != null;
+
             public void Cancel()
             {
+                if (_interval == Timeout.InfiniteTimeSpan)
```
DmitryLukyanov, line 181: let's add a comment about this condition

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -161,30 +161,40 @@ public void ThrowIfNotInitialized()
...
+
                 _cancellationTokenSource?.Cancel();
                 _cancellationTokenSource = null;
                 _maintenanceTask = null;
             }
 
             public void Start()
             {
+                if (_interval == Timeout.InfiniteTimeSpan)
+                    return;
```
DmitryLukyanov, line 192: as I know, we prefer using the full `if` form with brackets:

    if (true)
    {
          // action
    }

DmitryLukyanov (approved):
Some minor comments and then LGTM

PR #605, opened 2021-08-20
CSHARP-3809: Shut down mongocryptd to avoid timeouts after tests run 

BorisDog:


build.ps1
```diff
@@ -272,4 +272,8 @@ $cakeArguments += $ScriptArgs
 Write-Host "Running build script..."
 Invoke-Expression "& $CAKE_EXE_INVOCATION --bootstrap"
 Invoke-Expression "& $CAKE_EXE_INVOCATION $($cakeArguments -join " ")"
+
+# Shut down mongocryptd to avoid timeouts, tracked in CSHARP-3809
+Stop-Process -Name mongocryptd -Force
```
DmitryLukyanov, line 277: can we do this step in evergreen.yml? This file is taken from cake sources, so it's better to not add changes here

PR #606, opened 2021-08-23
CSHARP-3813 Suppress execution context flow when initializing ServerMonitor

andrewlock:
I recently ran into an issue caused by the `ServerMonitor` capturing the ambient execution context. This meant it captured a reference to an `AsyncLocal` variable, resulting in a memory leak. I don't believe it's necessary to capture the context here, so (where possible) I believe flow should be suppressed.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -152,8 +152,15 @@ namespace MongoDB.Driver.Core.Servers
                 // by using Task.Factory.StartNew we introduce a short delay before the MonitorServerAsync Task starts executing
                 // the delay is whatever time it takes for the new Task to be activated and scheduled
                 // and the delay is usually long enough for the test to get past the race condition (though not guaranteed)
-                _ = Task.Factory.StartNew(() => _ = MonitorServerAsync().ConfigureAwait(false)).ConfigureAwait(false);
-                _ = _roundTripTimeMonitor.RunAsync().ConfigureAwait(false);
+#if !NETSTANDARD1_5
+                using(_ = ExecutionContext.SuppressFlow())
+                {
+#endif
+                    _ = Task.Factory.StartNew(() => _ = MonitorServerAsync().ConfigureAwait(false)).ConfigureAwait(false);
+                    _ = _roundTripTimeMonitor.RunAsync().ConfigureAwait(false);
+#if !NETSTANDARD1_5
+                }
+#endif
             }
         }
 
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -153,12 +153,25 @@ namespace MongoDB.Driver.Core.Servers
                 // the delay is whatever time it takes for the new Task to be activated and scheduled
                 // and the delay is usually long enough for the test to get past the race condition (though not guaranteed)
 #if !NETSTANDARD1_5
-                using(_ = ExecutionContext.SuppressFlow())
+                var restoreFlow = false;
+                try
                 {
+                    if (!ExecutionContext.IsFlowSuppressed())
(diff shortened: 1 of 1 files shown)


src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -152,8 +152,15 @@ public void Initialize()
                 // by using Task.Factory.StartNew we introduce a short delay before the MonitorServerAsync Task starts executing
                 // the delay is whatever time it takes for the new Task to be activated and scheduled
                 // and the delay is usually long enough for the test to get past the race condition (though not guaranteed)
-                _ = Task.Factory.StartNew(() => _ = MonitorServerAsync().ConfigureAwait(false)).ConfigureAwait(false);
-                _ = _roundTripTimeMonitor.RunAsync().ConfigureAwait(false);
+#if !NETSTANDARD1_5
+                using(_ = ExecutionContext.SuppressFlow())
+                {
+#endif
```
tylerohlsen, line 158: Does this need to first check if the the ExecutionContext is already suppressed with `ExecutionContext.IsFlowSuppressed()`?

BorisDog:
@andrewlock thank you for reporting this.
Could you please create a Jira ticket for tracking this issue?

andrewlock:
Hi @BorisDog - I can't create a Jira unfortunately, as [the signup page](https://jira.mongodb.org/secure/Signup!default.jspa) says it's private: 
image

Is there another way for me to log it? Thanks

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
andrewlock: Great point @tylerohlsen, fixed in https://github.com/mongodb/mongo-csharp-driver/pull/606/commits/083250389345023545fb5e1893959e5d2d029f59

BorisDog:
Hi @andrewlock, we have created a ticket for this here [CSHARP-3813](https://jira.mongodb.org/browse/CSHARP-3813).

andrewlock:
Thanks @BorisDog! 🙂

DmitryLukyanov:
Hey, please be informed that we've added a status update in CSHARP-3813

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
Sweetog: @andrewlock Why is the `if !NETSTANDARD1_5` check? I would have expected the preprocessor directive to negate `NETSTANDARD` in general not just `1.5`

DmitryLukyanov:
This logic was rewritten in CSHARP-3761 and has been released in 2.15. Please check this behavior with 2.15 release

Sweetog:
Hi @DmitryLukyanov I have bumped Nuget package `MongoDB.Driver` from `2.12.3` to `2.15.0` and I still see massive memory consumption that may be some flavor of [this issue](https://jira.mongodb.org/browse/CSHARP-3813). *I understand that the rewrite of this area of code in `2.15.0` might make it so that is not possible*
image. 

It is unclear if perhaps I should [expect to see my .NET Core services using at least 512mb per service](https://stackoverflow.com/questions/62209648/bsonchunkpool-and-memory-leak) or that I have a separate memory leak issue entirely.

*I also just recently bumped from .NET Core 5 to .NET Core 6 but I do not think that is related*

JamesKovacs:
Hi, @Sweetog,

Thank you for reaching out about this issue. Reviewing the memory stats provided, the majority is strings and Datadog-related `MongoDbTags` objects. I don't see objects directly related to the MongoDB .NET/C# Driver (e.g. Objects of types in the `MongoDB.Driver.*` namespaces).

Please open a CSHARP ticket at https://jira.mongodb.org/browse/CSHARP with a minimal repro of the issue and a link to [CSHARP-3813](https://jira.mongodb.org/browse/CSHARP-3813). We will be happy to investigate further.

Sincerely,
James

PR #608, opened 2021-08-23
CSHARP-2813: Support Snappy for macOS and Linux.

DmitryLukyanov:


build.cake
```diff
@@ -564,28 +563,8 @@ Task("DumpGitVersion")
     {
         Information(gitVersion.Dump());
     });
-
-Task("PreparePackagingTests")
```
DmitryLukyanov, line 568: described below

build.cake
```diff
@@ -606,58 +585,41 @@ Task("PackagingProjectReferenceTests")
...
 
-Task("PackagingTests")
-    .IsDependentOn("PackagingProjectReferenceTests")
+Task("TestsPackaging")
+    .IsDependentOn("TestsPackagingProjectReference")
     .IsDependentOn("Package")
-    .IsDependentOn("PreparePackagingTests")
     .DoesForEach(
     () => 
-    {
```
DmitryLukyanov, line 615: described below

build.cake
```diff
@@ -708,25 +673,24 @@ Task("PackagingTests")
...
                             Configuration = configuration,
-                            ArgumentCustomization = args => args.Append($"-- RunConfiguration.TargetPlatform={testDetails.Bitness}")
+                            ArgumentCustomization = args => args.Append($"-- RunConfiguration.TargetPlatform={bitness}")
                         }
                     );
                 } 
                 break;
             case "console":
                 {
+                    if (moniker == "netcoreapp21")
```
DmitryLukyanov, line 683: this case is dotnet issue. To fix it we need add something like:

        <PropertyGroup 
      Condition="'$(OS)' == 'Windows_NT' and
                 '$(TargetFramework)' == 'netcoreapp2.1' and
                 '$(SelfContained)' != 'true'">
    <RunCommand Condition="'$(PlatformTarget)' == 'x86'">$(MSBuildProgramFiles32)\dotnet\dotnet</RunCommand>
    <RunCommand Condition="'$(PlatformTarget)' == 'x64'">$(ProgramW6432)\dotnet\dotnet</RunCommand>
        </PropertyGroup>

which is a bit tricky to add via CLI commands. I think it's possible to do it via creating a new custom target, but since `"netcoreapp21"` TF is already covered in `xunit`, I just disabled it

build.cake
```diff
@@ -757,16 +731,55 @@ Task("PackagingTests")
...
+                            Configuration = configuration
                         }
                     );
                 } 
                 break;
-            default: throw new NotSupportedException($"Packaging tests for {testDetails.Command} is not supported.");
+            default: throw new NotSupportedException($"Packaging tests for {command} is not supported.");
+        }
+
+        string ConfigureAndGetTestedDriverVersion(DirectoryPath directoryPath, string localNugetSourceName)
```
DmitryLukyanov, line 746: I moved this step before each test to allow configuring a local package source only as the last step, otherwise the required nugets like `FluentAssertions` won't be able to be consumed since they're not visible after adding a custom `nuget.config`. 
Before I didn't notice it because in my local tests I used cached nugets

build.cake
```diff
@@ -606,58 +585,41 @@ Task("PackagingProjectReferenceTests")
...
+            { ("netcoreapp21", "SDK", "x64", Command: "xunit") },
+            { ("netcoreapp30", "SDK", "x64", Command: "xunit") },
+            { ("net50", "SDK", "x64", Command: "xunit") }
         };
 
+        // run the same tests on x32 architecture
+        testDetails.AddRange(testDetails.ToArray().Select(i => (i.Moniker, i.CsprojType, "x86", Command: i.Command)));
         // run the same tests on console app
-        testDetails.AddRange(testDetails.ToArray().Select(i => (i.Moniker, i.CsprojType, i.Version, i.Bitness, Command: "console")));
+        testDetails.AddRange(testDetails.ToArray().Select(i => (i.Moniker, i.CsprojType, i.Bitness, Command: "console")));
```
JamesKovacs, line 608: Generating test cases this way is somewhat confusing. I would suggest a more standard technique using lists of TFs, project types, processor architectures, and commands. Then you can generate the test cases using LINQ (or nested for loops if you don't want to use LINQ).

```
var tfs = new[] {"net472", "netcoreapp21", "netcoreapp30", "net50"};
var projectTypes = new[] {"SDK"};
var procArchs = new[] {"x86", "x64"};
var commands = new[] {"xunit", "console"};
                                                                     
var testDetails =
  from tf in tfs
  from projectType in projectTypes
  from procArch in procArchs
  from command in commands
  select new { tf, projectType, procArch, command };
```

build.cake
JamesKovacs: Given that this is a `dotnet` bug, I don't see a problem with skipping it especially since we cover this case elsewhere.

build.cake
JamesKovacs: Good catch. Another option would be to configure the local nuget source first followed by a fallback to nuget.org. We would then get our local driver nuget package but still be able to use other nugets. The one potential problem I see with this solution is that if we misconfigure the local source, we would probably be testing our driver package from nuget.org rather than the local one.

src/MongoDB.Driver.Core/Core/Compression/Snappy/Snappy32NativeMethods.cs
```diff
@@ -106,20 +106,14 @@ private class Delegates32
...
-                    case OperatingSystemPlatform.Windows:
-                        return "snappy32.dll";
-                    case OperatingSystemPlatform.Linux: // TODO: add support for Linux and MacOS later
-                    case OperatingSystemPlatform.MacOS:
-                    default:
-                        throw new InvalidOperationException($"Snappy is not supported on the current platform: {currentPlatform}.");
-                }
-            }
+                    OperatingSystemPlatform.Windows => "snappy32.dll",
+                    _ => throw new InvalidOperationException($"Snappy is not supported on the current platform: {currentPlatform} and x32 bitness."),
```
JamesKovacs, line 115: I would suggest "... {currentPlatform} in 32-bit mode."

JamesKovacs (changes requested):
Some minor suggestions.

build.cake
DmitryLukyanov: yeah, it's painful that nuget doesn't support sources priority even though there are a lot requests about it. I think the current one is easiest and most straightforward, I would leave it as is

PR #615, opened 2021-08-27
CSHARP-3710: Specify 5.0 cursor behaviour.

DmitryLukyanov:
Some notes:
1. Added calculation batchSize based on limit for getMore command (previously it was done only for getMore wire protocol). This also was requested in the scope of CSHARP-3205 (data-lake), so I fixed json tests there.
2. Updated FindTests command monitoring step, this was required to make related tests pass, but it's also the scope of CSHARP-3790
So I will close affected tickets when the current one will be pushed.


src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
```diff
@@ -184,8 +184,8 @@ public BsonDocument PostBatchResumeToken
             get { return _postBatchResumeToken; }
         }
 
-        // methods
-        private int CalculateGetMoreProtocolNumberToReturn()
+        // private methods
+        private int CalculateGetMoreNumberToReturn()
```
DmitryLukyanov, line 188: I'm not 100% sure that we can reuse a getMore protocol version, but it looks like it should be possible

tests/MongoDB.Driver.Tests/Specifications/command-monitoring/CommandMonitoringTestRunner.cs
```diff
@@ -103,6 +103,11 @@ public void RunTestDefinition(JsonDrivenTestCase testCase)
         {
             var definition = testCase.Test;
 
+            if (CoreTestConfiguration.ServerVersion < new SemanticVersion(3, 2, 0))
```
DmitryLukyanov, line 106: since these old servers are going to be removed from supporting, it's easier just to disable it instead of fixing some bug regarding how to set `returnKey `

evergreen/evergreen.yml
```diff
@@ -1390,7 +1390,6 @@ buildvariants:
   display_name: "${version} ${compressor} ${topology} ${auth} ${ssl} ${os} "
   tags: ["tests-variant"]
   tasks:
-     - name: test-netstandard15
```
DmitryLukyanov, line 1393: will be reverted since it's already in master

rstam (approved):
LGTM

DmitryLukyanov:
[Merged](https://github.com/mongodb/mongo-csharp-driver/commit/9ad4d4c6b7c83a3b9e2277279e23938073104dbb).

PR #616, opened 2021-08-30
CSHARP-3290: LocalThreshold no longer accepts infinite timespan.

DmitryLukyanov:


JamesKovacs (approved):
LGTM. Verified that `LatencyLimitingServerSelector` already uses `Ensure.IsInfiniteOrGreaterThanOrEqualToZero` and considers the case of an infinite timespan.

PR #617, opened 2021-08-30
CSHARP-3786: Tests for client-side errors with snapshot reads use incorrect syntax for test.expectEvents.

DmitryLukyanov:
Simply tests sync.


BorisDog (approved):
LGTM

PR #621, opened 2021-09-03
CSHARP-3774: Parsing Enums as Strings is case-insensitive.

DmitryLukyanov:


src/MongoDB.Bson/Serialization/Serializers/EnumSerializer.cs
```diff
@@ -243,7 +243,15 @@ private TEnum ConvertInt64ToEnum(long value)
...
-            (TEnum)Enum.Parse(typeof(TEnum), value, true);
+        private TEnum ConvertStringToEnum(string value)
+        {
+            if (Enum.TryParse<TEnum>(value, ignoreCase: false, out var result))
+            {
+                return result;
+            }
+
+            // fall back to case-insensitive parse
+            return (TEnum)Enum.Parse(typeof(TEnum), value, ignoreCase: true);
```
DmitryLukyanov, line 254: the version with generic arguments is not available on all TFs

BorisDog (approved):
LGTM

PR #622, opened 2021-09-06
CSHARP-3800: Fix flaky RetryableReadsProseTests.PoolClearedError_read_retryablity_test

BorisDog:
The main root causes are fixed in CSHARP-3814, CSHARP-3815, CSHARP-3816.

[EG](https://evergreen.mongodb.com/version/61365c973e8e862904a19c92)

### DIFF
--- a/tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
+++ b/tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
@@ -109,7 +109,7 @@ namespace MongoDB.Driver.Tests.Specifications.retryable_reads
 
             // wait for 2 CommandSucceededEvent events, meaning that all other events should be received
             eventCapturer.WaitForOrThrowIfTimeout(
-                events => events.Count(e => e is CommandSucceededEvent) == 2,
+                events => events.OfType<CommandSucceededEvent>().Count() == 2,
                 eventsWaitTimeout);
 
             eventCapturer.Events.OfType<CommandStartedEvent>().Count().Should().Be(3);
--- a/tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
+++ b/tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
@@ -16,6 +16,8 @@
 using System;
 using System.Linq;
 using System.Net;
+using System.Threading.Tasks;
+using FluentAssertions;
 using MongoDB.Bson;
 using MongoDB.Bson.TestHelpers;
 using MongoDB.Bson.TestHelpers.XunitExtensions;
@@ -35,11 +37,11 @@ namespace MongoDB.Driver.Tests.Specifications.retryable_reads
     {
         [SkippableTheory]
         [ParameterAttributeData]
-        public void PoolClearedError_read_retryablity_test([Values(true, false)] bool async)
+        public async Task PoolClearedError_read_retryablity_test([Values(true, false)] bool async)
         {
             RequireServer.Check().Supports(Feature.FailPointsBlockConnection);
 
-            var heartbeatInterval = TimeSpan.FromMilliseconds(500);
+            var heartbeatInterval = TimeSpan.FromMilliseconds(50);
             var eventsWaitTimeout = TimeSpan.FromMilliseconds(5000);
 
             var failPointCommand = BsonDocument.Parse(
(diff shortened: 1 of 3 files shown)


tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
```diff
@@ -88,45 +90,34 @@ public void PoolClearedError_read_retryablity_test([Values(true, false)] bool as
...
-                    typeof(CommandSucceededEvent),
-
-                    // retry or second command succeeds
-                    typeof(ConnectionPoolCheckedOutConnectionEvent),
-                    typeof(CommandStartedEvent),
-                    typeof(CommandSucceededEvent)
-                },
+            // wait for 2 CommandSucceededEvent events, meaning that all other events should be received
+            eventCapturer.WaitForOrThrowIfTimeout(
+                events => events.Count(e => e is CommandSucceededEvent) == 2,
```
DmitryLukyanov, line 112: OfType?

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
```diff
@@ -88,45 +90,34 @@ public void PoolClearedError_read_retryablity_test([Values(true, false)] bool as
...
                 eventsWaitTimeout);
+
+            eventCapturer.Events.OfType<CommandStartedEvent>().Count().Should().Be(3);
```
DmitryLukyanov, line 115: I think the test description implies particular events order: https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.rst#poolclearederror-retryability-test

>Via CMAP monitoring, assert that a PoolClearedEvent is *then* emitted.

and this one:

>Via CMAP monitoring, assert that the second check out then fails due to a connection error.

DmitryLukyanov: similar for writes

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
```diff
@@ -88,45 +90,34 @@ public void PoolClearedError_read_retryablity_test([Values(true, false)] bool as
             var collection = database.GetCollection<BsonDocument>(DriverTestConfiguration.CollectionNamespace.CollectionName);
 
             eventCapturer.Clear();
-            ThreadingUtilities.ExecuteOnNewThreads(2, i =>
+
+            if (async)
             {
-                if (async)
+                await ThreadingUtilities.ExecuteTasksOnNewThreads(2, async __ =>
```
DmitryLukyanov, line 96: I'm surprised that we need to double it here: `__` :(

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
BorisDog: Done

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
BorisDog: `_` is used in child  scope...

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
BorisDog: There is no order guarantee here, as we got two concurrent processes.
Order of `ConnectionPoolClearedEvent` and `ConnectionPoolCheckingOutConnectionFailedEvent` is not defined, so it can't be verified.

DmitryLukyanov: is this described here: https://jira.mongodb.org/browse/DRIVERS-1785?

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
DmitryLukyanov: yeah, I thought that we can use `_` as temporary dummy value in all cases, but probably in all cases except method arguments

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
BorisDog: Yes, but for the same thread. We got two concurrent checkouts, so the observed order across multiple threads can't be quarantined.
Side note: Not the be confused with the fact that if `ConnectionPoolCheckingOutConnectionFailed` is received by **any** thread, the pool is already cleared for that failure (and maybe became ready again).

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
BorisDog: Yup, looks like the support for discard parameters in lambdas was introduced in C# 9. But not for single parameter lambdas for backward compatibility

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
DmitryLukyanov: I think `ThreadId` can be easily saved into an event tuple (event itself + threadId) since we have access to the delegate that save event and threadId is a static class

BorisDog: That won't help, as we got async version as well. I don't think we should overthink it, take a look at [Java implementation](https://github.com/stIncMale/mongo-java-driver/blob/JAVA-4034/driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java#L162-L175).

DmitryLukyanov: `waitForEvent` is blocking operation? If so, then it looks like they support some ordering logic. 
`as we got async version as well.` - we use `ThreadPerTaskScheduler` which means that it will be a separate thread as I understand

BorisDog: Consider `ConnectionPoolClearedEvent` and `ConnectionPoolCheckingOutConnectionFailedEvent` events race again:
`ConnectionPoolClearedEvent` pool is emitted **after** the pool is effectively cleared, the waitqueue is signaled as well. When waitqueue is signaled each waiting thread emits it's own `ConnectionPoolCheckingOutConnectionFailedEvent`, which effectively races with `ConnectionPoolClearedEvent`.
Therefore, the observer can receive those events in any order.

DmitryLukyanov: discussed offline. resolved.

DmitryLukyanov (approved):
LGTM

BorisDog:
Merged [here](https://github.com/mongodb/mongo-csharp-driver/commit/f2349adb8f8a02548229a0ccdb52c902a9722d9e)

PR #625, opened 2021-09-13
CSHARP-3789: Add BSON Binary subtype 7

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/patches/project/dot-net-driver

jyemin (approved):
Was BSON Binary subtype 6 from CSFLE, or is it related to subtype 7?

src/MongoDB.Bson/ObjectModel/BsonBinarySubType.cs
```diff
@@ -55,6 +55,10 @@ public enum BsonBinarySubType
         /// <summary>
         /// User defined binary data.
         /// </summary>
+        Column = 0x07,
+        /// <summary>
```
jyemin, line 59: This actually looks like it got mixed up.  The comments don't match the definitions.

jyemin (changes requested):

PR #626, opened 2021-09-22
CSHARP-3874: Upgrade to Cake 1.2.0 including latest build.sh and build.ps1 files

JamesKovacs:
These files are a direct copy from https://github.com/cake-build/cake.


build.config
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-CAKE_VERSION="1.1.0"
-DOTNET_VERSION="3.1.403"
+CAKE_VERSION=1.2.0
+DOTNET_VERSION=5.0.401
```
jyemin, line 3: What's the effect of updating the DOTNET_VERSION?

build.ps1
```diff
@@ -1,275 +1,154 @@
...
-    # Use integers because the enumeration values for TLS 1.2 and TLS 1.1 won't
-    # exist in .NET 4.0, even though they are addressable if .NET 4.5+ is
-    # installed (.NET 4.5 is an in-place upgrade).
-    # PowerShell Core already has support for TLS 1.2 so we can skip this if running in that.
-    if (-not $IsCoreCLR) {
-        [System.Net.ServicePointManager]::SecurityProtocol = 3072 -bor 768 -bor 192 -bor 48
-    }
-  } catch {
-    Write-Output 'Unable to set PowerShell to use TLS 1.2 and TLS 1.1 due to old .NET Framework installed. If you see underlying connection closed or trust errors, you may need to upgrade to .NET Framework 4.5+ and PowerShell v3'
+#!/usr/bin/env pwsh
```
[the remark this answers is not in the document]
JamesKovacs: Yes, this file is now identical to the official one from cakebuild.net.

DmitryLukyanov (approved):
LGTM

DmitryLukyanov:
I LGTM'ed but let's try to launch tests on all available platforms, at least I don't see windows with not net472

build.config
JamesKovacs: DOTNET_VERSION determines the version of the .NET Framework used to run Cake, not necessarily the version that we are compiling or running on.

BorisDog (approved):
LGTM,
Agree with @DmitryLukyanov that we should check all platforms combinations.

JamesKovacs:
Fair enough point. Everything is looking good on the patch build. I added variants 5.0 standalone for both Linux and MacOS. I'll fire off a complete suite to run overnight to make sure we don't break unexpectedly on some platform combination.

JamesKovacs:
The full patch build overnight is as green as `master` typically has been. Variants that timed out or erred out before are still doing so. Variants that were green before are still green. I'm going to merge based on the LGTMs and the fact that no new failures or timeouts were introduced by the changes.

PR #628, opened 2021-09-23
CSHARP-3876: Builders<T>.Filter.Regex(s => s.Name, keyword) doesn't work.

DmitryLukyanov:


rstam:
Do we need to update the JIRA ticket? 

This doesn't actually seem to have anything to do with Regex.

DmitryLukyanov:
yeah, the regex is just a place where the user noticed this issue. I don't mind to mention actual details in the ticket description. Will do it a bit later.

rstam (approved):
LGTM

PR #630, opened 2021-09-28
CSHARP-3225: Window Functions.

DmitryLukyanov:
LINQ2 approach

### DIFF
--- a/src/MongoDB.Driver/IAggregateFluent.cs
+++ b/src/MongoDB.Driver/IAggregateFluent.cs
@@ -323,6 +323,19 @@ namespace MongoDB.Driver
         /// <summary>
         /// Appends a #setWindowFields stage to the pipeline.
         /// </summary>
+        /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
+        /// <param name="output">The output definition.</param>
+        /// <param name="sortBy">The sortBy definition.</param>
+        /// <param name="outputWindowOptions">The output window options.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        IAggregateFluent<TNewResult> SetWindowFields<TNewResult>(
+            ProjectionDefinition<TResult, TNewResult> output,
+            SortDefinition<TResult> sortBy = null,
+            params AggregateOutputWindowOptionsBase<TNewResult>[] outputWindowOptions);
+
+        /// <summary>
+        /// Appends a #setWindowFields stage to the pipeline.
+        /// </summary>
         /// <typeparam name="TPartitionBy">The type of the partitionBy definition.</typeparam>
         /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
         /// <param name="partitionBy">The partitionBy definition.</param>
@@ -332,8 +345,8 @@ namespace MongoDB.Driver
         /// <returns>The fluent aggregate interface.</returns>
         IAggregateFluent<TNewResult> SetWindowFields<TPartitionBy, TNewResult>(
             AggregateExpressionDefinition<TResult, TPartitionBy> partitionBy,
-            SortDefinition<TResult> sortBy,
             ProjectionDefinition<TResult, TNewResult> output,
+            SortDefinition<TResult> sortBy = null,
             params AggregateOutputWindowOptionsBase<TNewResult>[] outputWindowOptions);
 
         /// <summary>
--- a/src/MongoDB.Driver/IAggregateFluent.cs
+++ b/src/MongoDB.Driver/IAggregateFluent.cs
@@ -321,6 +321,14 @@ namespace MongoDB.Driver
         IAggregateFluent<TNewResult> ReplaceWith<TNewResult>(AggregateExpressionDefinition<TResult, TNewResult> newRoot);
 
         /// <summary>
+        /// Appends a #setWindowFields stage to the pipeline.
+        /// </summary>
+        /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
+        /// <param name="setWindowFieldsDefinition">The $setWindowFields projection.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        IAggregateFluent<TNewResult> SetWindowFields<TNewResult>(ProjectionDefinition<TResult, TNewResult> setWindowFieldsDefinition);
+
+        /// <summary>
         /// Appends a skip stage to the pipeline.
         /// </summary>
         /// <param name="skip">The number of documents to skip.</param>
--- a/src/MongoDB.Driver/IAggregateFluent.cs
+++ b/src/MongoDB.Driver/IAggregateFluent.cs
@@ -323,10 +323,18 @@ namespace MongoDB.Driver
         /// <summary>
         /// Appends a #setWindowFields stage to the pipeline.
         /// </summary>
+        /// <typeparam name="TPartitionBy">The type of the partitionBy definition.</typeparam>
         /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
-        /// <param name="setWindowFieldsDefinition">The $setWindowFields projection.</param>
+        /// <param name="partitionBy">The partitionBy definition.</param>
+        /// <param name="sortBy">The sortBy definition.</param>
+        /// <param name="output">The output definition.</param>
+        /// <param name="outputWindowOptions">The output window options.</param>
         /// <returns>The fluent aggregate interface.</returns>
-        IAggregateFluent<TNewResult> SetWindowFields<TNewResult>(ProjectionDefinition<TResult, TNewResult> setWindowFieldsDefinition);
+        IAggregateFluent<TNewResult> SetWindowFields<TPartitionBy, TNewResult>(
+            AggregateExpressionDefinition<TResult, TPartitionBy> partitionBy,
+            SortDefinition<TResult> sortBy,
+            ProjectionDefinition<TResult, TNewResult> output,
+            params AggregateOutputWindowOptionsBase<TNewResult>[] outputWindowOptions);
 
         /// <summary>
         /// Appends a skip stage to the pipeline.
--- a/src/MongoDB.Driver/Linq/Translators/AggregateSetWindowFieldsTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Translators/AggregateSetWindowFieldsTranslator.cs
@@ -1,0 +1,91 @@
+﻿/* Copyright 2021-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+* 
+*/
+
+using System;
+using System.Linq;
+using System.Linq.Expressions;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Linq.Processors;
+
+namespace MongoDB.Driver.Linq.Translators
+{
+    internal static class AggregateSetWindowFieldsTranslator
+    {
+        public static RenderedProjectionDefinition<TResult> Translate<TPartitionBy, TDocument, TResult>(
+            Expression<Func<TDocument, TPartitionBy>> partitionByProjector,
+            SortDefinition<TDocument> sortDefinition,
+            Expression<Func<IGrouping<TPartitionBy, TDocument>, TResult>> outputProjector,
+            AggregateOutputWindowOptionsBase<TResult>[] outputWindowOptions,
+            IBsonSerializer<TDocument> parameterSerializer,
+            IBsonSerializerRegistry serializerRegistry)
+        {
+            var bindingContext = new PipelineBindingContext(serializerRegistry);
+
+            var keySelector = AggregateGroupTranslator.BindKeySelector(bindingContext, partitionByProjector, parameterSerializer);
+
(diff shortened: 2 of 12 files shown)


src/MongoDB.Driver/Linq/Translators/AggregateSetWindowFieldsTranslator.cs
```diff
@@ -0,0 +1,91 @@
...
+            var outputWithoutWindow = AggregateLanguageTranslator.Translate(boundGroupExpression, translationOptions: null).AsBsonDocument;
+
+            if (outputWindowOptions != null && outputWindowOptions.Length > 0)
+            {
+                foreach (var windowOptions in outputWindowOptions)
+                {
+                    var windowDocuments = windowOptions.Documents;
+
+                    var outputOptionFieldName = windowOptions.OutputWindowField.Render((IBsonSerializer<TResult>)projectionSerializer, serializerRegistry).FieldName;
+                    if (outputWithoutWindow.TryGetValue(outputOptionFieldName, out var outputWindow) && outputWindow is BsonDocument outputWindowDocument)
```
DmitryLukyanov, line 54: the main place where we enrich output fields by output window options if present

DmitryLukyanov:
ready for review

src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -320,6 +320,14 @@ public interface IAggregateFluent<TResult> : IAsyncCursorSource<TResult>
         /// <returns>The fluent aggregate interface.</returns>
         IAggregateFluent<TNewResult> ReplaceWith<TNewResult>(AggregateExpressionDefinition<TResult, TNewResult> newRoot);
 
+        /// <summary>
+        /// Appends a #setWindowFields stage to the pipeline.
+        /// </summary>
+        /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
+        /// <param name="setWindowFieldsDefinition">The $setWindowFields projection.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        IAggregateFluent<TNewResult> SetWindowFields<TNewResult>(ProjectionDefinition<TResult, TNewResult> setWindowFieldsDefinition);
```
rstam, line 329: I don't understand how a `ProjectionDefinition` can be the argument to `SetWindowFields`.

It doesn't look at all like the arguments to the `$setWindowFields` stage.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -1008,6 +1008,51 @@ public static class PipelineDefinitionBuilder
...
+        /// <typeparam name="TIntermediate">The type of the intermediate documents.</typeparam>
+        /// <typeparam name="TOutput">The type of the output documents.</typeparam>
+        /// <param name="pipeline">The pipeline.</param>
+        /// <param name="setWindowFieldsDefinition">The $setWindowFields projection.</param>
+        /// <returns>
+        /// A new pipeline with an additional stage.
+        /// </returns>
+        public static PipelineDefinition<TInput, TOutput> SetWindowFields<TInput, TIntermediate, TOutput>(
+            this PipelineDefinition<TInput, TIntermediate> pipeline,
+            ProjectionDefinition<TIntermediate, TOutput> setWindowFieldsDefinition)
```
rstam, line 1024: I don't understand how a `ProjectionDefinition` can be the argument to `SetWindowFields`.

It doesn't look at all like the arguments to the `$setWindowFields stage`.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1223,6 +1223,59 @@ public static class PipelineStageDefinitionBuilder
...
+
+        /// <summary>
+        /// Creates a $setWindowFields stage.
+        /// </summary>
+        /// <typeparam name="TInput">The type of the input documents.</typeparam>
+        /// <typeparam name="TOutput">The type of the output documents.</typeparam>
+        /// <param name="setWindowFieldsProjection">The $setWindowField definition.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TOutput> SetWindowFields<TInput, TOutput>(
+            ProjectionDefinition<TInput, TOutput> setWindowFieldsProjection)
```
rstam, line 1259: I don't understand how a `ProjectionDefinition` can be the argument to `SetWindowFields`.

It doesn't look at all like the arguments to the `$setWindowFields` stage.

rstam (commented):
Questions about whether it's appropriate to use a `ProjectionDefinition` as the argument to one of the overloads of `SetWindowFields`.

tests/MongoDB.Driver.Tests/AggregateFluentSetWindowFieldsTests.cs
```diff
@@ -0,0 +1,374 @@
...
+            {{
+                documents:[ ""unbounded"", ""current""]
+            }}
+        }}
+    }}
+}}
+");
+            var result = collection
+                .Aggregate()
+                .SetWindowFields<CakeSales>(setWindowFieldsStage)
```
rstam, line 58: I think this functionality is totally unnecessary.

If you are going to specify the entire SetWindowFields arguments as a single BsonDocument then what you actually should do is include `$setWindowFields` in the BsonDocument and use:

    .AppendStage(setWindowFieldsStage)

rstam (changes requested):

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
DmitryLukyanov: $setWindowFields works similar to $group [where ](https://github.com/mongodb/mongo-csharp-driver/blob/512005dc97c74722bd83c59e28c5b1edff2e5272/src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs#L677) we use `ProjectionDefinition`, I've implemented this stage taking $group as example.

tests/MongoDB.Driver.Tests/AggregateFluentSetWindowFieldsTests.cs
DmitryLukyanov: we can say the same for any stage we have :) so at the same time we can say that no reasons to not have this functionality :)

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
rstam: A `ProjectionDefinition` defines the shape of the *projected* RESULT.

It is not intended to carry the arguments to a stage like `$setWindowFields`. So argument values like `partitionBy` and `sortBy` do not belong in a `ProjectionDefinition`.

I suppose a `ProjectionDefinition` *could* be used for `output` argument to the `$setWindowsField`, althought it wouldn't be a standard projection because it is limited to "window operators" along with `window` arguments to the "window operator".

rstam (changes requested):

tests/MongoDB.Driver.Tests/AggregateFluentSetWindowFieldsTests.cs
rstam: I'm not objecting to having helpers for stages.

I'm just objecting that in this case we're essentially providing the *whole* stage as a BsonDocument so this helper is doing nothing useful.

An application should just use `AppendStage` if they're not going to use a type safe helper.

rstam (changes requested):

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
DmitryLukyanov: It's used for `group Id`, in Group stage [here](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs#L717) through `GroupExpressionProjection`. 
I think the whole description of this definition is: it's used to configure projection and requires input arguments to setup ait. For example like `groupId` in $group stage and `partitionBy` + `sortBy` in `setWindowFields`

rstam: In the `Group` and `Projection` helpers the `ProjectionDefinition` defines the RESULT.

If we use a `ProjectionDefinition` anywhere new it should ALSO define the RESULT.

It should not be used to carry arguments like `partitionBy` and `sortByt` that are NOT part of the RESULT.

tests/MongoDB.Driver.Tests/AggregateFluentSetWindowFieldsTests.cs
DmitryLukyanov: I just made a whole copy of how it's implemented in $group stage where we have exactly the same structure

rstam: I understand that you made a copy of Group but I don't think that is correct.

rstam (changes requested):

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
DmitryLukyanov: >In the Group and Projection helpers the ProjectionDefinition defines the RESULT.

not only, it's used to define `group Id` as well

tests/MongoDB.Driver.Tests/AggregateFluentSetWindowFieldsTests.cs
DmitryLukyanov: I kind of agreed about `sortBy`, originally i implemented it as a different argument, but then I decided that it's just an option that affects an output result almost the same as `partitionBy`, this is why I made this logic in the same way as `paritionBy`

rstam (changes requested):

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1223,6 +1223,107 @@ public static class PipelineStageDefinitionBuilder
...
+            Expression<Func<IGrouping<TPartitionBy, TInput>, TOutput>> output,
+            params AggregateOutputWindowOptionsBase<TOutput>[] outputWindowOptions)
+        {
+            Ensure.IsNotNull(output, nameof(output));
+            Ensure.IsNotNull(partitionBy, nameof(partitionBy));
+
+            var outputProjection = new SetWindowFieldsExpressionProjection<TInput, TPartitionBy, TOutput>(partitionBy, output);
+
+            return SetWindowFields<TInput, TPartitionBy, TOutput>(
+                partitionBy: new ExpressionAggregateExpressionDefinition<TInput, TPartitionBy>(partitionBy, ExpressionTranslationOptions.Default),
```
DmitryLukyanov, line 1249: I didn't really understand what `ExpressionTranslationOptions` does, but it looks not required for the current changes, @rstam do you have any ideas?

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1223,6 +1223,107 @@ public static class PipelineStageDefinitionBuilder
...
+        /// <typeparam name="TInput">The type of the input documents.</typeparam>
+        /// <typeparam name="TPartitionBy">The type of the partitionBy definition.</typeparam>
+        /// <typeparam name="TOutput">The type of the output documents.</typeparam>
+        /// <param name="partitionBy">The partitionBy definition.</param>
+        /// <param name="sortBy">The sortBy definition.</param>
+        /// <param name="output">The output definition.</param>
+        /// <param name="outputWindowOptions">The output window options.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TOutput> SetWindowFields<TInput, TPartitionBy, TOutput>(
+            AggregateExpressionDefinition<TInput, TPartitionBy> partitionBy,
```
DmitryLukyanov, line 1267: The main reason why I implemented the stage constructing not here, but on translator reason was that FieldExpression cannot handle cases like: `e => e.OrderDate.Year` correctly, so I needed to have a power of linq translator for that, but now I noticed that we already have a definition for this, so I just reused it

DmitryLukyanov:
ready for review

rstam:
I don't think we should adopt this work as-is. In these comments I'm going to focus mainly on the user level API.

The correct API to match the stage definition should be:

    SetWindowFields(
        AggregateExpressionDefinition partitionBy,
        SortDefinition sortBy,
        ProjectionDefinition output)

With possible convenience overloads:

    SetWindowFields(partitionBy, output) => SetWindowFields(partitionBy, sortBy: null, output)
    SetWindowFields(output) => SetWindowFields(partitionBy: null, sortBy: null, output)

Since partitionBy and sortBy are sometimes optional.

The problem with the above is that it doesn't provide any way to provide the window options for each output field.

You have worked around this by using an optional table of windowOptions:

    SetWindowFields(
        AggregateExpressionDefinition partitionBy,
        SortDefinition sortBy,
        ProjectionDefinition output,
        params AggregateOutputWindowOptionsBase[] windowOptions)

But that's a pretty awkward way of specifying window options, where each window option is matched by name with some field in the output projection.

While that's an interesting workaround, I don't think it's the API we want to provide.

We have discussed that in LINQ3 we could provide the window options using custom extension methods for the window operators. I say we want and do that.

rstam (changes requested):
See previous comment.

DmitryLukyanov:
>You have worked around this by using an optional table of windowOptions:
>
>SetWindowFields(
>    AggregateExpressionDefinition partitionBy,
>    SortDefinition sortBy,
>    ProjectionDefinition output,
>    params AggregateOutputWindowOptionsBase[] windowOptions)

>But that's a pretty awkward way of specifying window options, where each window option is matched by name with some field in >the output projection.

we've already discussed this in the slack, there is no other way in LINQ2, I don't think that it's a bad approach (even though the suggested LINQ3 approach is definitely better), since we just provide a way to add additional configuration via only possible approach. I think it's up to @jyemin whether we want to add this in LINQ2 or no.

JamesKovacs:
Closing this PR as we decided to pursue a different public API for this feature.

PR #646, opened 2021-10-10
CSHARP-3713: SelectMany doesn't support DefaultIfEmpty().

rstam:


rstam:
One other note: this PR is about `DefaultIfEmpty` inside an aggregation expression, NOT at the pipeline level.

In other words, the following is not supported:

    var queryable = collection.AsQueryable().SomeLinqQuery().DefaultIfEmpty();

the reason being that if the pipeline ahead of `DefaultIfEmpty` returns zero documents that's the end of the pipeline. There is no way  (that I can think of) to write a stage to detect an empty incoming pipeline and replace it with a single default document.

PR #649, opened 2021-10-13
CSHARP-3903: Change LinqProvider class to enum.

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/6167487c7742ae1a6b179367

PR #657, opened 2021-10-18
CSHARP-2071: Verify that the queries reported in this ticket work in LINQ3.

rstam:


rstam:
Note: an earlier version of these repros was already in master. This PR adds a new repro and makes minor changes to the tgwo existing one.

PR #665, opened 2021-10-22
CSHARP-3397: Support $merge and $out executing on secondaries.

rstam:
This is a proof of concept only... it's not a completely finished solution.

The main idea here is that $out is not special in and of it itself. What we need is the more general concept that some write operations can be executed on a secondary IF the server version is new enough.

To that end we can use a new `IMayUseSecondaryWriteOperation` interface to mark write operations that may (under the right conditions) be executed on a secondary. Operations that can execute on secondaries implement this interface.

So `AggregateToCollectionOperation` now implements `IMayUseSecondaryWriteOperation` and has a new `ReadPreference` property.

To accommodate the general notion that some operations may (under the right conditions) be executed on a secondary the following classes have been changed also:

**ChannelPinningHelper**: the `CreateReadWriteBinding` has a new `operation` parameter. If the `operation` parameter implements `IMayUseSecondaryWriteOperation` then the `ReadPreference` and `MinServerVersionToUseSecondary` values are copied from the operation to the binding being created.

**WritableServerBinding**: has new `ReadPreference` and `MinServerVersionToUseSecondary` properties, but these will only be set when the binding has been created for an operation that implements `IMayUseSecondaryOperation`, and these values are copied from such an operation.

**WritableServerSelector**; has new `ReadPreference` and `MinServerVersionToUseSecondary` properties, but these will only be set when selecting a server for a write operation that may (under the right conditions) be executed on a secondary. This is where the bulk of the remaining work is: the `SelectServers` method needs to be enhanced to look for these new properties and use them when selecting suitable servers (if they are not null).


tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedAggregateOperation.cs
```diff
@@ -0,0 +1,221 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Driver.Tests.UnifiedTestOperations
+{
+    public abstract class UnifiedAggregateOperation : IUnifiedEntityTestOperation
```
rstam, line 26: This abstract class and its subclasses are the combination of the old `UnifiedAggregateOnCollectionOperation` and `UnifiedAggregateOnDatabaseOperation`.

It has two new subclasses: `UnifiedDatabaseAggregateOperation` and `UnifiedCollectionAggregateOperation`.

These classes needed some work anyway and they weren't structured quite right, hence the refactoring.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedAggregateToCollectionOperation.cs
```diff
@@ -14,33 +14,120 @@
...
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Serializers;
 
 namespace MongoDB.Driver.Tests.UnifiedTestOperations
 {
-    public class UnifiedAggregateToCollectionOperation : IUnifiedEntityTestOperation
+    public abstract class UnifiedAggregateToCollectionOperation : IUnifiedEntityTestOperation
```
rstam, line 25: This class is now abstract and has two new subclasses: `UnifiedDatabaseAggregateToCollectionOperation` and `UnifiedCollectionAggregateToCollectionOperation`.

rstam:
New Evergreen patch build:

https://evergreen.mongodb.com/version/617de5b561837d6e4ce6b786

rstam:
Yet another Evergreen patch build:

https://evergreen.mongodb.com/version/617f1e0761837d6e4ce79494

tests/MongoDB.Driver.Tests/Specifications/crud/tests/unified/aggregate-write-readPreference.json
```diff
@@ -0,0 +1,457 @@
...
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "coll0",
+        "collectionOptions": {
+          "readPreference": {
+            "mode": "secondary",
```
rstam, line 49: This is `secondaryPreferred` in the specifications repo, but I think that's wrong.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -173,6 +175,17 @@ private void AddInitialData(IMongoClient client, BsonArray initialData)
                 var databaseName = dataItem["databaseName"].AsString;
                 var documents = dataItem["documents"].AsBsonArray.Cast<BsonDocument>().ToList();
 
+                var writeConcern = WriteConcern.WMajority;
+                SpinWait.SpinUntil(() => client.Cluster.Description.Type != ClusterType.Unknown, TimeSpan.FromSeconds(5)).Should().BeTrue();
```
rstam, line 179: If we don't SpinWait until the cluster type is known the if statement on line 181 might not see that it's a replica set.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -173,6 +175,17 @@ private void AddInitialData(IMongoClient client, BsonArray initialData)
...
+                var clusterDescription = client.Cluster.Description;
+                if (clusterDescription.Type == ClusterType.ReplicaSet)
+                {
+                    ServerDescription primary = null;
+                    SpinWait.SpinUntil(() => (primary = client.Cluster.Description.Servers.SingleOrDefault(s => s.Type == ServerType.ReplicaSetPrimary)) != null, TimeSpan.FromSeconds(5)).Should().BeTrue();
```
rstam, line 184: If we don't SpinWait until the primary has been discovered the replica set config on line 185 will probably be wrong.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -173,6 +175,17 @@ private void AddInitialData(IMongoClient client, BsonArray initialData)
...
+                    var n = primary.ReplicaSetConfig.Members.Count();
+                    writeConcern = new WriteConcern(n);
```
rstam, line 186: The only way to guarantee that the `initialData` has been propagated to all secondaries *before* the subsequent test is run on some random secondary is to figure out the size of the replica set and use `{ w : n }` when n is the number of members.

It's unfortunate that there is no `{ w : "all" }` write concern.

rstam:
New Evergreen patch build:

https://evergreen.mongodb.com/version/6181681e0305b96ff1f3f57e

DmitryLukyanov (commented):
Can you please also check the json tests in this related ticket https://jira.mongodb.org/browse/CSHARP-3925?

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
```diff
@@ -31,7 +32,7 @@ namespace MongoDB.Driver.Core.Operations
     /// <summary>
     /// Represents an aggregate operation that writes the results to an output collection.
     /// </summary>
-    public class AggregateToCollectionOperation : IWriteOperation<BsonDocument>
+    public class AggregateToCollectionOperation : IWriteOperation<BsonDocument>, IMayUseSecondaryCriteria
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
```diff
@@ -304,6 +326,22 @@ private void EnsureIsOutputToCollectionPipeline()
...
+        {
+            if (_readPreference == null)
+            {
+                return ReadPreference.Primary;
+            }
+
+            return selectedServer.Type switch
+            {
+                ServerType.LoadBalanced => _readPreference, // we can assume that any mongos behind a load balancer is 5.0+
+                ServerType.ReplicaSetSecondary => _readPreference, // we wouldn't have selected the secondary if it wasn't 5.0+
```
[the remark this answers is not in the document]
rstam: Turns out it is not possible to determine the effective read preference from the selected server alone.

Whether or not the server selector has to "fall back" to the primary (for older servers) is also a determining factor of the effective read preference.

src/MongoDB.Driver.Core/Core/Operations/IOperation.cs
```diff
@@ -47,11 +44,20 @@ public interface IReadOperation<TResult>
         Task<TResult> ExecuteAsync(IReadBinding binding, CancellationToken cancellationToken);
     }
 
+    /// <summary>
+    /// Represents a write operation without specifying the result type.
+    /// </summary>
+#pragma warning disable CA1040 // Avoid empty interfaces
+    public interface IWriteOperation
```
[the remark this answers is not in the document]
rstam: This interfaces is not needed after all.

rstam (commented):
I'm going to close this PR and start a new one.

rstam:
Closing and opening a new (considerably different) PR.

PR #667, opened 2021-10-29
CSHARP-3941: ToString for "FilterDefinition" and "SortDefinition"

bernd5:
Added a simple "ToString()" override for "FilterDefinition" and "SortDefinition"

### DIFF
--- a/src/MongoDB.Driver/FilterDefinition.cs
+++ b/src/MongoDB.Driver/FilterDefinition.cs
@@ -146,6 +146,15 @@ namespace MongoDB.Driver
         {
             return new NotFilterDefinition<TDocument>(op);
         }
+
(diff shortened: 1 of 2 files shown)


bernd5:
This fixes not a bug but provides a reasonable ToString implementation for these two classes similar to other Definition-classes (like Pipelinedefinition)

bernd5:
@BorisDog: I created a JIRA ticket...

rstam:
See comments in the JIRA ticket (including expanding the scope of this work to include all definition classes at the same time).

JamesKovacs:
With the addition of LINQ3, this PR requires more work. We have closed this PR and we will track the work in [CSHARP-3941](https://jira.mongodb.org/browse/CSHARP-3941).

PR #668, opened 2021-10-29
CSHARP-3942: Add [InternalsVisibleTo("MongoDB.Driver")] to Core.

DmitryLukyanov:


src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -190,12 +190,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\MongoDB.Shared\ApplicationNameHelper.cs" Link="Core\Configuration\ApplicationNameHelper.cs" />
-    <Compile Include="..\MongoDB.Shared\ConnectionStringConversions.cs" Link="Core\Misc\ConnectionStringConversions.cs" />
     <Compile Include="..\MongoDB.Shared\Hasher.cs" Link="Core\Misc\Hasher.cs" />
```
DmitryLukyanov, line 193: This is required for Bson level, so left it as is for now

src/MongoDB.Driver.GridFS/GridFSForwardOnlyDownloadStream.cs
```diff
@@ -41,7 +41,7 @@ internal class GridFSForwardOnlyDownloadStream<TFileId> : GridFSDownloadStreamBa
         private readonly BsonValue _idAsBsonValue;
         private readonly int _lastChunkNumber;
         private readonly int _lastChunkSize;
-        private readonly IncrementalMD5 _md5;
+        private readonly IncrementalHash _md5;
```
DmitryLukyanov, line 44: `IncrementalMD5` is not needed since we removed supporting netstandard1.5

src/MongoDB.Shared/CanonicalDisposableClass.cs
```diff
@@ -1,116 +0,0 @@
-﻿/* Copyright 2010-present MongoDB Inc.
```
DmitryLukyanov, line 1: these classes are not used

JamesKovacs (approved):
LGTM

src/MongoDB.Driver.Core/Properties/AssemblyInfo.cs
```diff
@@ -25,6 +25,8 @@
 // Required for most of the reflection usage in Xamarin.iOS/Xamarin.Mac.
 [assembly: Preserve(AllMembers = true)]
 
+[assembly: InternalsVisibleTo("MongoDB.Driver")]
+[assembly: InternalsVisibleTo("MongoDB.Driver.Legacy")]
```
rstam, line 29: Did we really need this for Legacy?

DmitryLukyanov: yes, because Legacy also uses classes like `MaxTimeHelper` and `ApplicationNameHelper`

rstam: So this is a misuse of `[InternalsVisibleTo]` that is not justified by the plan to unify `High` and `Core`.

On the other hand, we are planning to get rid of `MongoDB.Driver.Legacy` entirely so this misuse is also temporary.

PR #669, opened 2021-11-01
CSHARP-3948: Add support for FLE_MONGOCRYPTD_PORT environment variable.

DmitryLukyanov:


DmitryLukyanov:
I want to include a helper from this PR into KMIP, so it would be good to avoid merge changes later

rstam (approved):
LGTM

PR #670, opened 2021-11-01
CSHARP-3938: Sync'd max-staleness tests with spec repo

JamesKovacs:
Also mapped `PossiblePrimary` to `Unknown` in spec tests since .NET/C# driver does not consider `PossiblePrimary` as a replica set state.

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/Specifications/max-staleness/tests/README.rst
+++ b/tests/MongoDB.Driver.Core.Tests/Specifications/max-staleness/tests/README.rst
@@ -61,9 +61,9 @@ Each test specifies that it expects an error, or specifies two sets of servers:
   tests pass or fail deterministically.
 
 If the file contains ``error: true``, drivers MUST test that they throw an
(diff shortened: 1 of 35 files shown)


jyemin (commented):
Why are there 32 .json files changed, but only 4 .yml files?  Are they in sync now?  Were they out of sync before?

JamesKovacs:
> Why are there 32 .json files changed, but only 4 .yml files? Are they in sync now? Were they out of sync before?

Yes, those files were out of sync previously. I updated my `master` branch for the `specifications` repo using `git pull`. I then copied all the max-staleness tests via `rsync -r --delete $SPECROOT/source/max-staleness/tests/ tests/MongoDB.Driver.Core.Tests/Specifications/max-staleness/tests/`.

When I diff'd the files, I noticed that the element ordering of many of the .json tests had been updated but the test scenarios and data not changed. The `max-staleness` tests in our repo now matches what is in the `specifications` repository.

jyemin (approved):
LGTM

PR #675, opened 2021-11-09
CSHARP-3936: Support KMIP in client side field level encryption.

DmitryLukyanov:


.gitignore
```diff
@@ -47,7 +47,7 @@ Help
 src/packages
 
 # Nupkg artifacts
-*.nupkg
+#*.nupkg
```
DmitryLukyanov, line 50: will be reverted

NuGet.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+  <configuration>  
+	<packageSources>
+	  <add key="LocalPackages" value="libs" />
+	</packageSources>
+	<activePackageSource>
```
DmitryLukyanov, line 6: will be reverted

evergreen/evergreen.yml
```diff
@@ -303,6 +303,7 @@ functions:
           CLIENT_PEM=${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem \
           REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
           FRAMEWORK=${FRAMEWORK} \
+          ALLOW_NOSSL=true \
```
DmitryLukyanov, line 306: for kmip we need to pass client certificate regardless ssl options

evergreen/evergreen.yml
```diff
@@ -922,44 +964,6 @@ tasks:
           vars:
             FRAMEWORK: netstandard21
 
-    - name: test-kms-tls-expired-certificate-netstandard21
```
DmitryLukyanov, line 925: new tests cover this logic

evergreen/get-python-path.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Environment variables used as input:
+#   Venv                                             Venv path
+#   OS                                               The current operating system
+#
+# Environment variables produced as output:
+#   PYTHON                                           The python path
+
+# Don't write anything to output
+if [ -z "$Venv" ]; then
```
DmitryLukyanov, line 11: a new script that provides python path

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -175,7 +175,7 @@
   <ItemGroup>
     <PackageReference Include="DnsClient" Version="1.4.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2" PrivateAssets="All" />
-    <PackageReference Include="MongoDB.Libmongocrypt" Version="1.2.3" />
+    <PackageReference Include="MongoDB.Libmongocrypt" Version="9.9.93" />
```
DmitryLukyanov, line 178: the final version will be `1.3.0`

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
```diff
@@ -196,6 +214,10 @@ public override string ToString()
             {
                 sb.AppendFormat("SchemaMap : {0}, ", _schemaMap.ToJson(jsonWriterSettings));
             }
+            if (_tlsOptions != null)
+            {
+                sb.AppendFormat("TlsOptions: {0}", _tlsOptions.Select(t => new BsonDocument(t.Key, "<hidden>")).ToJson(jsonWriterSettings));
```
DmitryLukyanov, line 219: I'm open to suggestions here

src/MongoDB.Driver/Encryption/KmsProvidersHelper.cs
```diff
@@ -38,6 +38,24 @@ public static void EnsureKmsProvidersAreValid(IReadOnlyDictionary<string, IReadO
...
+            if (kmsProviderTlsSettings == null)
+            {
+                return;
+            }
+
+            foreach (var kmsProviderTlsSetting in kmsProviderTlsSettings)
+            {
+                var kmsProviderTlsSettingValue = Ensure.IsNotNull(kmsProviderTlsSetting.Value, nameof(kmsProviderTlsSetting.Value));
+
+                if (kmsProviderTlsSettingValue.ServerCertificateValidationCallback != null) // tlsInsecure
```
DmitryLukyanov, line 52: the spec says:

    Drivers SHOULD raise an error if insecure TLS options are set. The error MUST contain the message "Insecure TLS options prohibited". This includes options equivalent to the following URI options:

    tlsInsecure
    tlsAllowInvalidCertificates
    tlsAllowInvalidHostnames
    tlsDisableOCSPEndpointCheck
    tlsDisableCertificateRevocationCheck

In c# we supports only:

* tlsInsecure => via allowing anything in ServerCertificateValidationCallback. Since we don't have control over what user can place inside, I think it's better to not allow specifying this callback at all
* tlsDisableCertificateRevocationCheck => the default value in c# driver is tlsDisableCertificateRevocationCheck=true, so I think we should avoid this validation
* I think it probably makes sense to disable configuring SslSettings.ClientCertificateSelectionCallback even thought it's not required directly, thoughts?

DmitryLukyanov: Also, the spec has this requirement:

    Drivers MUST raise an error if the TLS options are set to disable TLS. The error MUST contain the message "TLS is required".

and `SslSettings.EnabledSslProtocols` has an option SslProtocols.None, but based on this doc https://docs.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols?view=net-5.0#fields it's still `Tls` option, so I didn't disable it

src/MongoDB.Driver/Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -218,18 +242,14 @@ private byte[] ProcessReadyState(CryptContext context)
...
-            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
-            ParseKmsEndPoint(request.Endpoint, out var host, out var port);
-
-            socket.Connect(host, port);
+            var endpoint = CreateKmsEndPoint(request.Endpoint);
 
-            using (var networkStream = new NetworkStream(socket, ownsSocket: true))
-            using (var sslStream = new SslStream(networkStream, leaveInnerStreamOpen: false))
+            var tlsStreamSettings = GetTlsStreamSettings(request.KmsProvider);
+            var sslStreamFactory = new SslStreamFactory(tlsStreamSettings, _networkStreamFactory);
```
DmitryLukyanov, line 248: reuse already existed logic to work with SslStream

src/MongoDB.Driver/Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -267,5 +283,28 @@ private async Task SendKmsRequestAsync(KmsRequest request, CancellationToken can
                 }
             }
         }
+
+        // nested type
+        private class NetworkStreamFactory : IStreamFactory
```
DmitryLukyanov, line 288: in theory, this can be replaced on [TcpStreamFactory](https://github.com/mongodb/mongo-csharp-driver/blob/4183a08781e1da461da73a9c7056c6518a143a2d/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs#L32), but that class has much more complicated logic that we used here and this logic wasn't required in this ticket, so I just created a simple nested implementation

src/MongoDB.Driver/SslSettings.cs
```diff
@@ -251,6 +252,18 @@ public override string ToString()
             return string.Format("{{{0}}}", string.Join(",", parts.ToArray()));
         }
 
+        // internal methods
+        internal SslStreamSettings ToSslStreamSettings()
```
DmitryLukyanov, line 256: no Core/Driver issues anymore 👍

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
```diff
@@ -23,6 +23,7 @@
 
 namespace MongoDB.Driver.Core.TestHelpers.Logging
 {
+    [DebuggerStepThrough]
```
DmitryLukyanov, line 26: This logic is described in CSHARP-3904, but there is one case that I didn't investigate/implement in this PR, so probably I won't close 3904 but just will add a note in the jira ticket

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingTestInvoker.cs
```diff
@@ -67,15 +69,15 @@ protected override async Task<decimal> InvokeTestMethodAsync(object testClassIns
                     throw new TestTimeoutException((int)timeout.TotalMilliseconds);
                 }
 
-                if (Aggregator.HasExceptions && testLoggable != null)
-                {
-                    var exception = Aggregator.ToException();
+                //if (Aggregator.HasExceptions && testLoggable != null)
```
DmitryLukyanov, line 72: will be reverted

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -52,6 +53,12 @@ public void Run(JsonDrivenTestCase testCase)
                 RequireEnvironment.Check().EnvironmentVariable("FLE_AWS_TEMPORARY_CREDS_ENABLED");
             }
 
+            if (testCase.Name.Contains("kmip"))
+            {
+                // kmip requires configuring kms mock server
+                RequireEnvironment.Check().EnvironmentVariable("KMS_MOCK_SERVERS_ENABLED");
```
DmitryLukyanov, line 59: skip because kmip tests require kms mock servers

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -40,6 +46,7 @@
 
 namespace MongoDB.Driver.Tests.Specifications.client_side_encryption.prose_tests
 {
+    [Trait("Category", "FLE")]
```
DmitryLukyanov, line 49: some of tests here still fail from time to time, investigating

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -961,33 +966,293 @@ public static IEnumerable<object[]> DeadlockTest_MemberData()
...
+                        "kmip" => new BsonDocument(), // empty doc
+                        _ => throw new Exception($"Unexpected kmsProvider {kmsProvider}."),
                     });
 
-                var exception = Record.Exception(() => CreateDataKey(clientEncryption, "aws", dataKeyOptions, async));
+                var exception = Record.Exception(() => CreateDataKey(clientEncryption, kmsProvider, dataKeyOptions, async));
+                AssertException(exception);
+            }
+
+            void AssertException(Exception exception)
```
DmitryLukyanov, line 1024: asserts exception based on OS, TF, Certificate configuration, KMS provider

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -961,33 +966,293 @@ public static IEnumerable<object[]> DeadlockTest_MemberData()
...
-                // .Net doesn't make difference between different certificate issues and throws the same exception for all cases.
-                // To ensure that we assert the expected case you need to configure a RemoteCertificateValidationCallback
-                // to the SslStream ctor in LibMongoCryptControllerBase.SendKmsRequest/SendKmsRequestAsync and assert
-                // sslPolicyErrors (for invalidHostname) and expiration dates (for expiredCertificate).
-                exception.Message.Should().Be("Encryption related exception: The remote certificate is invalid according to the validation procedure.", $"because {Environment.GetEnvironmentVariable("KMS_TLS_ERROR_TYPE")} EG configuration");
+                    Ensure.That(isCertificateExpired.GetValueOrDefault() || isInvalidHost.GetValueOrDefault(), $"Unexpected certificate issue detected for cert: {x509certificate2} and policyErrors: {policyErrors}.");
+
+                    return false;
+                });
+                clientEncryptionOptions._tlsOptions(tlsOptions); // avoid validation on serverCertificateValidationCallback
```
DmitryLukyanov, line 1255: I use reflection, because settings serverCertificateValidationCallback is disabled via public API

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1023,52 +1288,61 @@ public void ViewAreProhibitedTest([Values(false, true)] bool async)
         [SkippableTheory]
         [ParameterAttributeData]
         public void UnsupportedPlatformsTests(
-            [Values("local", "aws", "azure", "gcp")] string kmsProvider,
+            [Values("gcp")] string kmsProvider, // the rest kms providers are supported on all supported TFs
```
DmitryLukyanov, line 1291: this test is not longer complicated because most of the cases were gone for now

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -690,13 +693,15 @@ void ValidKmsEndpointConfigurator(string kt, Dictionary<string, object> ko)
                     case "gcp":
                         ko.Add("endpoint", "oauth2.googleapis.com:443");
                         break;
+                    case "kmip":
+                        //ko.Add("endpoint", "localhost:5698");
```
DmitryLukyanov, line 697: missed this, will fix in the next commit

DmitryLukyanov:
Ready for review, the only pending question is related to failing tests with unexpected exception:
>Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.

like [here]( https://evergreen.mongodb.com/task_log_raw/dot_net_driver_secure_tests__version~5.0_os~windows_64_topology~standalone_auth~auth_ssl~ssl_test_net472_patch_60ad0c84b164c0648eabe7dea72aa6d94f8289aa_618b3300c9ec4424bc57eb7d_21_11_10_02_49_12/0?type=T#L1661). It looks like it's related to mock server behavior, I asked Kevin maybe he faced with this, so will reach our him tommorow

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -567,6 +564,10 @@ EncryptionAlgorithm ParseAlgorithm(string algorithm)
         // gcp
         [InlineData("gcp", "cloudkms.googleapis.com:443", null, "parse error")]
         [InlineData("gcp", "example.com:443", "Invalid KMS response", null)]
+        // kmip
+        [InlineData("kmip", null, null, "$HostNotFound$")]
```
DmitryLukyanov, line 568: 1. I don't use enum itself, because it breaks a bit Test Explorer on .net framework, due some serialization issue
2. I use a code instead a message, because the message is platform dependent in some cases

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -294,6 +305,11 @@ private AutoEncryptionOptions ConfigureAutoEncryptionOptions(BsonDocument autoEn
                             kmsOptions.Add("privateKey", gcpPrivateKey);
                         }
                         break;
+                    case "kmip":
+                        {
+                            kmsOptions.Add("endpoint", "localhost:5698"); // mock server
```
JamesKovacs, line 310: Should this be hard-coded or read in from an environment variable?

evergreen/evergreen.yml
```diff
@@ -769,25 +803,33 @@ tasks:
...
           vars:
             FRAMEWORK: netstandard20
 
     - name: test-netstandard21
       commands:
+        - func: start-kms-mock-servers
+        - func: start-kms-kmip-server
         - func: bootstrap-mongo-orchestration
         - func: run-tests
           vars:
```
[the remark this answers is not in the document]
JamesKovacs: How should we run KMIP/KMS-related tests? Should they be included with all `test*` variants as done here? Or should they mimic the OCSP, CSFLE, GSSAPI, and other specialized tests and be their own variant(s)?

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
```diff
@@ -46,23 +47,27 @@ public class AutoEncryptionOptions
...
+        /// <param name="tlsOptions">The tls options.</param>
         public AutoEncryptionOptions(
             CollectionNamespace keyVaultNamespace,
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> kmsProviders,
             Optional<bool> bypassAutoEncryption = default,
             Optional<IReadOnlyDictionary<string, object>> extraOptions = default,
             Optional<IMongoClient> keyVaultClient = default,
-            Optional<IReadOnlyDictionary<string, BsonDocument>> schemaMap = default)
+            Optional<IReadOnlyDictionary<string, BsonDocument>> schemaMap = default,
+            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default)
```
JamesKovacs, line 58: I'm not wild about this API where `AutoEncryptionOptions` includes multiple separate dictionaries (`kmsProviders`, `extraOptions`, and `tlsOptions`), which should all have the same keys. This is probably not something we can change now unfortunately. Do we have code and tests to verify the expected keys are present in each?

JamesKovacs (changes requested):
See comments inline to discuss. Main point of discussion is around testing strategy. A few auxiliary questions about API and impl.

src/MongoDB.Driver/Encryption/KmsProvidersHelper.cs
DmitryLukyanov: The doc says:

>SslProtocols.None  => Allows the operating system to choose the best protocol to use, and to block protocols that are not secure. 

I didn't 100% understand what this means (but I inclined that TLS is presented in any case), but you say that it means that TLS can be disabled in this case? If so, I will change it in the next commit

src/MongoDB.Driver/Encryption/LibMongoCryptControllerBase.cs
DmitryLukyanov: >Are we sure we don't need the extra logic in TcpStreamFactory for proper operation on .NET Core?

we don't have evidences for this yet. Also TcpStreamFactory sets some options like ReadTimeout that is used in the driver, I doubt any driver set it here, so I incline to leave it as is until we have evidence that this is required logic, but I' don't insist :)

evergreen/add-certs-if-needed.sh
```diff
@@ -8,15 +8,20 @@ set -o errexit  # Exit the script with an error if any of the commands fail
 #     OCSP_TLS_SHOULD_SUCCEED Set to test OCSP. Values are true/false/nil
 #     OCSP_ALGORITHM          Set to test OCSP. Values are rsa/ecdsa/nil
 #     OS                      Set to access operating system
+#     ALLOW_NOSSL             Set to ignore check on SSL
 
 SSL=${SSL:-nossl}
 OCSP_TLS_SHOULD_SUCCEED=${OCSP_TLS_SHOULD_SUCCEED:-nil}
 OCSP_ALGORITHM=${OCSP_ALGORITHM:-nil}
+ALLOW_NOSSL=${ALLOW_NOSSL:-nil}
```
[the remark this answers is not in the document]
DmitryLukyanov: this option allows skipping the validation where we add the certificate to trusted storage only if ${SSL}="ssl", so this describes exactly what happens. I don't mind to change it, but `REQUIRE_CERTS` is not what this variable have control about. What about: `IGNORE_SSL_CHECK`?

evergreen/evergreen.yml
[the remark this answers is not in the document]
DmitryLukyanov: I will get back to this question

evergreen/add-certs-if-needed.sh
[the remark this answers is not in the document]
JamesKovacs: Given that there is no harm in adding certs even if they're unused, why don't we simplify the script to always add the certs and remove the need for this option?

src/MongoDB.Driver/Encryption/AutoEncryptionLibMongoController.cs
```diff
@@ -69,7 +69,8 @@ IMongoClient CreateInternalClient()
             : base(
                   Ensure.IsNotNull(cryptClient, nameof(cryptClient)),
                   Ensure.IsNotNull(keyVaultClient, nameof(keyVaultClient)),
-                  Ensure.IsNotNull(Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultNamespace, nameof(autoEncryptionOptions.KeyVaultNamespace)))
+                  Ensure.IsNotNull(Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).KeyVaultNamespace, nameof(autoEncryptionOptions.KeyVaultNamespace)),
+                  Ensure.IsNotNull(Ensure.IsNotNull(autoEncryptionOptions, nameof(autoEncryptionOptions)).TlsOptions, nameof(autoEncryptionOptions.TlsOptions)))
```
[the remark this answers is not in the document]
JamesKovacs: Fair enough. We can leave it as-is.

evergreen/add-certs-if-needed.sh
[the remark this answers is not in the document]
DmitryLukyanov: this file is used in other logic like in OCSP and other, I would not want to change anything there

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
JamesKovacs: If it is defined in the spec, then we should keep the same API - even if I don't particularly like it as a design choice. :)

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
```diff
@@ -216,7 +216,7 @@ public override string ToString()
             }
             if (_tlsOptions != null)
             {
-                sb.AppendFormat("TlsOptions: {0}", _tlsOptions.Select(t => new BsonDocument(t.Key, "<hidden>")).ToJson(jsonWriterSettings));
+                sb.AppendFormat("TlsOptions: {0}, ", _tlsOptions.ToJson(jsonWriterSettings));
```
JamesKovacs, line 219: I think hiding the `tlsOptions` - which could contain sensitive data - other than the key name is the correct thing to do even if it makes debugging more difficult. Why did you remove the `<hidden>` replacement code?

JamesKovacs (changes requested):
Some outstanding comments that need to be addressed. Please let me know if you have any questions or would like to discuss.

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
DmitryLukyanov: I described it here: https://github.com/mongodb/mongo-csharp-driver/pull/675#discussion_r746986108 so, to simplify logic, but I see that it gives some problem with serialization IntPtr like:

>MongoDB.Bson.BsonSerializationException : No serializer found for type System.IntPtr.

EG: https://evergreen.mongodb.com/build/dot_net_driver_kms_tls_mocked_tests_windows__version~5.0_os~windows_64_topology~standalone_ssl~nossl_patch_60ad0c84b164c0648eabe7dea72aa6d94f8289aa_618d6e7de3c33142e63b40c5_21_11_11_19_27_33
I will probably revert it

JamesKovacs: `SslStreamSettings` has 5 properties:
`CheckCertificateRevocation` - might be useful but currently only true is allowed for KMIP/KMS
`ClientCertificates` - we don't want to expose these
`ClientCertificateSelectionCallback` - delegate; only useful thing to show would be whether it is set or not
`EnabledSslProtocols` - might be useful to show these but we could potentially leak security information about whether support for older (and less secure) protocols is enabled 
`ServerCertificateValidationCallback` - delegate; only useful thing to show would be whether it is set or not

Of the 5 properties, they either have known values for KMIP/KMS or would leak security information. So I think `<hidden>` is a better choice rather than risking leaking sensitive information.

evergreen/evergreen.yml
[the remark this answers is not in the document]
DmitryLukyanov: changed

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
DmitryLukyanov: reverted

src/MongoDB.Driver/Encryption/KmsProvidersHelper.cs
JamesKovacs: Your interpretation is correct. `SslProtocols.None` is not what I expected and neither is `SslProtocols.Default`.

src/MongoDB.Driver/Encryption/LibMongoCryptControllerBase.cs
JamesKovacs: Let's leave it as it is here and revisit later if necessary.

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
JamesKovacs: I'm good with implementing CSHARP-3904 immediately. I just don't want to have `[DebuggerStepThrough]` changes mixed into the KMIP changes.

JamesKovacs (approved):
LGTM

.gitignore
DmitryLukyanov: done

NuGet.config
DmitryLukyanov: done

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
DmitryLukyanov: done

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingTestInvoker.cs
DmitryLukyanov: done

evergreen/evergreen.yml
```diff
@@ -307,6 +306,36 @@ functions:
...
+          AUTH=${AUTH} \
+          SSL=${SSL} \
+          MONGODB_URI="${MONGODB_URI}" \
+          TOPOLOGY=${TOPOLOGY} \
+          OS=${OS} \
+          CLIENT_PEM=${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem \
+          FRAMEWORK=${FRAMEWORK} \
+          TARGET="TestCsfleWithMockedKms" \
+            evergreen/run-tests.sh
+            evergreen/cleanup-test-resources.sh
```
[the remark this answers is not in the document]
DmitryLukyanov: done

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/6196d5d82fbabe028bed4f63

JamesKovacs (approved):
LGTM

PR #676, opened 2021-11-10
CSHARP-3963: MongoConnectionPoolPausedException should be resumable for change stream.

DmitryLukyanov:


src/MongoDB.Driver.Core/Core/Operations/RetryabilityHelper.cs
```diff
@@ -97,14 +97,6 @@ public static void AddRetryableWriteErrorLabelIfRequired(MongoException exceptio
             }
         }
 
-        public static void AddRetryableWriteErrorLabelOnCheckoutRetryableWrite(Exception exception)
```
DmitryLukyanov, line 100: not used

src/MongoDB.Driver.Core/Core/Operations/RetryabilityHelper.cs
```diff
@@ -123,6 +115,10 @@ public static bool IsResumableChangeStreamException(Exception exception, Semanti
             {
                 return true;
             }
+            if (exception is MongoConnectionPoolPausedException)
```
DmitryLukyanov, line 118: in theory we can add a retryable label to the exception itself in ctor. Previously we didn't find a good enough solution due argument that it's not so good to add a server error label to the client side exception, so I followed to the common approach in this PR

tests/MongoDB.Driver.Core.TestHelpers/CoreExceptionHelper.cs
```diff
@@ -30,69 +30,68 @@ public static Exception CreateException(string errorType)
         {
             switch (errorType)
             {
-
                 // Exception Types first:
-                case "IOException":
+                case nameof(IOException):
```
DmitryLukyanov, line 34: auto formatting + adding `nameof` which look like a good idea to ensure that we won't lose a sync in names

tests/MongoDB.Driver.Core.Tests/Core/Operations/RetryabilityHelperTests.cs
```diff
@@ -196,6 +197,7 @@ public void IsResumableChangeStreamException_should_return_expected_result_for_s
         [InlineData(typeof(MongoConnectionException), true)] // network exception
         [InlineData(typeof(MongoConnectionClosedException), false)]
         [InlineData(typeof(MongoCursorNotFoundException), true)]
+        [InlineData(typeof(MongoConnectionPoolPausedException), true)]
```
BorisDog, line 200: Where we able to fully reproduce this issue?
Should we add some more high level tests like `change-streams-resume-errorLabels.yml`?

DmitryLukyanov: I'm not sure, I think we can consider this in the follow-up ticket: https://jira.mongodb.org/browse/CSHARP-3964. This one is more about bug fixing before next release

BorisDog: Is it possible to somehow verify that this issue was solved e2e on a higher level before the release? Even manually by reproducing that with `change-streams-resume-errorLabels` or similar?

DmitryLukyanov: I think it's quite straightforward change, but I will try to do it tomorrow

BorisDog: Yeah I think this particular change is good in any case.
Just to be on the safe side, it would be great if we could reproduce the original issue before the release, not a blocker for LGTM I think.

DmitryLukyanov: yes, it works, some repro (obviously only for POC): https://github.com/DmitryLukyanov/mongo-csharp-driver/tree/csharp3963_test

BorisDog (approved):
LGTM

PR #678, opened 2021-11-12
CSHARP-3952: Make maxConnecting configurable.

DmitryLukyanov:


src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
```diff
@@ -152,6 +160,7 @@ internal bool IsPausable
         /// Returns a new ConnectionPoolSettings instance with some settings changed.
         /// </summary>
         /// <param name="maintenanceInterval">The maintenance interval.</param>
+        /// <param name="maxConnecting">The max connecting.</param>
```
BorisDog, line 163: nit: 'Maximum concurrent connections establishments.' or similar

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -46,6 +47,15 @@ public void constructor_should_throw_when_maintenanceInterval_is_negative()
             action.ShouldThrow<ArgumentException>().And.ParamName.Should().Be("maintenanceInterval");
         }
 
+        [Fact]
+        public void constructor_should_throw_when_maxConnecting_is_negative()
```
BorisDog, line 51: Let's validate `0` as well.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -189,6 +205,26 @@ public void With_maintenanceInterval_should_return_expected_result()
...
+#pragma warning disable 618
+            result.WaitQueueSize.Should().Be(subject.WaitQueueSize);
+#pragma warning restore 618
+            result.WaitQueueTimeout.Should().Be(subject.WaitQueueTimeout);
+        }
+
+        [Fact]
+        public void With_maxConnecting_should_return_expected_result()
+        {
+            var oldMaxConnecting = 1;
```
BorisDog, line 220: nit: my personal preference is to use `const` if possible, not critical though.

BorisDog (changes requested):
I think we need to add `maxConnecting` to `ConnectionString` as well.

JamesKovacs (changes requested):
These changes make `maxConnecting` configurable via code as follows:

```
var settings = new MongoClientSettings
{
    ClusterConfigurator = cb => cb.ConfigureConnectionPool(x => x.With(maxConnecting: 3))
};
```

I think this is sufficient for code-based configuration.

The URI spec has been updated (currently in PR) to include a `maxConnecting` URI option. Will you be adding `maxConnecting` URI parsing?

src/MongoDB.Driver.Core/Core/Configuration/ConnectionPoolSettings.cs
DmitryLukyanov: we usually don't really worry about such XML comments, I don't really agree with this decision, but I would follow to what we usually do

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: it's more letters :) I changed it, but mostly I would not worry about it

DmitryLukyanov:
This PR is a good example to spend few seconds more on even something that you think really simple. Originally I missed that we should not only make maxconnecting option public, but we also should add these values to the whole classes we have :( 

>I think this is sufficient for code-based configuration.

I would say it's better to have all options in the uri spec added to the all user-faced classes we have like MongoClientSettings, MongoUrl and etc.

>The URI spec has been updated (currently in PR) to include a maxConnecting URI option. Will you be adding maxConnecting URI parsing?

done

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -333,6 +334,14 @@ public bool IsResolved
             get { return _localThreshold; }
         }
 
+        /// <summary>
+        /// Gets the max connecting.
+        /// </summary>
```
JamesKovacs, line 339: Use description from spec.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -89,6 +100,7 @@ public void constructor_with_maintenanceInterval_should_initialize_instance()
             var subject = new ConnectionPoolSettings(maintenanceInterval: maintenanceInterval);
 
             subject.MaintenanceInterval.Should().Be(maintenanceInterval);
+            subject.MaxConnecting.Should().Be(__defaults.MaxConnecting);
```
JamesKovacs, line 103: We shouldn't be testing `maxConnecting` in the context of another property like `maintenanceInterval`. This applies to all assertions on `maxConnecting` in the below tests as well, except for `With_maxConnecting_should_return_expected_result` which specifically tests `maxConnecting` in the context of `With`.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -46,6 +47,16 @@ public void constructor_should_throw_when_maintenanceInterval_is_negative()
             action.ShouldThrow<ArgumentException>().And.ParamName.Should().Be("maintenanceInterval");
         }
 
+        [Theory]
+        [ParameterAttributeData]
+        public void constructor_should_throw_when_maxConnecting_is_negative([Values(0, -1)] int invalidMaxConnecting)
```
JamesKovacs, line 52: Change the test name to `_is_invalid` because zero isn't negative. :)

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -189,6 +206,26 @@ public void With_maintenanceInterval_should_return_expected_result()
...
 
             result.MaintenanceInterval.Should().Be(newMaintenanceInterval);
+            result.MaxConnecting.Should().Be(__defaults.MaxConnecting);
+            result.MaxConnections.Should().Be(subject.MaxConnections);
+            result.MinConnections.Should().Be(subject.MinConnections);
+#pragma warning disable 618
+            result.WaitQueueSize.Should().Be(subject.WaitQueueSize);
+#pragma warning restore 618
+            result.WaitQueueTimeout.Should().Be(subject.WaitQueueTimeout);
+        }
```
JamesKovacs, line 216: These additional assertions do not belong in this test. A test should (ideally) have one logical assertion.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -46,6 +47,16 @@ public void constructor_should_throw_when_maintenanceInterval_is_negative()
...
+        [Theory]
+        [ParameterAttributeData]
+        public void constructor_should_throw_when_maxConnecting_is_negative([Values(0, -1)] int invalidMaxConnecting)
+        {
+            var exception = Record.Exception(() => new ConnectionPoolSettings(maxConnecting: invalidMaxConnecting));
+
+            var e = exception.Should().BeOfType<ArgumentOutOfRangeException>().Subject;
+            e.ParamName.Should().Be("maxConnecting");
+        }
+
```
JamesKovacs, line 59: We should have a similar ctor test that verifies a few valid `maxConnecting` values.

JamesKovacs (changes requested):
Overall looking good. Just need to improve the XML comments on the new property and fix up some tests.

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
DmitryLukyanov: discussed offline, left: `Gets the maximum number of connections a pool may be establishing concurrently. Defaults to 2.`

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: we don't test maxConnecting itself here, we test that the only assigned(changed) option here is maintenanceInterval (if maintenanceInterval has been passed as income parameter), the rest options are unchanged (default)

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
DmitryLukyanov: described above, 
>A test should (ideally) have one logical assertion.

even though it makes sense, it also makes sense to know not only that we have changed what we wanted to change, but that we didn't change that we didn't want to do

src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -1294,6 +1306,16 @@ private bool EnsureTlsInsecureIsValid(bool value)
             return value;
         }
 
+        private static TValue EnsureValueValid<TValue>(TValue value, Func<TValue, bool> condition, string message)
+        {
+            if (!condition(value))
```
DmitryLukyanov, line 1311: we don't have this condition in all cases on this level, but the new json tests require it

BorisDog (approved):
LGTM.
Just needs a full EG, due to new spec tests.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
JamesKovacs: But we shouldn't be asserting on properties that we aren't testing. We should only assert on `subject.MaintenanceInterval` if that's the purpose of this test. Otherwise the test is over-specified and more fragile than it need be.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ConnectionPoolSettingsTests.cs
```diff
@@ -89,6 +100,24 @@ public void constructor_with_maintenanceInterval_should_initialize_instance()
...
+        [Fact]
+        public void constructor_with_maxConnecting_should_initialize_instance()
+        {
+            const int maxConnecting = 1;
+
+            var subject = new ConnectionPoolSettings(maxConnecting: maxConnecting);
+
+            subject.MaintenanceInterval.Should().Be(__defaults.MaintenanceInterval);
```
JamesKovacs, line 119: We shouldn't be asserting the value of `MaintenanceInterval` here for the same reason.

JamesKovacs (changes requested):
Only outstanding issue is the over-specified tests. If we are testing that setting `MaxConnecting` is successful, we don't need to verify that the other properties haven't been altered. Same with the other properties. If there are interrelationships between properties, then we would have tests to validate those interrelations.

JamesKovacs (approved):
Discussed over-specified tests via Slack. Given that the existing tests were like this, I can live with the fact that the tests are over-specified and revisit in the future if needed.

LGTM.

PR #680, opened 2021-11-15
CSHARP-3891: Add a load balancer test case for SRV resolving to 0 hosts.

DmitryLukyanov:


JamesKovacs (changes requested):
`loadBalanced-no-results` has both .yml and .json files, but the other 4 tests only contain .yml files. Where are the corresponding .json files?

DmitryLukyanov:
yml files were missed before, so it's just full possible sync for LB tests

JamesKovacs (approved):
Ah, I didn't realize we already had the .json files in our repo, but were missing the corresponding .yml files. LGTM then.

PR #681, opened 2021-11-15
CSHARP-3948: Follow-up changes.

DmitryLukyanov:


rstam (approved):
There are open issues with how mongocryptdSpawnArgs works, but for my purposes today I just need our tests to pass when FLE_MONGOCRYPTD_PORT is set, and with these changes they do.

PR #682, opened 2021-11-16
CSHARP-3397: Support $merge and $out executing on secondaries.

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/619427c1d1fe072144f5349b

rstam:
This PR is sufficiently different from the previous PR that it seemed better to just open a new PR.

The comments below should help understand the changes.

Noteworthy is that there are no longer any changes in High other than setting the new `ReadPreference` property of `AggregateToCollectionOperation`.

A good place to start is `WritableServerSelector`. It now takes an optional `IMayUseSecondaryCriteria` that is used when selecting a server for write operations (like $out) that can sometimes be sent to a secondary. Since it is too hard to pass the effective read preference back through multiple layers of indirection the effective read preference determined at server selection time is stored in the `_effectiveReadPreference` field and can be consulted later.

`AggregateToCollectionOperation` has minor changes to its `Execute` methods to handle the effective read preference.

`AggregateToCollectionOperation` uses the new overload of the `GetWriteChannelSource` method that takes a `mayUseSecondary` argument. This method is new to the `IWriteBinding` interface and all implementations must implement it, but it is only in `WritableServerBinding` that the significant change is found (the other bindings just have to implement the method in a trivial way because the other bindings don't do server selection and the effective read preference is baked in).

The new `GetWriteChannelSource` returns a `ChannelSourceWithEffectiveReadPreference` that contains both a channel and the effective read preference. It was not possible to use an `out effectiveReadPreference` parameter because `async` methods don't support `out` parameters.

Note: we *could* use a `Tuple` instead of the custom `ChannelSourceWithEffectiveReadPreference` class, but.... the `Tuple` form is *much* longer. Compare the following alternatives:

    ChannelSourceWithEffectiveReadPreference 
    Tuple<IChannelSourceHandle ChannelSource, ReadPreference EffectiveReadPreference>

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/WritableServerSelector.cs
```diff
@@ -42,26 +43,76 @@ public static WritableServerSelector Instance
...
             if (cluster.IsDirectConnection)
             {
                 return servers;
             }
 
-            return servers.Where(x => x.Type.IsWritable());
+            var serversList = servers.ToList(); // avoid multiple enumeration
+            if (CanUseSecondaries(cluster, serversList))
+            {
+                _effectiveReadPreference = _mayUseSecondary.ReadPreference;
```
rstam, line 82: This is where we "remember" the effective read preference determined during server selection.

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
```diff
@@ -239,12 +253,15 @@ public BsonDocument Execute(IWriteBinding binding, CancellationToken cancellatio
         {
             Ensure.IsNotNull(binding, nameof(binding));
 
-            using (var channelSource = binding.GetWriteChannelSource(cancellationToken))
+            var mayUseSecondary = new MayUseSecondary(_readPreference);
+            var (channelSource, effectiveReadPreference) = binding.GetWriteChannelSource(mayUseSecondary, cancellationToken);
```
rstam, line 257: While this is at a higher level than server selection... `GetWriteChannelSource` is *effectively* a server selection, and that's why we see an effective read preference surfaced here (though it's actually determined at a much lower level by the `WritableServerSelector`).

rstam:
I see some tests failing on Evergreen that didn't fail locally (I was testing on a sharded cluster with 2 mongoses). It's too hard to test against multiple configurations locally!

If reviewers want to wait until I resolve these failed tests that's fine.

rstam:
Looks like most of the replica set variants were failing because I calculated `{ w : 3 }` based on the number of members of the replica set instead of `{ w : 2 }` counting only the data-bearing nodes.

New Evergreen patch build:

https://evergreen.mongodb.com/version/619493f1e3c33129bc196b64

rstam:
I had an epiphany in my dreams: just as the new `IMayUseSecondaryCriteria` object can be used to carry the criteria *down* to the server selector it can be used to carry the effective read preference back *up* to the operation where we need to know it.

Doing this results in even less intrusive changes to the write bindings.

Making those changes now. You might want to wait for the next commit before reviewing.

rstam:
Done. Ready for review.

New Evergreen patch build:

https://evergreen.mongodb.com/version/619533e50305b967993af705

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
```diff
@@ -65,34 +65,60 @@ public IChannelSourceHandle GetReadChannelSource(CancellationToken cancellationT
...
         public IChannelSourceHandle GetWriteChannelSource(CancellationToken cancellationToken)
         {
             ThrowIfDisposed();
             var server = _cluster.SelectServerAndPinIfNeeded(_session, WritableServerSelector.Instance, cancellationToken);
-            return GetChannelSourceHelper(server);
+            return CreateServerChannelSource(server);
+        }
+
+        /// <inheritdoc/>
+        public IChannelSourceHandle GetWriteChannelSource(IMayUseSecondaryCriteria mayUseSecondary, CancellationToken cancellationToken)
```
DmitryLukyanov, line 88: we have tests for this class `WritableServerBindingTests`

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
```diff
@@ -65,34 +65,60 @@ public IChannelSourceHandle GetReadChannelSource(CancellationToken cancellationT
...
             ThrowIfDisposed();
             var server = await _cluster.SelectServerAndPinIfNeededAsync(_session, WritableServerSelector.Instance, cancellationToken).ConfigureAwait(false);
-            return GetChannelSourceHelper(server);
+            return CreateServerChannelSource(server);
+        }
+
+        /// <inheritdoc/>
+        public async Task<IChannelSourceHandle> GetWriteChannelSourceAsync(IMayUseSecondaryCriteria mayUseSecondary, CancellationToken cancellationToken)
+        {
+            if (_cluster.IsPinnedToServer(_session))
```
DmitryLukyanov, line 111: can you please explain this logic (I didn't find this in the spec)?

src/MongoDB.Driver.Core/Core/Clusters/ServerSelectors/WritableServerSelector.cs
```diff
@@ -42,26 +43,72 @@ public static WritableServerSelector Instance
...
+            if (_mayUseSecondary?.ReadPreference == null || servers.Count == 0)
+            {
+                return false;
+            }
+
+            switch (cluster.Type)
+            {
+                case ClusterType.ReplicaSet:
+                case ClusterType.Sharded:
+                    return servers.All(s => _mayUseSecondary.CanUseSecondary(s));
```
DmitryLukyanov, line 104: image

DmitryLukyanov: As I see, we check only equality to `ServerState.Connected` which probably effectively corresponds to Unknown serverType, but this logic is spread out across all driver levels(ServerMonitor/ServerDescription/Cluster) and not really clear. I would probably try to make this serverSelector less dependent on it

DmitryLukyanov:
As far as I can see, the failed EG tests related to this PR. It looks like these tests are not flaky and I never saw them failed previously, so it should be investigated I guess

src/MongoDB.Driver.Core/Core/Clusters/IClusterExtensions.cs
```diff
@@ -26,6 +26,11 @@ namespace MongoDB.Driver.Core.Clusters
     /// </summary>
     internal static class IClusterExtensions
     {
+        public static bool IsPinnedToServer(this ICluster cluster, ICoreSessionHandle session)
+        {
+            return GetPinnedServerIfValid(cluster, session) != null;
```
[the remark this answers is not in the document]
rstam: I think the link you provided only applies to "retryable write operations". And $out to collection is not retryable.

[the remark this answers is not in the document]
DmitryLukyanov: >Yes... but only if "in a transaction". And $out to secondary is not allowed in a transaction.

`GetPinnedServerIfValid` returns `not null` only in transaction as well, so if $out is not supported for transaction, then pinning cannot happen at all

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: I like `defensively` :), but in case if it doesn't trigger additional questions like why it's only here and not in LB mode, why we worry about it at all if pinning cannot happen for $out. Given that we know that pinning cannot happen for $out, I think we can just remove this condition

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
```diff
@@ -253,11 +268,12 @@ public async Task<BsonDocument> ExecuteAsync(IWriteBinding binding, Cancellation
...
             Ensure.IsNotNull(binding, nameof(binding));
 
-            using (var channelSource = await binding.GetWriteChannelSourceAsync(cancellationToken).ConfigureAwait(false))
+            var mayUseSecondary = new MayUseSecondary(_readPreference);
+            using (var channelSource = await binding.GetWriteChannelSourceAsync(mayUseSecondary, cancellationToken).ConfigureAwait(false))
             using (var channel = await channelSource.GetChannelAsync(cancellationToken).ConfigureAwait(false))
             using (var channelBinding = new ChannelReadWriteBinding(channelSource.Server, channel, binding.Session.Fork()))
             {
-                var operation = CreateOperation(channelBinding.Session, channel.ConnectionDescription);
+                var operation = CreateOperation(channelBinding.Session, channel.ConnectionDescription, mayUseSecondary.EffectiveReadPreference);
```
DmitryLukyanov, line 276: Given that we cannot control over which exactly secondary server will be selected let's say if we set ReadPreference=Secondary, what is this EffectiveReadPreference needed for?

DmitryLukyanov:
> As far as I can see, the failed EG tests related to this PR. It looks like these tests are not flaky and I never saw them failed previously, so it should be investigated I guess

missed.

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
rstam: It takes detailed analysis to convince ourselves that this method is never called when pinned to a server. There is no guarantee that some future change might result in this method being called (incorrectly) when pinned to a server.

I suggest we leave the defensive test. It not only protects us against calling this method incorrectly, it also documents that this method must not be called when pinned to a server.

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
rstam: The spec says we have to send the effective read preference to the server.

rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/61aa693c7742ae17f65eeb95

Not all green, but looks to me like the failures are all race conditions unrelated to this PR.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -173,10 +175,17 @@ private void AddInitialData(IMongoClient client, BsonArray initialData)
                 var databaseName = dataItem["databaseName"].AsString;
                 var documents = dataItem["documents"].AsBsonArray.Cast<BsonDocument>().ToList();
 
+                var writeConcern = WriteConcern.WMajority;
+                //if (DriverTestConfiguration.IsReplicaSet(client))
+                //{
+                //    var n = DriverTestConfiguration.GetReplicaSetNumberOfDataBearingMembers(client);
+                //    writeConcern = new WriteConcern(n);
+                //}
```
rstam, line 183: We don't understand why `{ w : n }` instead of `{ w : "majority" }` causes some snapshot tests to fail.

In any case, on Evergreen they are the same (for our purposes) because `{ w : "majority" }` is equivalent to `{ w : 2 }` in a PSA replica set.

src/MongoDB.Driver.Core/Core/Clusters/IClusterExtensions.cs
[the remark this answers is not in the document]
DmitryLukyanov: missed

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: I don't mind to leave this check and I agreed with this:
>There is no guarantee that some future change might result in this method being called (incorrectly) when pinned to a server.

so no guaranties that the same case won't be applied for LB mode in future changes as well. And this will be not easy to understand why we have 2 different places to work with pinned servers and only one of them has this condition (even thought it doesn't have a real value in both of this cases).

rstam: Can you resolve this comment?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
rstam: Oh I agree 100%.

The latest commit was an experiment to see if reverting back to `{ w : "majority" }` fixed the issues we were having with the snapshot read tests.

It did, so I will revert these changes (including removing the comment).

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: Well, I requested the same change in LB in the last comment or remove it in both cases. It also related to this unanswered comment https://github.com/mongodb/mongo-csharp-driver/pull/682#discussion_r755450349

rstam: Can you clarify what you mean? What specifically are you asking for?

I didn't realize you thought that previous comment was unanswered. I thought "missed" meant *you* had missed it. Either way, I don't understand what you want in that comment either.

src/MongoDB.Driver.Core/Core/Clusters/IClusterExtensions.cs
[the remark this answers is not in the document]
DmitryLukyanov: In the previous comment you wrote about why you don't think that pinning logic in LB should be "affected" by this change:
>Yes... but only if "in a transaction". And $out to secondary is not allowed in a transaction.

You may see that `GetPinnedServerIfValid` works exactly by the same workflow and it returns not null only in transaction (in fact these 2 logic (here and in LB) almost the same).

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: I meant that pinning logic in LB and here are almost the same. If we add such check here we should add it in the LB as well.

rstam: I honestly still don't understand what you are trying to say... but it did get me thinking a bit more about this.

Best I can guess is that you are saying that this method is being TOO specific in how it checks whether the session is pinned to a server or not.

What I do NOT think we should do is make any changes to any load balancing code. This PR is not about load balancing.

I've removed this new method entirely. See the corresponding changes in WritableServerBinding (where this method used to be called).

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
```diff
@@ -107,6 +133,11 @@ public void Dispose()
             }
         }
 
+        private bool IsSessionPinnedToServer()
+        {
+            return _session.IsInTransaction && _session.CurrentTransaction.PinnedServer != null;
+        }
```
rstam, line 139: In this new method we don't care HOW a session got pinned to a server (whether it's because of load balancing or not).

We simply check IF the session is pinned to a server.

rstam:
Note: github lost some comments we had written when I deleted some code.

In summary: I've removed the new `IClusterExtensions.IsPinnedToServer` helper method that was under discussion and added instead a more focused `WritableServerBinding.IsSessionPinnedToServer` instead.

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: Yes, now it doesn't matter how the server has been pinned. However it's not really equivalent change. If the transaction state is `session.CurrentTransaction.State == CoreTransactionState.Starting`, effectively it means that the server is not considered as pinned anymore. It's probably not really important, since I think the whole check here is added "just in case" and is not going to be called at all. So, I think we can leave it as is.

src/MongoDB.Driver.Core/Core/Bindings/WritableServerBinding.cs
DmitryLukyanov: >Best I can guess is that you are saying that this method is being TOO specific in how it checks whether the session is pinned to a server or not.

it's too specific of what cases should be checked. As I wrote, pinning logic in LB is almost the same with pinning logic in this class for sharded cluster (with the difference that in LB we additionally consider pinning a channel as well, but the server workflow is the same). Ideally we would even merge LB server pinning logic and this one, but it's a separate question.
I'm ok with the latest changes

DmitryLukyanov (approved):
LGTM

rstam:
I was verifying that the JSON tests added in this PR were still current and I discovered that they had changed slightly since they were pulled into our project.

I've gone ahead and updated them. The changes are minor and unrelated to the driver changes in the PR. No further changes were necessary.

New Evergreen patch build:

https://evergreen.mongodb.com/version/61ba3cf81e2d177b944bb9b5

JamesKovacs (approved):
LGTM. My main concern was disallowing $out to secondaries for mixed version topologies, but since that was an explicit impl decision, then I’m good with that. And I see that you have test coverage on the new helper functions. I wish I had some more insightful feedback, but all I can say is good job and thanks for all your hard work to get this over the finish line.

PR #683, opened 2021-11-18
CSHARP-3971: Make ExpressionNotSupportedException public.

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/6195a4cc2a60ed674990d553

src/MongoDB.Driver/Linq/ExpressionNotSupportedException.cs
```diff
@@ -16,9 +16,12 @@
 using System;
 using System.Linq.Expressions;
 
-namespace MongoDB.Driver.Linq.Linq3Implementation
+namespace MongoDB.Driver.Linq
```
rstam, line 19: Notice that the namespace and the directory changed also.

JamesKovacs (approved):
LGTM. Good catch on moving it to a common namespace.

PR #684, opened 2021-11-19
CSHARP-3904: Protect xunit/logging infrastructure by DebuggerStepThrough.

DmitryLukyanov:


tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingTestInvoker.cs
```diff
@@ -52,7 +54,18 @@ protected override async Task<decimal> InvokeTestMethodAsync(object testClassIns
         {
             var xUnitTestCase = Test.TestCase as IXunitTestCase;
             var timeoutMS = xUnitTestCase?.Timeout ?? 0;
-            var timeout = timeoutMS <= 0 ? CoreTestConfiguration.DefaultTestTimeout : TimeSpan.FromMilliseconds(timeoutMS);
+
+            TimeSpan timeout;
+            if (timeoutMS > 0)
+            {
+                timeout = TimeSpan.FromMilliseconds(timeoutMS);
```
BorisDog, line 61: I suggest that we override the timeout in any case if `Debugger.IsAttached`

DmitryLukyanov: If we specify it explicitly, then we want to assert it in some steps, it can break test, I would not do it

DmitryLukyanov: wait, no, it's for the whole test, not a test step,  then probably it makes sense

DmitryLukyanov: done

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingTestInvoker.cs
```diff
@@ -52,7 +54,10 @@ protected override async Task<decimal> InvokeTestMethodAsync(object testClassIns
         {
             var xUnitTestCase = Test.TestCase as IXunitTestCase;
             var timeoutMS = xUnitTestCase?.Timeout ?? 0;
-            var timeout = timeoutMS <= 0 ? CoreTestConfiguration.DefaultTestTimeout : TimeSpan.FromMilliseconds(timeoutMS);
+            var timeout = Debugger.IsAttached
```
BorisDog, line 57: nit: Personally I prefer more compact expressions like here. But in this case we could consider having the debugger-related logic separately, for better readability:

```
if (Debugger.IsAttached)
{
   timeout = Timeout.InfiniteTimeSpan
}
```

For your consideration.

BorisDog (approved):
LGTM

tests/MongoDB.Driver.Core.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingTestInvoker.cs
DmitryLukyanov: if you mean, reassignment timeout if `Debugger.IsAttached=true` after it's already initialized in line 55(above), I would prefer not do it, unless it's hard to do differently. But I never request it in PRs since it's definitely personal `nit` :)

JamesKovacs (approved):
LGTM

PR #685, opened 2021-11-19
CSHARP-3936: Disable cert configuring where it's not required.

DmitryLukyanov:


JamesKovacs (approved):
LGTM. Tests pass locally on my machine as well, which they didn't before.

PR #688, opened 2021-11-20
CSHARP-3974: Remove WithOptions and WithSessions.

rstam:


JamesKovacs (approved):
LGTM. Given that the user can provide session and aggregate options when calling `AsQueryable()`, I agree that we shouldn't add to our public API unnecessarily without a defined use case.

PR #689, opened 2021-11-20
CSHARP-3975: Move the string extension methods in MongoDBLinqExtensions to StringExtensions.

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/61992fef57e85a217dced19c

src/MongoDB.Driver/Linq/StringExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+*/
+
+using System;
+
+namespace MongoDB.Driver.Linq
+{
+    /// <summary>
+    /// This static class holds methods that can be used to express MongoDB specific operations in LINQ queries.
+    /// </summary>
+    public static class StringExtensions
```
rstam, line 23: The main thing to evaluate here is that we are introducing new public APIs, so changing them later is a breaking change.

JamesKovacs (approved):
LGTM.

I agree with you that the only question is API. `someString.StrLenBytes()` and `someString.SubstrBytes()` are odd method names for C#. They don’t match the common .NET APIs like `Count()`, `Length`, and `Substring`. If we were going to match the .NET API, we would probably use `someString.LengthBytes()` (or `someString.CountBytes()`) and `someString.SubstringBytes()`, but those are odd themselves. 

There is value in matching the server names in terms of discoverability and documentation.  I’m inclined to keep the server names even though they look odd in C#.

PR #696, opened 2021-11-26
CSHARP-3855: Remove code that implements support for server version 2.6.

DmitryLukyanov:
Some notes:

1. I didn't touch legacy.
2. Changes (mostly in auth tests) not related to 2.6 directly are caused by changes in some test helpers where I removed using server 2.6

Full EG: https://evergreen.mongodb.com/version/61a15ca3d6d80a0f59f0c904

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Authentication/MongoDBX509Authenticator.cs
+++ b/src/MongoDB.Driver.Core/Core/Authentication/MongoDBX509Authenticator.cs
@@ -80,7 +80,6 @@ namespace MongoDB.Driver.Core.Authentication
         {
             Ensure.IsNotNull(connection, nameof(connection));
             Ensure.IsNotNull(description, nameof(description));
-            EnsureUsernameIsNotNullOrNullIsSupported(connection, description);
 
             if (description.HelloResult.SpeculativeAuthenticate != null)
             {
@@ -103,7 +102,6 @@ namespace MongoDB.Driver.Core.Authentication
         {
             Ensure.IsNotNull(connection, nameof(connection));
             Ensure.IsNotNull(description, nameof(description));
-            EnsureUsernameIsNotNullOrNullIsSupported(connection, description);
 
             if (description.HelloResult.SpeculativeAuthenticate != null)
             {
@@ -159,16 +157,6 @@ namespace MongoDB.Driver.Core.Authentication
         {
             var message = string.Format("Unable to authenticate username '{0}' using protocol '{1}'.", _username, Name);
             return new MongoAuthenticationException(connection.ConnectionId, message, ex);
-        }
-
-        private void EnsureUsernameIsNotNullOrNullIsSupported(IConnection connection, ConnectionDescription description)
-        {
-            var serverVersion = description.ServerVersion;
-            if (_username == null && !Feature.ServerExtractsUsernameFromX509Certificate.IsSupported(serverVersion))
-            {
-                var message = $"Username cannot be null for server version {serverVersion}.";
-                throw new MongoConnectionException(connection.ConnectionId, message);
-            }
         }
     }
 }
--- a/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
+++ b/src/MongoDB.Driver.Core/Core/Misc/Feature.cs
@@ -23,11 +23,15 @@ namespace MongoDB.Driver.Core.Misc
     public class Feature
     {
         #region static
+        private static readonly Feature __aggregate = new Feature("Aggregate", new SemanticVersion(2, 2, 0));
         private static readonly Feature __aggregateAccumulator = new Feature("AggregateAccumulator", new SemanticVersion(4, 3, 4));
         private static readonly Feature __aggregateAddFields = new Feature("AggregateAddFields", new SemanticVersion(3, 4, 0));
+        private static readonly Feature __aggregateAllowDiskUse = new Feature("AggregateAllowDiskUse", new SemanticVersion(2, 6, 0));
         private static readonly Feature __aggregateBucketStage = new Feature("AggregateBucketStage", new SemanticVersion(3, 3, 11));
         private static readonly Feature __aggregateComment = new Feature("AggregateComment", new SemanticVersion(3, 6, 0, "rc0"));
         private static readonly Feature __aggregateCountStage = new Feature("AggregateCountStage", new SemanticVersion(3, 3, 11));
+        private static readonly Feature __aggregateCursorResult = new Feature("AggregateCursorResult", new SemanticVersion(2, 6, 0));
+        private static readonly Feature __aggregateExplain = new Feature("AggregateExplain", new SemanticVersion(2, 6, 0));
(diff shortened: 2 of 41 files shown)


src/MongoDB.Driver.Core/Core/Authentication/MongoDBX509Authenticator.cs
```diff
@@ -160,15 +158,5 @@ private MongoAuthenticationException CreateException(IConnection connection, Exc
             var message = string.Format("Unable to authenticate username '{0}' using protocol '{1}'.", _username, Name);
             return new MongoAuthenticationException(connection.ConnectionId, message, ex);
         }
-
-        private void EnsureUsernameIsNotNullOrNullIsSupported(IConnection connection, ConnectionDescription description)
-        {
-            var serverVersion = description.ServerVersion;
-            if (_username == null && !Feature.ServerExtractsUsernameFromX509Certificate.IsSupported(serverVersion))
```
DmitryLukyanov, line 167: this cannot happen starting from `3, 3, 12`, changed this because of some failed tests

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -626,6 +637,17 @@ public void ThrowIfNotSupported(SemanticVersion serverVersion)
             }
         }
 
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            var message = $"First supported version: {_firstSupportedVersion}";
```
DmitryLukyanov, line 643: looks like suitable change, but I'm ok to revert it

tests/MongoDB.Driver.Core.Tests/Core/Authentication/MongoAWSAuthenticatorTests.cs
```diff
@@ -214,102 +210,6 @@ static MongoAWSAuthenticatorTests()
             sentMessages[1].Should().Be(GetExpectedSaslContinueCommandMessage(actualRequestId1, expectedClientSecondMessage, expectedServerApiString));
         }
 
-        [Theory]
-        [ParameterAttributeData]
-        public void Authenticate_should_send_serverApi_with_command_wire_protocol_if_serverApi_is_provided(
```
DmitryLukyanov, line 219: after all changes in auth where I replaced using 2.6 server on 3.6 in test helpers, now we cannot use query protocol here, so this test has no sense any more, similar logic in other auth tests

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -112,6 +112,7 @@ public class Feature
         /// <summary>
         /// Gets the aggregate feature.
         /// </summary>
+        [Obsolete("This property will be removed in a later release.")]
```
rstam, line 115: Why deprecate these?

I realize that they refer to features that are all implemented in 3.6+, but is that enough reason to deprecate them?

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
rstam: I think this change is fine but I would suggest slightly different wording and adding the name of the feature:

    "${_name} feature added in {_firstSupportedVersion}."

or

    "${_name} feature added in {_firstSupportedVersion} and removed in {_supportRemovedVersion}."

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
DmitryLukyanov: because they're not used at all? I think having properties with zero references in the code looks weird. 
NOTE: I've added one more commit where I just removed these properties, we discuss it on Friday with James and can discuss it again with the whole team today :)

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -104,15 +95,8 @@ public class Feature
...
         private static readonly Feature __transactions = new Feature("Transactions", new SemanticVersion(4, 0, 0));
-        private static readonly Feature __userManagementCommands = new Feature("UserManagementCommands", new SemanticVersion(2, 6, 0));
         private static readonly Feature __views = new Feature("Views", new SemanticVersion(3, 3, 11));
         private static readonly Feature __wildcardIndexes = new Feature("WildcardIndexes", new SemanticVersion(4, 1, 6));
-        private static readonly Feature __writeCommands = new Feature("WriteCommands", new SemanticVersion(2, 6, 0));
-
-        /// <summary>
-        /// Gets the aggregate feature.
-        /// </summary>
-        public static Feature Aggregate => __aggregate;
```
rstam, line 115: This is a `public` class. Removing these properties is a breaking change.

DmitryLukyanov: yes, but it's core, as I know (and as we discussed with @JamesKovacs ), we're ok with this now

rstam: OK. So long as we know we're making a breaking change on purpose.

There are probably users who have used this class. We've probably even *told* them to at some point.

DmitryLukyanov: what is a use case for users to look at this class?

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
rstam: Same as our own use case of course... to detect whether a feature is supported by a particular version of the server.

DmitryLukyanov: so since users usually work only with one server, for them it's effectively means: `if (true) .. / if (false) ..`? And Given than we no longer support servers < 3.6, the condition effectively means:

     if (true)
     {
          ...
     }

     if (false)
     {
           throw exception
     }

I'm surprised that we could say for users to use this class. But If it's a real case, probably it makes sense to leave only `obsolete` attribute without actual removing this property

DmitryLukyanov: cc: @JamesKovacs

DmitryLukyanov:
ready for review

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
DmitryLukyanov: reverted removed fields

DmitryLukyanov: It's probably better making the whole class obsolete a bit later to avoid unrelated changes in the code now

rstam (approved):
LGTM. Looks great.

Note: I only reviewed the changes to tests lightly. They look reasonable, but I'm relying on you and the fact that tests are passing.

PR #699, opened 2021-12-01
CSHARP-3978: CancellationTokenSource dispose in SemaphoreSlimSignalable and MaintenanceHelper

BorisDog:
- Added disposing of linked `cancellationTokenSource` and `signalCancellationTokenSource` in `SemaphoreSlimSignalable`
- Added disposing of `cancellationTokenSource` in `MaintenanceHelper`

[EG](https://evergreen.mongodb.com/version/61a7d2d361837d77de087981)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
+++ b/src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
@@ -32,24 +32,22 @@ namespace MongoDB.Driver.Core.Misc
         private readonly struct CancellationContext : IDisposable
         {
             public bool IsSignaled { get; }
-            public CancellationToken CancellationToken { get; }
             public CancellationTokenSource LinkedCancellationTokenSource { get; }
-            public CancellationTokenSource SignaledCancellationTokenSource { get; }
+            public CancellationTokenSource SignalCancellationTokenSource { get; }
 
-            public static CancellationContext Signaled { get; } = new CancellationContext(true, default, default, default);
+            public static CancellationContext Signaled { get; } = new CancellationContext(true, default, default);
 
             public CancellationContext(
                 bool isSignaled,
                 CancellationTokenSource linkedCancellationTokenSource,
-                CancellationTokenSource signaledCancellationTokenSource,
-                CancellationToken cancellationToken)
+                CancellationTokenSource signaledCancellationTokenSource)
             {
                 IsSignaled = isSignaled;
-                CancellationToken = cancellationToken;
                 LinkedCancellationTokenSource = linkedCancellationTokenSource;
-                SignaledCancellationTokenSource = signaledCancellationTokenSource;
+                SignalCancellationTokenSource = signaledCancellationTokenSource;
             }
 
+            public CancellationToken CancellationToken => LinkedCancellationTokenSource.Token;
             public void Dispose() => LinkedCancellationTokenSource?.Dispose();
         }
 
@@ -136,7 +134,7 @@ namespace MongoDB.Driver.Core.Misc
 
         public SemaphoreWaitResult WaitSignaled(TimeSpan timeout, CancellationToken cancellationToken)
         {
-            using var cancellationContext = GetCancelationTokenContext(cancellationToken);
+            using var cancellationContext = GetCancellationTokenContext(cancellationToken);
 
             if (cancellationContext.IsSignaled)
             {
@@ -145,13 +143,12 @@ namespace MongoDB.Driver.Core.Misc
 
             try
             {
-
                 var entered = _semaphore.Wait(timeout, cancellationContext.CancellationToken);
                 return entered ? SemaphoreWaitResult.Entered : SemaphoreWaitResult.TimedOut;
             }
             catch (OperationCanceledException)
             {
-                if (IsSignaled(cancellationContext.SignaledCancellationTokenSource, cancellationToken))
+                if (IsSignaled(cancellationContext.SignalCancellationTokenSource, cancellationToken))
                 {
                     return SemaphoreWaitResult.Signaled;
                 }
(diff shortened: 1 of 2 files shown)


src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -29,6 +29,30 @@ public enum SemaphoreWaitResult
             Entered
         }
 
+        private readonly struct CancellationContext : IDisposable
```
DmitryLukyanov, line 32: As I understand we have reproduction of this case (that James's confirmed), then we need to add a test for this

JamesKovacs: We do have a repro. Unfortunately it is not easy to codify into a test. Basically we execute a large number of operations and make sure that `GC.GetTotalMemory(true)` doesn't increase over time. Turning that into an automated integration test isn't straightforward. Longer term it would be desirable to create one or more smoke tests that execute some workload and verify that common resources (memory, connections, sockets, threads, etc.) remain within some reasonable bound. This doesn't exist today and is outside the scope of this fix.

DmitryLukyanov: I'm ok to skip it now, but I think we should implement it in one or another way, so can create a separate ticket for this.

DmitryLukyanov: note: it can be just a unit test for `SemaphoreSlimSignalable.cs`

JamesKovacs: I agree that we should think about how to validate situations like this. This one was particularly sneaky because `SemaphoreSlimSignalable` was introduced in 2.13.0, but didn't have a noticeable affect until 2.14.0 when the wait queue was refactored to use it.

We do have tests around this class. See `SemaphoreSlimSignalableTests`. Unfortunately the resource leak in this case happens due to local variables not being disposed correctly. Since the variables are locals, it is difficult (if not impossible) to unit test. I'm open to ideas - both that can be implemented now and larger efforts that can be captured in a new ticket.

DmitryLukyanov: I think testing it should be possible somehow :), let's just create a ticket and will try to work on this soon

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -174,31 +200,58 @@ public void Release()
         public void Dispose()
         {
             _semaphore.Dispose();
+            _signalCancellationTokenSource.Dispose();
```
DmitryLukyanov, line 203: Should it be also cancelled before this line? I don't really see the whole logic here, but as I know it's recommended to call Cancel and then Dispose.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -174,31 +200,58 @@ public void Release()
...
         }
 
-        private (CancellationTokenSource TokenSourceLinked, CancellationTokenSource SignalTokenSource, bool Signaled) GetLinkedTokenAndCheckForSignaled(CancellationToken cancellationToken)
+        private CancellationContext GetCancelationTokenContext(CancellationToken cancellationToken)
```
JamesKovacs, line 206: `GetCancelationTokenContext` => `GetCancellationTokenContext`

JamesKovacs (changes requested):
Minor naming change for now. You mentioned on Slack that some additional `Dispose` calls are needed. Once those are in place, I'll retest the CSHARP-3978 repro.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
BorisDog: I think we can address this in in scope of stress/smoke tests as @JamesKovacs  mentioned.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
BorisDog: Yes in general. But we need to be cautious with that, as `Cancel` can lead to a deadlock sometimes.
In this case not need to cancel as `_semaphore` is the only consumer of this token and it's disposed in a line before.
Also `_signalCancellationTokenSource` can be already disposed, and then `Cancel` with throw.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
BorisDog: Thanks, done.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
```diff
@@ -174,31 +197,46 @@ public void Release()
...
-                return (default, default, true);
+                return CancellationContext.Signaled;
             }
 
-            var tokenSourceLinked = CancellationTokenSource.CreateLinkedTokenSource(
-                signalTokenSource.Token,
-                cancellationToken);
+            try
+            {
+                var cancellationLinkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(
```
BorisDog, line 214: Reverted using `signalTokenSource` directly for now, as it leads to a deadlock.

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
DmitryLukyanov: I would prefer to have this sooner  :), but we can discuss on the next triage

DmitryLukyanov (approved):
LGTM

src/MongoDB.Driver.Core/Core/Misc/SemaphoreSlimSignalable.cs
JamesKovacs: Let's file a CSHARP ticket and triage it on Monday. Volunteer?

JamesKovacs (approved):
LGTM

BorisDog:
[Merged](https://github.com/mongodb/mongo-csharp-driver/commit/f4fd11d2c3efea1844ce0979b97381749173b73e)

PR #704, opened 2021-12-06
CSHARP-3985: Support multiple SerializerRegistry(s)

tmm360:


sanych-sun:
Thank you for your contribution. However having multiple Serializer Registries is much wider problem, because historically we used direct call to static registry a lot. We are going to review the approach in scope of upcoming major release.

PR #706, opened 2021-12-09
CSHARP-3990: Use `throw;` instead of `throw ex;` when re-throwing an unwrapped exception in BinaryConnection.

rstam:


rstam:
Evergreen patch build:

https://evergreen.mongodb.com/version/61b280d3306615750a497fe5

DmitryLukyanov (approved):
Interesting issue. LGTM

PR #708, opened 2021-12-14
CSHARP-3996: Fix cleanup-test-resources on Linux and Macos.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/61b7e7463627e00295e8b335

### DIFF
--- a/evergreen/cleanup-test-resources.sh
+++ b/evergreen/cleanup-test-resources.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 set -o xtrace   # Write all commands first to stderr
 
 # Environment variables used as input:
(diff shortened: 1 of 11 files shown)


JamesKovacs (changes requested):
We are missing the shebang (`#!`) for `get-python-path.sh`, `set-temp-fle-aws-creds.sh`, and `set-virtualenv.sh` as well. We also have a mix of `#!/bin/bash` and `#!/usr/bin/env bash`. We should standardize on the latter. Let's update CSHARP-3996 to fix all these issues.

DmitryLukyanov:
EG: https://evergreen.mongodb.com/version/61b893fda4cf470453bbcda5

DmitryLukyanov:
ready for review

JamesKovacs (approved):
LGTM

PR #709, opened 2021-12-14
CSHARP-3997: Suppress execution context flow when starting background tasks

kevingosse:
When the first instance of `MongoClient` is created, it initializes the connection to the server, which in turn starts a few background tasks (notably `RoundTripMonitor.RunAsync`). Because the execution context flow is not suppressed, this may result in unrelated objects being captured for an arbitrary amount of time.

This PR also removes usages of `.ConfigureAwait(false)` that are useless in this context (because the task is not awaited), and switches to `Task.Run` which has safer defaults than `Task.Factory.StartNew`.


DmitryLukyanov:
This logic was rewritten in CSHARP-3761 and has been released in 2.15. Please check this behavior with 2.15 release

PR #711, opened 2021-12-24
CSHARP-3154: fixed registration of DiscriminatorConvention for object type

tmm360:


adelinowona:
@tmm360 Thanks for your contribution, unfortunately will not be pursuing this work, find more details in the associated ticker for this PR.

PR #712, opened 2021-12-24
CSHARP-3980: Implement $explain modifier for CommandWireProtocol (OP_MSG).

DmitryLukyanov:
This is initial PR, there are some pending questions that might require some changes + adding more tests.
Some notes/questions:

0. This PR is placed on top of CSHARP-3858 which is not merged yet. I will rebase it as soon as the previous PR will be merged.
1. I'm not sure about current public explain API, so it would be good to discuss it.
2. do we want adding explain functionality for operations like?:
* findAndModify (and other similar)
* changeStream, probably there are more cases

3. I think logic in `FindCommandOperation` and `FindOperation` should be merged. Will do it in the final version.
4. Aggregate provides 2 ways to configure explain: 
* Explain command: https://docs.mongodb.com/manual/reference/command/explain/
* Explain argument in aggregate command itself: https://docs.mongodb.com/manual/reference/command/aggregate/#command-fields
In this PR I left only the second approach(the first one is used in Legacy) since it would be good to have the same code path for all explain "workflow"s. Also, only the second approach allows configuring verbosity, so this approach looks like the best. However I'm going to confirm that it doesn't give us some issues either performance or maybe missed functionality (probably only in January).
5. I'm also going to confirm whether I understand the format of server response for explain correctly (in other words there is no edge cases and etc).
6. In the scope of this PR, I also replaced serialization logic in AsyncCursor and Find in the same manner as it's done for aggregation. So, now aggregate and find and getmore use the same serializer. I'm not sure why it wasn't done in this way originally. So, we can discuss it.


src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
```diff
@@ -358,183 +354,69 @@ internal BsonDocument CreateCommand(ConnectionDescription connectionDescription,
...
-            return new ReadCommandOperation<AggregateResult>(databaseNamespace, command, serializer, MessageEncoderSettings)
-            {
-                RetryRequested = _retryRequested // might be overridden by retryable read context
-            };
-        }
 
-        private AsyncCursor<TResult> CreateCursor(IChannelSourceHandle channelSource, IChannelHandle channel, BsonDocument command, AggregateResult result)
-        {
-            if (result.CursorId.HasValue)
+            if (_explainVerbosity.HasValue)
```
DmitryLukyanov, line 375: Configuring explain in aggregation

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
```diff
@@ -358,183 +354,69 @@ internal BsonDocument CreateCommand(ConnectionDescription connectionDescription,
...
-                result.PostBatchResumeToken,
+                cursorContext.PostBatchResumeToken,
                 _batchSize,
                 null, // limit
                 _resultSerializer,
                 MessageEncoderSettings,
                 _maxAwaitTime);
         }
 
-        private AsyncCursor<TResult> CreateCursorFromInlineResult(BsonDocument command, AggregateResult result)
```
DmitryLukyanov, line 418: I'm going to confirm it (additionally to this [discussion](https://mongodb.slack.com/archives/C72LB5RPV/p1640358821319100?thread_ts=1640358021.318900&cid=C72LB5RPV)). But it looks like we no longer have a way to use this code path since the only approach that allowed cursorId be omitted is configuring `UseCursor=false` that we don't support anymore.

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
```diff
@@ -358,183 +354,69 @@ internal BsonDocument CreateCommand(ConnectionDescription connectionDescription,
...
-
         private void EnsureIsReadOnlyPipeline()
         {
             if (Pipeline.Any(s => { var n = s.GetElement(0).Name; return n == "$out" || n == "$merge"; }))
             {
                 throw new ArgumentException("The pipeline for an AggregateOperation contains a $out or $merge operator. Use AggregateOutputToCollectionOperation instead.", "pipeline");
             }
         }
-
-        private class AggregateResult
```
DmitryLukyanov, line 442: I've created a global class CursorContext instead

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
```diff
@@ -358,183 +354,69 @@ internal BsonDocument CreateCommand(ConnectionDescription connectionDescription,
...
-        {
-            public BsonTimestamp AtClusterTime;
-            public long? CursorId;
-            public CollectionNamespace CollectionNamespace;
-            public BsonDocument PostBatchResumeToken;
-            public TResult[] Results;
-        }
-
-        private class AggregateResultDeserializer : SerializerBase<AggregateResult>
```
DmitryLukyanov, line 451: Now it's named `CursorContextDeserializer`

src/MongoDB.Driver.Core/Core/Operations/AsyncCursor.cs
```diff
@@ -294,16 +351,16 @@ private CursorBatch<TDocument> ExecuteGetMoreCommand(IChannelHandle channel, Can
                 throw new MongoCursorNotFoundException(channel.ConnectionDescription.ConnectionId, _cursorId, command);
             }
 
-            return CreateCursorBatch(result);
+            return new CursorBatch<TDocument>(result);
```
DmitryLukyanov, line 354: it makes sense to call all deserializer logic on lower layers. It also allows reusing this logic in different places

src/MongoDB.Driver.Core/Core/Operations/ExplainOperation.cs
```diff
@@ -160,18 +170,41 @@ private ReadCommandOperation<BsonDocument> CreateReadOperation()
...
-                command,
-                BsonDocumentSerializer.Instance,
-                _messageEncoderSettings);
+            using var bsonDocumentReader = new BsonDocumentReader(explainedResult);
+            var context = BsonDeserializationContext.CreateRoot(bsonDocumentReader);
+            var result = _resultSerializer.Deserialize(context);
+
+            CollectionNamespace collectionNamespace;
+            var command = explainedResult["command"].AsBsonDocument;
+            if (!command.TryGetValue("aggregate", out var collectionName) &&
```
DmitryLukyanov, line 185: here we get a `collectionNamespace`  value from explain response. I'm going to check/confirm which logic is better here

src/MongoDB.Driver.Core/Core/Operations/FindOperation.cs
```diff
@@ -505,6 +519,12 @@ internal FindCommandOperation<TDocument> CreateFindCommandOperation()
                     switch (element.Name)
                     {
                         case "$comment": comment = _comment ?? value.AsString; break;
+                        case "$explain": explainVerbosity =
+                                _explainVerbosity ??
+                                (Enum.TryParse<ExplainVerbosity>(value.ToString(), ignoreCase: true, out var verbosity)
```
DmitryLukyanov, line 524: We can remove `Enum.TryParse` logic since previosuly `_modifiers["$explain"]` didn't allow setting verbosity (it was just false/true flag). 
I checked with old servers that setting `_modifiers["$explain"]` corresponds to the verbosity: `ExplainVerbosity.AllPlansExecution`. so we can leave only this part. Thoughts?

src/MongoDB.Driver/ProjectionDefinition.cs
```diff
@@ -272,6 +273,47 @@ public override RenderedProjectionDefinition<TProjection> Render(IBsonSerializer
         }
     }
 
+    /// <summary>
+    /// Transforms provided TProjection to BsonDocument.
+    /// </summary>
+    /// <typeparam name="TSource">The type of the source.</typeparam>
+    /// <typeparam name="TProjection">The type of the projection.</typeparam>
+    internal sealed class ExplainBsonDocumentProjectionDefinition<TSource, TProjection> : ProjectionDefinition<TSource, BsonDocument>
```
DmitryLukyanov, line 281: this needs to convert a projection with custom type in TProjection (that might be configured by a user) to the same projection but with replacing TProjection on the same value but projected to BsonDocument

src/MongoDB.Driver.Core/Core/Operations/FindCommandOperation.cs
```diff
@@ -558,23 +545,45 @@ public async Task<IAsyncCursor<TDocument>> ExecuteAsync(RetryableReadContext con
...
+        private IExecutableInRetryableReadContext<CursorContext<TDocument>> CreateOperation(RetryableReadContext context)
         {
             var command = CreateCommand(context.Channel.ConnectionDescription, context.Binding.Session);
-            var operation = new ReadCommandOperation<BsonDocument>(
-                _collectionNamespace.DatabaseNamespace,
-                command,
-                __findCommandResultSerializer,
-                _messageEncoderSettings)
+
+            if (_explainVerbosity.HasValue)
```
DmitryLukyanov, line 557: configuring Explain for FindCommandOperation

src/MongoDB.Driver.Core/Core/Operations/FindOperation.cs
```diff
@@ -38,6 +38,7 @@ public class FindOperation<TDocument> : IReadOperation<IAsyncCursor<TDocument>>,
         private readonly CollectionNamespace _collectionNamespace;
         private string _comment;
         private CursorType _cursorType;
+        private ExplainVerbosity? _explainVerbosity;
```
DmitryLukyanov, line 41: I think this class should be merged with `FindCommandOperation`. Will do it in the final commit.

src/MongoDB.Driver/FindFluent.cs
```diff
@@ -114,6 +116,46 @@ public override Task<long> CountDocumentsAsync(CancellationToken cancellationTok
...
+#pragma warning restore 618
+                // Projection is handled below
+                ReturnKey = _options.ReturnKey,
+                ShowRecordId = _options.ShowRecordId,
+                Skip = _options.Skip,
+                Sort = _options.Sort,
+            };
+            if (_options.Projection != null)
+            {
+                newOptions.Projection = new ExplainBsonDocumentProjectionDefinition<TDocument, TProjection>(_options.Projection);
```
DmitryLukyanov, line 151: Replace TProjection on BsonDocument

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
DmitryLukyanov: I left only logic that already presented in previous version, explain in aggregate is temporary reverted. I will create a ticket to proceed with these changes there

src/MongoDB.Driver/ProjectionDefinition.cs
DmitryLukyanov: This logic is reverted for the scope of this PR

src/MongoDB.Driver.Core/Core/Operations/FindOperation.cs
DmitryLukyanov: This is not so easy as I expected originally, I think we can implement it in the scope of the follow-up ticket

src/MongoDB.Driver.Core/Core/Operations/ExplainOperation.cs
```diff
@@ -160,18 +171,19 @@ private ReadCommandOperation<BsonDocument> CreateReadOperation()
...
-            return new WriteCommandOperation<BsonDocument>(
-                _databaseNamespace,
-                command,
-                BsonDocumentSerializer.Instance,
-                _messageEncoderSettings);
+            using var bsonDocumentReader = new BsonDocumentReader(explainedResult);
+            var context = BsonDeserializationContext.CreateRoot(bsonDocumentReader);
+            var result = _resultSerializer.Deserialize(context);
+
+            return new CursorBatch<TResult>(
```
DmitryLukyanov, line 184: Originally I had tried to parse explain response to get collectionName which will allow us having collectionName value all the time, however $explain responses are not the same in different servers and parsing logic became too complicated eventially, so it's better to handle lack of collectionNamespace before creating cursor

DmitryLukyanov:
The last commit https://github.com/mongodb/mongo-csharp-driver/pull/712/commits/e8a5ebaf7206a97fe13131be6911df403195ea87 reverts changes in any user faced logic that had not been implemented before. This ticket https://jira.mongodb.org/browse/CSHARP-4008 tracks reverted changes

DmitryLukyanov:
ready for review

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Operations/AggregateOperation.cs
DmitryLukyanov: but they're related to the CursorBatch, aren't they?

rstam (changes requested):
Waiting for changes that were discussed in a Zoom meeting with James.

rstam:
This is a large PR and it's hard to work through the comments so far. Here's a summary of the changes I'm requesting (a few might be new):

replace `description.ServerVersion` with `description.MaxWireVersion` (not `description.WireVersionRange`)
Use `MaxWireVersion` everywhere instead of `WireVersionRange` (min of range is irrelevant)

use `int` for `MinWireVersion` and `MaxWireVersion` instead of `WireVersion` class

`WireVersion` can be a static class with the following helper methods:

    WireVersion.ToServerVersion(wireVersion) // approximate server version
    WireVersion.ToWireVersion(serverVersion)
 
Remove ALL uses of `ServerVersion` in the driver (OK to use `ServerVersion` in some form in tests, but see warning below)

Add `ConnectionDescription.MinWireVersion` and `ConnectionDescription.MaxWireVersion` instead of `ConnectionDescription.WireVersionRange` (ease of use and readability)

Remove `ConnectionDescription.BuildInfoResult` property entirely
returning `null` or a gutted emulated value is already backward breaking

Implement `ConnectionDescription.ServerVersion` to return an estimated server version based on MaxWireVersion

But now tests can't use `ConnectionDescription.ServerVersion` because it's no longer accurate
Tests need to run `buildInfo` explicitly: refactor `CoreTestConfiguration.ServerVersion` to lazily run `buildInfo` when first referenced

All `Feature` classes (including subclasses) should use wire version only and not server version
Remove server version entirely from `Feature` classes

NOTE: I'm assuming we're resolving the contradictory decisions about whether to call `buildInfo` or not by deciding NOT to call it anymore

DmitryLukyanov:
ready for review

tests/MongoDB.Driver.Legacy.Tests/Jira/CSharp3980Tests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [Fact]
+        public void Cursor_Explain_should_throw()
+        {
+            var collection = GetCollection();
+            var query = new QueryDocument("X", 1);
+            var cursor = collection.Find(query);
+
+            var exception = Record.Exception(() => cursor.Explain());
+
+            exception.Should().BeOfType<NotSupportedException>();
```
rstam, line 51: `cursor.Explain` should actually work now so this test needs to be altered.

I'm assuming it doesn't pass for you?

DmitryLukyanov: done

rstam (changes requested):
Something went wrong in the rebasing on master.

The following files need to be deleted:

FindCommandOperation.cs
FindOpcodeOperation.cs
FindOpcodeOperationTests.cs

Everything else looks good.

DmitryLukyanov:
ready for review

rstam (approved):
LGTM

DmitryLukyanov:
merged.

PR #718, opened 2022-01-12
CSHARP-3928: Support 'let' option for multiple CRUD commands.

DmitryLukyanov:


src/MongoDB.Driver/MongoCollectionBase.cs
```diff
@@ -680,7 +697,8 @@ private async Task<ReplaceOneResult> ReplaceOneAsync(FilterDefinition<TDocument>
             {
                 var bulkWriteOptions = new BulkWriteOptions
                 {
-                    BypassDocumentValidation = options.BypassDocumentValidation
+                    BypassDocumentValidation = options.BypassDocumentValidation,
+                    Let = options.Let
```
DmitryLukyanov, line 701: this is part of follow up ticket: https://jira.mongodb.org/browse/DRIVERS-2027

src/MongoDB.Driver/FindFluent.cs
```diff
@@ -131,6 +131,7 @@ public override Task<long> CountDocumentsAsync(CancellationToken cancellationTok
                 Comment = _options.Comment,
                 CursorType = _options.CursorType,
                 Hint = _options.Hint,
+                Let = _options.Let,
```
DmitryLukyanov, line 134: `FindFluent` overrides `ToString` method. However it looks like the output there should be something that we can paster into the shell. But I don't see anything to configure `let` at least in `mongo` shell process from 53 server:

    MongoDB Enterprise replset:PRIMARY> db.test.find().help()
    find(<predicate>, <projection>) modifiers
            .sort({...})
            .limit(<n>)
            .skip(<n>)
            .batchSize(<n>) - sets the number of docs to return per getMore
            .collation({...})
            .hint({...})
            .readConcern(<level>)
            .readPref(<mode>, <tagset>)
            .count(<applySkipLimit>) - total # of objects matching query. by default ignores skip,limit
            .size() - total # of objects cursor would return, honors skip,limit
            .explain(<verbosity>) - accepted verbosities are {'queryPlanner', 'executionStats', 'allPlansExecution'}
            .min({...})
            .max({...})
            .maxTimeMS(<n>)
            .comment(<comment>)
            .tailable(<isAwaitData>)
            .noCursorTimeout()
            .allowPartialResults()
            .returnKey()
            .showRecordId() - adds a $recordId field to each returned object
            .allowDiskUse() - allow using disk in completing the query

    Cursor methods
            .toArray() - iterates through docs and returns an array of the results
            .forEach(<func>)
            .map(<func>)
            .hasNext()
            .next()
            .close()
            .objsLeftInBatch() - returns count of docs left in current batch (when exhausted, a new getMore will be issued)
            .itcount() - iterates through documents and counts them
            .getClusterTime() - returns the read timestamp for snapshot reads
            .pretty() - pretty print each document, possibly over multiple lines

src/MongoDB.Driver/MongoCollectionBase.cs
```diff
@@ -190,7 +190,11 @@ private DeleteResult DeleteMany(FilterDefinition<TDocument> filter, DeleteOption
             };
             try
             {
-                var result = bulkWriteFunc(new[] { model });
+                var bulkWriteOptions = options == null ? null : new BulkWriteOptions
```
rstam, line 193: `options` will never be null (see line 184).

rstam: Same for several other places in this file.

rstam (changes requested):
This looks great. Very good attention to detail.

Only a few minor comments.

src/MongoDB.Driver/MongoCollectionBase.cs
DmitryLukyanov: true, done

src/MongoDB.Driver/ReplaceOptions.cs
```diff
@@ -96,5 +98,14 @@ public bool IsUpsert
             get { return _isUpsert; }
             set { _isUpsert = value; }
         }
+
+        /// <summary>
+        /// Gets or sets a let document.
```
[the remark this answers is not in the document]
DmitryLukyanov: done

rstam (approved):
LGTM

PR #723, opened 2022-01-26
CSHARP-3984: Remove BinaryConnection.DropBox.

DmitryLukyanov:


src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -721,6 +644,23 @@ private bool AnyMessageNeedsToBeCompressed(IEnumerable<RequestMessage> messages)
             compressedMessageEncoder.WriteMessage(compressedMessage);
         }
 
+        private IByteBuffer EnsureResponseToIsCorrect(IByteBuffer message, int responseTo)
+        {
+            var receivedResponseTo = GetResponseTo(message);
+            if (receivedResponseTo != responseTo)
+            {
+                throw new InvalidOperationException($"The received responseTo {receivedResponseTo} doesn't match with expected {responseTo}."); // should not be reached
```
rstam, line 652: Can we word the message more like many other messages we already have?

    throw new InvalidOperationException($"Expected responseTo to be {responseTo} but was {receivedResponseTo}."); // should not be reached

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
```diff
@@ -721,6 +644,23 @@ private bool AnyMessageNeedsToBeCompressed(IEnumerable<RequestMessage> messages)
...
+            }
+
+            return message;
+
+            int GetResponseTo(IByteBuffer message)
+            {
+                var backingBytes = message.AccessBackingBytes(8);
```
rstam, line 659: I think it should be: 

    message.AccessBackingBytes(4);

Not 8.

rstam (changes requested):
Nice to see it wasn't hard to remove the Dropbox!

Minor changes requested.

JamesKovacs (approved):
LGTM

src/MongoDB.Driver.Core/Core/Connections/BinaryConnection.cs
rstam: Good catch!

Sorry for the mis-direction.

DmitryLukyanov: nice find @JamesKovacs !

rstam (approved):
LGTM

tests/MongoDB.Driver.Core.TestHelpers/CoreTestConfiguration.cs
```diff
@@ -75,6 +75,21 @@ public static DatabaseNamespace DatabaseNamespace
             get { return __databaseNamespace.Value; }
         }
 
+        public static ICluster FullyConnectedCluster
+        {
+            get
+            {
+                var timeout = TimeSpan.FromSeconds(30);
+                var cluster = __cluster.Value;
+                if (!SpinWait.SpinUntil(() => cluster.Description.Servers.Any(s => s.Type == ServerType.ReplicaSetPrimary), timeout))
```
rstam, line 84: This isn't what the name of the property says...

We should either change the name of the property or spin until the cluster is fully connected (presumably that means that the status of every server in the cluster is "connected").

rstam: This might be stretching the definition of what a property normally does. Properties normally do almost zero work and return quickly.

Maybe this should be a method (or several):

    var cluster = DriverTestConfiguration.GetClusterWithConnectedPrimary();
    var cluster = DriverTestConfiguration.GetClusterWithConnectedSecondary();
    var cluster = DriverTestConfiguration.GetFullyConnectedCluster();

rstam (changes requested):

PR #725, opened 2022-01-31
CSHARP-4001: Upgrade to DnsClient.NET 1.6.0, which includes the fix for duplicate DNS responses.

JamesKovacs:
Straightforward upgrade to DnsClient.NET. I was able to connect to MongoDB Atlas using `mongodb+srv://` in a test project. Open to suggestions of anything else I should verify before merging.

### DIFF
--- a/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
+++ b/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
@@ -175,7 +175,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="DnsClient" Version="1.4.0" />
+    <PackageReference Include="DnsClient" Version="1.6.0" />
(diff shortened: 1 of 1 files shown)


DmitryLukyanov (approved):
I'm not familiar with the external PR, but can we implement a driver-side test that emulates the initial issue and check that the issue has been fixed?

Otherwise, LGTM

JamesKovacs:
Testing this in the driver itself is difficult because it requires configuring DNS to return duplicate packets for a single request. I ran our repro which uses a simple Python script to return duplicate responses. DnsClient.NET 1.4.0 exhibits the bad behaviour of confusing the duplicate TXT response with the expected SRV response. DnsClient.NET 1.6.0 successfully retries in the face of duplicate TXT records and returns the correct SRV response.

Now why don't we incorporate this test into our test suite? Because this test is verifying that one of our dependencies (DnsClient.NET) behaves correctly in the face of DNS errors. And tests for handling these types of DNS errors were included in the DnsClient.NET PR that we submitted.

PR #727, opened 2022-02-10
CSHARP-3927: Drivers should check out an implicit session only after checking out a connection

BorisDog:
[EG](https://evergreen.mongodb.com/version/620544200305b97e469d3658)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
+++ b/src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
@@ -39,7 +39,7 @@ namespace MongoDB.Driver.Core.Bindings
         private bool _isCommitTransactionInProgress;
         private readonly IOperationClock _operationClock = new OperationClock();
         private readonly CoreSessionOptions _options;
-        private readonly ICoreServerSession _serverSession;
+        private readonly Lazy<ICoreServerSession> _serverSession;
         private BsonTimestamp _snapshotTime;
 
         // constructors
@@ -49,13 +49,32 @@ namespace MongoDB.Driver.Core.Bindings
         /// <param name="cluster">The cluster.</param>
         /// <param name="serverSession">The server session.</param>
         /// <param name="options">The options.</param>
+        [Obsolete("This constructor is deprecated. Avoid using CoreSession directly.")]
         public CoreSession(
             ICluster cluster,
             ICoreServerSession serverSession,
             CoreSessionOptions options)
+            : this(cluster, options: options)
+        {
+            Ensure.IsNotNull(serverSession, nameof(serverSession));
+            _serverSession = new Lazy<ICoreServerSession>(() => serverSession);
+        }
+
+        internal CoreSession(
+            ICluster cluster,
+            ICoreServerSessionPool serverSessionPool,
+            CoreSessionOptions options)
+             : this(cluster, options)
+        {
+            Ensure.IsNotNull(serverSessionPool, nameof(serverSessionPool));
+            _serverSession = new Lazy<ICoreServerSession>(() => serverSessionPool.AcquireSession());
+        }
+
+        private CoreSession(
+           ICluster cluster,
+           CoreSessionOptions options)
         {
             _cluster = Ensure.IsNotNull(cluster, nameof(cluster));
-            _serverSession = Ensure.IsNotNull(serverSession, nameof(serverSession));
             _options = Ensure.IsNotNull(options, nameof(options));
         }
 
@@ -75,13 +94,13 @@ namespace MongoDB.Driver.Core.Bindings
         public CoreTransaction CurrentTransaction => _currentTransaction;
 
         /// <inheritdoc />
-        public BsonDocument Id => _serverSession.Id;
+        public BsonDocument Id => _serverSession.Value.Id;
 
         /// <inheritdoc />
         public bool IsCausallyConsistent => _options.IsCausallyConsistent;
 
         /// <inheritdoc />
-        public bool IsDirty => _serverSession.IsDirty;
+        public bool IsDirty => _serverSession.Value.IsDirty;
 
         /// <inheritdoc />
         public bool IsImplicit => _options.IsImplicit;
@@ -120,7 +139,7 @@ namespace MongoDB.Driver.Core.Bindings
         public CoreSessionOptions Options => _options;
 
         /// <inheritdoc />
-        public ICoreServerSession ServerSession => _serverSession;
+        public ICoreServerSession ServerSession => _serverSession.Value;
 
         /// <inheritdoc />
         public BsonTimestamp SnapshotTime => _snapshotTime;
@@ -270,7 +289,7 @@ namespace MongoDB.Driver.Core.Bindings
         /// <inheritdoc />
         public long AdvanceTransactionNumber()
         {
-            return _serverSession.AdvanceTransactionNumber();
+            return _serverSession.Value.AdvanceTransactionNumber();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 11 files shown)


tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
```diff
@@ -40,5 +48,61 @@ public void Snapshot_and_causal_consistent_session_is_not_allowed()
             var exception = Record.Exception(() => mongoClient.StartSession(sessionOptions));
             exception.Should().BeOfType<NotSupportedException>();
         }
+
+        [SkippableFact]
+        public async Task Ensure_server_session_are_allocated_only_on_connection_checkout()
```
BorisDog, line 53: Prose test as described in the spec will not work with current implementation. Currently drivers will make their own judgment on the testing strategy.

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -52,11 +52,33 @@ public sealed class CoreSession : ICoreSession
...
+                serverSession: serverSession,
+                serverSessionProvider: null,
+                options: options)
+        {
+        }
+
+        internal CoreSession(
+            ICluster cluster,
+            ICoreServerSession serverSession,
+            ICoreServerSessionProvider serverSessionProvider,
```
rstam, line 66: The previous two parameters are mutually exclusive. Instead consider refactoring the constructors as follows so that only a `serverSession` or a `serverSessionPool` is provided:

    public CoreSession(
        ICluster cluster,
        ICoreServerSession serverSession,
        CoreSessionOptions options)
        : this(
            cluster: cluster,
            serverSession: new Lazy<ICoreServerSession>(() => serverSession, isThreadSafe: true),
            options: options)
    {
    }

    internal CoreSession(
        ICluster cluster,
        ICoreServerSessionPool serverSessionPool,
        CoreSessionOptions options)
        : this(
            cluster: cluster,
            serverSession: new Lazy<ICoreServerSession>(() => serverSessionPool.AcquireSession(), isThreadSafe: true),
            options: options)
    {
    }

    internal CoreSession(
        ICluster cluster,
        Lazy<ICoreServerSession> serverSession,
        CoreSessionOptions options)
    {
        _cluster = Ensure.IsNotNull(cluster, nameof(cluster));
        _serverSession = Ensure.IsNotNull(serverSession, nameof(serverSession));
        _options = Ensure.IsNotNull(options, nameof(options));
    }

NOTE: above assumes the recommendation to not introduce the new `ICoreSessionProvider` interface is accepted.

src/MongoDB.Driver.Core/Core/Bindings/ICoreServerSessionPool.cs
```diff
@@ -15,18 +15,20 @@
 
 namespace MongoDB.Driver
 {
-    /// <summary>
-    /// A server session pool.
-    /// </summary>
-    internal interface ICoreServerSessionPool
+    internal interface ICoreServerSessionProvider
```
rstam, line 18: I don't see any value in splitting `ICoreServerSessionPool` into two.

I think we should leave this interface unchanged.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -329,8 +329,7 @@ public async Task<IServer> SelectServerAsync(IServerSelector selector, Cancellat
         public ICoreSessionHandle StartSession(CoreSessionOptions options)
         {
             options = options ?? new CoreSessionOptions();
-            var serverSession = AcquireServerSession();
-            var session = new CoreSession(this, serverSession, options);
+            var session = new CoreSession(this, null, _serverSessionPool, options);
```
rstam, line 332: After refactoring constructors to not have mutually exclusive parameters:

    var session = new CoreSession(this, _serverSessionPool, options);

src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
```diff
@@ -230,8 +230,7 @@ public ICoreSessionHandle StartSession(CoreSessionOptions options = null)
             ThrowIfDisposed();
 
             options = options ?? new CoreSessionOptions();
-            var serverSession = AcquireServerSession();
-            var session = new CoreSession(this, serverSession, options);
+            var session = new CoreSession(this, null, _serverSessionPool, options);
```
rstam, line 233: After refactoring constructors to not have mutually exclusive parameters:

    var session = new CoreSession(this, _serverSessionPool, options);

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -52,11 +52,33 @@ public sealed class CoreSession : ICoreSession
         public CoreSession(
             ICluster cluster,
             ICoreServerSession serverSession,
+            CoreSessionOptions options) : this(
```
rstam, line 55: Formatting suggestion: we like to put the call to `this` on the following line.

Reasons:

1. It is logically separate from the parameter preceding it
2. Important things shouldn't be hidden near the end of the line

rstam (changes requested):
Overall looks good. Some changes requested.

src/MongoDB.Driver.Core/Core/Bindings/ICoreServerSessionPool.cs
BorisDog: Yeah not sure what was the initial justification for splitting, reverted.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs

src/MongoDB.Driver.Core/Core/Clusters/LoadBalancedCluster.cs
BorisDog: Done.

src/MongoDB.Driver/ClientSessionHandle.cs
```diff
@@ -78,7 +76,18 @@ internal ClientSessionHandle(IMongoClient client, ClientSessionOptions options,
         public ClientSessionOptions Options => _options;
 
         /// <inheritdoc />
-        public IServerSession ServerSession => _serverSession;
+        public IServerSession ServerSession
+        {
+            get
+            {
+                if (_serverSession == null)
```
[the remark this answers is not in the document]
BorisDog: 1. No, sessions are not thread safe in general.
2. Specifically `ServerSession` wrapper does not introduce additional state, so having multiple copies is safe.
3. General note: I could not find any usages of `ServerSession` in our code or Spec, so this property would be a candidate for removal.

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
BorisDog: Good point, exposing mutually exclusive parameters even internally is not a good idea. Added additional private constructor.
Please note that `Lazy<ICoreServerSession>` will never be `null`, and the underlying objects are not validated in constructor. So the new private constructor still contains the both params, as I'd prefer to have all validations grouped together.

src/MongoDB.Driver.Core/Core/Bindings/ICoreServerSessionPool.cs
```diff
@@ -20,7 +20,6 @@ namespace MongoDB.Driver
     /// </summary>
     internal interface ICoreServerSessionPool
     {
-        // methods
```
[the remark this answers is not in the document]
BorisDog: Done.

rstam (approved):
LGTM

PR #729, opened 2022-02-21
CSHARP-3761: Fix issue with ObjectDisposedException.

DmitryLukyanov:
Additional EG links:
* https://evergreen.mongodb.com/patch/621427b732f4176b1fd74270
* https://evergreen.mongodb.com/version/621427de5623435871e65ebe

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Misc/ThreadHelper.cs
+++ b/src/MongoDB.Driver.Core/Core/Misc/ThreadHelper.cs
@@ -23,7 +23,19 @@ namespace MongoDB.Driver.Core.Misc
     {
         public static void Sleep(TimeSpan timeout, CancellationToken cancellationToken)
         {
-            Task.Delay(timeout, cancellationToken).Wait(cancellationToken);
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver.Core/Core/Misc/ThreadHelper.cs
```diff
@@ -23,7 +23,19 @@ internal static class ThreadHelper
...
+            WaitTask(Task.Delay(timeout), cancellationToken);
+        }
+
+        public static void WaitTask(Task task, CancellationToken cancellationToken)
+        {
+            try
+            {
+                task.Wait(cancellationToken);
+            }
+            catch (ObjectDisposedException)
```
BorisDog, line 35: Looks like there is no need to catch `ObjectDisposedException` here, as the simple usage of the token like here won't result in `DisposedException`. So `Sleep` method can remain as is.

BorisDog (changes requested):

BorisDog (approved):
LGTM

PR #738, opened 2022-03-02
CSHARP-3225: Skip SetWindowFields tests on older servers.

rstam:


rstam:
Do you want me to run a full Evergreen patch?

JamesKovacs (approved):
LGTM

JamesKovacs:
Verifying that tests run on 5.0+ and don't run on pre-5.0 should be sufficient. There is nothing OS or topology specific in the tests. So a single variant of each should suffice.

PR #742, opened 2022-03-07
CSHARP-4069: Fix flaky Misbehaved_async_method_should_not_deadlock_server_selection test.

DmitryLukyanov:


JamesKovacs (approved):
LGTM

FYI - I looked for other tests like `CSharp2564Tests` that depended on finding a primary within a timeout, but couldn't find any - which surprised me.

PR #749, opened 2022-03-09
CSHARP-4087: Fixed API docs warnings.

JamesKovacs:


JamesKovacs:
Note that only CSHARP-4087 should be included in this PR. CSHARP-4089 (which this PR is built on top of) is part of https://github.com/mongodb/mongo-csharp-driver/pull/748.

JamesKovacs:
Merged manually.

PR #753, opened 2022-03-21
CSHARP-4079: Implement positional update operators in LINQ3.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4079Tests.cs
```diff
@@ -0,0 +1,195 @@
...
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.TestHelpers.XunitExtensions;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
+{
+    public class CSharp4079Tests : Linq3IntegrationTest
```
rstam, line 26: These tests are the best place to see what the new public API looks like.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4079Tests.cs
```diff
@@ -0,0 +1,195 @@
...
+            rendered.Should().Be("{ $inc : { 'Grades.$[].Questions.$[score]' : 2 } }");
+        }
+
+        // the following examples are from the server documentation:
+        // hhttps://docs.mongodb.com/manual/reference/operator/update/positional-filtered/
+
+        [Fact]
+        public void Filtered_positional_update_operator_update_all_array_elements_that_match_array_filter_example()
+        {
+            var update = Builders<Student1>.Update.Set(s => s.Grades.AllMatchingElements("element"), 100);
```
JamesKovacs, line 130: It would be much nicer if the user could specify an array filter as a lambda. Then the driver could transform this into a filter name and append the array filter to the update options. That is total scope creep and this syntax is sufficient for now.

JamesKovacs (approved):
LGTM. I really like the syntax. Much clearer and expressive than LINQ2's -1 magic. One minor correction in the URL noted below.

PR #757, opened 2022-03-26
CSHARP-4042: Rename "Versioned API" to "Stable API" in documentation

BorisDog:


evergreen/evergreen.yml
```diff
@@ -930,7 +930,7 @@ tasks:
         - func: run-aws-auth-test-with-aws-EC2-credentials
         # ECS test is skipped untill testing on Linux becomes possible
 
-    - name: versioned-api-tests-net472
+    - name: stable-api-tests-net472
```
DmitryLukyanov, line 933: We don't change our cake and yml scripts, I would say this one also can be left alone

DmitryLukyanov (commented):
Question and then LGTM

evergreen/evergreen.yml
BorisDog: I think that we should update the naming for better hygiene. We could do this also as part of [DRIVERS-2216](https://jira.mongodb.org/browse/DRIVERS-2216) (but that's also not a perfect fit) so I don't see a reason not do it just now, like [python](https://github.com/mongodb/mongo-python-driver/pull/867/files) did.

DmitryLukyanov: EG.yml works in pair with sh/cake scripts that were not updated. This is why it confuses a bit.

As for changes in different drivers, mostly they updated more code than we do (and than was requested as I see). For example mentioned python applied changes in comments, some other drivers changed examples and etc.. So I don't know whether we need to do anything other than docs. I will rely on you here

DmitryLukyanov (approved):
LGTM

PR #758, opened 2022-03-29
CSHARP-4115: Case insensitive string comparison support for LINQ translator

tspence:
The LINQ translator does not recognize calls to `String.Compare(a, b, StringComparison.OrdinalIgnoreCase)` in the LINQ translator.  This means that it's not possible to create a LINQ query that does a case insensitive string comparison; only case sensitive string equals works today.

The goal is to convert this:
`String.Compare(a, "value", StringComparison.OrdinalIgnoreCase) >= 0`

Into this:
`{ $expr: { $gte: [ { $strcasecmp: [ \"$a\", \"value\" ] }, 0 ]} }`

This patch implements this feature in a similar method to the way String.Equals is implemented today.

Unfortunately, as I'm a new contributor, I haven't yet figured out how to run the tests prior to submitting a PR.  It looks like there are environment variables and AWS keys that must be set up for the tests to function correctly - is it possible for me to run these tests standalone?

Potential improvements:
* Modify the "CompareTo" translator to also support CompareTo calls of type String; today it only recognizes generic $cmp
* Add a function into FilterBuilder called `.StrCaseCmp` to allow casual users to make these calls via shorthand


JamesKovacs:
Hi, @tspence, 

Thank you for the PR. I see that @rstam already replied on [CSHARP-4115](https://jira.mongodb.org/browse/CSHARP-4115) explaining the challenges with `StringComparison.OrdinalIgnoreCase`. I've closed this PR for the reasons explained in that ticket. This is an area that we are actively trying to improve at the moment and appreciate your contribution.

Sincerely,
James

PR #760, opened 2022-03-30
CSHARP-4062: Projection of constant with no known serializer is not supported.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersRegistry.cs
```diff
@@ -40,7 +40,7 @@ public IBsonSerializer GetSerializer(Expression expression, IBsonSerializer defa
             var possibleSerializers = _registry.TryGetValue(expression, out var knownSerializers) ? knownSerializers.GetPossibleSerializers(expressionType) : new HashSet<IBsonSerializer>();
             return possibleSerializers.Count switch
             {
-                0 => defaultSerializer ?? throw new InvalidOperationException($"Cannot find serializer for {expression}."),
+                0 => defaultSerializer ?? BsonSerializer.LookupSerializer(expressionType),
```
JamesKovacs, line 43: As much as I don't like comments, this is a subtle change that I think deserves some explanation of why we have to fall back to the serializer registry.

rstam: Done

PR #766, opened 2022-04-11
CSHARP-4108: Inherited comment causes OperationFailure on MongoDB <4.4.0

BorisDog:
Also
CSHARP-4109: Test getMore with comment on all topologies


rstam (approved):
Nice simple change. LGTM.

PR #767, opened 2022-04-13
CSHARP-4136: CSFLE badQueries and types errorContains failing to match on mongocryptd 6.0.0-alpha

BorisDog:
[EG](https://evergreen.mongodb.com/version/625713e10305b93ec0ffe5b2)


DmitryLukyanov (approved):
LGTM if EG is green regarding CSFLE

PR #768, opened 2022-04-13
CSHARP-4126: Replace parameter names that aren't valid server variable names with generated variable names.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Misc/NameGenerator.cs
```diff
@@ -40,7 +40,7 @@ public string GetVarName(string symbolName)
                 return symbolName;
             }
 
-            return $"_v{_varCounter++}";
+            return $"v__{_varCounter++}";
```
rstam, line 43: This code existed to replace parameter names that weren't valid server variable names with a generated variable name.

Unfortunately, we were generating a variable name that was itself not valid on the server. The server does not allow leading underscores.

I think we should use double underscores in our generated variable names to reduce the likelihood of colliding with a user variable name.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/NameGenerator.cs
```diff
@@ -52,7 +52,7 @@ static bool IsValidVarName(string name)
 
                 static bool IsValidFirstChar(char c)
                 {
-                    return c == '_' || IsBetween(c, 'a', 'z') || IsBetween(c, 'A', 'Z');
+                    return IsBetween(c, 'a', 'z');
```
rstam, line 55: The server is even stricter than I realized about server variable names.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4126Tests.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Fact]
+        public void Test()
+        {
+            var collection = GetCollection<C>();
+            CreateCollection(
+                collection,
+                new C { Id = 1, A = new[] { 1, -2, -3 } });
+
+            var queryable = collection.AsQueryable()
+                .Select(_p0 => new { Id = _p0.Id, A = _p0.A.Select(_p1 => Math.Abs(_p1)) });
```
rstam, line 34: Note: `_p0` doesn't consume a generated variable name because it is replaced with "ROOT" instead.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Linq3IntegrationTest.cs
```diff
@@ -51,6 +51,11 @@ protected void CreateCollection<TDocument>(IMongoCollection<TDocument> collectio
             }
         }
 
+        protected void CreateCollection<TDocument>(IMongoCollection<TDocument> collection, params TDocument[] documents)
```
rstam, line 54: New convenience overload.

JamesKovacs (approved):
LGTM

PR #770, opened 2022-04-18
CSHARP-4135: Add CSFLE spec test for auto encryption on a collection with no jsonSchema.

DmitryLukyanov:
As far as I understood, the auto CSFLE behavior with no schema can be different and sometimes libmongocrypt can throw exception. Kevin mentioned this case: `an aggregate with $lookup stage to another collection`. But for this particular test case, no schema means no encryption, even though some inner machinery still will be involved


BorisDog (approved):
LGTM. It would be great to have a full EG before merging.

PR #773, opened 2022-04-19
CSHARP-1950: Allow string or regular expressions in values for $in with string.

rstam:


src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -291,6 +292,50 @@ public FilterDefinition<TDocument> AnyNin<TItem>(Expression<Func<TDocument, IEnu
             return AnyNin(new ExpressionFieldDefinition<TDocument>(field), values);
         }
 
+        /// <summary>
+        /// Creates an in filter for a string array field.
+        /// </summary>
+        /// <param name="field">The field.</param>
+        /// <param name="values">The values.</param>
+        /// <returns>An in filter.</returns>
+        public FilterDefinition<TDocument> AnyStringIn(FieldDefinition<TDocument, IEnumerable<string>> field, IEnumerable<StringOrRegularExpression> values)
```
rstam, line 301: This is like `AnyIn` except that `field` is `FieldDefinition<TDocument, IEnumerable<string>>` instead of `FieldDefinition<TDocument, IEnumerable<TItem>>` and `values` is `IEnumerable<StringOrRegularExpression>` instead of `IEnumerable<TItem>`.

Similarly for other new methods below.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -2327,6 +2416,85 @@ public override BsonDocument Render(IBsonSerializer<TDocument> documentSerialize
         }
     }
 
+    internal abstract class StringInOrNinFilterDefinition<TDocument, TField> : FilterDefinition<TDocument>
```
rstam, line 2419: This is different from a regular `$in` (or `$nin`) in that `values` is `IEnumerable<StringOrRegularExpression>` instead of `IEnumerable<string>`.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -2327,6 +2416,85 @@ public override BsonDocument Render(IBsonSerializer<TDocument> documentSerialize
...
+                var regularExpressionSerializer = BsonRegularExpressionSerializer.Instance;
+
+                bsonWriter.WriteStartDocument();
+                bsonWriter.WriteName(renderedField.FieldName);
+                bsonWriter.WriteStartDocument();
+                bsonWriter.WriteName(_operator);
+                bsonWriter.WriteStartArray();
+                foreach (var value in _values)
+                {
+                    if (value?.Type == typeof(BsonRegularExpression))
```
rstam, line 2458: Serialize `BsonRegularExpression` as a `BsonRegularExpression`. Do *not* convert to `string`.

src/MongoDB.Driver/StringOrRegularExpression.cs
```diff
@@ -0,0 +1,85 @@
...
+using System;
+using System.Text.RegularExpressions;
+using MongoDB.Bson;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Represents a value that is either a string or a BsonRegularExpression.
+    /// </summary>
+    public class StringOrRegularExpression
```
rstam, line 25: I searched for general purpose discriminated unions and found some stuff but in the end I don't want a dependency on any third party Nuget package for this.

This is similar to what some examples on the web might call `OneOf<string, BsonRegularExpression>`.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
JamesKovacs: One slight annoyance I found is that common use cases such as only using only string or only regex arguments requires the use of `new StringOrRegularExpression[] { }`. Changing the type signature to:

`public FilterDefinition<TDocument> AnyStringIn(FieldDefinition<TDocument, IEnumerable<string>> field, params StringOrRegularExpression[] values)`

allows us to call the method like this:
`var builder = Builders<C>.Filter.StringIn(x => x.Name, "A", "B");`
-- or --
`var builder = Builders<C>.Filter.StringIn(x => x.Name, new Regex("^start"), new Regex("end$"));`

The disadvantage of this technique is it breaks from the way `In` is defined:
`public FilterDefinition<TDocument> In<TField>(FieldDefinition<TDocument, TField> field, IEnumerable<TField> values)`

What do you think?

JamesKovacs (changes requested):
This works in `Builders<T>`, but is not supported in LINQ. For example, the following results in a compile error:
```
var query = coll.AsQueryable().Where(x => x.Name.In(new StringOrRegularExpression[] {new Regex("^start"), "A"}));
```
And `x.Name.StringIn` doesn't exist as a string extension method.

It would be ideal if we supported params-style syntax, but that would be a departure from the method signature for `In`:

```
var query = coll.AsQueryable().Where(x => x.Name.StringIn(new Regex("^start"), "A"));
```

src/MongoDB.Bson/Serialization/Serializers/StringSerializer.cs
```diff
@@ -22,6 +22,16 @@ namespace MongoDB.Bson.Serialization.Serializers
...
     public class StringSerializer : SealedClassSerializerBase<string>, IRepresentationConfigurable<StringSerializer>
     {
+        #region static
+        private static readonly StringSerializer __instance = new StringSerializer();
+
+        // public static properties
+        /// <summary>
+        /// Gets a cached instance of a default string serializer.
+        /// </summary>
+        public static StringSerializer Instance => __instance;
```
rstam, line 32: This cached instance allows us to avoid creating a new one every time we need one.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
rstam: Added `params` overloads.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/StringInOrNinMethodToFilterTranslator.cs
```diff
@@ -0,0 +1,92 @@
...
+        private static readonly MethodInfo[] __stringInMethods =
+        {
+            StringMethod.AnyStringInWithEnumerable,
+            StringMethod.AnyStringInWithParams,
+            StringMethod.StringInWithEnumerable,
+            StringMethod.StringInWithParams,
+        };
+
+        // public static methods
+        public static AstFilter Translate(TranslationContext context, MethodCallExpression expression)
```
rstam, line 51: This is the implementation of the new LINQ support for $in and $nin.

src/MongoDB.Driver/Linq/StringExtensions.cs
```diff
@@ -22,6 +23,50 @@ namespace MongoDB.Driver.Linq
     /// </summary>
     public static class StringExtensions
     {
+        /// <summary>
+        /// Returns true if any value in s is present in values (corresponds to the $in filter operator).
+        /// </summary>
+        /// <param name="s">The values to test.</param>
+        /// <param name="values">The values to test against.</param>
+        /// <returns>True if any value in s is present in values.</returns>
+        public static bool AnyStringIn(this IEnumerable<string> s, IEnumerable<StringOrRegularExpression> values)
```
rstam, line 32: These are the new string extension methods used to support $in and $nin in LINQ.

tests/MongoDB.Driver.Tests/Jira/CSharp1950BuilderTests.cs
```diff
@@ -0,0 +1,494 @@
...
+                "{ \"S\" : { \"$in\" : [] } }");
+        }
+
+        [Fact]
+        public void StringIn_with_field_name_and_one_regex_should_work()
+        {
+            Assert(
+                builder => builder.StringIn("S", new BsonRegularExpression("^a")),
+                "{ \"S\" : { \"$in\" : [/^a/] } }",
+                "a1", "a2");
```
rstam, line 43: The arguments to Assert are:

1. the filter to test
2. the expected rendered filter
3. the expected results (third argument omitted if no results expected)

All tests do the same thing, with the exception of the filter and the expected results. All the logic common to all tests is in the Assert helper method.

tests/MongoDB.Driver.Tests/Jira/CSharp1950LinqTests.cs
```diff
@@ -0,0 +1,525 @@
...
+                "{ $match : { S : { $in : [] } } }");
+        }
+
+        [Fact]
+        public void StringIn_with_array_of_one_regex_should_work()
+        {
+            Assert(
+                queryable => queryable.Where(x => x.S.StringIn(new BsonRegularExpression("^a"))),
+                "{ $match : { S : { $in : [/^a/] } } }",
+                "a1", "a2");
```
rstam, line 51: The three arguments to Assert are:

1. the queryable to test
2. the expected pipeline stage
3. the expected results (third argument is omitted if no results are expected)

tests/MongoDB.Driver.Tests/Jira/CSharp1950LinqTests.cs
```diff
@@ -0,0 +1,525 @@
...
+
+namespace MongoDB.Driver.Tests.Jira
+{
+    public class CSharp1950LinqTests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void StringIn_with_no_arguments_should_work()
+        {
+            Assert(
+                queryable => queryable.Where(x => x.S.StringIn()),
```
rstam, line 33: The tests in this file test the newly added LINQ support for $in and $nin you requested.

So this is a good file to see what the public API for $in and $nin in LINQ looks like.

JamesKovacs (approved):
LGTM

PR #777, opened 2022-04-25
Set __isBufferRented before returning

matthid:
This prevents a rare case where the flag stays `true` when an exception occurs (for example OOM)

### DIFF
--- a/src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
+++ b/src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
@@ -79,26 +79,32 @@ namespace MongoDB.Bson.IO
             {
                 throw new ArgumentOutOfRangeException(nameof(size), size, $"Valid buffer size range is [1..{MaxAllocationSize}] bytes.");
             }
-
-            __isBufferRented = true;
-
+            
             if (__threadId == default)
             {
                 __threadId = Thread.CurrentThread.ManagedThreadId;
             }
-
+            
+            // Allocate space
+            RentedBuffer ret;
             if (size > MaxSize)
             {
-                return new RentedBuffer(__threadId, new byte[size]);
+                ret = new RentedBuffer(__threadId, new byte[size]);
             }
-
-            if (__buffer == null || __buffer.Length < size)
+            else
             {
-                var newSize = size <= MinSize ? MinSize : PowerOf2.RoundUpToPowerOf2(size);
-                __buffer = new byte[newSize];
+                if (__buffer == null || __buffer.Length < size)
+                {
+                    var newSize = size <= MinSize ? MinSize : PowerOf2.RoundUpToPowerOf2(size);
+                    __buffer = new byte[newSize];
+                }
+
+                ret = new RentedBuffer(__threadId, __buffer);
             }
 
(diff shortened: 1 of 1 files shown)


JamesKovacs:
Hi, @matthid,

Thank you for submitting this PR. Please file an issue at https://jira.mongodb.org/browse/CSHARP that describes the problem being fixed. It would also help to have unit tests to demonstrate that the fix behaves as intended.

Sincerely,
James

matthid:
hi @JamesKovacs, see https://jira.mongodb.org/browse/CSHARP-4159
Sadly this is quite hard to reproduce, you might be able to mock some part of the framework to do it (like make `Thread.CurrentThread.ManagedThreadId` throw an exception), I'm not sure you have to infrastructure in place to do it and I don't think it is worth building for this particular issue.
The good thing is that the problem and fix should be quite obvious from code inspection and there are other ways to fix it (like a try-finally) so you should feel free to fix it in a different way.

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
```diff
@@ -79,26 +79,32 @@ public static RentedBuffer RentBuffer(int size)
...
-
+            
             if (__threadId == default)
             {
                 __threadId = Thread.CurrentThread.ManagedThreadId;
             }
-
+            
+            // Allocate space
+            RentedBuffer ret;
```
DmitryLukyanov, line 89: please rename this variable to `rentedBuffer`.

MindForks:
I've faced the same problem when __isBufferRented never unset. Will the fix be merged?

src/MongoDB.Bson/IO/ThreadStaticBuffer.cs
matthid: Please go ahead and change as you'd like - just as I mentioned initially. 
The PR was more or less just to demonstrate the issue and a potential solution.

Schanckopotamus:
@matthid I started seeing this being thrown in our Prod system as well. During a spike in load. Was wondering your thoughts on how to properly handle this. Was thinking about disposing the whole connection and starting over but I don't see that as a `try-finally` solution. Was wondering your thoughts on that since you mentioned it above.

matthid:
@Schanckopotamus The only thing you can do to keep your app running is to dispose the whole thread this happens in, but this is not viable depending on which threads you use (like the main thread). The easiest is to restart the application when this happens (gracefully shutdown / failfast).
The easiest would be if the mongo guys would just fix it.

Schanckopotamus:
@matthid Yeah that was my fear that I would have to kill the thread. I might be able to figure something out. I'm running a separate hosted service in an asp.net core Worker service for each ChangeStream created. I might be able to do something there, I'll have to research (haven't had an issue yet to make me dig into restarting a hosted service), until this is fixed in the Driver. 

Man it's gonna be a gross solution I bet, ugh.

Thank you for the insight.

BorisDog:
@matthid thank you for submitting a PR for this issue. Merged [here](https://github.com/mongodb/mongo-csharp-driver/commit/a87a6c5122bc0991abf493aa6d68e6f8a8fc1b7b).

PR #771, opened 2022-04-18
CSHARP-4034: Change streams support for user-facing PIT pre- and post-images

BorisDog:
[EG](https://spruce.mongodb.com/version/625db17e7742ae58d2d83865)

PR #778, opened 2022-04-25
CSHARP-3946: Allow limit in $filter expression.

rstam:


src/MongoDB.Driver.Core/Core/Misc/WireVersion.cs
```diff
@@ -80,6 +80,14 @@ internal static class WireVersion
         /// Wire version 15.
         /// </summary>
         public const int Server52 = 15;
+        /// <summary>
+        /// Wire version 16.
+        /// </summary>
+        public const int Server53 = 16;
```
rstam, line 86: I only needed Server60 but I didn't want to leave a gap here so I added Server53 also.

rstam: Boris also added Server53 and Server60 in his PR which I reviewed this morning:

https://github.com/mongodb/mongo-csharp-driver/pull/771

He did it correctly (I missed other required changes). 

I will wait for his PR to be pushed to master and after that rebase this PR on his changes.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -586,5 +587,18 @@ public static decimal StandardDeviationSample<TSource>(this IEnumerable<TSource>
...
+
+        /// <summary>
+        /// Filters a sequence of values based on a predicate and limits the number of results.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source" />.</typeparam>
+        /// <param name="source">The source values.</param>
+        /// <param name="predicate">The predicate.</param>
+        /// <param name="limit">The limit.</param>
+        /// <returns>The filtered results.</returns>
+        public static IEnumerable<TSource> WhereWithLimit<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate, long limit)
```
JamesKovacs, line 599: The syntax `WhereWithLimit` doesn’t feel very LINQ-like to me:
```
var queryable = collection.AsQueryable()
     .Select(x => x.IA.WhereWithLimit(i => i >= 2, 2));
```
I would express this more naturally as:
```
var queryable = collection.AsQueryable()
    .Select(x => x.IA.Where(i => i>=2).Take(2));
```
How hard would it be to transform the `Where`/`Take` expression into a `$filter` with limit? I know we usually map one LINQ call into 1 or more pieces of MQL, but I don’t think we take multiple LINQ calls and emit a single MQL function.

JamesKovacs (commented):
Let's discuss the complexity of using `Where`/`Take` syntax rather than `WhereWithLimit`.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
rstam: I had mentioned in Slack before putting up the PR why I chose a new method name instead of detecting the `Where/Take` pattern.

We are discussing it further.

In summary: `$filter` with `limit` is *only* supported on the latest server. Using a new method name like this allows the user to opt-in to `$filter` with `limit`. Detecting the `Where/Take` pattern would translate to `$filter` with `limit` without knowing whether the server supported or not. That would be a breaking change.

rstam: Result of offline discussion: keep current design but overload `Where` instead of using `WhereWithLimit`.

Also, limit should be `int` not `long`, and add test showing `limit` can be computed server-side.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/WhereMethodToAggregationExpressionTranslator.cs
```diff
@@ -28,22 +28,32 @@ public static AggregationExpression Translate(TranslationContext context, Method
             var method = expression.Method;
             var arguments = expression.Arguments;
 
-            if (method.Is(EnumerableMethod.Where))
+            if (method.IsOneOf(EnumerableMethod.Where, MongoEnumerableMethod.WhereWithLimit))
             {
                 var sourceExpression = arguments[0];
                 var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
+                var itemSerializer = ArraySerializerHelper.GetItemSerializer(sourceTranslation.Serializer);
```
rstam, line 35: Minor refactoring: we only need to computer `itemSerializer` once (removed `predicateParameterSerializer`).

JamesKovacs (approved):
LGTM. Good to merge once you rebase on Boris' PR.

PR #779, opened 2022-04-25
CSHARP-4127: Language specific examples for AWS Lambda.

DmitryLukyanov:
How to configure AWS auth for atlas cluster:
1. Get your AWS_* (aws_access_key_id, aws_secret_access_key, aws_session_token) credentials in Single Sign-on page:
image

Pay attention that values should be regenerated from time to time.
2. Configure credentials folder here: `c:\Users\{user_name}\.aws\`
3. Get your arn via `get-caller-identity`:
   
       $ ./aws sts get-caller-identity
        {
            "UserId": "blablabla:[dmitry.lukyanov@mongodb.com](mailto:dmitry.lukyanov@mongodb.com)",
            "Account": "%ID_VALUE%",
            "Arn": "arn:aws:sts::%ID_VALUE%:assumed-role/%ROLE_NAME%/[dmitry.lukyanov@mongodb.com](mailto:dmitry.lukyanov@mongodb.com)"
         }
pay attention on %ROLE_NAME%
4. list all roles via:

         $ ./aws iam list-roles
         {
           "Roles": [
           {
                 "Path": "..",
                 "RoleName": "%ROLE_NAME%",
                 "Arn": "arn:aws...:
            ...

in the provided roles, search for a record with a RoleName equal to %ROLE_NAME% and record his `arn`.
5. In your atlas cluster, create a new user with AWS authentication and set AWS IAM Role ARN from #4.
6. Then configure a mongoClient in the same way as it's done in this PR with `MONGODB-AWS` auth credentials

### DIFF
--- a/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
+++ b/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
@@ -15,6 +15,7 @@
 
 #if NETCOREAPP3_1_OR_GREATER
 using Amazon.Lambda.Core;
+using MongoDB.Bson;
 using MongoDB.Driver;
 using System;
 
@@ -41,8 +42,10 @@ namespace LambdaTest
 
         public string HandleRequest(ILambdaContext context)
         {
-            var databases = MongoClient.ListDatabaseNames();
-            return string.Join(",", databases);
+            var database = MongoClient.GetDatabase("db");
+            var collection = database.GetCollection<BsonDocument>("coll");
+            var result = collection.Find(FilterDefinition<BsonDocument>.Empty).First();
+            return result.ToString();
         }
         // End AWS Lambda Example 1
     }
@@ -78,8 +81,10 @@ namespace LambdaTest
 
         public string HandleRequest(ILambdaContext context)
         {
-            var databases = MongoClient.ListDatabaseNames();
-            return string.Join(",", databases);
+            var database = MongoClient.GetDatabase("db");
+            var collection = database.GetCollection<BsonDocument>("coll");
+            var result = collection.Find(FilterDefinition<BsonDocument>.Empty).First();
+            return result.ToString();
         }
     }
 }
--- a/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
+++ b/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
@@ -27,15 +27,15 @@ namespace LambdaTest
     public class ShareMongoClientLambdaHandler
     {
         // Start AWS Lambda Example 1
-        private MongoClient MongoClient { get; set; }
-        private MongoClient CreateMongoClient()
+        private static MongoClient MongoClient { get; set; }
+        private static MongoClient CreateMongoClient()
         {
             var mongoClientSettings = MongoClientSettings.FromConnectionString($"<MONGODB_URI>");
             mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1, strict: true);
             return new MongoClient(mongoClientSettings);
         }
 
-        public ShareMongoClientLambdaHandler()
+        static ShareMongoClientLambdaHandler()
         {
             MongoClient = CreateMongoClient();
         }
@@ -52,8 +52,8 @@ namespace LambdaTest
 
     public class ConnectUsingAwsIamAuthenticatorLambdaHandler
     {
-        private MongoClient MongoClient { get; set; }
-        private MongoClient CreateMongoClient()
+        private static MongoClient MongoClient { get; set; }
+        private static MongoClient CreateMongoClient()
         {
             // Start AWS Lambda Example 2
             string username = Environment.GetEnvironmentVariable("AWS_ACCESS_KEY_ID");
@@ -74,7 +74,7 @@ namespace LambdaTest
             // End AWS Lambda Example 2
         }
 
-        public ConnectUsingAwsIamAuthenticatorLambdaHandler()
+        static ConnectUsingAwsIamAuthenticatorLambdaHandler()
         {
             MongoClient = CreateMongoClient();
         }
--- a/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
+++ b/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
@@ -27,8 +27,8 @@ namespace LambdaTest
     public class ShareMongoClientLambdaHandler
     {
         // Start AWS Lambda Example 1
-        private static MongoClient MongoClient { get; set; }
-        private static MongoClient CreateMongoClient()
+        private MongoClient MongoClient { get; set; }
+        private MongoClient CreateMongoClient()
         {
             var mongoClientSettings = MongoClientSettings.FromConnectionString($"<MONGODB_URI>");
             mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1, strict: true);
@@ -52,8 +52,8 @@ namespace LambdaTest
 
     public class ConnectUsingAwsIamAuthenticatorLambdaHandler
     {
-        private static MongoClient MongoClient { get; set; }
-        private static MongoClient CreateMongoClient()
+        private MongoClient MongoClient { get; set; }
+        private MongoClient CreateMongoClient()
         {
             // Start AWS Lambda Example 2
             string username = Environment.GetEnvironmentVariable("AWS_ACCESS_KEY_ID");
--- a/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
+++ b/tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
@@ -1,0 +1,86 @@
+﻿/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


JamesKovacs (changes requested):
See comment on DRIVERS-2018. We need to expand our example slightly to include some minimal CRUD operations otherwise we aren't guaranteed to be successfully talking to a server. We might have just established connection pool, but not successfully connected to a cluster from AWS Lambda.

https://jira.mongodb.org/browse/DRIVERS-2018?focusedCommentId=4504303&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4504303

DmitryLukyanov:
it was decided to use `find` instead listdatabases.

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public class ConnectUsingAwsIamAuthenticatorLambdaHandler
+    {
+        private static MongoClient MongoClient { get; set; }
+        private static MongoClient CreateMongoClient()
+        {
+            // Start AWS Lambda Example 2
+            string username = Environment.GetEnvironmentVariable("AWS_ACCESS_KEY_ID");
+            string password = Environment.GetEnvironmentVariable("AWS_SECRET_ACCESS_KEY");
```
JamesKovacs, line 60: Password should be stored in a `SecureString`.

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
```diff
@@ -0,0 +1,91 @@
...
+            string awsSessionToken = Environment.GetEnvironmentVariable("AWS_SESSION_TOKEN");
+
+            var awsCredentials =
+                new MongoCredential("MONGODB-AWS", new MongoExternalIdentity(username), new PasswordEvidence(password))
```
JamesKovacs, line 64: `PasswordEvidence` has a ctor that accepts a `SecureString`.

JamesKovacs (changes requested):
Minor change requested.

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
JamesKovacs: Microsoft is no longer recommending the use of `SecureString`.
https://docs.microsoft.com/en-us/dotnet/api/system.security.securestring?view=net-6.0
And this rule:
https://github.com/dotnet/platform-compat/blob/master/docs/DE0001.md

Let's skip the use of `SecureString` that I initially suggested and keep the example easier to follow as you have it.

JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
```diff
@@ -0,0 +1,91 @@
...
+        private static MongoClient CreateMongoClient()
+        {
+            var mongoClientSettings = MongoClientSettings.FromConnectionString($"<MONGODB_URI>");
+            mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1, strict: true);
+            return new MongoClient(mongoClientSettings);
+        }
+
+        public ShareMongoClientLambdaHandler()
+        {
+            MongoClient = CreateMongoClient();
```
BorisDog, line 40: Is it guaranteed that the ctor will be called once/not concurrently?
Or should it be `static ShareMongoClientLambdaHandler` ?

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
```diff
@@ -0,0 +1,91 @@
...
+        // End AWS Lambda Example 1
+    }
+
+    public class ConnectUsingAwsIamAuthenticatorLambdaHandler
+    {
+        private static MongoClient MongoClient { get; set; }
+        private static MongoClient CreateMongoClient()
+        {
+            // Start AWS Lambda Example 2
+            string username = Environment.GetEnvironmentVariable("AWS_ACCESS_KEY_ID");
```
BorisDog, line 59: Are environment variables part of the Aws Lambda auth process?
If not, maybe for simplicity just use dummy values: `var username = "<username>"`

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: Recommended way I found was about regular ctor, see for example here https://blog.steadycoding.com/using-singletons-in-net-core-in-aws-lambda/. The note what I didn't mention before, that there are 2 modes for AWS Lambda called "Cold" and "Warm" start ("Hot is a bit different): https://medium.com/@danielmanchev/cold-warm-and-hot-start-in-aws-lambda-bc8d64f28575. So the actually reused instance will be only after `warm` start. Each `Cold` start will create new instance of lambda handler

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: it's not specified in our envs, but I saw it in other examples exactly in this form, so I would leave it as here

BorisDog: In this case if it's not required, wondering if `Environment.GetEnvironmentVariable` adds any additional value for the sample. Maybe we'd better keep it as simple as possible...

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
BorisDog: From the first glance I fail to see the reason for initializing a static property in regular ctor. Do we know why is this the recommended way, and is it AWS recommendation as well?
In [AWS .Net samples](https://github.com/awsdocs/aws-lambda-developer-guide/blob/main/sample-apps/ec2-spot/src/ec2spot/Function.cs) static ctors are used.
Probably we'd need to look deeper into this.

DmitryLukyanov: As far as I understand, both ways can be used. This is what created by VS "Simple S3 function" project:
image

BorisDog: Yes, note that `S3Client` in this case is not static. The initialization of a static field in non-static constructor is what puzzles me, but this does not mean it's not working though. I think we have to be able to explain such unusual combination, or follow the official samples.

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
BorisDog: Looks like it's only one of the options, used in case of "build agent process", not sure what it is :)
Also I don't think we should prioritize simplifying the usage of this code in our driver.

Not a big issue, I just think it introduces some "noise" to the sample, and not sure we got any use for it.

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
```diff
@@ -0,0 +1,91 @@
...
+            mongoClientSettings.ServerApi = new ServerApi(ServerApiVersion.V1, strict: true);
+            return new MongoClient(mongoClientSettings);
+        }
+
+        public ShareMongoClientLambdaHandler()
+        {
+            MongoClient = CreateMongoClient();
+        }
+
+        public string HandleRequest(ILambdaContext context)
```
BorisDog, line 43: In all samples a came across, I see additional `inputType` parameter in Handler function:
https://docs.aws.amazon.com/lambda/latest/dg/csharp-handler.html

Can it be omitted, and do you happen to have some official sample without it?

DmitryLukyanov: it can be omitted

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: it's true, I forgot that suggested MongoClient is static. I think it can work in any way, but I will look at this one more time :(

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: >to have some official sample without it?

I just saw that it works without this argument

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: I think it's fine to take what is suggested in the VS templates (ie official place). So I removed static modifier

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: At the very least these variables present in the examples that we needed to mimic: https://www.mongodb.com/docs/atlas/manage-connections-aws-lambda/. So I think we even have no way to not set it

tests/MongoDB.Driver.Examples/AwsLambda/AwsLambdaExamples.cs
DmitryLukyanov: I asked this in this issue just in case https://github.com/awsdocs/aws-lambda-developer-guide/issues/360

BorisDog (approved):
LGTM

PR #781, opened 2022-04-26
CSHARP-4138: Added new allowDiskUse tests.

JamesKovacs:
Sync'd new `aggregate-allowdiskuse.[json|yml]`. No production code changes necessary as we use `bool? AllowDiskUse` throughout the driver. If `AllowDiskUse` isn't set explicitly, we use the server's default, which is false on pre-6.0 and true on 6.0 or later.


DmitryLukyanov (approved):
LGTM

PR #783, opened 2022-04-27
CSHARP-4124: Updated docs.mongodb.com to www.mongodb.com/docs and similar redirect fixes.

JamesKovacs:
Fixing URLs in our `gh-pages` branch. You can browse the live deployment on my fork here:

https://jameskovacs.github.io/mongo-csharp-driver/


1.11/getting_started/index.html
```diff
@@ -348,7 +348,7 @@ <h2 id="introduction">Introduction</h2>
 <h2 id="downloading-the-c-driver">Downloading the C# Driver</h2>
```
DmitryLukyanov, line 348: broken link: 
image

DmitryLukyanov (approved):
I briefly looked at documentation and found only one broken link, see whether you want to fix it now or not.
Maybe it makes sense to search for this pattern "\index.md" (with backslash) in other places to ensure that it doesn't present anywhere else

Other then this LGTM

1.11/getting_started/index.html
JamesKovacs: That's a pre-existing glitch as it is present in the live site that no one has ever noticed/reported. I might as well fix it along with everything else. Thanks for catching that. And I'll search for `\index.md` to see if there are any other similar errors.

JamesKovacs: I took a closer look. The `%5c` encoded `\` is being rendered by Hugo. Given that it touches all the pages - both current and historical - as well as the Hugo templates, I've created [CSHARP-4158](https://jira.mongodb.org/browse/CSHARP-4158) to address the issue.

PR #784, opened 2022-04-28
CSHARP-4155: Test against MongoDB 6.0

JamesKovacs:


evergreen/evergreen.yml
```diff
@@ -1330,6 +1330,10 @@ axes:
         display_name: "latest"
         variables:
            VERSION: "latest"
+      - id: "6.0"
```
DmitryLukyanov, line 1333: we have a number of envs where we run only `5.0` servers like `csfle-with-mocked-kms-tests-windows` or where servers are defined explicitly like in `secure-tests-macOS` or compressions, we may update them too

DmitryLukyanov (approved):
Question and then LGTM

evergreen/evergreen.yml
JamesKovacs: Good catch. I've updated the test matrix.

evergreen/evergreen.yml
```diff
@@ -1466,7 +1470,7 @@ buildvariants:
      - name: test-netstandard21
 
 - matrix_name: "tests-zlib-compression"
-  matrix_spec: { compressor : "zlib", auth: "noauth", ssl: "nossl", version: ["3.6", "4.0", "4.2", "4.4", "5.0", "latest"], topology: "standalone", os: "windows-64" }
+  matrix_spec: { compressor : "zlib", auth: "noauth", ssl: "nossl", version: "*", topology: "standalone", os: "windows-64" }
```
DmitryLukyanov, line 1473: I've scheduled all variants to check that it works (but it should as I see)

PR #785, opened 2022-04-28
CSHARP-4156: Test against MongoDB quarterly rapid releases

JamesKovacs:
Similar to CSHARP-4155 except adding the quarterly rapid releases to our test matrix.

### DIFF
--- a/evergreen/evergreen.yml
+++ b/evergreen/evergreen.yml
@@ -1330,6 +1330,10 @@ axes:
         display_name: "latest"
         variables:
            VERSION: "latest"
+      - id: "rapid"
+        display_name: "rapid"
+        variables:
+           VERSION: "rapid"
       - id: "6.0"
         display_name: "6.0"
         variables:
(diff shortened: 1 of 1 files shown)


evergreen/evergreen.yml
```diff
@@ -1330,6 +1330,14 @@ axes:
         display_name: "latest"
         variables:
            VERSION: "latest"
+      - id: "rapid"
```
DmitryLukyanov, line 1333: same as in 4155 :) :
>we have a number of envs where we run only 5.0 servers like csfle-with-mocked-kms-tests-windows or where servers are defined explicitly like in secure-tests-macOS or compressions, we may update them too

DmitryLukyanov (approved):
Question and then LGTM

evergreen/evergreen.yml
JamesKovacs: Rebased on latest CSHARP-4155 and pushed updates.

DmitryLukyanov (approved):
LGTM

PR #786, opened 2022-05-01
CSHARP-3915: Implement Densify builder and LINQ methods.

rstam:


src/MongoDB.Driver/AggregateFluent.cs
```diff
@@ -106,6 +106,22 @@ public override IAggregateFluent<AggregateCountResult> Count()
...
+            DensifyRange range,
+            IEnumerable<FieldDefinition<TResult>> partitionByFields = null)
+        {
+            return WithPipeline(_pipeline.Densify(field, range, partitionByFields));
+        }
+
+        public override IAggregateFluent<TResult> Densify(
+            FieldDefinition<TResult> field,
+            DensifyRange range,
+            params FieldDefinition<TResult>[] partitionByFields)
```
rstam, line 120: AFAIK it's good practice to provide general purpose IEnumerable version and also the params version.

src/MongoDB.Driver/DensifyRange.cs
```diff
@@ -0,0 +1,489 @@
...
+using System;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Represents a range for the $densify stage.
+    /// </summary>
+    public abstract class DensifyRange
```
rstam, line 25: All the classes in this file need to be carefully reviewed for naming and usage. They are public API.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/LambdaExpressionExtensions.cs
```diff
@@ -24,7 +26,16 @@ internal static class LambdaExpressionExtensions
...
+            var parameterExpression = fieldSelectorLambda.Parameters.Single();
+            if (parameterSerializer.ValueType != parameterExpression.Type)
+            {
+                throw new ArgumentException($"ValueType '{parameterSerializer.ValueType.FullName}' of parameterSerializer does not match parameter type '{parameterExpression.Type.FullName}'.", nameof(parameterSerializer));
+            }
+            var parameterSymbol = context.CreateSymbolWithVarName(parameterExpression, varName: "ROOT", parameterSerializer, isCurrent: true);
+            var lambdaContext = context.WithSymbol(parameterSymbol);
+            var lambdaBody = ConvertHelper.RemoveConvertToBaseType(fieldSelectorLambda.Body);
+            var fieldSelectorTranslation = ExpressionToAggregationExpressionTranslator.Translate(lambdaContext, lambdaBody);
+
```
rstam, line 38: The above changes are to support fields defined using `Expression<Func<TDocument, object>>` instead of `Expression<Func<TDocument, TField>>`.

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -464,6 +464,58 @@ public static Task<int> CountAsync<TSource>(this IMongoQueryable<TSource> source
...
+        /// <param name="range">The range.</param>
+        /// <param name="partitionByFields">The partition by fields.</param>
+        /// <returns>The densified sequence of values.</returns>
+        public static IMongoQueryable<TSource> Densify<TSource>(
+            this IMongoQueryable<TSource> source,
+            Expression<Func<TSource, object>> field,
+            DensifyRange range,
+            IEnumerable<Expression<Func<TSource, object>>> partitionByFields = null)
+        {
+            return Densify(source, field, range, partitionByFields?.ToArray());
```
rstam, line 482: Normally the params version would delegate to the IEnumerable version, but in this case it was easier to go in the other direction.

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -464,6 +464,58 @@ public static Task<int> CountAsync<TSource>(this IMongoQueryable<TSource> source
...
+        {
+            Expression quotedPartitionByFields;
+            if (partitionByFields == null || partitionByFields.Length == 0)
+            {
+                quotedPartitionByFields = Expression.Constant(null, typeof(Expression<Func<TSource, object>>[]));
+            }
+            else
+            {
+                quotedPartitionByFields = Expression.NewArrayInit(typeof(Expression<Func<TSource, object>>), partitionByFields.Select(pf => Expression.Quote(pf)));
+            }
```
rstam, line 508: How to quote an array of expressions looks easy now that I've figured it out, but figuring it out was quite hard. There are NO standard LINQ methods that have a list of Expressions that need to be quoted so there was no existing code that I could find to model this after.

The key here  is `NewArrayInit`, and the special case for `null`.

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -2821,5 +2873,15 @@ public static IMongoQueryable<TSource> Where<TSource>(this IMongoQueryable<TSour
...
+
+        private static MethodInfo GetMethodInfo<T1, T2, T3, T4>(Func<T1, T2, T3, T4> f, T1 unused1, T2 unused2, T3 unused3)
+        {
+            return f.GetMethodInfo();
+        }
+
+        private static MethodInfo GetMethodInfo<T1, T2, T3, T4, T5>(Func<T1, T2, T3, T4, T5> f, T1 unused1, T2 unused2, T3 unused3, T4 unused4)
+        {
+            return f.GetMethodInfo();
+        }
```
rstam, line 2885: These are just like the existing methods above but for more parameters.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -381,6 +381,103 @@ public static class PipelineStageDefinitionBuilder
...
+                        {
+                            operatorName, new BsonDocument
+                            {
+                                { "field", field.Render(s, sr, linqProvider).FieldName },
+                                { "partitionByFields", () => new BsonArray(renderedPartitionByFields), partitionByFields != null && renderedPartitionByFields.Count > 0 },
+                                { "range", range.Render() }
+                            }
+                        }
+                    };
+                    return new RenderedPipelineStageDefinition<TInput>(operatorName, document, s);
```
rstam, line 418: The above is where a builder based $densify stage bets build.

The LINQ version of Densify is handled in the LINQ translator.

tests/MongoDB.Driver.Tests/Jira/CSharp3915Tests.cs
```diff
@@ -0,0 +1,328 @@
...
+        {
+            var collection = CreateCoffeeCollection();
+            var subject = collection.Aggregate();
+
+            IAggregateFluent<CoffeeInfo> aggregate;
+            if (usingExpressions)
+            {
+                aggregate = subject.Densify(
+                    field: x => x.Altitude,
+                    range: DensifyRange.Numeric(DensifyBounds.Full, step: 200),
```
rstam, line 124: Note the use of `DensifyBounds.Full` as the public API for the "full" bounds.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -117,6 +117,48 @@ public static class IAggregateFluentExtensions
...
+        /// <param name="range">The range.</param>
+        /// <param name="partitionByFields">The partition by fields.</param>
+        /// <returns>
+        /// The fluent aggregate interface.
+        /// </returns>
+        public static IAggregateFluent<TResult> Densify<TResult>(
+            this IAggregateFluent<TResult> aggregate,
+            Expression<Func<TResult, object>> field,
+            DensifyRange range,
+            IEnumerable<Expression<Func<TResult, object>>> partitionByFields = null)
```
rstam, line 135: This has to be `Expression<Func<TResult, object>>` because each field is of a different type.

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -299,6 +300,11 @@ public class Feature
         [Obsolete("This property will be removed in a later release.")]
         public static Feature CurrentOpCommand => __currentOpCommand;
 
+        /// <summary>
+        /// Gets the $densify stage feature.
+        /// </summary>
+        public static Feature DensifyStage => __densifyStage;
```
rstam, line 306: We've also often used a more verbose convention that would have resulted in this feature being named `AggregateDensifyStage`. Not sure whether it's worth using the longer name (all stages are aggregate sub-features).

JamesKovacs: I don't think there is any ambiguity around `DensifyStage`. There isn't a need to prefix it with "aggregate".

src/MongoDB.Driver/DensifyRange.cs
JamesKovacs: I overall like the naming. It is clear and concise.

My only hesitation is that we are encountering a proliferation of of time-related enumerations:
- `DensifyDateTimeUnit`
- `TimeSeriesGranularity`
- `DateTimeUnit` <-- Smart enum, not simple enum
- `TimeSpanUnits`

Unfortunately while there is overlap in the enumeration value names, there isn't a one-to-one correspondence between them. Each server function has a different set of valid enumeration values. We will have to maintain separate though similar enumerations so that we can provide accurate Intellisense and avoid  allowing users to call server functions with invalid enumeration values.

The only thing I'm left scratching my head over is `DensifyDateTimeUnit` versus `DateTimeUnit`. Those two enumerations are very different but the naming suggests otherwise. Should we rename `DensifyDateTimeUnit` to `DensifyTimeUnit` to better match the `$densify` server documentation? Though I do like `DensifyDateTimeUnit` since the enumeration contains values related to time (millisecond, second, minute, hour) as well as dates (week, month, etc.). Throwing this out there for discussion.

JamesKovacs (changes requested):
Minor changes and discussion requested.

src/MongoDB.Driver/DensifyRange.cs
rstam: I also was concerned about the potential duplication between enum types, but as you pointed out they don't all align perfectly so we seem to have no choice but to have different unique enums.

src/MongoDB.Driver/DensifyRange.cs
```diff
@@ -0,0 +1,489 @@
...
+        internal static void EnsureIsValidNumericType()
+        {
+            switch (Type.GetTypeCode(typeof(TNumber)))
+            {
+                case TypeCode.Decimal:
+                case TypeCode.Double:
+                case TypeCode.Int16:
+                case TypeCode.Int32:
+                case TypeCode.Single:
+                    break;
```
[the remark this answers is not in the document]
rstam: Int64 definitely feels like an omission.

The unsigned types don't exist server-side. Do we want to support then anyway?

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Stages/AstDensifyStage.cs
```diff
@@ -0,0 +1,61 @@
...
+        private readonly DensifyRange _range;
+
+        public AstDensifyStage(
+            string fieldPath,
+            DensifyRange range,
+            IEnumerable<string> partitionByFieldPaths = null)
+        {
+            _fieldPath = Ensure.IsNotNull(fieldPath, nameof(fieldPath));
+            _range = Ensure.IsNotNull(range, nameof(range));
+            _partitionByFieldPaths = partitionByFieldPaths == null ? null : partitionByFieldPaths.AsReadOnlyList();
```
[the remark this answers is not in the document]
rstam: True. Will change in next commit.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -381,6 +381,103 @@ public static class PipelineStageDefinitionBuilder
...
+            DensifyRange range, // can be null
+            IEnumerable<FieldDefinition<TInput>> partitionByFields = null)
+        {
+            Ensure.IsNotNull(field, nameof(field));
+            Ensure.IsNotNull(range, nameof(range));
+
+            const string operatorName = "$densify";
+            var stage = new DelegatedPipelineStageDefinition<TInput, TInput>(
+                operatorName,
+                //(s, sr, linqProvider) => new RenderedPipelineStageDefinition<TInput>(operatorName, new BsonDocument(operatorName, filter.Render(s, sr, linqProvider)), s));
```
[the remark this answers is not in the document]
rstam: Oops. It is. Will remove in next commit.

src/MongoDB.Driver/Linq/MongoQueryable.cs
rstam: Looking at this code today I'm wondering if `NewArrayInit` is really the right thing to do here.

I can't think of anything else that's better, but this is uncharted territory (because we're quoting multiple expressions).

JamesKovacs: Seems reasonable to me though you've thought about it more deeply than I have. I can't think of a reason why it would be problematic. Let's run with it until we find a problem. Hopefully we never do.

JamesKovacs (approved):
LGTM

PR #790, opened 2022-05-05
CSHARP-4149: Add FLE 2 behavior for CreateCollection() and Collection.Drop().

DmitryLukyanov:


src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
```diff
@@ -79,6 +85,12 @@ public class AutoEncryptionOptions
         /// </value>
         public bool BypassAutoEncryption => _bypassAutoEncryption;
 
+        /// <summary>
+        /// Gets the encrypted fields map.
+        /// Supplying an encryptedFieldsMap provides more security than relying on an encryptedFields obtained from the server. It protects against a malicious server advertising a false encryptedFields.
```
DmitryLukyanov, line 90: this xml comment is required in the spec

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -678,14 +707,69 @@ private CreateCollectionOperation CreateCreateCollectionOperation<TDocument>(str
...
+                collectionNamespace,
+                effectiveEncryptedFields,
+                messageEncoderSettings,
+                (dco) =>
+                {
+                    dco.WriteConcern = _settings.WriteConcern;
+                });
+        }
+
+        private async Task<IWriteOperation<BsonDocument>> CreateDropCollectionOperationAsync(string name, DropCollectionOptions options, IClientSessionHandle session, CancellationToken cancellationToken)
```
DmitryLukyanov, line 742: it's unfortunate that creating collection step require calling a server, so now we need to have 2 versions of this method: sync and async

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            var testException = new Exception("test");
+
+            var healthyOperation1 = CreateHealthyOperation(new BsonDocument("operation", 1));
+            var faultyOperation2 = CreateFaultyOperation(testException);
+            var healthyOperation3= CreateHealthyOperation(new BsonDocument("operation", 3));
+
+            var subject = new CompositeWriteOperation<BsonDocument>((healthyOperation1.Object, IsMainOperation: false), (faultyOperation2.Object, IsMainOperation: false), (healthyOperation3.Object, IsMainOperation: true));
+
+            var resultedException = async
+                ? Record.Exception(() => subject.ExecuteAsync(Mock.Of<IWriteBinding>(), CancellationToken.None).GetAwaiter().GetResult())
```
BorisDog, line 72: minor: I think it's more "correct" and in some sense more safer, to execute tasks in tests via regular await pattern.
```
var resultedException = async
                ? await Record.ExceptionAsync(() => subject.ExecuteAsync(Mock.Of<IWriteBinding>(), CancellationToken.None))
                : Record.Exception(() => subject.Execute(Mock.Of<IWriteBinding>(), CancellationToken.None));
```
That being said, functionally this does not matter a lot in this specific case, but still we might want to prefer this approach by default.

tests/MongoDB.Driver.Tests/AutoEncryptionOptionsTests.cs
```diff
@@ -99,8 +101,53 @@ public void Ctor_should_handle_kmsProviderOptions_type_correctly(Type optionType
...
+        [Theory]
+        [InlineData(null, null, null, false)]
+        [InlineData("db.test", null, null, false)]
+        [InlineData(null, "db.test", null, false)]
+        [InlineData("db.test", "db.test1", null, false)]
+        [InlineData("db.test", "db.test", "db.test", true)]
+        [InlineData("db.test", "db.test;db.test1", "db.test", true)]
+        [InlineData("db.test;db.test1", "db.test;db.test1", "db.test, db.test1", true)]
+        [InlineData("db.test1;db.test", "db.test;db.test1", "db.test1, db.test", true)]
+        public void constructor_should_handle_schemaMap_and_encryptedFieldsMap_type_correctly(string schemaMapKey, string encryptedFieldsMapKey, string errorMessage, bool shouldFail)
```
BorisDog, line 113: minor: `shouldFail` parameter can be skipped, having just `errorMessage`

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -229,6 +229,15 @@ private AutoEncryptionOptions ConfigureAutoEncryptionOptions(BsonDocument autoEn
                     case "keyVaultNamespace":
                         autoEncryptionOptions = autoEncryptionOptions.With(keyVaultNamespace: CollectionNamespace.FromFullName(option.Value.AsString));
                         break;
+                    case "encryptedFieldsMap":
+                        var encryptedFieldsMap = new Dictionary<string, BsonDocument>();
+                        var encryptedFieldsMapDocument = option.Value.AsBsonDocument;
+                        foreach (var encryptedFieldsMapElement in encryptedFieldsMapDocument.Elements)
```
BorisDog, line 235: minor, for your consideration:
```
var encryptedFieldsMapDocument = option.Value.AsBsonDocument;
var encryptedFieldsMap = encryptedFieldsMapDocument.Elements.ToDictionary(e => e.Name, e => e.Value.AsBsonDocument);
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -229,6 +229,15 @@ private AutoEncryptionOptions ConfigureAutoEncryptionOptions(BsonDocument autoEn
...
+                        {
+                            encryptedFieldsMap.Add(encryptedFieldsMapElement.Name, encryptedFieldsMapElement.Value.AsBsonDocument);
+                        }
+                        autoEncryptionOptions = autoEncryptionOptions.With(encryptedFieldsMap:encryptedFieldsMap);
```
BorisDog, line 239: minor: space around `:`

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
```diff
@@ -51,17 +95,12 @@ public class CreateCollectionOperation : IWriteOperation<BsonDocument>
         private WriteConcern _writeConcern;
 
         // constructors
-        /// <summary>
-        /// Initializes a new instance of the <see cref="CreateCollectionOperation"/> class.
-        /// </summary>
-        /// <param name="collectionNamespace">The collection namespace.</param>
-        /// <param name="messageEncoderSettings">The message encoder settings.</param>
-        public CreateCollectionOperation(
+        internal CreateCollectionOperation(
```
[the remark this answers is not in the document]
DmitryLukyanov: I think we agreed that we can change majority of classed in the Core that are not supposed to be used by users - this class is definitely not supposed. But I think I will revert this change. The idea behind here that initially I had implementation without CompositeOperation and it required few constructors

src/MongoDB.Driver.Core/Core/Operations/DropCollectionOperation.cs
```diff
@@ -29,8 +31,48 @@ namespace MongoDB.Driver.Core.Operations
...
+            }
+
+            DropCollectionOperation CreateInnerDropOperation(string collectionName)
+                => new DropCollectionOperation(new CollectionNamespace(collectionNamespace.DatabaseNamespace.DatabaseName, collectionName.ToString()), messageEncoderSettings);
+        }
+        #endregion
+
         // fields
         private readonly CollectionNamespace _collectionNamespace;
+        private BsonDocument _encryptedFields;
```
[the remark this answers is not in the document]
DmitryLukyanov: for operation itself - nope, but we pass it to the factory method. It would be good to have ability to see what values was provided to the class via a property

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
DmitryLukyanov: We don't use `value` for a new classes only for old that already have it. As I know we planned to remove this tag in 3.0.
`remarks` sounds good to me

src/MongoDB.Driver/MongoDatabaseBase.cs
```diff
@@ -142,6 +141,12 @@ public virtual void DropCollection(IClientSessionHandle session, string name, Ca
...
         }
 
+        /// <inheritdoc />
+        public virtual void DropCollection(string name, DropCollectionOptions options, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+
+        /// <inheritdoc />
+        public virtual void DropCollection(IClientSessionHandle session, string name, DropCollectionOptions options, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+
         /// <inheritdoc />
         public abstract Task DropCollectionAsync(string name, CancellationToken cancellationToken = default(CancellationToken));
```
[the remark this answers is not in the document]
DmitryLukyanov: the idea is to avoid BCs, not sure how healthy is this idea nowadays :) In 2 words, in past we expected that someone can inherit his classes from MongoCollectionBase class, so first methods in this class were abstract, but then to allow users compile their code after upgrading driver reference, we started to add new methods as virtual, they could still not use new methods, but at least they would not need to implement their bodies immediately

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -644,26 +664,35 @@ private CreateCollectionOperation CreateCreateCollectionOperation<TDocument>(str
...
+            var collectionNamespace = new CollectionNamespace(_databaseNamespace, name);
+
+            _ = EncryptedCollectionHelper.TryGetEffectiveEncryptedFields(collectionNamespace, options.EncryptedFields, _client.Settings?.AutoEncryptionOptions?.EncryptedFieldsMap, out var effectiveEncryptedFields);
+            var messageEncoderSettings = GetMessageEncoderSettings(withEncryption: effectiveEncryptedFields != null);
+
+            return CreateCollectionOperation.CreateEncryptedCreateCollectionOperationIfConfigured(
+                collectionNamespace,
+                effectiveEncryptedFields,
+                messageEncoderSettings,
+                createCollectionOperationConfigurator: ccp =>
```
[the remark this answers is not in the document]
DmitryLukyanov: typo, will fix

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -678,14 +707,69 @@ private CreateCollectionOperation CreateCreateCollectionOperation<TDocument>(str
...
+            var collectionNamespace = new CollectionNamespace(_databaseNamespace, name);
+
+            options = options ?? new DropCollectionOptions();
+
+            var autoEncryptionOptions = _client.Settings?.AutoEncryptionOptions;
+            if (!EncryptedCollectionHelper.TryGetEffectiveEncryptedFields(collectionNamespace, options.EncryptedFields, autoEncryptionOptions?.EncryptedFieldsMap, out var effectiveEncryptedFields))
+            {
+                if (autoEncryptionOptions != null)
+                {
+                    var listCollectionOptions = new ListCollectionsOptions() { Filter = $"{{ name : '{collectionNamespace.CollectionName}' }}" };
```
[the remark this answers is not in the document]
DmitryLukyanov: yes and if the condition in line 749 false

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
DmitryLukyanov: it will require making test async. It's not big deal I think, but since almost all our code is written differently, not sure whether I want to start using it now, but I don't have strong opinion here, just consistency to the most of our code base

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
```diff
@@ -0,0 +1,75 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using MongoDB.Bson;
+
+namespace MongoDB.Driver.Encryption
+{
+    internal static class EncryptedCollectionHelper
+    {
+        private static BsonDocument __additionalCreateIndexDocument = new BsonDocument("__safeContent__", 1);
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Operations/CompositeWriteOperation.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Bindings;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Operations
+{
+    internal class CompositeWriteOperation<TResult> : IWriteOperation<TResult>
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Operations/CompositeWriteOperation.cs
```diff
@@ -0,0 +1,64 @@
...
+                var itemResult = operationInfo.Operation.Execute(binding, cancellationToken);
+                if (operationInfo.IsMainOperation)
+                {
+                    result = itemResult;
+                }
+            }
+
+            return result;
+        }
+        public async Task<TResult> ExecuteAsync(IWriteBinding binding, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
```diff
@@ -30,12 +31,55 @@ namespace MongoDB.Driver.Core.Operations
     /// </summary>
     public class CreateCollectionOperation : IWriteOperation<BsonDocument>
     {
+        #region static
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CreateCollectionOperation"/> class.
+        /// </summary>
+        /// <param name="collectionNamespace">The collection namespace.</param>
+        /// <param name="encryptedFields">The encrypted feilds.</param>
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Operations/DropCollectionOperation.cs
```diff
@@ -29,8 +31,48 @@ namespace MongoDB.Driver.Core.Operations
...
+                    (CreateInnerDropOperation(EncryptedCollectionHelper.GetAdditionalCollectionName(encryptedFields, collectionNamespace, HelperCollectionForEncryption.Ecc)), IsMainOperation: false),
+                    (CreateInnerDropOperation(EncryptedCollectionHelper.GetAdditionalCollectionName(encryptedFields, collectionNamespace, HelperCollectionForEncryption.Ecos)), IsMainOperation: false));
+            }
+            else
+            {
+                return mainOperation;
+            }
+
+            DropCollectionOperation CreateInnerDropOperation(string collectionName)
+                => new DropCollectionOperation(new CollectionNamespace(collectionNamespace.DatabaseNamespace.DatabaseName, collectionName.ToString()), messageEncoderSettings);
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver/Encryption/AutoEncryptionOptions.cs
DmitryLukyanov: I see that we use `value` here, so will add this too, done

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -291,28 +292,48 @@ public override Task CreateCollectionAsync(IClientSessionHandle session, string
 
         public override void DropCollection(string name, CancellationToken cancellationToken)
         {
-            UsingImplicitSession(session => DropCollection(session, name, cancellationToken), cancellationToken);
+            UsingImplicitSession(session => DropCollection(session, name, options: null, cancellationToken), cancellationToken);
+        }
+
+        public override void DropCollection(string name, DropCollectionOptions options, CancellationToken cancellationToken = default)
+        {
+            UsingImplicitSession(session => DropCollection(session, name, options: options, cancellationToken), cancellationToken);
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/AutoEncryptionOptionsTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/AutoEncryptionOptionsTests.cs
```diff
@@ -99,8 +101,53 @@ public void Ctor_should_handle_kmsProviderOptions_type_correctly(Type optionType
...
+                exception.Should().BeNull();
+            }
+
+            Dictionary<string, BsonDocument> CreateMap(string key)
+            {
+                var dummyMapValue = new BsonDocument();
+                var map = new Dictionary<string, BsonDocument>();
+                if (key != null)
+                {
+                    foreach (var keyItem in key.Split(';'))
```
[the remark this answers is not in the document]
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/AutoEncryptionOptionsTests.cs
```diff
@@ -213,5 +357,15 @@ private object CreateInstance(Type type)
             };
             return kmsProviders;
         }
+
+        private Dictionary<string, TValue> GetDictionary<TValue>(params (string Key, TValue Document)[] map)
+        {
+            if (map.Length == 0)
+            {
+                return null;
```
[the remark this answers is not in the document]
DmitryLukyanov: It was something for a test part that no longer exists, removed

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public static void EnsureCollectionsValid(IReadOnlyDictionary<string, BsonDocument> schemaMap, IReadOnlyDictionary<string, BsonDocument> encryptedFieldsMap)
+        {
+            if (schemaMap == null || encryptedFieldsMap == null || schemaMap.Count == 0 || encryptedFieldsMap.Count == 0)
+            {
+                return;
+            }
+
+            var joined = schemaMap.Join(encryptedFieldsMap, o => o.Key, i => i.Key, (kv, i) => new { KeyValue = kv, Inner = i }).ToList();
+            if (joined.Count > 0)
```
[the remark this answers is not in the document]
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
BorisDog: So should we check for `(autoEncryptionOptions.EncryptedFieldsMap != null)` instead?

src/MongoDB.Driver/MongoDatabaseBase.cs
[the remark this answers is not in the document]
BorisDog: I see, it's a good point. BC here would also break Analyzer.

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
BorisDog: Yes it will, we got lots of tests returning `Task` now. And I think the new tests that use tasks should be written with the usual TPL approach.
Not a big deal in this case at all, for your consideration.

src/MongoDB.Driver.Core/Core/Operations/DropCollectionOperation.cs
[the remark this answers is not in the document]
BorisDog: Not sure that we should store non-used data.  Also it's a public class, and it would be harder to change our mind here.
Maybe `internal` if you find it useful in some cases? Relying on you here.

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
DmitryLukyanov: I misread spec, I saw it required exactly this condition, but I see that it doesn't: https://github.com/mongodb/specifications/pull/1183/files#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR679. So yeah, it should be `autoEncryptionOptions.EncryptedFieldsMap != null`. Will change it

src/MongoDB.Driver.Core/Core/Operations/DropCollectionOperation.cs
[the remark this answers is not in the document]
DmitryLukyanov: From consumer's for this class POV, this field is used. We don't send this to the server, but we provide it to a factory method that is almost like ctor in this meaning. I'm ok to make it internal

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
DmitryLukyanov: I think to say that we will use it for all new tests further, we need to agree that we're ok to change implementation (on tasks) for a lot of old tests. Not sure that we're ok with this at this point, so I would probably not apply this suggestion at least for now

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
DmitryLukyanov: done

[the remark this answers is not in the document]
BorisDog: minor: I'd combine the two conditions into single statement.

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
BorisDog: Not sure that it's crucial to change all the old tests. But lots of new new tests are definitely using the usual TPL oriented approach.

BorisDog (approved):
LGTM, two minor comments for you considerations.

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -644,26 +664,35 @@ private CreateCollectionOperation CreateCreateCollectionOperation<TDocument>(str
...
-                UsePowerOf2Sizes = options.UsePowerOf2Sizes,
-                ValidationAction = options.ValidationAction,
-                ValidationLevel = options.ValidationLevel,
-                Validator = validator,
-                WriteConcern = _settings.WriteConcern
-            };
-#pragma warning restore
+            var collectionNamespace = new CollectionNamespace(_databaseNamespace, name);
+
+            _ = EncryptedCollectionHelper.TryGetEffectiveEncryptedFields(collectionNamespace, options.EncryptedFields, _client.Settings?.AutoEncryptionOptions?.EncryptedFieldsMap, out var effectiveEncryptedFields);
```
rstam, line 669: How can you ignore the return value? What if `TryGetEffectiveEncryptedFields` returns `false`?

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
rstam: I think the only reason OLD tests don't return Task is that when they were written xUnit did not yet support that.

And there was probably some period of time between when xUnit started supporting async tests and when we became aware of it and started using that.

I think it's fine to use async tests going forward. No need to change old tests. Perhaps if we work on on old tests for some reason we can change them little by little.

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -291,28 +292,48 @@ public override Task CreateCollectionAsync(IClientSessionHandle session, string
 
         public override void DropCollection(string name, CancellationToken cancellationToken)
         {
-            UsingImplicitSession(session => DropCollection(session, name, cancellationToken), cancellationToken);
+            UsingImplicitSession(session => DropCollection(session, name, options: null, cancellationToken), cancellationToken);
```
rstam, line 295: I would avoid duplicating even small amounts of logic. Simply provide the default value for the missing parameter and delegate to the next method:

    DropCollection(name, options: null, cancellationToken);

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -291,28 +292,48 @@ public override Task CreateCollectionAsync(IClientSessionHandle session, string
...
             Ensure.IsNotNullOrEmpty(name, nameof(name));
-            var operation = CreateDropCollectionOperation(name);
+            var operation = CreateDropCollectionOperation(name, options, session, cancellationToken);
             ExecuteWriteOperation(session, operation, cancellationToken);
         }
 
         public override Task DropCollectionAsync(string name, CancellationToken cancellationToken)
         {
-            return UsingImplicitSessionAsync(session => DropCollectionAsync(session, name, cancellationToken), cancellationToken);
+            return UsingImplicitSessionAsync(session => DropCollectionAsync(session, name, options: null, cancellationToken), cancellationToken);
```
rstam, line 318: Same here, avoid duplicating even the smallest amount of logic:

    return DropCollectionAsync(name, options: null, cancellationToken);

src/MongoDB.Driver/MongoDatabaseImpl.cs
rstam: This "looks" like a bug. Consider a new helper method instead:

    var effectiveEncryptedFields = EncryptedCollectionHelper.GetEffectiveEncryptedFields(collectionNamespace, options.EncryptedFields, _client.Settings?.AutoEncryptionOptions?.EncryptedFieldsMap);

src/MongoDB.Driver/MongoDatabaseImpl.cs
```diff
@@ -824,7 +908,7 @@ private MessageEncoderSettings GetMessageEncoderSettings()
 #pragma warning disable 618
             if (BsonDefaults.GuidRepresentationMode == GuidRepresentationMode.V2)
             {
-                messageEncoderSettings.Add(MessageEncoderSettingsName.GuidRepresentation, _settings.GuidRepresentation);
+                messageEncoderSettings.Add(MessageEncoderSettingsName.GuidRepresentation, withEncryption ? GuidRepresentation.Unspecified : _settings.GuidRepresentation);
```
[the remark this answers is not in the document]
rstam: I don't think you need a `withEncryption` parameter at all, you can just do:

    messageEncoderSettings.Add(MessageEncoderSettingsName.GuidRepresentation, GuidRepresentation.Unspecified);

whether encryption is configured or not.

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+                if (encryptedFieldsMap != null)
+                {
+                    return encryptedFieldsMap.TryGetValue(collectionNamespace.ToString(), out effectiveEncryptedFields);
+                }
+
+                return false;
+            }
+
+            return true;
+        }
```
rstam, line 65: I found the logic of the above method somewhat confusing and it took me a long time to convince myself that it could be correct. I think the following refactoring expresses the intent more clearly:

    public static bool TryGetEffectiveEncryptedFields(CollectionNamespace collectionNamespace, BsonDocument encryptedFields, IReadOnlyDictionary<string, BsonDocument> encryptedFieldsMap, out BsonDocument effectiveEncryptedFields)
    {
        if (encryptedFields != null)
        {
            effectiveEncryptedFields = encryptedFields;
            return true;
        }

        if (encryptedFieldsMap != null)
        {
            return encryptedFieldsMap.TryGetValue(collectionNamespace.ToString(), out effectiveEncryptedFields);
        }

        effectiveEncryptedFields = null;
        return false;
    }

What do you think?

src/MongoDB.Driver/MongoDatabaseImpl.cs
rstam: The new helper method could be:

    public static BsonDocument GetEffectiveEncryptedFields(CollectionNamespace collectionNamespace, BsonDocument encryptedFields, IReadOnlyDictionary<string, BsonDocument> encryptedFieldsMap)
    {
        if (TryGetEffectiveEncryptedFields(collectionNamespace, encryptedFields, encryptedFieldsMap, out var effectiveEncryptedFields))
        {
            return effectiveEncryptedFields;
        }
        else
        {
            return null;
        }
    }

I realize there's a bit of redundancy in this proposed helper method but it's very clear and doesn't rely on any side effects of the `TryGetEffectiveEncryptedFields` method.

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
```diff
@@ -30,12 +31,48 @@ namespace MongoDB.Driver.Core.Operations
...
+
+            createCollectionOperationConfigurator?.Invoke(mainOperation);
+
+            if (encryptedFields != null)
+            {
+                return new CompositeWriteOperation<BsonDocument>(
+                    (CreateInnerCollectionOperation(EncryptedCollectionHelper.GetAdditionalCollectionName(encryptedFields, collectionNamespace, HelperCollectionForEncryption.Esc)), IsMainOperation: false),
+                    (CreateInnerCollectionOperation(EncryptedCollectionHelper.GetAdditionalCollectionName(encryptedFields, collectionNamespace, HelperCollectionForEncryption.Ecc)), IsMainOperation: false),
+                    (CreateInnerCollectionOperation(EncryptedCollectionHelper.GetAdditionalCollectionName(encryptedFields, collectionNamespace, HelperCollectionForEncryption.Ecos)), IsMainOperation: false),
+                    (mainOperation, true),
```
rstam, line 56: Suggestion:

    (mainOperation, IsMainOperation: true),

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
rstam: And it is correct to only do this when `GuidRepresentationMode` is `V2`.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
DmitryLukyanov: this line is already quite long with configureawait, not a big deal, but I would probably leave it as is

tests/MongoDB.Driver.Core.Tests/Core/Operations/CompositeWriteOperationTests.cs
DmitryLukyanov: then, let it be so :)

DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
DmitryLukyanov: done

src/MongoDB.Driver/MongoDatabaseImpl.cs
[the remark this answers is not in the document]
DmitryLukyanov: discussed offline

rstam (approved):
LGTM

PR #791, opened 2022-05-07
CSHARP-3947: Update serverless testing for load balancer fronting single proxy.

DmitryLukyanov:
NOTE: We will need updating EG env variables for release branches as well


src/MongoDB.Driver.Core/Core/Servers/LoadBalancedServer.cs
```diff
@@ -105,12 +105,12 @@ protected override void InitializeSubClass()
...
 
-            // propagate event to upper levels, this will be called only once
-            TriggerServerDescriptionChanged(this, eventArgs);
-
             // mark pool as ready, start the connection creation thread.
             // note that the pool can not be paused after it was marked as ready in LB mode.
             ConnectionPool.SetReady();
+
+            // propagate event to upper levels, this will be called only once
+            TriggerServerDescriptionChanged(this, eventArgs);
```
DmitryLukyanov, line 113: The case presented before this change:
1. We've created cluster that then has created a LB server.
2. One line above ConnectionPool.SetReady();  we've published a server description, so now we're able to select this server. But if it happens immediately after publishing a pool is still paused which then will trigger the above exception

JamesKovacs (approved):
LGTM

PR #789, opened 2022-05-04
CSHARP-4095: Use count command for estimatedDocumentCount.

JamesKovacs:
Mostly straightforward PR. Comments inline below.

PR #792, opened 2022-05-09
CSHARP-4094: Add 'comment' option to EstimatedDocumentCountOptions.

JamesKovacs:
CSHARP-4094 changes split from CSHARP-4095 (PR #789).


tests/MongoDB.Driver.Tests/EstimatedDocumentCountOptionsTests.cs
```diff
@@ -27,6 +27,7 @@ public void constructor_should_initialize_instance()
         {
             var result = new EstimatedDocumentCountOptions();
 
+            result.Comment.Should().BeNull();
```
BorisDog, line 30: We probably need to have similar to `AggregateOperationTests` comment tests in `EstimatedDocumentCountOperationTests` as well:
[Comment_get_and_set_should_work](https://github.com/mongodb/mongo-csharp-driver/blob/7308368ff2235d3166ff3b1748be9ec3e6e546b0/tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs#L190)
[CreateCommand_should_return_expected_result_when_Comment_is_set](https://github.com/mongodb/mongo-csharp-driver/blob/7308368ff2235d3166ff3b1748be9ec3e6e546b0/tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateOperationTests.cs#L405)

BorisDog (changes requested):
Looks good, just some comment for unit tests.

tests/MongoDB.Driver.Tests/EstimatedDocumentCountOptionsTests.cs
JamesKovacs: `EstimatedDocumentCountOperationsTests` already includes the test `Comment_get_and_set_should_work`. Added new test named `CreateCountOperation_should_return_expected_result_when_Comment_is_set`.

BorisDog: Oh sorry, missed the `Comment_get_and_set_should_work`.
`CreateCountOperation_should_return_expected_result_when_Comment_is_set`: the intention was to check that comment field is sent on the wire. But that is probably `CountOperation` tests responsibility, and I don't see the comment tests there. I've probably missed that in the previous PR.
So we could test the actual wire representation here (like `CreateCommand_should_return_expected_result_when_ReadConcern_is_set` does) or add `CountOperation` tests  in another ticket.

BorisDog (approved):
LGTM, please notice the additional test comment.

tests/MongoDB.Driver.Tests/EstimatedDocumentCountOptionsTests.cs
JamesKovacs: Added test to `CountOperationTests` to ensure wire representation of comment is correct. Testing it as part of `EstimatedDocumentCountOperationsTests` would have required digging through at least two private methods, which I felt would be fragile going forward.

PR #799, opened 2022-05-20
CSHARP-4065: Don't enable SDAM heartbeat for integration tests in CMAP spec.

DmitryLukyanov:
Waiting for EG before being ready for review


DmitryLukyanov:
not ready for review

DmitryLukyanov:
close this for now since current implementation gives some additional issues and it's unclear when we will have chance to look at it again

PR #803, opened 2022-05-26
CSHARP-4181: Reproduced using LINQ2. Works in LINQ3.

rstam:


rstam:
Probably not worth pushing to master unless we want evidence in master that it works in LINQ3.

JamesKovacs (approved):
LGTM

rstam:
Closed without merging to master because the repro showed that it already works in LINQ3.

PR #804, opened 2022-06-01
adds remarks when zero or null for IFindFluent.Limit 

leandromoh:


leandromoh:
FYI @DmitryLukyanov @rstam

src/MongoDB.Driver/IFindFluent.cs
```diff
@@ -112,6 +112,9 @@ public interface IFindFluent<TDocument, TProjection> : IAsyncCursorSource<TProje
         /// <summary>
         /// Limits the number of documents.
         /// </summary>
+        /// <remarks>
+        /// Limit equals to zero or null is equivalent to setting no limit.
```
JamesKovacs, line 116: This is a reasonable request to improve our documentation and is similar to the note in our [shell documentation on limit](https://www.mongodb.com/docs/manual/reference/method/cursor.limit/). The remark would read more naturally as "Limit of zero or null is..."

JamesKovacs (changes requested):
Thank you for contributing to the MongoDB .NET/C# Driver by clarifying our documentation. Please implement the minor wording change requested and I will merge your change into our source tree.

src/MongoDB.Driver/IFindFluent.cs
leandromoh: adjust done @JamesKovacs ! thanks for review!

JamesKovacs (approved):
LGTM

PR #820, opened 2022-06-14
CSHARP-3529: Support $topN et al in GroupBy.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstAccumulatorExpression.cs
```diff
@@ -13,55 +13,9 @@
...
 */
 
-using MongoDB.Bson;
-using MongoDB.Driver.Core.Misc;
-using MongoDB.Driver.Linq.Linq3Implementation.Ast.Visitors;
-
 namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
 {
-    internal sealed class AstAccumulatorExpression : AstExpression
+    internal abstract class AstAccumulatorExpression : AstExpression
```
rstam, line 18: `AstAccumulatorExpression` is now abstract.

The old `AstAccumulatorExpression` has been renamed `AstUnaryAccumulatorExpression` and is now a subclass of `AstAccumulatorExpression`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickAccumulatorExpression.cs
```diff
@@ -0,0 +1,150 @@
...
+
+using System;
+using System.Linq;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Visitors;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
+{
+    internal sealed class AstPickAccumulatorExpression : AstAccumulatorExpression
```
rstam, line 24: New subclass of the now abstract `AstAccumulatorExpression` class.

This class supports all the `$topN`-like operators.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickExpression.cs
```diff
@@ -0,0 +1,164 @@
...
+
+using System;
+using System.Linq;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Visitors;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
+{
+    internal sealed class AstPickExpression : AstAccumulatorExpression
```
rstam, line 24: Instances of this class in an AST are usually temporary and will be optimized into `AstPickAccumulatorExpression` by the `AstGroupPipelineOptimizer`.

rstam: Oops.. this class should derive from `AstExpression`, not `AstAccumulatorExpression`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickOperator.cs
```diff
@@ -0,0 +1,66 @@
...
+            return @operator switch
+            {
+                AstPickOperator.Bottom => "$bottom",
+                AstPickOperator.BottomN => "$bottomN",
+                AstPickOperator.FirstN => "$firstN",
+                AstPickOperator.LastN => "$lastN",
+                AstPickOperator.MaxN => "$maxN",
+                AstPickOperator.MinN => "$minN",
+                AstPickOperator.Top => "$top",
+                AstPickOperator.TopN => "$topN",
```
rstam, line 45: Most of these aren't supported by the server unless they're in a `$group` stage.

We've considered using strings like `$pseudoBottom`. Haven't decided yet...

Thoughts?

We'd also discussed writing tests that define what happens when these methods are used in the wrong context. That's not done yet.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupPipelineOptimizer.cs
```diff
@@ -309,14 +310,36 @@ public override AstNode VisitMapExpression(AstMapExpression node)
...
-                    var accumulatorExpression = AstExpression.AccumulatorExpression(AstAccumulatorOperator.Push, rewrittenArg);
+                    var accumulatorExpression = AstExpression.UnaryAccumulator(AstUnaryAccumulatorOperator.Push, rewrittenArg);
                     var accumulatorFieldName = _accumulators.AddAccumulatorExpression(accumulatorExpression);
                     return AstExpression.GetField(root, accumulatorFieldName);
                 }
 
                 return base.VisitMapExpression(node);
             }
 
+            public override AstNode VisitPickExpression(AstPickExpression node)
```
rstam, line 321: This is the new optimizer method that moves `AstPickExpressions` to the `$group` stage.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,253 @@
...
+                        var keyTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, keyExpression);
+                        if (!IsValidKey(keyTranslation))
+                        {
+                            throw new ExpressionNotSupportedException(keyExpression, expression, because: "key must be a reference the _id field");
+                        }
+
+                        var nLambda = (LambdaExpression)nExpression;
+                        var keyParameter = nLambda.Parameters.Single();
+                        var keySymbol = context.CreateSymbol(keyParameter, keyTranslation.Serializer, isCurrent: true);
+                        var nContext = context.WithSingleSymbol(keySymbol);
```
rstam, line 128: The `WithSingleSymbol` is important because the lambda to compute `n` can *only* refer to the `key` (`_id`) and nothing else.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        private static object GetSortByDefinition(Expression sortByExpression, Expression expression)
+        {
+            if (sortByExpression.NodeType == ExpressionType.Constant)
+            {
+                return sortByExpression.GetConstantValue<object>(sortByExpression);
+            }
+
+            // we get here when the PartialEvaluator couldn't fully evalute the SortDefinition
```
rstam, line 197: We've talked about this awhile back.

This is how we avoid having to make the `PartialEvaluator` know about `SortDefinition`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,253 @@
...
+
+            return false;
+        }
+
+        private static AstSortFields TranslateSortByDefinition(Expression expression, Expression sortByExpression, object sortByDefinition, IBsonSerializer documentSerializer)
+        {
+            var methodInfoDefinition = typeof(PickMethodToAggregationExpressionTranslator).GetMethod(nameof(TranslateSortByDefinitionGeneric), BindingFlags.Static | BindingFlags.NonPublic);
+            var documentType = documentSerializer.ValueType;
+            var methodInfo = methodInfoDefinition.MakeGenericMethod(documentType);
+            return (AstSortFields)methodInfo.Invoke(null, new object[] { expression, sortByExpression, sortByDefinition, documentSerializer });
```
rstam, line 229: We have to use reflection to call `TranslateSortByDefinitionGeneric` because we don't know `TDocument` until runtime.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/TranslationContext.cs
```diff
@@ -41,7 +41,7 @@ public static TranslationContext Create(Expression expression, IBsonSerializer s
 
         // private fields
         private TranslationContextData _data;
-        private readonly KnownSerializersRegistry _knownKnownSerializersRegistry;
+        private readonly KnownSerializersRegistry _knownSerializersRegistry;
```
rstam, line 44: Fixing a typo I hadn't noticed before.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -57,6 +57,64 @@ public static TSource AllMatchingElements<TSource>(this IEnumerable<TSource> sou
...
+        /// <param name="sortBy">The sort order.</param>
+        /// <param name="selector">The selector.</param>
+        /// <param name="key">The key (needed to infer the key type and to determine the key serializer).</param>
+        /// <param name="n">The number of results to return.</param>
+        /// <returns>The bottom n results.</returns>
+        public static IEnumerable<TResult> BottomN<TSource, TKey, TResult>(
+            this IEnumerable<TSource> source,
+            SortDefinition<TSource> sortBy,
+            Func<TSource, TResult> selector,
+            TKey key,
```
rstam, line 112: Take note of the `key` parameter here.

Turns out not only do we need it to infer the `TKey` type, we also need it to discover the serializer for the key.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3529Tests.cs
```diff
@@ -0,0 +1,1706 @@
...
+                                Builders<GameScore>.Sort.Descending(g => g.Score),
+                                e => new { P = e.PlayerId, S = e.Score })
+                        });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $match : { GameId : 'G1' } }",
+                "{ $group : { _id : '$GameId', __agg0 : { $bottom : { sortBy : { Score : -1 }, output : { P : '$PlayerId', S : '$Score' } } } } }",
+                "{ $project : { Id : '$_id', PlayerId : '$__agg0', _id : 0 } }");
```
rstam, line 52: These pipelines don't *exactly* match the server documentation because of how we move accumulators to the `$group` stage and that we have to project away the `_id` to match the POCO coming out of the pipeline.

JamesKovacs (changes requested):
Overall looks great. See my comment about throwing if the `AstPickExpression` cannot be optimized into a `$group`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickOperator.cs
```diff
@@ -0,0 +1,74 @@
...
+        FirstNAccumulator,
+        FirstNArray,
+        LastNAccumulator,
+        LastNArray,
+        MaxNAccumulator,
+        MaxNArray,
+        MinNAccumulator,
+        MinNArray,
+        TopAccumulator,
+        TopNAccumulator
```
rstam, line 33: There are "accumulator" and "array" versions of these operators. 

Depending on the operator there might only be an "accumulator" version (no "array" version).

It is helpful to have different operators because the accumulator and array versions of the operators take different arguments and are rendered differently.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,312 @@
...
+
+            if (method.IsOneOf(__pickMethods))
+            {
+                var sourceExpression = arguments[0];
+                var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
+                var itemSerializer = ArraySerializerHelper.GetItemSerializer(sourceTranslation.Serializer);
+
+                if (method.IsOneOf(__accumulatorOnlyMethods) && !IsGroupingSource(sourceTranslation.Ast))
+                {
+                    throw new ExpressionNotSupportedException(expression, because: $"{method.Name} can only be used as an accumulator with GroupBy");
```
rstam, line 109: This is how we throw an exception when a method like `Bottom` is used in a way that is not supported.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,312 @@
...
+                else
+                {
+                    resultSerializer = selectorTranslation.Serializer;
+                }
+
+                AstPickOperator @operator;
+                var sourceAst = sourceTranslation.Ast;
+                var @as = selectorParameterSymbol.Var;
+                var selectorAst = selectorTranslation.Ast;
+                if (IsGroupingSource(sourceTranslation.Ast))
```
rstam, line 161: The accumulator and array versions look the same in LINQ. The only difference is whether the `source` is a reference to `$_elements` or not (i.e. a "grouping" source).

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,312 @@
...
+                }
+                else
+                {
+                    @operator = GetArrayOperator(method);
+                    sourceAst = AstExpression.Map(
+                        input: sourceAst,
+                        @as: selectorParameterSymbol.Var,
+                        @in: selectorAst);
+                    @as = null;
+                    selectorAst = null;
```
rstam, line 173: Even though the accumulator and array versions look the same in LINQ, they are *not* the same in MQL.

In MQL we have to combine the `source` and the `selector` into a new source that combines them.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3529Tests.cs
```diff
@@ -0,0 +1,1852 @@
...
+            results[1].GameScores.Should().Equal(10);
+        }
+
+        [Fact]
+        public void FirstN_array_operator_should_work()
+        {
+            var collection = CreateDocumentsWithArrayCollection();
+            var queryable = collection
+                .AsQueryable()
+                .Select(x => x.A.FirstN(e => e.X, 2));
```
rstam, line 626: As currently implemented `FirstN` (and others) look the same in LINQ whether being applied to a grouping or to an array.

Alternatively, we *could* choose to use a *different* API for the array versions.

For example, instead of:

```Select(x => x.A.FirstN(e => e.X, 2)```

we could require:

```Select(x => x.A.Select(e => e.x).FirstN(2)```

that is, the array version of `FirstN` would take only an `n` parameter (besides the `source`), just like the underlying MQL `$firstN` array operator.

File: src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs (hunk 15 above)
Comment (rstam) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs:173]: Unless we change the public API for the array version. See another comment about this in the tests.

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickOperator.cs (hunk 12 above)
Comment (JamesKovacs) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickOperator.cs:33]: Good realization.

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstPickOperator.cs [hunk 17]
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    internal static class AstPickOperatorExtensions
+    {
+        public static string Render(this AstPickOperator @operator)
+        {
+            return @operator switch
+            {
+                AstPickOperator.BottomAccumulator => "$bottomAccumulator",
```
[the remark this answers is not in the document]
JamesKovacs: My concern with `$bottomAccumulator` is that it isn't immediately obvious to the average user that this is a pseudo-operator that will get replaced. `$bottomPlaceholder` is better. Another option could be `$bottom(placeholder)`. Not sure which I prefer. Or some other nomenclature instead of `placeholder`. Dummy, fake, temporary? But all those have other meanings that might be more confusing.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3529Tests.cs
JamesKovacs: I honestly like the symmetry between the grouping and array versions. I think the array version is more discoverable as `x.A.FirstN(e => e.X, 2)` than the proposed alternative.

JamesKovacs (approved):
LGTM. Only outstanding question is whether to use `$bottomPlaceholder`, `$bottom(placeholder)`, or some other formulation that indicates it is not a real operator supported by the server.

JamesKovacs (approved):
LGTM

PR #822, opened 2022-06-16
CSHARP-4214: Fixed Build target.

alkampfergit:
We found a problem in target build for msbuild that seems
to break web project publishing in MsBuild. The modification
done at the target includes multiple times the same file
making publishing from MsBuild commandline to fail.

Details are in jira https://jira.mongodb.org/browse/CSHARP-4214

I'm not entirely sure of the fix, previous version of the file was really simpler, it includes explicitly the dll that should be included, this version uses overlapping wildcard that generates issue.

### DIFF
--- a/src/MongoDB.Driver.Core/MongoDB.Driver.Core.targets
+++ b/src/MongoDB.Driver.Core/MongoDB.Driver.Core.targets
@@ -7,7 +7,7 @@
       <CompressionOsxRuntimesPath>runtimes/osx/native</CompressionOsxRuntimesPath>
     </PropertyGroup>
 
-    <!--snappy-->
+    <!--snappy/zstd-->
     <!--win-->
     <ItemGroup>
       <!--x32/x64-->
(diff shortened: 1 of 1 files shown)


src/MongoDB.Driver.Core/MongoDB.Driver.Core.targets
```diff
@@ -32,27 +32,4 @@
         <Link>%(FileName)%(Extension)</Link>
       </Content>
     </ItemGroup>
-
```
DmitryLukyanov, line 35: can you also please update line 10 to ""

alkampfergit: Done!

DmitryLukyanov (approved):
LGTM,

It looks like the issue you're seeing is related to `msbuild` _preview_ version. I don't see it with released version: `17.0.0.52104`, but changes itself looks harmless and correct too.

PR #831, opened 2022-06-27
CSHARP-4234: Add AppendStage extension method to IMongoQueryable.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Stages/AstUniversalStage.cs
```diff
@@ -0,0 +1,54 @@
...
+* limitations under the License.
+*/
+
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Visitors;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Stages
+{
+    internal sealed class AstUniversalStage : AstStage
```
rstam, line 22: I struggled with the name for this class...

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -66,6 +67,29 @@ public static Task<bool> AnyAsync<TSource>(this IMongoQueryable<TSource> source,
...
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source" />.</typeparam>
+        /// <typeparam name="TResult">The type of the result values returned by the appended stage.</typeparam>
+        /// <param name="source">A sequence of values.</param>
+        /// <param name="stage">The stage to append.</param>
+        /// <param name="resultSerializer">The result serializer.</param>
+        /// <returns>The queryable with a new stage appended.</returns>
+        public static IMongoQueryable<TResult> AppendStage<TSource, TResult>(
+            this IMongoQueryable<TSource> source,
+            PipelineStageDefinition<TSource, TResult> stage,
+            IBsonSerializer<TResult> resultSerializer = null)
```
rstam, line 82: The result serializer *might* be available from the result of rendering the stage, but it might not be.

It can be provided here when it is not otherwise discoverable by the LINQ translator.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Stages/AstUniversalStage.cs
JamesKovacs: What about `AstGenericStage`?

JamesKovacs (changes requested):
Minor naming discussion around `AstUniversalStage`, which you didn't care for. What about `AstGenericStage` instead?

JamesKovacs (approved):
LGTM.

As discussed over Slack, `AstGenericStage` has the possibility to get confused with .NET generics. Other options either sounded odd (`AstNonspecificStage`) or were potentially confusing (`AstBsonDocumentStage` might be confused with the upcoming `$documents`). `AstUniversalStage` is the best option we could come up with.

PR #833, opened 2022-06-28
CSHARP-2107: Verified that the issue is no longer present in LINQ3.

rstam:
No changes to push to master.

Just verifying that the issue is no longer present.


JamesKovacs (approved):
LGTM

PR #838, opened 2022-07-01
CSHARP-1555: Verify that issue is not present in LINQ3.

rstam:


rstam:
The original repro in the ticket registered these conventions:

```
var conventionPack = new ConventionPack
{
    new IgnoreIfNullConvention(true),
    new NoIdMemberConvention()
};
ConventionRegistry.Register("GlobalConventions", conventionPack, t => true);
```

But we can't do that in our test suite because it would alter global state.

Instead I used the `[BsonIgnoreIfNull]` attribute on each member of the `Person` class.

I did NOT use the equivalent of `NoIdMemberConvention` because that convention is not appropriate for documents stored in a collection which MUST have an `_id`.

JamesKovacs (approved):
LGTM

PR #840, opened 2022-07-01
CSHARP-1906: Case insensitive searches in LINQ.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp1906Tests.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public class CSharp1906Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Using_ToLower_should_work()
+        {
+            var collection = CreateCollection();
+            var lowerCaseValues = new[] { "abc", "def" }; // ensure all are lower case at compile time
+            var queryable = collection.AsQueryable()
+                .Where(c => lowerCaseValues.Contains(c.S.ToLower()));
```
rstam, line 33: This is the approach suggested in the JIRA ticket.

They claim that this usage should *imply* using regular expressions. I disagree.

Nonetheless, this usage can work (now that we support `$expr` filters) as long as the string array is all lower case at compile time (or at least before being passed to the query).

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp1906Tests.cs
```diff
@@ -0,0 +1,78 @@
...
+            results.Select(x => x.Id).Should().Equal(1, 2);
+        }
+
+        [Fact]
+        public void Using_regular_expression_should_work()
+        {
+            var collection = CreateCollection();
+            var regularExpresssion = new StringOrRegularExpression[] { new Regex("ABC", RegexOptions.IgnoreCase), new Regex("DEF", RegexOptions.IgnoreCase) };
+            var queryable = collection.AsQueryable()
+                .Where(c => c.S.StringIn(regularExpresssion));
```
rstam, line 48: The `StringIn` method is new in LINQ3. It is specifically for expressing `$in` filters against an array of strings or regular expressions.

This is the official way to do regular expression based filters now.

PR #842, opened 2022-07-04
CSHARP-4209: Integrate with .NET Logging library

BorisDog:
This is draft PR, and does not contain unit tests for logging helpers


src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+            "MongoDB.Bson.Tests",
+            "MongoDB.Driver.Tests",
+            "MongoDB.Driver.Core.Tests"
+        };
+
+        private static readonly string __specCategoryPrefix = typeof(LogCategories).FullName;
+
+        private const string PrefixSpec = "MongoDB";
+        private const string PrefixInternal = "MongoDB.Internal";
+        private const string PrefixUnittests = "MongoDB.Unittests";
```
JamesKovacs, line 41: I assume that `PrefixUnittests` will be used for all our tests, not just the unit tests. Wouldn't this better named `PrefixTests` with a value of `MongoDB.Tests`?

tests/MongoDB.Driver.Core.TestHelpers/Logging/XUnitLogger.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions.Internal;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.TestHelpers.Logging
+{
+    //[DebuggerStepThrough]
```
JamesKovacs, line 24: Should this still be commented?

JamesKovacs (changes requested):
Overall looking great. My main concern is that previously the driver would not create new event objects unless we had handlers registered. Now we create those event objects every time and they may never be used if no logger is present. I'm not sure if the additional allocation pressure is a concern or not, but we should do some quick tests to verify whether it has a noticeable effect on driver memory usage and/or performance. Happy to discuss more. Curious as to your thoughts.

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
BorisDog: Agreed, done and added tests.

src/MongoDB.Driver.Core/Core/Clusters/Cluster.cs
```diff
@@ -223,9 +224,13 @@ public virtual void Initialize()
             ThrowIfDisposed();
             if (_state.TryChange(State.Initial, State.Open))
             {
+                _clusterEventsLogger.LogDebug("Initialized");
+
                 if (_settings.CryptClientSettings != null)
                 {
                     _cryptClient = CryptClientCreator.CreateCryptClient(_settings.CryptClientSettings);
+
+                    _clusterEventsLogger.LogDebug("CryptClient created. Shared library version {SharedLibraryVersion}", _cryptClient.CryptSharedLibraryVersion);
```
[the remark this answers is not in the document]

BorisDog (commented):
@JamesKovacs Memory pressure is an important aspect to track and validate. In our case, for better or for worse the events entities are structs, so at least we don't have to worry about that.

tests/MongoDB.Driver.Core.TestHelpers/Logging/XUnitOutputAccumulator.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private readonly HashSet<string> _categoriesToExclude;
+        private readonly List<LogEntry> _logs;
+        private readonly ITestOutputHelper _output;
+
+        public XUnitOutputAccumulator(ITestOutputHelper output, string[] logCategoriesToExclude)
+        {
+            _categoriesToExclude = logCategoriesToExclude?.Any() == true ? new HashSet<string>(logCategoriesToExclude) : null;
+            _logs = new List<LogEntry>();
+            _output = Ensure.IsNotNull(output, nameof(output));
```
DmitryLukyanov, line 35: I would say accumulator should not be responsible for output and better leave output logic outside. Thoughts?

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -214,18 +187,20 @@ public void Clear(ObjectId serviceId)
 
             // generation increment can happen outside lock, as _serviceStates is threadsafe
             // and currently we allow dispose to start during generation increment.
-            _clearingEventHandler?.Invoke(new ConnectionPoolClearingEvent(_serverId, _settings, serviceId));
+            _eventsLogger.LogAndPublish(new ConnectionPoolClearingEvent(_serverId, _settings, serviceId));
```
DmitryLukyanov, line 190: here and in other places, previously we didn't create event if it wasn't requested. Now we do it a lot, not sure how bad it is

src/MongoDB.Driver.Core/Core/Clusters/DnsMonitor.cs
```diff
@@ -57,7 +65,8 @@ public DnsMonitor(IDnsMonitoringCluster cluster, IDnsResolver dnsResolver, strin
             _service = "_mongodb._tcp." + _lookupDomainName;
             _state = DnsMonitorState.Created;
 
-            eventSubscriber?.TryGetEventHandler(out _sdamInformationEventHandler);
+            _eventsLogger = logger.ToEventsLogger(eventSubscriber, _service);
+            _isSdamEventTracked = _eventsLogger.IsEventTracked<SdamInformationEvent>();
```
DmitryLukyanov, line 69: Can you describe what this "IsEventTracked" method is supposed to do? It will be true not only when we subscribed on events, but also if `  _isInformationEnabled = Logger?.IsEnabled(LogLevel.Information) == true;` is true.

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
```diff
@@ -17,12 +17,14 @@
 using System.Diagnostics;
 using System.IO;
 using System.Threading;
+using Microsoft.Extensions.Logging;
```
DmitryLukyanov, line 20: we should use a new logic for previous `ConfigureSdamLogging` logging

DmitryLukyanov (commented):
Initial comments. I didn't go through all places yet. But it provides some questions to think on

src/MongoDB.Driver.Core/Core/Events/EventType.cs
```diff
@@ -0,0 +1,74 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+namespace MongoDB.Driver.Core.Events
+{
+    internal enum EventType
```
DmitryLukyanov, line 18: let's keep eye on this enum. Maybe we will find a better approach eventually

src/MongoDB.Driver.Core/Core/Logging/LoggerFactoryCategoryDecorator.cs
```diff
@@ -0,0 +1,37 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using Microsoft.Extensions.Logging;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Logging
+{
+    internal sealed class LoggerFactoryCategoryDecorator : ILoggerFactory
```
DmitryLukyanov, line 21: let's create a ticket to investigate downsides of using BeginScope instead decorator

src/MongoDB.Driver.Core/Core/Logging/LoggerDecorator.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace MongoDB.Driver.Core.Logging
+{
+    internal class LoggerDecorator<T>
+    {
+        public ILogger<T> Logger { get; }
+        public string DecorationName { get; }
+        public string DecorationValue { get; }
+
+        public LoggerDecorator(ILogger<T> logger, string decorationName, string decorationValue)
```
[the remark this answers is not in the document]
DmitryLukyanov: I'm ok to think about it in the scope of different ticket

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
DmitryLukyanov: not answered

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
```diff
@@ -26,29 +28,36 @@ namespace MongoDB.Driver.Core.TestHelpers.Logging
...
 
-            LoggerFactory = new XUnitLoggerFactory(_loggerBase);
+            LoggerFactory = (new XUnitLoggerFactory(_output)).DecorateCategories();
             Logger = LoggerFactory.CreateLogger<LoggableTestClass>();
         }
 
         protected ILogger<LoggableTestClass> Logger { get; }
         protected ILoggerFactory LoggerFactory { get; }
         protected LogLevel MinLogLevel { get; set; }
+        public LogEntry[] Logs => _output.Logs;
```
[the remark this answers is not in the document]
DmitryLukyanov: Just to recall previously unanswered points: 
1. Sooner or later you will need to break this style to add a backed field.
2. With this change, you have no longer a clear split between external and internal class API

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
```diff
@@ -26,29 +28,36 @@ namespace MongoDB.Driver.Core.TestHelpers.Logging
...
 
         protected ILogger<TCategory> CreateLogger<TCategory>() => LoggerFactory.CreateLogger<TCategory>();
 
+        protected virtual void Dispose(bool disposing) { }
```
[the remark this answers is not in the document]
DmitryLukyanov: not answered

tests/MongoDB.Driver.Core.TestHelpers/Logging/XUnitLoggerFactory.cs
```diff
@@ -14,31 +14,24 @@
...
         }
 
-        public ILogger<TCategory> CreateLogger<TCategory>() =>
-            new TypedLoggerDecorator<TCategory>(_loggerBase);
-    }
-
-    public class EmptyLoggerFactory : ILoggerFactory
-    {
-        public static ILoggerFactory Instance { get; } = new EmptyLoggerFactory();
+        public void AddProvider(ILoggerProvider provider) { }
```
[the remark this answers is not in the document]
DmitryLukyanov: yeah, I missed that it's test side logic

src/MongoDB.Driver.Core/Core/Connections/CommandEventHelper.cs
```diff
@@ -542,7 +517,7 @@ private void ProcessCommandReplyMessage(CommandState state, ReplyMessage<RawBson
 
         private void ProcessQueryReplyMessage(CommandState state, ReplyMessage<RawBsonDocument> replyMessage, ConnectionId connectionId)
         {
-            if (_succeededEvent != null)
+            if (_shouldTrackSucceeded)
```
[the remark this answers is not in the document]
DmitryLukyanov: not answered

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
DmitryLukyanov: well, we can make it obsolete at the very least. Let's create a ticket for this

tests/MongoDB.Driver.Core.TestHelpers/Logging/LoggableTestClass.cs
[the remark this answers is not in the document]
DmitryLukyanov: 1. Well, when it will happen, we may already had code base that will require changes. I would say it differently, let's use a new styling approach when we will know that the previous approach doesn't work for us for some reason to avoid any unnecessary changes.
2. it's not what I meant, unless you're planning to create a private properties. With this approach any property can be used as internal class API and external which gives some "small" pain.

After all, I disagree with this change, but also don't want spending time on this one more time

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
BorisDog: Yeah good idea. Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -57,7 +57,7 @@ public sealed class UnifiedTestRunner : IDisposable
 
         public void Run(JsonDrivenTestCase testCase)
         {
-            _logger.Debug("Running {0}", testCase.Name);
+            _logger?.LogDebug("Running {0}", testCase.Name);
```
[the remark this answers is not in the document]
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
DmitryLukyanov: can you post a link on ticket?

src/MongoDB.Driver/ClusterRegistry.cs
```diff
@@ -52,15 +53,19 @@ public static ClusterRegistry Instance
...
+#pragma warning disable CS0618 // Type or member is obsolete
             var builder = new ClusterBuilder()
                 .ConfigureCluster(settings => ConfigureCluster(settings, clusterKey))
                 .ConfigureServer(settings => ConfigureServer(settings, clusterKey))
                 .ConfigureConnectionPool(settings => ConfigureConnectionPool(settings, clusterKey))
                 .ConfigureConnection(settings => ConfigureConnection(settings, clusterKey))
                 .ConfigureTcp(settings => ConfigureTcp(settings, clusterKey))
-                .ConfigureSdamLogging(settings => ConfigureSdamLogging(settings, clusterKey));
+                .ConfigureSdamLogging(settings => ConfigureSdamLogging(settings, clusterKey))
+                .ConfigureLoggingFactory(_ => loggerFactory);
```
DmitryLukyanov, line 67: todo

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+        private static readonly string[] __driverNamespaces = new []
+        {
+            "MongoDB.Bson",
+            "MongoDB.Driver",
+            "MongoDB.Driver.Core"
+        };
+
+        private static readonly string[] __driverTestsNamespaces = new[]
+        {
+            "MongoDB.Bson.Tests",
```
DmitryLukyanov, line 32: this looks like wrong logic given that now our driver code is aware of how it's used, ie it should not know about tests at all

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace MongoDB.Driver.Core.Logging
+{
+    internal static class LogCatergoryHelper
+    {
+        private static readonly string[] __driverNamespaces = new []
+        {
+            "MongoDB.Bson",
+            "MongoDB.Driver",
+            "MongoDB.Driver.Core"
```
DmitryLukyanov, line 27: MongoDB.Driver.GridFS?

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
DmitryLukyanov: Can it be not `driverTests`, but something like `driverUsed` or something similar?

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -127,15 +139,18 @@ private void MonitorServer()
...
                     IConnection toDispose;
                     lock (_lock)
                     {
                         toDispose = _roundTripTimeConnection;
                         _roundTripTimeConnection = null;
                     }
+                    var connectionId = toDispose?.ConnectionId;
                     toDispose?.Dispose();
+
+                    _logger.LogDebug(ex, StructuredLogsTemplates.Message_ConnectionId, "Monitoring exception", connectionId);
```
DmitryLukyanov, line 153: RTT monitoring exception

DmitryLukyanov: I still don't understand where we use `LogDebug` and where `LogException` or something. Can we define clear rules here?

DmitryLukyanov: what will happen if connectionIs is null?

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
DmitryLukyanov: this also should include logic like in this class `SdamLoggingSettings`

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
```diff
@@ -113,11 +116,25 @@ public ClusterBuilder ConfigureConnectionPool(Func<ConnectionPoolSettings, Conne
...
 
+        /// <summary>
+        /// Configures the logging factory.
+        /// </summary>
+        /// <param name="configurator">The logging factory configurator delegate.</param>
+        /// <returns>A reconfigured cluster builder.</returns>
+        [CLSCompliant(false)]
+        public ClusterBuilder ConfigureLoggingFactory(Func<ILoggerFactory, ILoggerFactory> configurator)
+        {
+            Ensure.IsNotNull(configurator, nameof(configurator));
```
DmitryLukyanov, line 127: let's add a blank line after ensure as in other methods here

src/MongoDB.Driver/ClusterRegistry.cs
```diff
@@ -52,15 +53,19 @@ public static ClusterRegistry Instance
         private readonly Dictionary<ClusterKey, ICluster> _registry = new Dictionary<ClusterKey, ICluster>();
 
         // methods
-        private ICluster CreateCluster(ClusterKey clusterKey)
+        private ICluster CreateCluster(ClusterKey clusterKey, ILoggerFactory loggerFactory)
         {
+
```
DmitryLukyanov, line 58: empty line

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -78,7 +78,7 @@ public MongoClient(MongoClientSettings settings)
         {
             _settings = Ensure.IsNotNull(settings, nameof(settings)).FrozenCopy();
             _linqProvider = _settings.LinqProvider;
-            _cluster = ClusterRegistry.Instance.GetOrCreateCluster(_settings.ToClusterKey());
+            _cluster = ClusterRegistry.Instance.GetOrCreateCluster(_settings.ToClusterKey(), _settings.LoggerFactory);
```
DmitryLukyanov, line 81: I would say we need going though a regular workflow passing arguments from highlevel settings to low onces. Ie we need to do it through `ClusterKey`

src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -448,6 +451,20 @@ public TimeSpan LocalThreshold
             }
         }
 
+        /// <summary>
+        /// Gets or sets the logger factory
+        /// </summary>
+        [CLSCompliant(false)]
+        public ILoggerFactory LoggerFactory
```
DmitryLukyanov, line 458: we have tests for this class

src/MongoDB.Driver/ClusterRegistry.cs
```diff
@@ -174,14 +179,14 @@ private TcpStreamSettings ConfigureTcp(TcpStreamSettings settings, ClusterKey cl
...
-        internal ICluster GetOrCreateCluster(ClusterKey clusterKey)
+        internal ICluster GetOrCreateCluster(ClusterKey clusterKey, ILoggerFactory loggerFactory)
         {
             lock (_lock)
             {
                 ICluster cluster;
                 if (!_registry.TryGetValue(clusterKey, out cluster))
                 {
-                    cluster = CreateCluster(clusterKey);
+                    cluster = CreateCluster(clusterKey, loggerFactory);
```
DmitryLukyanov, line 189: I think we need working through clusterKey as the rest of options here

DmitryLukyanov (changes requested):

src/MongoDB.Driver/ClusterRegistry.cs
BorisDog: Done.

src/MongoDB.Driver/MongoClient.cs
BorisDog: Done.

src/MongoDB.Driver/MongoClientSettings.cs
BorisDog: Added tests.

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
DmitryLukyanov: >BTW similar to our InternalsVisibleTo.

this is not too good workaround that we probably can remove when we will have better code structure (remove Core). At least we can consider this

>but we are making a compromise in favor of simplicity.

I'm ok with simplicity here, but renaming this naming from "test" into "notdriver" (or something) looks simple too. See how difficult to do it, if not too easy, we can consider this work later

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
DmitryLukyanov: >There is no LogException. Exception object can be optionality passed in regardless of verbosity.

I meant, do we call anything other than `LogDebug`? If so, how do we decide it? If no so, then why?

DmitryLukyanov (approved):
LGTM

See some remaining comments. I assume no issues on EG

PR #846, opened 2022-07-11
CSHARP-4245: Dispose CancellationTokenSource in tests

BorisDog:


tests/MongoDB.Driver.Core.Tests/Core/Clusters/DnsClientWrapperTests.cs
```diff
@@ -93,7 +93,7 @@ public void ResolveSrvRecords_should_throw_when_cancellation_is_already_requeste
         {
             var subject = CreateSubject();
             var service = "_mongodb._tcp.test5.test.build.10gen.cc";
-            var cts = new CancellationTokenSource();
+            using var cts = new CancellationTokenSource();
```
JamesKovacs, line 96: Since we are updating these lines anyways, let's go with our usual coding standard of not using abbreviations.

`cts` => `cancellationTokenSource`

Same elsewhere.

tests/MongoDB.Driver.GridFS.Tests/DelegatingStreamTests.cs
```diff
@@ -146,14 +145,13 @@ public void CopyToAsync_should_call_wrapped_stream()
             var subject = new DelegatingStream(mockStream.Object);
             var mockDestination = new Mock<Stream>();
             var bufferSize = 1;
-            var cancellationToken = new CancellationTokenSource().Token;
             var task1 = Task.FromResult<object>(null);
-            mockStream.Setup(s => s.CopyToAsync(mockDestination.Object, bufferSize, cancellationToken)).Returns(task1);
+            mockStream.Setup(s => s.CopyToAsync(mockDestination.Object, bufferSize, default)).Returns(task1);
```
JamesKovacs, line 149: Isn't this changing the meaning of the test? Previously we were asserting that the passed-in `CancellationTokenSource` was passed correctly. But since `CancellationTokenSource` is a class `default(CancellationTokenSource` is simply `null`. Thus we cannot tell if the `CancellationTokenSource` is being correctly passed through or if we are simply explicitly passing `null` through the call chain.

tests/MongoDB.Driver.Legacy.Tests/DefaultLegacyOperationExecutorTests.cs
```diff
@@ -32,11 +31,10 @@ public void ExecuteReadOperation_should_execute_operation()
             var subject = CreateSubject();
             var binding = new Mock<IReadBinding>().Object;
             var mockOperation = new Mock<IReadOperation<BsonDocument>>();
-            var cancellationToken = new CancellationTokenSource().Token;
 
-            var result = subject.ExecuteReadOperation(binding, mockOperation.Object, cancellationToken);
+            var result = subject.ExecuteReadOperation(binding, mockOperation.Object, default);
 
-            mockOperation.Verify(m => m.Execute(binding, cancellationToken), Times.Once);
+            mockOperation.Verify(m => m.Execute(binding, default), Times.Once);
```
JamesKovacs, line 37: Same concern around the use of `default` as with the previous file.

JamesKovacs (changes requested):
Very nice use of `using var` syntax. My one concern is around the use of `default` in certain mocking scenarios. See my comment below for more details. Plus a minor point about using `cancellationTokenSource` rather than `cts` (even though `cts` was in the original).

tests/MongoDB.Driver.GridFS.Tests/DelegatingStreamTests.cs
BorisDog: Fixed.

tests/MongoDB.Driver.Legacy.Tests/DefaultLegacyOperationExecutorTests.cs
BorisDog: Fixed.

tests/MongoDB.Driver.Core.Tests/Core/Clusters/DnsClientWrapperTests.cs
BorisDog: Done.

JamesKovacs (approved):
LGTM

PR #849, opened 2022-07-22
Skunkworks: Update .editorconfig to match coding convention for C# driver

dgolub:
This PR updates the .editorconfig file to match the coding conventions used in the C# driver.  The changes eliminated something like 95% of the 10,000+ errors from running code analysis on this project.  Tweaks were made to the source code to address the remaining errors.  These changes consisted of fixing portions of the code that were inconsistent with the conventions used elsewhere and removing suppressions of warnings that didn't exist in the first place.  (Perhaps some of these were genuine warnings in earlier C# versions, but more recent compiler versions may be smart enough to recognize that these warnings were superfluous.)  With the entirety of this PR in place, code analysis passes on the project with no errors.

I've set the severity for a number of the analysis rules to `refactoring`, which means that no error is produced for violating them, but IntelliSense offers you the option to change them in the code editor.  At least some of these might make sense to consider enabling in the future, given that they would represent code quality improvements and modernizations to take full advantage of some of the features added in more recent C# versions.  However, given that this PR is already not tiny and Skunkworks is of limited duration, I'm leaving them disabled for now.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -39,17 +39,17 @@ csharp_indent_labels = one_less_than_current
 csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:suggestion
 
 # avoid this. unless absolutely necessary
-dotnet_style_qualification_for_field = false:suggestion
-dotnet_style_qualification_for_property = false:suggestion
-dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_field = false:refactoring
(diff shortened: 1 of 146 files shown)


BorisDog:
Thanks @dgolub for this cleanup.
We are going to close this PR as won't do at this moment. The codebase had diverged significantly since this PR, and we'll be doing further significant refactoring for 3.0 release. We are planning to address this issue in [CSHARP-4408](https://jira.mongodb.org/browse/CSHARP-4408), we plan to unify editor configs across all our C# repos.

PR #850, opened 2022-07-22
Skunkworks: Tweak CSS for printing

dgolub:
This PR adds a printer-specific section (`@media print`) to the CSS for the reference documentation to fix a number of things that were causing the pages to not look right when printed.  The appearance when viewing on screen is unaffected.  I've tested by running locally with

```
hugo server --baseUrl=http://localhost --buildDrafts --watch
```

There is an issue where the CSS used in the browser reverts back to the original version when the page is refreshed or the user navigates to a different page.  I'm not sure exactly why.  Before merging, we would need to ensure that the issue does not occur when viewing the documentation for real or fix whatever is causing it.

Before:
C# Print Before

After:
C# Print After

### DIFF
--- a/Docs/reference/themes/mongodb/static/css/overrides.css
+++ b/Docs/reference/themes/mongodb/static/css/overrides.css
@@ -115,3 +115,26 @@ a code {
 .jsEnabled #search {
   visibility: visible;
 }
+
+@media print {
+  a[href]:after {
+    content: none !important;
+  }
+
+  .edit-link,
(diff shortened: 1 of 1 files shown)


JamesKovacs:
Given that your changes are surrounded by `@media print`, I'm not sure how refreshing or browsing to another page could affect the layout. Sounds like a browser idiosyncrasy rather than a CSS problem on our end. I'll merge and we can tweak further if we notice additional problems.

PR #853, opened 2022-07-27
CSHARP-4271: Replace Snappy and Zstd with managed implementations

adelinowona:


src/MongoDB.Driver.Core/Core/Compression/SnappyCompressor.cs
```diff
@@ -35,15 +34,9 @@ public void Compress(Stream input, Stream output)
             var uncompressedSize = (int)(input.Length - input.Position);
             var uncompressedBytes = new byte[uncompressedSize]; // does not include uncompressed message headers
             input.ReadBytes(uncompressedBytes, offset: 0, count: uncompressedSize, CancellationToken.None);
-            var maxCompressedSize = SnappyCodec.GetMaxCompressedLength(uncompressedSize);
+            var maxCompressedSize = Snappier.Snappy.GetMaxCompressedLength(uncompressedSize);
```
DmitryLukyanov, line 37: let's mention `Snappier` ns in the top of the page, same in other places

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
```diff
@@ -30,23 +29,27 @@ internal class ZstandardCompressor : ICompressor
 
         public ZstandardCompressor(Optional<int> compressionLevel = default)
         {
+            if (compressionLevel.HasValue && (compressionLevel.Value <= 0 || compressionLevel.Value >= 23))
```
DmitryLukyanov, line 32: can you provide a link where you found these restrictions?

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -199,6 +80,10 @@
     <Compile Include="..\MongoDB.Shared\Hasher.cs" Link="Core\Misc\Hasher.cs" />
   </ItemGroup>
 
+  <ItemGroup>
```
DmitryLukyanov, line 83: is it really required?

src/MongoDB.Driver.Core/Core/Compression/Zstandard/PinnedBufferWalker.cs
```diff
@@ -1,71 +0,0 @@
...
-* limitations under the License.
-*/
-
-using System;
-using System.Runtime.InteropServices;
-using MongoDB.Driver.Core.Misc;
-
-namespace MongoDB.Driver.Core.Compression.Zstandard
-{
-    internal class PinnedBufferWalker : IDisposable
```
DmitryLukyanov, line 22: please close this ticket as well https://jira.mongodb.org/browse/CSHARP-2959

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -15,28 +15,57 @@
...
+
+        // private static fields
+        private static readonly byte[] __bigMessage = GenerateBigMessage(135000);
         private static string __testMessage = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz";
 
-        [Theory]
-        [InlineData(CompressorType.Snappy)]
-        [InlineData(CompressorType.ZStandard)]
-        public void Compressor_should_read_the_previously_written_message(CompressorType compressorType)
+
```
DmitryLukyanov, line 51: double blank line

src/MongoDB.Driver.Core/Core/NativeLibraryLoader/LibraryLoader.cs
```diff
@@ -1,79 +0,0 @@
...
-* limitations under the License.
-*/
-
-using System;
-using System.Runtime.InteropServices;
-using MongoDB.Driver.Core.Misc;
-
-namespace MongoDB.Driver.Core.NativeLibraryLoader
-{
-    internal class LibraryLoader
```
DmitryLukyanov, line 22: please close this ticket as well https://jira.mongodb.org/browse/CSHARP-3693

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -15,28 +15,57 @@
...
+        What here shall miss, our toil shall strive to mend.";
+
+        // private static fields
+        private static readonly byte[] __bigMessage = GenerateBigMessage(135000);
         private static string __testMessage = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz";
 
-        [Theory]
-        [InlineData(CompressorType.Snappy)]
-        [InlineData(CompressorType.ZStandard)]
-        public void Compressor_should_read_the_previously_written_message(CompressorType compressorType)
```
DmitryLukyanov, line 36: I would probably recommend to leave all previous tests in this file. And additionally put tests from the files you've removed

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -56,6 +85,92 @@ public void Compressor_should_read_the_previously_written_message(CompressorType
...
+        }
+
+        [Theory]
+        [InlineData(1, "40,181,47,253,0,72,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39")]
+        [InlineData(6, "40,181,47,253,0,88,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39")]
+        [InlineData(15, "40,181,47,253,0,96,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39")]
+        [InlineData(21, "40,181,47,253,0,128,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39")]
+        public void Zstandard_compress_should_generate_expected_bytes_for_different_compression_levels(int compressionLevel, string expectedBytes)
+        {
+            var testMessage = "abcdefghijklmnopqrstuvwxyz0123456789 abcdefghijklmnopqrstuvwxyz0123456789 abcdefghijklmnopqrstuvwxyz0123456789";
```
DmitryLukyanov, line 120: you already have global `__testMessage` , since the value in this test is copied from another test, you can modify `__testMessage` by the value in line 120 (to not change generated bytes)

DmitryLukyanov (commented):
Initial review related to our code base. I didn't look at external nugets yet

src/MongoDB.Driver.Core/Core/Compression/SnappyCompressor.cs
JamesKovacs: Agreed. We should have `using Snappier;` in the using directives at the top to eliminate the need to include the namespace every time we use `Snappy`.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
```diff
@@ -30,23 +29,27 @@ internal class ZstandardCompressor : ICompressor
...
+
             _compressionLevel = compressionLevel.WithDefault(_defaultCompressionLevel);
         }
 
         public CompressorType Type => CompressorType.ZStandard;
 
         public void Compress(Stream input, Stream output)
         {
-            using (var zstandardStream = new ZstandardStream(output, CompressionMode.Compress, _compressionLevel))
+            using (var zstandardStream = new ZstdSharp.CompressionStream(output, _compressionLevel))
```
JamesKovacs, line 44: Same as with `Snappier`. We should include `using ZstdSharp;` in the using directives at the top.

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -34,6 +34,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0" />
+    <PackageReference Include="Snappier" Version="1.0.0" />
+    <PackageReference Include="ZstdSharp.Port" Version="0.6.1" />
```
JamesKovacs, line 38: I would place these package references with the other 3rd-party NuGet packages such as DnsClient.NET below. (See line 59.)

JamesKovacs (changes requested):
This is looking very good. Minor feedback below.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
adelinowona: The facebook zstd library states that it only supports compression levels 1 through 22. Here is a link: https://facebook.github.io/zstd/zstd_manual.html#:~:text=The%20library%20supports%20regular%20compression%20levels%20from%201%20up%20to%20ZSTD_maxCLevel()%2C%0A%20%20which%20is%20currently%2022.

The ZstdSharp.Port library I am using is a port of the facebook zstd library so it also only supports compression levels 1-22. The library does not throw an error if you pass in a compression level outside the 1-22 range. Instead it will clamp the value to be within that range. We could leave as it was but I thought we would want to be explicit here so that the user gets an error if they provide unsupported compression levels.

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
adelinowona: This was automatically added to the project file when I added the managed libraries through nuget on VS but I think can remove it.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
DmitryLukyanov: Effectively it says:
>The library supports regular compression levels from 1 up to ZSTD_maxCLevel(), which is currently 22. Levels >= 20,

so 22 is only current version. I would validate on calculated value instead. NOTE: we actually don't have a way to configure this option in the driver yet

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
adelinowona: I actually didn't delete this test. I just renamed it to Snappy as it was only testing the Snappy compressor after I wrote a similar tests for the Zstandard compressor that uses different compression levels. 

For the files I removed, I looked through the tests and only migrated some tests (those that didn't test the internal implementation details of the unmanaged compressors).

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
DmitryLukyanov: yes, I think this part is not needed

DmitryLukyanov:
Just to mention that there is a note that older TFs can have problems with performance on windows for snappier: https://github.com/brantburnett/Snappier#other-projects

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
adelinowona: Oh I had assumed the driver supported setting the compression level for zstd as it does for zlib. From your point, I also agree that we should use the calculated value but since the driver doesn't support passing a compression level for the zstd compressor, I think we can then omit this for now?

DmitryLukyanov: sgtm

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -56,6 +85,92 @@ public void Compressor_should_read_the_previously_written_message(CompressorType
...
+                (input, output) =>
+                {
+                    input.Position = 0;
+                    var resultBytes = input.ReadBytes((int)input.Length);
+                    resultBytes.Should().Equal(messageBytes);
+                });
+        }
+
+        [Theory]
+        [InlineData(1, "40,181,47,253,0,72,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39")]
```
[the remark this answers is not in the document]
adelinowona: I don't know but I can try to look into it

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -46,8 +47,7 @@ public class CompressorsTests
 
         // private static fields
         private static readonly byte[] __bigMessage = GenerateBigMessage(135000);
-        private static string __testMessage = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz";
-
+        private static string __testMessage = "abcdefghijklmnopqrstuvwxyz0123456789 abcdefghijklmnopqrstuvwxyz0123456789 abcdefghijklmnopqrstuvwxyz0123456789";
```
JamesKovacs, line 50: This can be a constant and moved to private constants above.

JamesKovacs: Why was `__testMessage` changed from a-z repeated 3 times to a-z0-9<space> repeated 3 times? Changing `__testMessage` necessitated changing an unrelated zlib test.

JamesKovacs (changes requested):
Minor feedback.

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -56,6 +86,79 @@ public void Compressor_should_read_the_previously_written_message(CompressorType
...
+                    var compressor = GetCompressor(CompressorType.ZStandard, compressionLevel);
+                    compressor.Compress(input, output);
+                    lengths.Add((int)output.Length);
+                }
+            }
+            lengths.Should().BeInDescendingOrder();
+        }
+
+        [Fact]
+        public void Zstandard_compress_should_throw_when_output_stream_is_null()
```
DmitryLukyanov, line 151: also please put tests in this file in alphabetical order based on their names

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -73,7 +176,7 @@ public void Zlib_should_generate_expected_compressed_bytes()
                     var result = string.Join(",", resultBytes);
                     result
                         .Should()
-                        .Be("120,156,74,76,74,78,73,77,75,207,200,204,202,206,201,205,203,47,40,44,42,46,41,45,43,175,168,172,74,36,67,6,0,0,0,255,255,3,0,21,79,33,94");
+                        .Be("120,156,74,76,74,78,73,77,75,207,200,204,202,206,201,205,203,47,40,44,42,46,41,45,43,175,168,172,50,48,52,50,54,49,53,51,183,176,84,72,164,150,34,0,0,0,0,255,255,3,0,228,159,39,197");
```
DmitryLukyanov, line 179: why these bytes were changed?

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
DmitryLukyanov: I suggested this since we can use this `__testMessage` in zstandard. So this needed to not change expected bytes for zstandard, but I missed that this change will require updating expected result for zlib

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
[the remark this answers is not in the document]
adelinowona: Ok so I did some investigating:
the last 3 bytes seem to be way of ending the compressed stream. There is a function called ZSTD_endStream that is used to end the stream. The unmanaged implementation in the driver actually called that function via Flush when it flushed the zstandard stream (https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs#:~:text=zstandardStream.Flush()%3B) you can follow the call stack for Flush to see it.

So I tried to also Flush the zstandard stream using the Library we are using I got this:
`"40,181,47,253,0,72,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39,40,181,47,253,32,0,1,0,0"`

The Flush in the library basically writes a null buffer to the compressed stream. You can see it has the 3 bytes (1, 0, 0) at the end. If I don't call Flush then those bytes won't be there.

However, I don't think any of these matters as I compressed a message with and without Flushing the stream using the library and then decompressed the compressed bytes using the unmanaged code implementation and it produced the right uncompressed output. I also compressed the message using the unmanaged code implementation and decompressed it using the library with success. Therefore, whether those bytes are there or not doesn't seem matter as far as I can tell.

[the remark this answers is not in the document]
DmitryLukyanov: I think it's better to call Flush. Effectively it means that currently we ignore some information (last 3 bytes in this test). Maybe this information is "unimportant technical" details, maybe it's "technical" only for cases we checked in tests, and will be important in other cases. In any case it looks safer to work with all data compressor prepared for us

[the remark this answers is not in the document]
adelinowona: SGTM

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -131,6 +161,79 @@ public void Zlib_should_throw_exception_if_the_level_is_out_of_range([Values(-2,
             e.ParamName.Should().Be("compressionLevel");
         }
 
+        [Theory]
+        [InlineData(1, "40,181,47,253,0,72,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39,40,181,47,253,32,0,1,0,0")]
```
DmitryLukyanov, line 165: given that you don't call flush, how did you get this result? Also, now the difference is even bigger than before, do you know why?

adelinowona: I updated my commit with flushing as for some reason it didn't have the addition of flushing. Flushing writes a null buffer to the compressed stream and you can see that before the (1, 0, 0) bytes, the bytes (40,181,47,253) inserted are similar to the bytes at the start of the result bytes. My guess was the original zstd implementation do something to eliminate those starting bytes and just have the (1, 0, 0) bytes and the library we are using doesn't do it.

DmitryLukyanov: well, I'm a bit surprised that we still have such difference. Can you please check what bytes are generated for this test with other zstd 3rd party nuget like with this one https://github.com/bp74/Zstandard.Net ?

adelinowona: I tested and it returns the original expected bytes but I think that is to be expected since other third party libraries are wrappers and not re-implementations. I think ZstdSharp.Port is the only C# port of zstd and it is even recommended on the zstd website: https://facebook.github.io/zstd/#:~:text=https%3A//github.com/oleg%2Dst/ZstdSharp. Everything listed there is said to support the zstd format specification and are interoperable: https://facebook.github.io/zstd/#:~:text=All%20versions%20support%20the%20Zstandard%20format%20specification%20and%20are%20therefore%20interoperable.

adelinowona:
>

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
```diff
@@ -37,7 +36,7 @@ public ZstandardCompressor(Optional<int> compressionLevel = default)
 
         public void Compress(Stream input, Stream output)
         {
-            using (var zstandardStream = new ZstandardStream(output, CompressionMode.Compress, _compressionLevel))
+            using (var zstandardStream = new CompressionStream(output, _compressionLevel))
             {
                 input.EfficientCopyTo(zstandardStream);
                 zstandardStream.Flush();
```
JamesKovacs, line 42: I did some investigation. The `01 00 00` is a data block denoting the last block of a frame. Without it the frame is considered corrupted though many non-strict implementations will still decode it. If you only call `Flush` or `Close` on the `CompressionStream`, this last block is not written. If you call both, it is.

```
zstandardStream.Flush();
zstandardStream.Close();
```

I would expect `Dispose` (which is implicitly called via the `using` block`) should flush and close the stream, but with this library it does not. At least it doesn't write out the final data block in the frame for some reason.

With both flush and close, the output bytes from managed zstd exactly matches the unmanaged version. You'll have to update the tests to account for the additional `01 00 00`.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
```diff
@@ -46,7 +45,7 @@ public void Compress(Stream input, Stream output)
 
         public void Decompress(Stream input, Stream output)
         {
-            using (var zstandardStream = new ZstandardStream(input, CompressionMode.Decompress))
+            using (var zstandardStream = new DecompressionStream(input))
             {
                 zstandardStream.CopyTo(output);
             }
```
JamesKovacs, line 51: Given that we had to flush/close the `CompressionStream`, I would suggest explicitly calling `Close` on the `DecompressionStream`. Since we are reading from the `DecompressionStream` there is nothing to flush and in fact calling `Flush` is invalid.

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
JamesKovacs: If there are differences in the test output between the original unmanaged zstd impl and the new managed zstd impl, let's make sure that we can account for them.

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
```diff
@@ -131,6 +161,79 @@ public void Zlib_should_throw_exception_if_the_level_is_out_of_range([Values(-2,
...
+        [InlineData(6, "40,181,47,253,0,88,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39,40,181,47,253,32,0,1,0,0")]
+        [InlineData(15, "40,181,47,253,0,96,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39,40,181,47,253,32,0,1,0,0")]
+        [InlineData(21, "40,181,47,253,0,128,109,1,0,84,2,97,98,99,100,101,102,103,104,105,106,107,108,109,110,111,112,113,114,115,116,117,118,119,120,121,122,48,49,50,51,52,53,54,55,56,57,32,1,0,53,132,170,39,40,181,47,253,32,0,1,0,0")]
+        public void Zstandard_compress_should_generate_expected_bytes_for_different_compression_levels(int compressionLevel, string expectedBytes)
```
JamesKovacs, line 169: The output bytes are identical regardless of compression level since zstd decides to write the bytes out raw (aka uncompressed). If we are going to assert that different compression levels output the expected bytes, we should choose an input that produces different outputs depending on compression level. I would also be good with asserting that compression level 6 produces the expected output - which happens to be the raw ASCII text with a magic number and some headers.

JamesKovacs (changes requested):
Home stretch here. See my comments, but let's also discuss over Slack.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
JamesKovacs: As discussed over Zoom, updating to Zstandard.Port 0.6.2 resolves this issue and only `Flush` is required.

src/MongoDB.Driver.Core/Core/Compression/ZstandardCompressor.cs
JamesKovacs: This shouldn't be necessary after updating to Zstandard.Port 0.6.2.

tests/MongoDB.Driver.Core.Tests/Core/Compression/CompressorsTests.cs
adelinowona: Removed this test as it showed itself as disposable given that the compressed format for the ZstdSharp.Port library is now fixed and although the output bytes are still different for different compression levels on input that is compressible, I have tested that other 3rd party wrappers of zstd can decompress the data compressed with the library without error and vice versa. The evergreen tests also show that we can communicate with server without errors.

DmitryLukyanov: so bytes are still different than before? I would prefer to not remove this test. At the very list it's a good prove for now and later that upgrading a new version of nuget actually works as expected

JamesKovacs: Testing whether the compressed bytes match some expected output is more about testing the compression library than testing our integration with it. We considered compressing and decompressing to validate that the input isn't modified, but that once again is testing whether the compression library does its job. The real test is whether Zstandard.Port can compress data that the server understands and whether it can decompress data from the server. That is already proven out via our Zstandard variants in EG. In summary, I'm OK with removing these tests because we verify this functionality already via our integration tests.

DmitryLukyanov: > the compression library does its job

which is not bad idea given that we're working with 3rd part library which we don't know how close 100% reliability it is. I would prefer to have a control over this, but if you're ok with this, I can rely on you

JamesKovacs: If there are meaningful inconsistencies that result in our driver not being able to decompress a response or compressing a request incorrectly, then our integration tests will catch it. So I'm good with leaving this to our integration testing via snappy and zstd variants rather than explicitly checking payloads.

JamesKovacs (approved):
LGTM

JamesKovacs:
Note that we do have quite a few broken tests, but they are related to server changes (e.g. `latest`) or flaky tests. We need to fix these but they don't have to hold up integration of the compressor work IMHO.

DmitryLukyanov (approved):
LGTM

adelinowona:
Squash-merged manually due to a minor merge conflict.

PR #854, opened 2022-07-28
CSHARP-4272: Avoids null reference exceptions when BsonClassMap::CreateInstance() returns null

MatteoSp:
Prevents BsonClassMapSerializer from throwing a NullReferenceException when BsonClassMap::CreateInstance() returns null. That may happen if one has configured a creator on the bson map and that creator fails silently returning nulls instead of object instances (think of DI containers).


BorisDog:
Thanks @MatteoSp 
Merged [here](https://github.com/mongodb/mongo-csharp-driver/commit/c91a16661d4079bd75a94b0c9e63aa809a835dbb)

PR #856, opened 2022-07-29
CSHARP-4259: Spec test "change stream resumes after StaleShardVersion" fails on server version > 6.0.

DmitryLukyanov:


BorisDog (approved):
LGTM

PR #859, opened 2022-07-31
CSHARP-4284: add Rewrap prose test.

DmitryLukyanov:


tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2149,6 +2177,24 @@ private EventCapturer CreateEventCapturer(string commandNameFilter = null)
             }
         }
 
+        private RewrapManyDataKeyResult RewrapManyDataKey(
```
BorisDog, line 2180: I'd suggest `ValueTask<RewrapManyDataKeyResult>` approach instead of ` .GetAwaiter().GetResult()`.
So we are don't mix TPL mechanism and regular sync execution.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1522,6 +1522,52 @@ void KmsProviderEndpointConfigurator(string kmsProviderName, Dictionary<string,
...
+            RequirePlatform
+                .Check()
+                .SkipWhen(() => kmsProviderFilter.Contains("gcp"), SupportedOperatingSystem.Linux, SupportedTargetFramework.NetStandard20)  // gcp is supported starting from netstandard2.1
+                .SkipWhen(() => kmsProviderFilter.Contains("gcp"), SupportedOperatingSystem.MacOS, SupportedTargetFramework.NetStandard20);
+            if (kmsProviderFilter.Contains("kmip"))
+            {
+                RequireEnvironment.Check().EnvironmentVariable("KMS_MOCK_SERVERS_ENABLED", isDefined: true);
+            }
+
+            string value = "test";
```
BorisDog, line 1546: minor: `const string value = "test" `?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
```diff
@@ -202,7 +228,7 @@ public static void ConfigureDefaultExtraOptions(Dictionary<string, object> extra
 
         public static IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> GetKmsProviders(string filter = null) =>
             __kmsProviders
-                .Where(kms => filter == null || kms.Key == filter)
+                .Where(kms => filter == null || filter.Split(new[] { KmsProviderFilterDelimiter }, StringSplitOptions.None).Contains(kms.Key))
```
BorisDog, line 231: maybe just `params string[] filters` ?

BorisDog (commented):
Minor comments.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
```diff
@@ -27,6 +26,7 @@ namespace MongoDB.Driver.Tests.Specifications.client_side_encryption
 {
     public static class EncryptionTestHelper
     {
+        private const string KmsProviderFilterDelimiter = ";";
```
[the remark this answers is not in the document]
DmitryLukyanov: given `System.String` TF limitations, it doesn't work well for `net472` unfortunately, in particular `char` separator doesn't work with `string.Join` :(

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
DmitryLukyanov: I wanted to use it too, but if we add `params` here, we will need do it up the stack too and in particular in methods like:

>  private DisposableMongoClient ConfigureClientEncrypted(
// other arguments
            Func<AutoEncryptionOptions, AutoEncryptionOptions> autoEncryptionOptionsConfigurator = null,
            params string[] filter)

which doesn't work well if we skip some arguments and use multi filter:
image

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: It's probably a good idea for new test files (or at least for more independent test classes), but in this file the rest of helper methods and tests itself are written differently which means that if we will need to call this helper method in other tests we will need updating at least test definition itself and possibly another used test helpers. So I would prefer to leave  this class written as "old style" to avoid unrelated to the current PR changes in other places

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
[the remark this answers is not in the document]
BorisDog: :(

BorisDog (approved):
LGTM

PR #863, opened 2022-08-09
CSHARP-4287: Only call Socket.IOControl on Windows

JamesKovacs:
Only call Socket.IOControl with IOControlCode.KeepAliveValues on Windows as non-Windows platforms will always throw PlatformNotSupportedExceptions.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
@@ -33,7 +33,6 @@ namespace MongoDB.Driver.Core.Connections
     {
         // fields
         private readonly TcpStreamSettings _settings;
-        private static readonly bool __supportsIOControl = OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows;
 
         // constructors
         public TcpStreamFactory()
@@ -260,30 +259,29 @@ namespace MongoDB.Driver.Core.Connections
 
             var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
 
-            if (__supportsIOControl)
+            try
             {
-                // Reviewing the .NET source, Socket.IOControl for IOControlCode.KeepAlivesValue will
-                // throw PlatformNotSupportedException on all platforms except for Windows.
-                // https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs#L1346
-                var keepAliveValues = new KeepAliveValues
+                if (OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows)
                 {
-                    OnOff = 1,
-                    KeepAliveTime = 120000, // 120 seconds in milliseconds
-                    KeepAliveInterval = 10000 // 10 seconds in milliseconds
-                };
-                socket.IOControl(IOControlCode.KeepAliveValues, keepAliveValues.ToBytes(), null);
-            }
-            else
-            {
-                // most platforms should support this call to SetSocketOption, but just in case call it in a try/catch also
-                try
+                    // Reviewing the .NET source, Socket.IOControl for IOControlCode.KeepAlivesValue will
+                    // throw PlatformNotSupportedException on all platforms except for Windows.
+                    // https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs#L1346
+                    var keepAliveValues = new KeepAliveValues
+                    {
+                        OnOff = 1,
+                        KeepAliveTime = 120000, // 120 seconds in milliseconds
+                        KeepAliveInterval = 10000 // 10 seconds in milliseconds
+                    };
+                    socket.IOControl(IOControlCode.KeepAliveValues, keepAliveValues.ToBytes(), null);
+                }
+                else
                 {
                     socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
                 }
-                catch (PlatformNotSupportedException)
-                {
-                    // ignore PlatformNotSupportedException
-                }
+            }
+            catch (PlatformNotSupportedException)
+            {
+                // ignore PlatformNotSupportedException
             }
 
             return socket;
--- a/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
@@ -33,6 +33,7 @@ namespace MongoDB.Driver.Core.Connections
     {
         // fields
         private readonly TcpStreamSettings _settings;
+        private static readonly bool __supportsIOControl = OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows;
 
         // constructors
         public TcpStreamFactory()
@@ -259,7 +260,7 @@ namespace MongoDB.Driver.Core.Connections
 
             var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
 
-            if (OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows)
+            if (__supportsIOControl)
             {
                 // Reviewing the .NET source, Socket.IOControl for IOControlCode.KeepAlivesValue will
                 // throw PlatformNotSupportedException on all platforms except for Windows.
--- a/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
@@ -259,9 +259,11 @@ namespace MongoDB.Driver.Core.Connections
 
             var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
 
-            // not all platforms support IOControl
-            try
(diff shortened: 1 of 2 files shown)


JamesKovacs:
Here's a patch build:
https://spruce.mongodb.com/version/62f1a0073066153bcb189396/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
```diff
@@ -259,9 +259,11 @@ private Socket CreateSocket(EndPoint endPoint)
 
             var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
 
-            // not all platforms support IOControl
-            try
+            if (OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows)
             {
+                // Reviewing the .NET source, Socket.IOControl for IOControlCode.KeepAlivesValue will
```
DmitryLukyanov, line 264: can we rely on undocumented behavior? As I understand this can be changed at further versions. Can you please explain what's the issue with the previous workflow? As I understand, both behaviors will be the same (for old and new versions) eventually

JamesKovacs: I agree with you that I don't relying on undocumented behaviour and that it could change in a future version. That said, this portion of the PAL hasn't changed in 5 years.

https://github.com/dotnet/runtime/blame/main/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs#L1346

The reason why is twofold:

First exceptions are a lot slower than if statements. A quick test shows that you can thrown ~48 exceptions per millisecond compared to evaluating 20K+ branches per millisecond. Given we are only catching a single exception per created connection and connections are being pooled, this is less of a concern than if we were in a tight loop. Still why add the extra delay when we know `Socket.IOControl` is going to fail on non-Windows systems?

The second reason is that some monitoring frameworks will report on exception behaviour including caught exceptions. Thus avoiding the expected `PlatformNotSupportedExceptions` on non-Windows platforms cleans up the monitoring stats, which I understand is the original ask in CSHARP-4287.

The overall behaviour before and after this proposed change should be identical, but connection creation will be slightly faster on non-Windows systems with less monitoring noise after the change.

Happy to discuss further if you still feel this is a bad idea.

src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
```diff
@@ -259,9 +259,11 @@ private Socket CreateSocket(EndPoint endPoint)
 
             var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
 
-            // not all platforms support IOControl
-            try
+            if (OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows)
```
BorisDog, line 262: Optional:
As we are in performance optimization area, can change the `OperatingSystemHelper` to evaluate `OperatingSystemHelper.CurrentOperatingSystem` once?
Looks that at least for Unix it involves string comparisons.

src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
DmitryLukyanov: I consider the previous logic after exception as fallback, which means using throwing exception as workflow is one of expected ways to do it. 

So my concerns:
1. Now we have different workflow. Previously any `PlatformNotSupportedException` was suppressed. We also additionally had a fallback mechanism that also always suppressed  `PlatformNotSupportedException`. Now windows doesn't suppress any exceptions.
2. There is a number of discussions like https://github.com/dotnet/runtime/issues/25555#issuecomment-375087989 or even approved PR like https://github.com/dotnet/corefx/pull/29963. For example, it looks like SetSocletOptions can be set for windows as well: https://github.com/dotnet/corefx/pull/29963/files#diff-d2470117f55a00b4e04f47d3411800d9c4b49df7619e21a0e1a8b48b6d93a4bdR52 
3. Can we rely say that `IOControl` always works only on windows? What about let's say mono on linux? Or some similar weird or legacy setup? What about old TFs?

JamesKovacs: Point #1...
The previous logic:
- On Windows, try `IOControl` - which would always succeed as all versions of Windows supports this. .NET Framework, .NET Core, and Mono all call `ioctl` Win32 API.
- On non-Windows, try IOControl, throw, try SetSocketOption for KeepAlive, which succeeds on all platforms (according to Stephen Toub https://github.com/dotnet/runtime/issues/25555#issuecomment-375087989). Even if it fails, we catch the `PlatformNotSupportedException` and ignore. 

The new logic:
- On Windows, set via `IOControl`. We don't fall back to `SetSocketOption`, but we don't need to because `IOControl` will always succeed on Windows.
- On non-Windows, skip `IOControl` (because we know it will fail) and use `SetSocketOption`. Although it should work on all platforms (according to Stephen Toub above), we catch `PlatformNotSupportedException` as before and ignore.

In summary the new logic is equivalent to the old for all practical purposes.

Point #2...
`SetSocketOption` can be used on Windows but all it does is enable keep alives, whereas `IOControl` enables keep alives and sets both the keepalives time and internal. So if `IOControl` works - which it will on Windows - it does everything that `SetSocketOption` does and more.

The more interesting point is other .NET impls like Mono. Mono partially implements `IOControl` on non-Windows operating systems:

```
/* See Socket.IOControl, WSAIoctl documentation in MSDN. The common options between UNIX
 * and Winsock are FIONREAD, FIONBIO and SIOCATMARK. Anything else will depend on the system
 * except SIO_KEEPALIVE_VALS which is properly handled on both windows and linux. */
```

See https://github.com/mono/mono/blob/main/mcs/class/System/System.Net.Sockets/Socket.cs#L2763-L2765.

We don't officially support Mono, nor do we test on it. So we shouldn't be basing our decision of whether to call `Socket.IOControl` due to partial support in Mono.

Point #3...
Mono partially implements `IOControl`. Some future .NET 7, 8, 9, ???, ... might support it on non-Windows operating systems. But historically it hasn't supported it in the last 5 years nor have any changes been made in this code in the last 5 years. If a future version does support it, we can always remove the platform check.

As for other TFs... .NET Framework only runs on Windows. So that's a moot point. We support back to .NET Core 2.1, which doesn't support `IOControl` on non-Windows. .NET 6 (latest) doesn't support `IOControl` on non-Windows. Mono partially supports it, but see point 2 above.

Does that sufficiently address your questions/concerns? Or are you still hesitant about this change?

src/MongoDB.Driver.Core/Core/Connections/TcpStreamFactory.cs
JamesKovacs: `OperatingSystemHelper.CurrentOperatingSystem` returns an enum. Internally it queries `RuntimeInformation.IsOsPlatform(OSPlatform)`, which appears to be a struct-based comparison. That said, it would make sense to cache the result since it cannot change in a running application.

src/MongoDB.Driver.Core/Core/Misc/OperatingSystemHelper.cs
```diff
@@ -20,37 +20,44 @@ namespace MongoDB.Driver.Core.Misc
 {
     internal enum OperatingSystemPlatform
     {
+        Unsupported = -1,
```
DmitryLukyanov, line 23: https://github.com/mongodb/mongo-csharp-driver/pull/863/files#diff-eaad3cedb006cb73527906b86198fe064b4b0666fc37ea60b2600dc469c54604R23 why do we need this option given that we will never able to actually use this value?

DmitryLukyanov: now if we write code like:
>if (WINDOWS)
{
}
else
{
  // here we may have unsupported. We have a guard for this, but it still looks a bit unexpected
}

JamesKovacs: Because it is a bad idea to throw an exception in a static constructor (or any constructor). So we need to remember that we detected an unsupported OS so we can throw in `CurrentOperatingSystem` as we always have previously.

DmitryLukyanov: agreed, but what about something like: 

    public static OperatingSystemPlatform CurrentOperatingSystem =>__cachedValue ?? (__cachedValue =  DetermineOperatingSystem)
    
 or thread safe way with using `Lazy<OperatingSystemPlatform>(DetermineOperatingSystem..` as we do in our helpers and other code

BorisDog: Not sure that we should be changing the PlatformNotSupportedException throwing behaviour in CurrentOperatingSystem, with Lazy<> approach.
It's not recommended to throw in properties, but I think this one is fine, and we should leave the same behavior.

DmitryLukyanov: >It's not recommended to throw in properties,

I don't see problems with lazy here. We just postpone arguments validating in ctor to a first usage given we don't want calculating/validating provided data in ctor :), so it's exactly what is lazy designed for.

JamesKovacs: Switched `OperatingSystemHelper` to use `Lazy<T>`.

DmitryLukyanov (approved):
LGTM

PR #867, opened 2022-08-16
CSHARP-4272: Validate return value of bsonClassMap.CreateInstance() in BsonClassMapSerializer

BorisDog:


src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
```diff
@@ -139,6 +139,11 @@ public TClass DeserializeClass(BsonDeserializationContext context)
                 // for mutable classes we deserialize the values directly into the result object
                 document = (TClass)_classMap.CreateInstance();
 
+                if (document == null)
+                {
+                    throw new InvalidOperationException($"{nameof(BsonClassMap)} did not provide an instance of {typeof(TClass).Name}.");
```
JamesKovacs, line 144: Reviewing the exceptions thrown in `BsonClassMapSerializer`, I think that `BsonSerializationException` is more appropriate than `InvalidOperationException`. `BsonSerializationException` is used when a creator map is unexpectedly null or similar. `InvalidOperationException` is thrown when we attempt to perform an invalid operation such as setting an Id when there is no setter.

JamesKovacs (changes requested):
Minor change with regard to the type of exception to throw.

src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
BorisDog: Done.

JamesKovacs (approved):
LGTM

PR #869, opened 2022-08-22
CSHARP-4293: Update with aggregation pipeline fails when used on a OfTypeMongoCollection (OfType<T>).

DmitryLukyanov:


src/MongoDB.Driver/FilteredMongoCollectionBase.cs
```diff
@@ -351,60 +351,62 @@ protected virtual UpdateDefinition<TDocument> AdjustUpdateDefinition(UpdateDefin
...
+                                ArrayFilters = updateOneModel.ArrayFilters,
+                                Collation = updateOneModel.Collation,
+                                Hint = updateOneModel.Hint,
+                                IsUpsert = updateOneModel.IsUpsert
+                            };
+                        default:
+                            throw new MongoInternalException("Request type is invalid.");
+                    }
+                })
+                .ToList();
```
JamesKovacs, line 409: It appears that the only change is the addition of `ToList()` on the end. This makes the diff confusing to review.

JamesKovacs (changes requested):
Just need to fix the diff as it looks like more was changed than actually was due to reformatting.

src/MongoDB.Driver/FilteredMongoCollectionBase.cs
DmitryLukyanov: done

JamesKovacs (approved):
LGTM

DmitryLukyanov:
This provided solution is incorrect, instead converting a pipeline argument to regular update document, we should emulate `$SetOnInsert` with aggregate pipeline stage. This was recommended by query team:

    rs:PRIMARY> db.mycoll.update({}, [{$set: {x: {$cond: [{$eq: [{$type: "$_id"}, "missing"]}, 1, 2]}}}], {upsert: true})
    WriteResult({
            "nMatched" : 0,
            "nUpserted" : 1,
            "nModified" : 0,
            "_id" : ObjectId("630561dceb783de551833de9")
    })
    rs:PRIMARY> db.mycoll.find()
    { "_id" : ObjectId("630561dceb783de551833de9"), "x" : 1 }
    rs:PRIMARY> db.mycoll.update({}, [{$set: {x: {$cond: [{$eq: [{$type: "$_id"}, "missing"]}, 1, 2]}}}], {upsert: true})
    WriteResult({ "nMatched" : 1, "nUpserted" : 0, "nModified" : 1 })
    rs:PRIMARY> db.mycoll.find()
    { "_id" : ObjectId("630561dceb783de551833de9"), "x" : 2 }

DmitryLukyanov:
ready for review

JamesKovacs (changes requested):
Overall looking good, but we have failures on 3.6 and 4.0. From the EG logs:
```
FAILURE: MongoDB.Driver.MongoCommandException : Command update failed: BSON field 'update.updates.u' is the wrong type 'array', expected type 'object'.
```

JamesKovacs (approved):
LGTM

PR #870, opened 2022-08-26
Update ObjectSerializer.cs

ajinkyagx:
When object has value converted to decimal by system, the TypeCode assigned to it becomes TypeCode.Decimal and the switch case where TypeCode.Object is handled is ignored which is the only place where decimal is being added for object .Because of that, right now when the object is added to a collection instead of being added as "$numberDecimal" it gets processed as an non primitive object, the SerializeDiscriminatedValue method is called and the object gets serialized as "_t":"System.Decimal", "_v": value.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
@@ -256,6 +256,10 @@ namespace MongoDB.Bson.Serialization.Serializers
                             case TypeCode.String:
                                 bsonWriter.WriteString((string)value);
                                 return;
+
+                            case TypeCode.Decimal:
+                                bsonWriter.WriteDecimal128((decimal)value);
+                                return;
                         }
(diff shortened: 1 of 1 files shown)


JamesKovacs:
Hi, @ajinkyagx,

Thank you for your PR. In order to consider accepting this PR, we require a few additional things:

1. File a CSHARP ticket in [our JIRA project](https://jira.mongodb.org/browse/CSHARP] clearly explaining the problem along with a repro. Be sure to link to your PR.
2. In your PR, add tests to `ObjectSerializerTests` demonstrating that the issue has been fixed.

We look forward to reviewing your updated PR.

Sincerely,
James

JamesKovacs (changes requested):
Please add tests and file a JIRA ticket.

JamesKovacs:
We have closed this as there has been no response on this PR. If this is still an issue for you, please file a JIRA ticket and add the tests as requested. Thank you.

PR #871, opened 2022-08-30
35% performance improvement for class deserialization

troepolik:
35% performance improvement for huge dto classes deserialization.
1) optimized create instance - without reflection
2) optimized set values to properties- without reflection, boxing, interface virtualization impact (for Deserialize Call)
3) optimized collection deserialization- in most cases we can just return prepared List, in some other cases - removed reflection
4) cache IsReadonly
5) dont call SetRequiredFields if required fields does not exists (flag of required fields exist was cached)
6) seporate deserialization of classes with default constructor and classes with creatorMap (to get rid of null checks)
7) add generic BsonMemberMap<TClass> for some optimizations above
8) fix one test , that used local culture in Parse method and failed
var expectedResult = decimal.Parse(expectedResultString, CultureInfo.InvariantCulture); -- pass culture here

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonClassMap.cs
+++ b/src/MongoDB.Bson/Serialization/BsonClassMap.cs
@@ -29,37 +29,42 @@ namespace MongoDB.Bson.Serialization
     /// <summary>
     /// Represents a mapping between a class and a BSON document.
     /// </summary>
-    public class BsonClassMap
+    public abstract class BsonClassMap
     {
+
+#pragma warning disable CA1051
+#pragma warning disable CA1707
+#pragma warning disable CS1591
+#pragma warning disable CS3008
         // private static fields
-        private readonly static Dictionary<Type, BsonClassMap> __classMaps = new Dictionary<Type, BsonClassMap>();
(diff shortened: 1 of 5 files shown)


DmitryLukyanov:
Thank you for your PR. In order to consider accepting this PR, we require filing a CSHARP ticket in [our JIRA project](https://jira.mongodb.org/browse/CSHARP) clearly explaining the problem along with a repro. Be sure to link to your PR.

troepolik:
@DmitryLukyanov I Tried, but get error:
image

troepolik:
Any way, I created test project here:
https://github.com/troepolik/testMongoDBPerformance

just set reference to BSON project, sure you set release mode and run, and press key to repeate several times to see statistics, then switch to my optimized project. 

Sugestion: press repeat in moment when CPU% is low.

BorisDog:
Thank you for your research and the PR.

Closing this PR as we are going to consider those great ideas as part of a general serialization improvement effort [CSHARP-3230](https://jira.mongodb.org/browse/CSHARP-3230).

PR #873, opened 2022-09-02
CSHARP-4189: Upgraded build scripts and Cake to support Apple M1 including fixing failing tests.

JamesKovacs:
Full EG run of all variants:
https://spruce.mongodb.com/version/631131a7d1fe0729bd02f85c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


tests/AstrolabeWorkloadExecutor/AstrolabeWorkloadExecutor.csproj
```diff
@@ -5,8 +5,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net472</TargetFrameworks>
-    <TargetFrameworks Condition="'$(IsWindows)'!='true'">netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net6.0;net472</TargetFrameworks>
+    <TargetFrameworks Condition="'$(IsWindows)'!='true'">netcoreapp2.1;netcoreapp3.1;net6.0</TargetFrameworks>
```
JamesKovacs, line 9: I only added .NET 6 to the test project compile targets so as not to increase testing time too much while still being able to run tests on M1. .NET 6 is an LTS release whereas .NET 5 is not. .NET 6 EOL is 2024-11-12 whereas .NET 5 has already reached EOL on 2022-05-10.

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

tests/MongoDB.Bson.Tests/Serialization/Options/RepresentationConverterTests.cs
```diff
@@ -235,8 +235,6 @@ public void TestAllowTruncationFalse()
         {
             var converter = new RepresentationConverter(false, false);
 
-            Assert.Throws<TruncationException>(() => converter.ToDouble(long.MaxValue));
-            Assert.Throws<TruncationException>(() => converter.ToDouble(ulong.MaxValue));
             Assert.Throws<TruncationException>(() => converter.ToInt16((double)1.5));
```
JamesKovacs, line 238: I removed the truncation tests for `long.MaxValue` and `ulong.MaxValue` because the results are processor dependent. Our truncation test relies on odd x64 behaviour, which isn't present on arm64. Notably:
```
public double ToDouble(long value)
{
    var doubleValue = (double)value;
    if (value != (long)doubleValue)
    {
        if (!_allowTruncation) { throw new TruncationException(); }
    }
    return doubleValue;
}
```
On x64, converting `long.MaxValue` to a `double` results in a value of `long.MaxValue+1` (because a double doesn't have enough bits of precision to accurately represent the value). Converting back to a `long` causes x64 to return `long.MinValue` thus throwing the `TruncationException`. On arm64, it clamps the value to `long.MaxValue` and thus doesn't throw the `TruncationException`. While x64 is correctly indicating truncation due to rounding, it is doing it for a very odd reason - because `double` values exceeding `long.MaxValue` are interpreted as `long.MinValue`.

The same behaviour happens with C/C++ and other languages. Rather than trying to work around this processor-dependent behaviour, I recommend not testing these quirky edge cases. Users shouldn't be depending on this behaviour regardless.

tests/MongoDB.Driver.Core.Tests/Core/Configuration/SslStreamSettingsTests.cs
```diff
@@ -63,7 +63,8 @@ public void constructor_with_checkCertificateRevocation_should_initialize_instan
         [Fact]
         public void constructor_with_clientCertificates_should_initialize_instance()
         {
-            var clientCertificates = new[] { new X509Certificate() };
+            var certificate = Array.Empty<byte>();
+            var clientCertificates = new[] { new X509Certificate(certificate) };
```
JamesKovacs, line 67: Parameterless `X509Certificate` constructor has been marked as obsolete. Instantiating with an empty byte array is the recommended replacement.

tests/MongoDB.Driver.Core.Tests/Core/Connections/TcpStreamFactoryTests.cs
```diff
@@ -207,7 +207,7 @@ public void Constructor_should_throw_an_ArgumentNullException_when_tcpStreamSett
                 stream = subject.CreateStream(endPoint, CancellationToken.None);
             }
 
-            var socketProperty = typeof(NetworkStream).GetProperty("Socket", BindingFlags.NonPublic | BindingFlags.Instance);
+            var socketProperty = typeof(NetworkStream).GetProperty("Socket", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
```
JamesKovacs, line 210: `NetworkStream.Socket` was changed from `protected` to `public` in .NET 5.

tests/MongoDB.Driver.Legacy.Tests/Jira/CSharp112Tests.cs
```diff
@@ -93,8 +93,7 @@ public void TestDeserializeDouble()
             {
                 0x7eeeeeeeeeeeeeee,
                 0xfeeeeeeeeeeeeeee,
-                Int64.MinValue + 1, // need some low order bits to see data loss
-                Int64.MaxValue
+                Int64.MinValue + 1  // need some low order bits to see data loss
```
JamesKovacs, line 96: Same processor-specific behaviour as earlier where we were inadvertently depending on x64 CPUs returning `Int64.MinValue` when trying to cast a value greater than `Int64.MaxValue` (which happened due to rounding error when converting the value to a double).

tests/MongoDB.Driver.Tests/OcspIntegrationTests.cs
```diff
@@ -70,7 +70,11 @@ public void MongoClientShouldRespectCertificateStatusAndTlsInsecure()
                 secureClientException.Should().BeOfType<TimeoutException>();
                 var message = secureClientException.Message;
                 // The exception will lack this message if the heartbeat doesn't fire
+#if NET6_0_OR_GREATER
+                message.Should().Contain("The remote certificate is invalid because of errors in the certificate chain");
+#else
                 message.Should().Contain("The remote certificate is invalid according to the validation procedure.");
+#endif
```
JamesKovacs, line 77: .NET 6 improved TLS error messages. Hence the need to assert on different error messages when the test assembly is compiled for `net6.0` TFM.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1267,6 +1267,14 @@ void RunTestCase(IMongoCollection<BsonDocument> explicitCollectionFromEncryptedC
 
             void AssertException(Exception exception)
             {
+#if NET6_0_OR_GREATER
+                const string invalidCertificateError = "The remote certificate was rejected by the provided RemoteCertificateValidationCallback.";
+                const string http404Error = "KMS response parser error with status 404, error: 'Unexpected extra HTTP content'";
+#else
+                const string invalidCertificateError = "The remote certificate is invalid according to the validation procedure.";
+                const string http404Error = "HTTP status=404";
+#endif
```
JamesKovacs, line 1276: Same as above. Improved TLS error messages starting with .NET 6.

tests/MongoDB.Driver.Tests/SslSettingsTests.cs
```diff
@@ -184,7 +184,7 @@ public void TestServerCertificateValidationCallback()
 
         private string GetTestCertificateFileName()
         {
-            var codeBase = typeof(SslSettingsTests).GetTypeInfo().Assembly.CodeBase;
+            var codeBase = typeof(SslSettingsTests).GetTypeInfo().Assembly.Location;
```
JamesKovacs, line 187: `CodeBase` is obsolete and shouldn't be used anymore. `Location` should be used instead. This is only in tests so I'm not concerned with end user impact when running on old TFMs. Our tests pass with this change for all TFMs we test on.

JamesKovacs (commented):
Comments explaining required changes.

tests/MongoDB.Bson.Tests/Exceptions/BsonExceptionTests.cs
```diff
@@ -31,9 +31,11 @@ public void Serialization_should_work()
             var formatter = new BinaryFormatter();
             using (var stream = new MemoryStream())
             {
+#pragma warning disable SYSLIB0011
```
DmitryLukyanov, line 34: I guess you added this #pragma manually. Consider format that is added by VS helper, this is how it looks like for `obselete` warnings: `#pragma warning disable CS0618 // Type or member is obsolete` (with a comment that is also generated by VS)

evergreen/evergreen.yml
```diff
@@ -1505,10 +1510,10 @@ buildvariants:
     - ubuntu1804-test
   tasks:
     - name: compile
-- name: macos1015-compile
-  display_name: "macOS 10.15 compile"
+- name: macos1100-compile
```
DmitryLukyanov, line 1513: nit: you can remove it :) (and other compile variants)

build.cake
```diff
@@ -57,18 +59,18 @@ Task("Restore")
     .Does(() =>
     {
         // disable parallel restore to work around apparent bugs in restore
-        var restoreSettings = new DotNetCoreRestoreSettings
+        var restoreSettings = new DotNetRestoreSettings
```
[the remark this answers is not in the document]
JamesKovacs: CakeBuild renamed `DotNetCore[TASK]` to `DotNet[TASK]` to align with the name change from .NET Core M.m to .NET M. It was simply a rename along with `[Obsolete] public class DotNetCore[TASK]` added for backwards compatibility.

build.cake
```diff
@@ -815,19 +828,31 @@ Task("TestsPackaging")
...
 
 Setup<BuildConfig>(
-    setupContext => 
+    setupContext =>
     {
+        var targetPlatform = RuntimeInformation.OSArchitecture switch
+        {
+            Architecture.Arm64 => "arm64",
+            Architecture.X64 => "x64",
+            var unknownArchitecture => throw new Exception($"Unknown CPU architecture: {unknownArchitecture}")
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

tests/MongoDB.Bson.Tests/Exceptions/BsonExceptionTests.cs
JamesKovacs: Yes, I added these manually. I'll add the reason as a comment.

evergreen/evergreen.yml
JamesKovacs: I'll pick up CSHARP-4230 next and remove all compile variants in a separate PR.

tests/MongoDB.Bson.Tests/Exceptions/BsonExceptionTests.cs
```diff
@@ -31,9 +31,11 @@ public void Serialization_should_work()
             var formatter = new BinaryFormatter();
             using (var stream = new MemoryStream())
             {
+#pragma warning disable SYSLIB0011 // BinaryFormatter serialization is obsolete
```
DmitryLukyanov, line 34: VS generates this message: `#pragma warning disable SYSLIB0011 // Type or member is obsolete` :)

build.cake
```diff
@@ -446,6 +458,7 @@ Task("TestCsfleWithMongocryptd")
 Task("TestCsfleWithMongocryptdNet472").IsDependentOn("TestCsfleWithMongocryptd");
 Task("TestCsfleWithMongocryptdNetStandard20").IsDependentOn("TestCsfleWithMongocryptd");
 Task("TestCsfleWithMongocryptdNetStandard21").IsDependentOn("TestCsfleWithMongocryptd");
+Task("TestCsfleWithMongocryptdNet60").IsDependentOn("TestCsfleWithMongocryptd");
```
DmitryLukyanov, line 461: given that we don't call it, why do we need this?

DmitryLukyanov (approved):
Some comments and then LGTM

build.cake
JamesKovacs: Consistency with other targets. In most cases we call the base target, which runs all TFMs in the test assembly.

JamesKovacs: Discussed with Dima over Slack. Looks like in most cases we run tests with `FRAMEWORK` set so only a specific TFM is run. This isn't done consistently. Notably the OCSP tests don't specify `FRAMEWORK` and run all TFMs in the test assemblies.

PR #874, opened 2022-09-06
CSHARP-4230: evergreen.yml cleanup

JamesKovacs:
Removed `compile` variants along with unused `compile` and `test` tasks. `publish-snapshot` called `evergreen/publish.sh`, which only echoed "Publishing .NET driver" and did nothing else.

I've also removed `compile` variants from running on EG patch requests.


DmitryLukyanov (approved):
LGTM, Assuming EG has no issues :)

PR #875, opened 2022-09-09
CSHARP-4316: Fixed translation of types similar to Nullable<T>

floatW0lf:


JamesKovacs (commented):
Thank you for your PR. I would like to apologize for missing the fact that you included the PR when you filed CSHARP-4316. I dove straight into repro'ing and fixing the issue without realizing you had already done the majority of the legwork. I've rebased my changes (mostly some additional tests and a bit of cleanup) on top of yours. I'll give credit where credit is due when I push your changes and mine. Thank you again for your contribution.

PR #876, opened 2022-09-12
CSHARP-4316: Ensure that LINQ3 only applies Nullable property translators to Nullables types and not properties with similar names.

JamesKovacs:
We were accidentally matching all properties named `Value` or `HasValue` even if they weren't on `Nullable<T>` types.

### DIFF
--- a/tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4316Tests.cs
+++ b/tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4316Tests.cs
@@ -1,0 +1,60 @@
+﻿using System.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
+{
+    public class CSharp4316Tests : Linq3IntegrationTest
+    {
+        [Fact]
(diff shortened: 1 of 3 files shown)


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4316Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+using FluentAssertions;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
+{
+    public class CSharp4316Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Property_named_Value_should_work_when_not_Nullable_type()
```
BorisDog, line 26: Would be good to test `HasValue` path as well.

JamesKovacs: Good suggestion. Added. I've also rebased my changes on top of the community-provided PR by Vladimir (PR #875).

BorisDog (approved):
LGTM

PR #877, opened 2022-09-14
CSHARP-4317: Added integration test demonstrating that CSHARP-4316 fix also resolved dictionary key/value projection issue.

JamesKovacs:
New integration test only.


BorisDog (approved):
LGTM

PR #879, opened 2022-09-15
CSHARP-4278: Unified runner for logging tests

BorisDog:


tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedEntityMap.cs
```diff
@@ -260,6 +286,21 @@ private void ThrowIfDisposed()
                 throw new ObjectDisposedException(nameof(UnifiedEntityMap));
             }
         }
+
+        public void AddEntities(UnifiedEntityMap unifiedEntityMap)
+        {
```
DmitryLukyanov, line 291: I've already implemented this logic in SDAM test runner, so I think we don't need this now, we can discuss it

DmitryLukyanov: https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/8db024528079dbd80244b4aaf2d174db8c12258a#diff-c111e924835b5b0fbf76667b327e1e7e9efe1a7463337a802e73a9f0fe1dd877R51

DmitryLukyanov: To simplify our work, we can defer refactoring I did to my next PR

src/MongoDB.Driver.Core/Core/Events/IEvent.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace MongoDB.Driver.Core.Events
```
DmitryLukyanov, line 1: copyright

src/MongoDB.Driver.Core/Core/Logging/LogCatergoryHelper.cs
```diff
@@ -13,13 +13,18 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Linq;
 using MongoDB.Driver.Core.Misc;
 
 namespace MongoDB.Driver.Core.Logging
 {
     internal static class LogCatergoryHelper
     {
+        private static readonly IDictionary<Type, string> __catergories = new ConcurrentDictionary<Type, string>();
```
DmitryLukyanov, line 26: filename contains typo

src/MongoDB.Driver.Core/Core/Logging/StructuredLogsTemplates.cs
```diff
@@ -13,24 +13,183 @@
...
+        public const string Reason = nameof(Reason);
+        public const string ServerHost = nameof(ServerHost);
+        public const string ServerPort = nameof(ServerPort);
+        public const string ServerConnectionId = nameof(ServerConnectionId);
+        public const string ServiceId = nameof(ServiceId);
+        public const string SharedLibraryVersion = nameof(SharedLibraryVersion);
+        public const string WaitQueueTimeoutMS = nameof(WaitQueueTimeoutMS);
+        public const string WaitQueueSize = nameof(WaitQueueSize);
+
+        public static readonly string ConnectionId_Message = $"{{{DriverConnectionId}}} {{{Message}}}";
```
DmitryLukyanov, line 55: can we use properties instead public fields?

src/MongoDB.Driver.Core/Core/Logging/StructuredLogsTemplatesCluster.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using Microsoft.Extensions.Logging;
+using MongoDB.Driver.Core.Events;
+
+namespace MongoDB.Driver.Core.Logging
+{
+    internal static partial class StructuredLogsTemplates
+    {
+        private static string[] __clusterCommonParams = new[]
+            {
```
DmitryLukyanov, line 24: I think better formatting is without tabs (same in other) like here https://github.com/mongodb/mongo-csharp-driver/pull/879/files#diff-8642a2a81535506783eea5811d0946bb45708157d8b1c7cade74e716e9605de8L322:

     private static string[] __clusterCommonParams = new[]
     {

src/MongoDB.Driver.Core/Core/Logging/StructuredLogsTemplatesCluster.cs
```diff
@@ -0,0 +1,94 @@
...
+                ClusterId,
+                Message,
+            };
+
+        private static string ClusterCommonParams(params string[] @params) => Concat(__clusterCommonParams, @params);
```
DmitryLukyanov, line 29: alphabetical order?

src/MongoDB.Driver.Core/Core/Misc/EndPointHelper.cs
```diff
@@ -157,6 +157,9 @@ public static EndPoint Parse(string value)
             return endPoint;
         }
 
+        internal static (string host, int port) GetHostAndPort(this EndPoint endPoint) =>
```
DmitryLukyanov, line 160: I think better format is with capital letter: 

    internal static (string Host, int Port)

tests/MongoDB.Driver.Core.TestHelpers/Logging/ILoggingService.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Microsoft.Extensions.Logging;
```
DmitryLukyanov, line 1: copyright

tests/MongoDB.Driver.Core.TestHelpers/Logging/LogEntry.cs
```diff
@@ -27,6 +27,7 @@ public sealed class LogEntry
         public DateTime Timestamp { get;  }
         public LogLevel LogLevel { get; }
         public string Category { get; }
+        public int ClusterId { get; }
```
DmitryLukyanov, line 30: should it really be here? We def can have logging before cluster creating.

tests/MongoDB.Driver.Core.TestHelpers/Logging/NullLoggingService.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.Extensions.Logging;
```
DmitryLukyanov, line 1: copyright

tests/MongoDB.Driver.Tests/Specifications/command-monitoring/CommandLoggingUnifiedTestRunner.cs
```diff
@@ -0,0 +1,65 @@
...
+using MongoDB.Driver.Core.TestHelpers.Logging;
+using MongoDB.Driver.Tests.UnifiedTestOperations;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MongoDB.Driver.Tests.Specifications.command_monitoring
+{
+    public class CommandLoggingUnifiedTestRunner : LoggableTestClass
+    {
+        // public constructors
```
DmitryLukyanov, line 28: it's not necessary

tests/MongoDB.Driver.Tests/Specifications/command-monitoring/CommandLoggingUnifiedTestRunner.cs
```diff
@@ -0,0 +1,65 @@
+﻿/* Copyright 2021-present MongoDB Inc.
```
DmitryLukyanov, line 1: 2010

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedLogMatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                "error" => LogLevel.Error,
+                "warning" => LogLevel.Warning,
+                "informational" or "notice" => LogLevel.Information,
+                "debug" => LogLevel.Debug,
+                "trace" => LogLevel.Trace,
+                _ => throw new ArgumentOutOfRangeException(nameof(logLevel), logLevel)
+            };
+
+        public void AssertLogsMatch(LogEntry[] actualLogs, BsonArray expectedLogs)
```
DmitryLukyanov, line 70: let's not mix static and non static logic

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
```diff
@@ -119,6 +120,10 @@ private void AssertValuesMatch(BsonValue actual, BsonValue expected, bool isRoot
                                 actualDocument.Names.Should().Contain(expectedName);
                                 AssertExpectedType(actualDocument[expectedName], operatorValue);
                                 continue;
+                            case "$$matchAsDocument":
+                                var parsedDocument = BsonDocument.Parse(actualDocument[expectedName].AsString);
+                                AssertValuesMatch(parsedDocument, operatorValue, false);
+                                continue;
```
DmitryLukyanov, line 126: can you explain this line?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
```diff
@@ -0,0 +1,65 @@
...
+        public UnifiedCloseClientOperation(IMongoClient client)
+        {
+            _client = client;
+        }
+
+        public OperationResult Execute(CancellationToken cancellationToken)
+        {
+            try
+            {
+                _client.Cluster.Dispose();
```
DmitryLukyanov, line 36: client here is Disposable, why not just client.Dispose?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
```diff
@@ -0,0 +1,65 @@
...
+                return OperationResult.FromResult(null);
```
DmitryLukyanov, line 37: OperationResult.Empty()

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly UnifiedEntityMap _entityMap;
+
+        public UnifiedCloseClientOperationBuilder(UnifiedEntityMap entityMap)
+        {
+            _entityMap = entityMap;
+        }
+
+        public UnifiedCloseClientOperation Build(string targetClientId, BsonDocument arguments)
+        {
+            var client = _entityMap.GetClient(targetClientId);
```
DmitryLukyanov, line 60: we usually add validation on arguments. In your case, validation that there is no any argument to make sure that we don't silently ignore it

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            return new(client);
```
DmitryLukyanov, line 62: oh, this syntax is interesting, but it looks like for bigger methods it can hide what actual class we create, but for such small ones this looks ok

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCreateEntitiesOperation.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public OperationResult Execute(CancellationToken cancellationToken)
+        {
+            try
+            {
+                var newEntityMap = new UnifiedEntityMapBuilder(null, _entityMap.LoggerFactory).Build(_entitiesArray);
+                _entityMap.AddEntities(newEntityMap);
+
+                return OperationResult.FromResult(null);
```
DmitryLukyanov, line 41: OperationResult.Empty

I've postponed the rest refactoring to unified sdam test runner PR

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCreateEntitiesOperation.cs
```diff
@@ -0,0 +1,69 @@
...
+        private readonly UnifiedEntityMap _entityMap;
+
+        public UnifiedCreateEntitiesOperationBuilder(UnifiedEntityMap entityMap)
+        {
+            _entityMap = entityMap;
+        }
+
+        public UnifiedCreateEntitiesOperation Build(BsonDocument arguments)
+        {
+            var entities = arguments["entities"].AsBsonArray;
```
DmitryLukyanov, line 64: we usually add validation on arguments. In your case, validation that there is no any argument to make sure that we don't silently ignore it

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -95,7 +98,7 @@ public void Run(JsonDrivenTestCase testCase)
 
             var schemaSemanticVersion = SemanticVersion.Parse(schemaVersion);
             if (schemaSemanticVersion < new SemanticVersion(1, 0, 0) ||
-                schemaSemanticVersion > new SemanticVersion(1, 8, 0))
+                schemaSemanticVersion > new SemanticVersion(1, 11, 0))
```
DmitryLukyanov, line 101: it will be at least `1,12`

src/MongoDB.Driver.Core/Core/Events/ClusterAddedServerEvent.cs
```diff
@@ -40,6 +40,8 @@ public ClusterAddedServerEvent(ServerId serverId, TimeSpan duration)
             _timestamp = DateTime.UtcNow;
         }
 
+        EventType IEvent.Type => EventType.ClusterAddedServer;
```
DmitryLukyanov, line 43: we put explicit implementation to the end (in this case to the end of events) with this comment: 

    // explicit interface implementations

tests/MongoDB.Driver.Tests/Specifications/change-streams/ChangeStreamUnifiedTestRunner.cs
```diff
@@ -37,7 +37,7 @@ public ChangeStreamUnifiedTestRunner(ITestOutputHelper testOutputHelper)
         [ClassData(typeof(TestCaseFactory))]
         public void Run(JsonDrivenTestCase testCase)
         {
-            using (var runner = new UnifiedTestRunner(LoggerFactory))
+            using (var runner = new UnifiedTestRunner(loggingService: this))
```
DmitryLukyanov, line 40: do you think it's really better? Is only loggerFactory needed for a test runner and the rest is garbage there?

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
```diff
@@ -319,7 +319,7 @@ private IServerMonitorFactory CreateServerMonitorFactory()
                 serverMonitorStreamFactory,
                 new EventAggregator(),
                 _clusterSettings.ServerApi,
-                _loggerFactory);
+                null);
```
DmitryLukyanov, line 322: I would recommend adding arguments name for cases when it's not clear from the context (like this one)

src/MongoDB.Driver.Core/Core/Events/IEvent.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.TestHelpers/Logging/LogEntry.cs
BorisDog: Yes for simplicity. It's commonly used parameter, so cashing it for optimization (if exists).

tests/MongoDB.Driver.Tests/Specifications/command-monitoring/CommandLoggingUnifiedTestRunner.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedLogMatcher.cs
BorisDog: Split to UnifiedLogHelper.cs.

tests/MongoDB.Driver.Core.TestHelpers/Logging/ILoggingService.cs
BorisDog: Done.

tests/MongoDB.Driver.Core.TestHelpers/Logging/NullLoggingService.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/Specifications/command-monitoring/CommandLoggingUnifiedTestRunner.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
BorisDog: `$$matchAsDocument` means that the string value has to be parsed as bson document and validated as a document.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCloseClientOperation.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCreateEntitiesOperation.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedCreateEntitiesOperation.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/Specifications/change-streams/ChangeStreamUnifiedTestRunner.cs
BorisDog: We need `loggingService.Logs` as well.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
DmitryLukyanov: I meant `continue`

src/MongoDB.Driver.Core/Core/Events/ClusterAddedServerEvent.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Configuration/ClusterBuilder.cs
BorisDog: Done.

JamesKovacs (changes requested):
Why not use `Dictionary<Type, Delegate>` for `EventsPublisher.__eventHandlers`? This would remove the need for the `EventType` enum, `IEvent` interface, explicit interface impl on dozens of event classes, ... It would also remove the possibility of a mismatch between the `Event` derived class and its `IEvent.Type` when we add more events in the future. What do you think?

BorisDog:
> 
@JamesKovacs Yes, that would make the code simpler. The downside is that we would be introducing two dictionary look-ups for each event and/or log. Beforehand each event handler was resolved once, so the idea was not to introduce unnecessary look-ups on the hot path. Furthermore I think lookup by type is not the most efficient one.

Introducing internal `IEvent` interface is actually an old idea [CSHARP-3433](https://jira.mongodb.org/browse/CSHARP-3433) that introduces more structure to event handling code. We do have a way to achieve the same performance without additional interface, but interface was the most straightforward approach.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Bson;
+using MongoDB.Driver.Core;
+using MongoDB.Driver.Tests.UnifiedTestOperations.Matchers;
+
+namespace MongoDB.Driver.Tests.UnifiedTestOperations
+{
+    public class UnifiedWaitForEventOperation : IUnifiedEntityTestOperation
```
BorisDog, line 27: @DmitryLukyanov  new op.

BorisDog (commented):
Rebased and incorporated recent spec log tests spec changes.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
DmitryLukyanov: It's also implemented in my commit :) 
I think it should be inherited from `IUnifiedSpecialTestOperation` because exception in this step means failed test. We will never assert this exception

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -0,0 +1,96 @@
...
+        private readonly UnifiedEntityMap _entityMap;
+
+        public UnifiedWaitForEventOperationBuilder(UnifiedEntityMap entityMap)
+        {
+            _entityMap = entityMap;
+        }
+
+        public UnifiedWaitForEventOperation Build(BsonDocument arguments)
+        {
+            var clientId = arguments["client"].AsString;
```
DmitryLukyanov, line 86: Validate arguments

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -43,33 +41,19 @@ public class UnifiedWaitForEventOperation : IUnifiedEntityTestOperation
...
 
-        public Task<OperationResult> ExecuteAsync(CancellationToken cancellationToken)
-        {
-            return Task.FromResult(Execute(cancellationToken));
-        }
-
-        private bool DoEventsMatch(IEnumerable<object> events) =>
-            events
+            bool DoEventsMatch(IEnumerable<object> events) =>
+                events
```
DmitryLukyanov, line 52: first tab formatting was better

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -86,8 +70,13 @@ public UnifiedWaitForEventOperation Build(BsonDocument arguments)
             var clientId = arguments["client"].AsString;
             var eventObject = arguments["event"].AsBsonDocument;
             var count = arguments["count"].AsInt32;
-            var eventCapturer = _entityMap.EventCapturers[clientId];
 
+            if (arguments.ElementCount != 3)
```
DmitryLukyanov, line 74: This will work, I would recommend using the same validation/assignment approach as in other operations: https://github.com/mongodb/mongo-csharp-driver/blob/master/tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedReplaceOneOperation.cs#L89-L117

DmitryLukyanov: at the very least if count will be missed we won't get nullreference exception

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
BorisDog: Fixed.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
```diff
@@ -210,8 +210,8 @@ public void Initialize()
             {
                 if (_poolState.TransitionState(State.Paused))
                 {
-                    _eventsLogger.LogAndPublish(new ConnectionPoolOpeningEvent(_serverId, _settings));
-                    _eventsLogger.LogAndPublish(new ConnectionPoolOpenedEvent(_serverId, _settings));
+                    _eventsLogger.LogAndPublish(new ConnectionPoolOpeningEvent(_serverId, _settings), _connectionFactory.ConnectionSettings);
```
DmitryLukyanov, line 213: q: why do we show `connection` settings in a event related to connection pool? Looks wrong to me at first glance

src/MongoDB.Driver.Core/Core/Events/EventsPublisher.cs
```diff
@@ -17,26 +17,27 @@ public EventsPublisher(IEventSubscriber eventSubscriber)
...
-            var eventHandler = _eventHandlers[(int)eventType];
+            var eventType = (int)@event.Type;
+            var eventHandler = _eventHandlers[eventType];
 
             if (eventHandler == null)
             {
-                _eventSubscriber.TryGetEventHandler<T>(out var registeredHandler);
+                _eventSubscriber.TryGetEventHandler<TEvent>(out var registeredHandler);
                 eventHandler = registeredHandler ?? __eventHandlerNull;
                 _eventHandlers[(int)eventType] = eventHandler;
```
DmitryLukyanov, line 35: nit: not needed cast, eventType is already `int`

src/MongoDB.Driver.Core/Core/Events/EventsPublisher.cs
```diff
@@ -17,26 +17,27 @@ public EventsPublisher(IEventSubscriber eventSubscriber)
...
-        public bool IsEventTracked<T>()
+        public bool IsEventTracked<TEvent>() where TEvent : struct, IEvent
         {
-            _eventSubscriber.TryGetEventHandler<T>(out var handler);
+            _eventSubscriber.TryGetEventHandler<TEvent>(out var handler);
             return handler != null;
         }
 
-        public void Publish<T>(EventType eventType, T @event)
+        public void Publish<TEvent>(TEvent @event) where TEvent : struct, IEvent
```
DmitryLukyanov, line 26: nit: I think it's enough to have only: `where TEvent : IEvent`, now we consider as `event` anything that is inherited from IEvent (which is a goal of this interface), no need to have more strict restriction

src/MongoDB.Driver.Core/Core/Events/IEvent.cs
DmitryLukyanov: not done

tests/MongoDB.Driver.Core.TestHelpers/Logging/NullLoggingService.cs
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace MongoDB.Driver.Core.TestHelpers.Logging
+{
+    public sealed class NullLoggingService : ILoggingService
+    {
+        public ILoggerFactory LoggerFactory { get; } = NullLoggerFactory.Instance;
```
DmitryLukyanov, line 23: we put static properties before ctor, but regular after

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenWaitForEventTest.cs
```diff
@@ -89,7 +89,8 @@ protected override void SetArgument(string name, BsonValue value)
         private void Wait()
         {
             var eventCondition = MapEventNameToCondition(_event);
-            Func<IEnumerable<object>, bool> eventsConditionWithFilterByCount = (events) => events.Count(eventCondition) >= _count;
+            Func<IEnumerable<object>, bool> eventsConditionWithFilterByCount = (events) =>
```
DmitryLukyanov, line 92: As far as I can see, this is the same logic as before, why did you change it?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedLogHelper.cs
```diff
@@ -0,0 +1,61 @@
+﻿/* Copyright 2020-present MongoDB Inc.
```
DmitryLukyanov, line 1: 2010

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            _eventCapturer.WaitForOrThrowIfTimeout(
+                DoEventsMatch,
+                TimeSpan.FromSeconds(10),
+                timeout => $"Waiting for {_count} {_eventCapturer} exceeded the timeout {timeout}.");
+
+            bool DoEventsMatch(IEnumerable<object> events) =>
+                events
+                    .Where(e => _unifiedEventMatcher.DoEventsMatch(e, _eventDocument))
+                    .Take(_count)
```
DmitryLukyanov, line 54: what's the point in this line? Should we just assert that number of events equals to expected?

DmitryLukyanov (approved):
Some minor questions and then LGRM

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
BorisDog: Events information is unchanged, for logs we need the `MaxIdleTime` parameter.

src/MongoDB.Driver.Core/Core/Events/EventsPublisher.cs
BorisDog: Good catch.

src/MongoDB.Driver.Core/Core/Events/IEvent.cs
BorisDog: Right, done now.

src/MongoDB.Driver.Core/Core/Events/EventsPublisher.cs
BorisDog: You are right, removed.
We do need it other places where where `new TEvent` is used.

src/MongoDB.Driver.Core/Core/Events/EventsPublisher.cs
BorisDog: Good catch.

tests/MongoDB.Driver.Core.TestHelpers/Logging/NullLoggingService.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenWaitForEventTest.cs
BorisDog: I caught this when implementing UnifiedWaitForEvent op.
This is optimization to avoid iterating over all existing events, as opposed to up first instance (in the common case)

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedLogHelper.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
BorisDog: Yes we should. The match is relatively complex op, so this avoids iterating over all events, only up to `_count`.
Also we need **at least** `_count`.

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
BorisDog: Changed.

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenWaitForEventTest.cs
DmitryLukyanov: Given that we have no tons of events and it's only in test, I think more clear logic (ie 1 step instead 2) is more important. But it will leave this up to you

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.cs
DmitryLukyanov: Changed?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
```diff
@@ -0,0 +1,85 @@
...
+            _eventDocument = eventDocument;
+            _unifiedEventMatcher = unifiedEventMatcher;
+        }
+
+        public void Execute()
+        {
+            _eventCapturer.WaitForOrThrowIfTimeout(
+                DoEventsMatch,
+                TimeSpan.FromSeconds(10),
+                timeout => $"Waiting for {_count} {_eventCapturer} exceeded the timeout {timeout}.");
```
DmitryLukyanov, line 49: I think it would be good to mention how many expected events we actually caught in the exception

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
DmitryLukyanov: Yes, we need `at least`, but I think it's helpful to provide how many expected events we actually caught (similar to what we do currently in master)

tests/MongoDB.Driver.Tests/JsonDrivenTests/JsonDrivenWaitForEventTest.cs
BorisDog: We do have lots events sometimes, and also lots of tests, so this all adds up.
Also the logic is quite self explanatory.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedWaitForEventOperation.cs
BorisDog: Done.

DmitryLukyanov (approved):
LGTM

src/MongoDB.Driver.Core/Core/Logging/EventsLogger.cs
```diff
@@ -16,501 +16,74 @@
...
-        public void LogAndPublish(ServerHeartbeatStartedEvent @event)
-        {
-            Logger?.LogInformation(Id_Message_ConnectionId, Id, "Heartbeat started", @event.ConnectionId);
 
-            _eventsPublisher?.Publish(EventType.ServerHeartbeatStarted, @event);
+            _eventsPublisher?.Publish(@event);
         }
 
-        public void LogAndPublish(ServerHeartbeatSucceededEvent @event)
+        public void LogAndPublish<TEvent, TArg>(TEvent @event, TArg arg) where TEvent : struct, IEvent
```
JamesKovacs, line 60: I like how this generic method cleaned up and consolidated the dozens of overloads we had previously. Very nice work!

JamesKovacs (approved):
LGTM. I especially liked the refactoring to use generic methods instead of dozens of overloads.

PR #881, opened 2022-09-16
CSHARP-4274: Added DisambiguatedPaths to ChangeStreamDocument.

JamesKovacs:
Changes should be self-explanatory. The remarks are copied from the spec update that added the new field.


BorisDog (approved):
LGTM

PR #885, opened 2022-09-28
CSHARP-4220: Add support for $documents stage.

rstam:


src/MongoDB.Driver/AggregateExpressionDefinition.cs
```diff
@@ -14,15 +14,29 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
 using MongoDB.Driver.Linq.Linq3Implementation.Translators;
 
 namespace MongoDB.Driver
 {
+    /// <summary>
+    /// Implemented by classes that have a DocumentSerializer property.
+    /// </summary>
+    /// <typeparam name="TDocument">The type of the document.</typeparam>
+    public interface IHasDocumentSerializer<TDocument>
```
rstam, line 32: Let's talk about this. Not sure this is the best approach.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -477,6 +478,53 @@ public static class PipelineStageDefinitionBuilder
...
+                throw new ArgumentException("Documents cannot be of type NoPipelineInput.", nameof(documents));
+            }
+
+            const string operatorName = "$documents";
+            var stage = new DelegatedPipelineStageDefinition<NoPipelineInput, TDocument>(
+                operatorName,
+                (s, sr, linqProvider) =>
+                {
+                    var renderedDocuments = documents.Render(NoPipelineInputSerializer.Instance, sr, linqProvider);
+                    documentSerializer ??= (documents as IHasDocumentSerializer<TDocument>)?.DocumentSerializer ?? sr.GetSerializer<TDocument>();
```
rstam, line 503: This is where I need the new `IHasDocumentSerializer` interface.

Let's talk about this.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -159,6 +159,22 @@ public static class IAggregateFluentExtensions
...
 
+        /// <summary>
+        /// Appends a $documents stage to the pipeline.
+        /// </summary>
+        /// <param name="aggregate">The aggregate.</param>
+        /// <param name="documents">The documents.</param>
+        /// <param name="documentSerializer">The document serializer.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        public static IAggregateFluent<TResult> Documents<TResult>(
+            this IAggregateFluent<NoPipelineInput> aggregate,
```
rstam, line 170: Note the use of `NoPipelineInput` here so that `Documents` can only be called as the first stage.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -477,6 +478,52 @@ public static class PipelineStageDefinitionBuilder
...
+        /// <typeparam name="TDocument">The type of the documents.</typeparam>
+        /// <param name="documents">The documents.</param>
+        /// <param name="documentSerializer">The document serializer.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<NoPipelineInput, TDocument> Documents<TDocument>(
+            IEnumerable<TDocument> documents,
+            IBsonSerializer<TDocument> documentSerializer = null)
+        {
+            var aggregateExpression = new DocumentsAggregateExpressionDefinition<TDocument>(documents, documentSerializer);
+            return Documents(aggregateExpression, documentSerializer);
```
rstam, line 524: By passing the `documentSerializer` to *both* the `DocumentsAggregateExpressionDefinition` and the `Documents` method we no longer need the `IHasDocumentSerializer` interface.

rstam (commented):
I'm pretty sure I found a way that the originally proposed `IHasDocumentSerializer` interface is not actually needed.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Linq3IntegrationTest.cs
```diff
@@ -58,7 +58,7 @@ protected void CreateCollection<TDocument>(IMongoCollection<TDocument> collectio
             CreateCollection(collection, (IEnumerable<TDocument>)documents); ;
         }
 
-        protected IMongoClient GetClient(LinqProvider linqProvider)
+        protected IMongoClient GetClient(LinqProvider linqProvider = LinqProvider.V3)
```
JamesKovacs, line 61: Given that `GetClient` and `GetCollection` can request either the LINQ2 or LINQ3 `MongoClient`, should this test base class still be called `Linq3IntegrationTest`? Should we rename it to `LinqIntegrationTest`?

JamesKovacs (changes requested):
`$documents` impl looks great. Only a minor question around whether `Linq3IntegrationTest` is still the best name or not given it can handle both LINQ2 and LINQ3 integration tests.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Linq3IntegrationTest.cs
JamesKovacs: Discussed over Slack...

Robert:
>This class is never used in testing LINQ2 itself.
>It is only used when testing LINQ3.
>At most it is sometimes used when testing against LINQ3 when we want to write a test to compare what the LINQ3 implementation does to what the LINQ2 implementation did. This doesn't happen often. Most of the time we only care what LINQ3 does.
>We're never going to use this class to write integration tests against LINQ2.
>So I think the name is still OK.

James:
>Valid points. I’m ok with leaving the name as-is. But I’ll have to keep in mind when writing LINQ3 tests that I can swap the provider for comparison purposes. It’s a useful trick when seeing how LINQ2 executed the same expression.

PR #888, opened 2022-09-30
CSHARP-4304: Handle conversions to an interface.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4304Tests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $project : { _outer : '$$ROOT', _id : 0 } }",
+                "{ $lookup : { from : 'Child', localField : '_outer.ChildId', foreignField : '_id', as : '_inner' } }",
+                "{ $unwind : '$_inner' }",
+                "{ $project : { _v : '$_inner', _id : 0 } }");
+
+            var results = queryable.ToList();
```
rstam, line 42: I think we should do a full blown integration test with data, but that's going to take awhile to figure out and I wanted to let you look at the work so far.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConvertExpressionToAggregationExpressionTranslator.cs
```diff
@@ -46,8 +46,20 @@ public static AggregationExpression Translate(TranslationContext context, UnaryE
...
 
-                var ast = AstExpression.Convert(operandTranslation.Ast, expressionType);
-                var serializer = context.KnownSerializersRegistry.GetSerializer(expression);
+                var ast = operandTranslation.Ast;
+                IBsonSerializer serializer;
+                if (expressionType.IsInterface)
+                {
+                    // when an expression is cast to an interface it's a no-op as far as we're concerned
+                    // and we can just use the serializer for the concrete type and members not defined in the interface will just be ignored
+                    serializer = operandTranslation.Serializer;
```
rstam, line 55: Your suggestion is a one liner (not counting the if statement)!

I like it when fixes are simple.

Hopefully simple is correct in this case. I think it is, see if you can think of any way this could be wrong in other scenarios.

JamesKovacs: Where this will almost certainly break down is explicit interface implementations. I would suggest we don't support this for now and evaluate the use case if/when a user files a ticket complaining.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4304Tests.cs
JamesKovacs: I agree that we should have a full-blown integration test. The one-liner approach discussed above is sufficient for now (and possibly forever). Let's get the integration test done so we can get this in.

JamesKovacs (changes requested):
I think the approach is a good one. Just need to turn the unit test into an integration test and we are good to go.

JamesKovacs (approved):
LGTM

PR #889, opened 2022-09-30
CSHARP-2646: DateTimeOffsetSerializer should be forgiving of the actual numeric types it encounters during deserialization.

rstam:


JamesKovacs (approved):
LGTM.

PR #890, opened 2022-10-03
CSHARP-4169: Move unified spec tests to  root folder.

JamesKovacs:
Relocated all spec tests to `$/specifications` to match the layout of `mongodb/specifications`. Added `LinkedBase` to our existing `EmbeddedResource` directives in the `csproj` files to keep the Solution Explorer layout the same. I made some minor updates to test runners to ensure the correct unified tests were sourced.


JamesKovacs:
NOTE: Although "Files changed" says 1,719, the vast majority are renames. Actual real changes were only in 4 `csproj` files and a handful of test runners.

tests/MongoDB.Driver.Core.Tests/Specifications/server-discovery-and-monitoring/ServerDiscoveryAndMonitoringTestRunner.cs
```diff
@@ -616,10 +619,9 @@ protected override IEnumerable<JsonDrivenTestCase> CreateTestCases(BsonDocument
 
             protected override bool ShouldReadJsonDocument(string path)
             {
-                var loadBalancerTestDirectory = $"{Path.PathSeparator}load-balanced{Path.PathSeparator}";
                 return base.ShouldReadJsonDocument(path) &&
                        !MonitoringPrefixes.Any(prefix => path.StartsWith(prefix)) &&
-                       !path.Contains(loadBalancerTestDirectory);  // load balancer support not yet implemented
+                       !path.StartsWith(IntegrationTestPrefix);
```
JamesKovacs, line 624: `loadBalancerTestDirectory` evaluated to `:load-balanced:` but the test paths were `.load_balanced.`. Thus we've been running the load balancer spec tests successfully for months now since the code to ignore them was incorrect. 🤷‍♂️  I decided to simply remove the condition since the load balancer tests are running successfully.

tests/MongoDB.Driver.Core.Tests/Specifications/server-discovery-and-monitoring/ServerDiscoveryAndMonitoringTestRunner.cs
```diff
@@ -606,6 +606,9 @@ private class TestCaseFactory : JsonDrivenTestCaseFactory
                 "MongoDB.Driver.Core.Tests.Specifications.server_discovery_and_monitoring.tests.legacy_hello.monitoring."
             };
 
+            // Integration tests are run by ServerDiscoveryAndMonitoringIntegrationTestRunner in MongoDB.Driver.Tests
```
DmitryLukyanov, line 609: did we discuss avoid placing unrelated to the current assembly tests?

tests/MongoDB.Driver.Tests/Specifications/read-write-concern/OperationTestRunner.cs
```diff
@@ -44,7 +44,7 @@ public void Run(JsonDrivenTestCase testCase)
         public class TestCaseFactory : JsonDrivenTestCaseFactory
         {
             // protected properties
-            protected override string PathPrefix => "MongoDB.Driver.Tests.Specifications.read_write_concern.tests.";
+            protected override string PathPrefix => "MongoDB.Driver.Tests.Specifications.read_write_concern.tests.operation.";
```
DmitryLukyanov, line 47: operation? Where it's came from?

tests/MongoDB.Driver.Core.Tests/Specifications/server-discovery-and-monitoring/ServerDiscoveryAndMonitoringTestRunner.cs
DmitryLukyanov: I guess it's done for simplicity, but then it also means that the initial goal is not reached. Maybe it's not a big deal though

JamesKovacs: I only included top-level specs necessary for each `csproj`. But there were a few examples - like SDAM and Read/Write Concern - where we had some subdirectories of `tests` that were only used in `Core` while others were used in `Driver`. Rather than getting really specific with `<EmbeddedResource/>` directives, I decided it was easier to omit certain test subdirectories via `PathPrefix` shenanigans as we do elsewhere.

tests/MongoDB.Driver.Tests/Specifications/read-write-concern/OperationTestRunner.cs
JamesKovacs: `read-write-concern` contains `connection-string`, `document`, and `operation` subdirectories. The first two are handled by `Core.Tests` whereas the last one is handled by `Driver.Tests`. Rather than including all and then omitting `connection-string` and `document` it was easier to only run the spec tests in the `operation` subdirectory.

DmitryLukyanov: I see, thanks

rstam:
Note regarding:

> Added LinkedBase to our existing EmbeddedResource directives in the csproj files to keep the Solution Explorer layout the same

My interpretation is slightly different. LinkBase PRIMARILY affects the path of the embedded resources. It results in the path of the embedded resources being the same as they were before the JSON files were moved (in the file system), so that the test runners can continue to find them without requiring changes to the test runner.

What an IDE does with this information varies from IDE to IDE.

In Visual Studio 2022 it appears to continue to display the JSON test files in the same locations as before (deep inside the individual test projects). In Rider it displays them in the root of the Solution Explorer (according to James).

rstam (approved):
LGTM

JamesKovacs:
>In Visual Studio 2022 it appears to continue to display the JSON test files in the same locations as before (deep inside the individual test projects). In Rider it displays them in the root of the Solution Explorer (according to James).

To clarify, Rider also displays the JSON files in their original locations just as VS2022 does.

DmitryLukyanov (approved):
LGTM

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
```diff
@@ -103,4 +103,10 @@
     <Folder Include="Specifications\transactions-convenient-api\tests\" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\..\specifications\atlas-data-lake-testing\AtlasDataLakeTestRunner.cs">
```
DmitryLukyanov, line 107: can you please clarify what did you change it?

JamesKovacs: I diff'd our `$/specifications` directory with the `mongodb/specifications` repo and noticed two discrepancies:
1. In ours, the spec was called `atlas-data-lake` but in the spec repo it is called `atlas-data-lake-testing`.
2. In ours, the spec was called `bson` but in the spec repo it is called `bson-decimal128`.

I renamed our two directories to match the directory names in the spec repo. None of the tests were modified or updated. The only change was in the runners to account for the new directory name.

JamesKovacs: Oh! Good catch. That `cs` file should not have been moved, only the JSON files. Looks like I fat-fingered something. I'll fix it.

BorisDog (approved):
LGTM.
Nice usage of `LinkedBase`. Excited to see this done!

JamesKovacs:
NOTE: Force-pushed after rebase and squash. Verified that total tests before and after the re-organization are the same.

PR #891, opened 2022-10-04
CSHARP-4279: Logs truncation

BorisDog:
Part of [DRIVERS-1204](https://jira.mongodb.org/browse/DRIVERS-1204)

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
@@ -1,0 +1,101 @@
+/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace MongoDB.Driver.Core.Configuration
+{
+    /// <summary>
+    /// Represents the settings for logging.
+    /// </summary>
+    public sealed class LoggingSettings : IEquatable<LoggingSettings>
+    {
+        /// <summary>
+        /// Gets the logger factory.
+        /// </summary>
+        [CLSCompliant(false)]
+        public ILoggerFactory LoggerFactory { get; }
+
+        /// <summary>
+        /// Gets the maximum document size in chars
+        /// </summary>
+        public int MaxDocumentSize { get; }
+
+        // constructors
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LoggingSettings"/> class.
+        /// </summary>
+        /// <param name="loggerFactory">The logger factory.</param>
+        /// <param name="maxDocumentSize">The maximum document size in chars.</param>
+        [CLSCompliant(false)]
+        public LoggingSettings(
+            ILoggerFactory loggerFactory = default,
+            Optional<int> maxDocumentSize = default)
+        {
+            LoggerFactory = loggerFactory;
+            MaxDocumentSize = maxDocumentSize.WithDefault(MongoInternalDefaults.Logging.MaxDocumentSize);
+        }
+
+        // public operators
+        /// <summary>
+        /// Determines whether two <see cref="LoggingSettings"/> instances are equal.
+        /// </summary>
+        /// <param name="lhs">The LHS.</param>
+        /// <param name="rhs">The RHS.</param>
+        /// <returns>
+        ///   <c>true</c> if the left hand side is equal to the right hand side; otherwise, <c>false</c>.
+        /// </returns>
+        public static bool operator ==(LoggingSettings lhs, LoggingSettings rhs)
+        {
+            return object.Equals(lhs, rhs); // handles lhs == null correctly
+        }
+
+        /// <summary>
+        /// Determines whether two <see cref="LoggingSettings"/> instances are not equal.
+        /// </summary>
+        /// <param name="lhs">The LHS.</param>
+        /// <param name="rhs">The RHS.</param>
+        /// <returns>
+        ///   <c>true</c> if the left hand side is not equal to the right hand side; otherwise, <c>false</c>.
+        /// </returns>
+        public static bool operator !=(LoggingSettings lhs, LoggingSettings rhs)
+        {
+            return !(lhs == rhs);
+        }
+
+        // public methods
+        /// <summary>
+        /// Determines whether the specified <see cref="LoggingSettings" /> is equal to this instance.
+        /// </summary>
+        /// <param name="rhs">The <see cref="LoggingSettings" /> to compare with this instance.</param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="LoggingSettings" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
+        public bool Equals(LoggingSettings rhs)
+        {
+            return
+                rhs != null &&
+                LoggerFactory == rhs.LoggerFactory &&
+                MaxDocumentSize == rhs.MaxDocumentSize;
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) => Equals(obj as LoggingSettings);
+
+        /// <inheritdoc/>
+        public override int GetHashCode() => base.GetHashCode();
+    }
+}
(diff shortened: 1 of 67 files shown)


rstam:
As I start to review this PR I'm finding a number of naming issues that I would have raised earlier if I had been part of earlier PRs.

As a general rule class names should not be plural (e.g., `string` not `strings`, `Guid` not `Guids`) with a few exceptions like `XyzSettings`.

Accordingly, I would like to see the following names changed in the new logging code:

EventsLogger => EventLogger
StructuredLogsTemplates => StructuredLogTemplate
LoggingFactory => LoggerFactory (because it returns an `ILoggerFactory` not an `ILoggingFactory`)
EventsPublisher => EventPublisher
EventsLogsFormattingOptions => EventLogFormattingOptions
Extentions => Extensions (spelling)
LogsTemplateProvider => LogTemplateProvider

It would probably make sense to make these changes in a separate PR addressing only the name changes and then rebase this PR on master after pushing those changes to master.

rstam (changes requested):
See comment about name changes that should be done as a separate PR before this one/

rstam:
Boris agrees the renamings should be in a separate PR but he wants to do it after this one, so I'm resuming review of this ignoring the naming issues for now.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+* limitations under the License.
+*/
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace MongoDB.Driver.Core.Configuration
+{
+    /// <summary>
+    /// Represents the settings for using SSL.
```
rstam, line 22: /// Represents the settings for logging.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+    public sealed class LoggingSettings : IEquatable<LoggingSettings>
+    {
+        /// <summary>
+        /// Gets the logger factory.
+        /// </summary>
+        [CLSCompliant(false)]
+        public ILoggerFactory LoggerFactory { get; }
+
+        /// <summary>
+        /// Gets the maximal document size in chars
```
rstam, line 33: /// Gets the maximum document size in chars

"maximal" doesn't sound quite right to me.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        public int MaxDocumentSize { get; }
+
+        // constructors
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LoggingSettings"/> class.
+        /// </summary>
+        /// <param name="loggerFactory">The logging factory.</param>
+        /// <param name="maxDocumentSize">The maximal document size in chars.</param>
```
rstam, line 42: /// <param name="loggerFactory">The logger factory.</param>
        /// <param name="maxDocumentSize">The maximum document size in chars.</param>

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        public bool Equals(LoggingSettings rhs)
+        {
+            if (object.ReferenceEquals(rhs, null))
+            {
+                return false;
+            }
+
+            return
+                LoggerFactory == rhs.LoggerFactory &&
+                MaxDocumentSize == rhs.MaxDocumentSize;
```
rstam, line 96: Optionally could be simplified to:

            return
                rhs != null && // or equivalently: !object.ReferenceEquals(rhs, null)
                LoggerFactory == rhs.LoggerFactory &&
                MaxDocumentSize == rhs.MaxDocumentSize;

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        /// <summary>
+        /// Determines whether the specified <see cref="System.Object" /> is equal to this instance.
+        /// </summary>
+        /// <param name="obj">The <see cref="System.Object" /> to compare with this instance.</param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
+        /// </returns>
```
rstam, line 105: Instead of repeating the doc comments from the base class just use:

        /// <inheritdoc/>

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        public override bool Equals(object obj) => Equals(obj as LoggingSettings);
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance.
+        /// </returns>
```
rstam, line 113: /// <inheritdoc/>

src/MongoDB.Driver.Core/Core/Logging/LoggerExtentions.cs
```diff
@@ -27,6 +27,10 @@ public static EventsLogger<T> ToEventsLogger<T>(this ILogger<T> logger, IEventSu
             where T : LogCategories.EventCategory =>
             new EventsLogger<T>(eventSubscriber, logger);
 
+        public static EventsLogger<T> ToEventsLogger<T>(this IEventSubscriber eventSubscriber)
+            where T : LogCategories.EventCategory =>
+            new EventsLogger<T>(eventSubscriber, null);
```
rstam, line 32: Consider:

        public EventsLogsFormattingOptions(int maxCommandDocumentSize)
        {
            MaxDocumentSize = Ensure.IsGreaterThanOrEqualToZero(maxCommandDocumentSize, nameof(maxCommandDocumentSize));
        }

It seems weird to use `=>` when there is no value being returned.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Thanks, fixed.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Good point, done.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Configuration/LoggingSettings.cs
BorisDog: Done.

src/MongoDB.Driver.Core/Core/Logging/LoggerExtentions.cs
BorisDog: Done.

rstam (approved):
LGTM

src/MongoDB.Driver.Core/Core/Logging/EventsLogger.cs
```diff
@@ -39,19 +43,22 @@ public EventsLogger(IEventSubscriber eventSubscriber, ILogger<T> logger)
         private LogLevel GetEventVerbosity<TEvent>() where TEvent : struct, IEvent =>
             StructuredLogsTemplates.GetTemplateProvider(new TEvent().Type).LogLevel;
 
-        public void LogAndPublish<TEvent>(TEvent @event) where TEvent : struct, IEvent
-            => LogAndPublish(null, @event);
+        public void LogAndPublish<TEvent>(TEvent @event, bool skipLog = false) where TEvent : struct, IEvent
+          => LogAndPublish(null, @event, skipLog);
```
rstam, line 47: It's called `skipLogging` everywhere else

rstam: Not all overloads of `LogAndPublish` take the new `skipLogging` parameter.

I assume that's intentional?

rstam (changes requested):
Probably LGTM but I had some minor questions.

src/MongoDB.Driver.Core/Core/Logging/EventsLogger.cs
BorisDog: Renamed.
The other overload is not used in places with where skipping logging is required. So for simplicity added only where it is necessary.
Side note: I believe that this parameter will be removed in [CSHARP-3823](https://jira.mongodb.org/browse/CSHARP-3823)

PR #896, opened 2022-10-07
CSHARP-4322: Run legacy Client Side Encryption tests on serverless.

DmitryLukyanov:


JamesKovacs (changes requested):
This seems like a good solution, but we should get Andreas' changes pulled into the main branch of `mongodb-labs` rather than depending on an arbitrary branch in his fork. I assume that you'll make these changes before merging with `master`?

DmitryLukyanov:
> This seems like a good solution, but we should get Andreas' changes pulled into the main branch of `mongodb-labs` rather than depending on an arbitrary branch in his fork. I assume that you'll make these changes before merging with `master`?

yes, I also added some comments to his PR, so his PR is not 100% ready yet

JamesKovacs (approved):
LGTM

PR #897, opened 2022-10-11
CSHARP-4331: Make LINQ3 the default LinqProvider.

rstam:
All tests are passing just by finding everywhere that was still defaulting to LINQ2 and changing the default to LINQ3. For the most part the failing tests were failing because they were asserting on LINQ2 expected results (like filters and stages).

Before we flip the switch though we might should refactor those tests to test against both LINQ2 and LINQ3, or perhaps only LINQ3 where appropriate.

### DIFF
--- a/tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
+++ b/tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
@@ -34,24 +34,24 @@ namespace MongoDB.Driver.Tests
     public static class DriverTestConfiguration
     {
         // private static fields
-        private static Lazy<MongoClient> __client;
         private static Lazy<MongoClient> __clientWithMultipleShardRouters;
         private static CollectionNamespace __collectionNamespace;
         private static DatabaseNamespace __databaseNamespace;
         private static Lazy<IReadOnlyList<IMongoClient>> __directClientsToShardRouters;
+        private static Lazy<MongoClient> __linq2Client;
         private static Lazy<MongoClient> __linq3Client;
 
         // static constructor
         static DriverTestConfiguration()
         {
-            __client = new Lazy<MongoClient>(() => new MongoClient(GetClientSettings()), true);
+            __linq2Client = new Lazy<MongoClient>(CreateLinq2Client, isThreadSafe: true);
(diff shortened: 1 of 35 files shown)


rstam:
@JamesKovacs this PR can wait until you get back since it's not going into 2.18.0.

rstam:
See also:

https://jira.mongodb.org/browse/CSHARP-4358

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
```diff
@@ -50,6 +51,7 @@ static DriverTestConfiguration()
                 () => CreateDirectClientsToHostsInConnectionString(CoreTestConfiguration.ConnectionStringWithMultipleShardRouters).ToList().AsReadOnly(),
                 isThreadSafe: true);
             __collectionNamespace = new CollectionNamespace(__databaseNamespace, "testcollection");
+            __linq2Client = new Lazy<MongoClient>(CreateLinq2Client, isThreadSafe: true);
```
JamesKovacs, line 54: Do we need `Client`, `Linq2Client`, and `Linq3Client`? Even if we keep all three to avoid a cascading change to hundreds of tests, shouldn't `__linq3Client = __client` rather than creating two separate instances on the LINQ3 `MongoClient`?

tests/MongoDB.Driver.Tests/AggregateFluentBucketTests.cs
```diff
@@ -39,7 +40,7 @@ public class AggregateFluentBucketTests
         static AggregateFluentBucketTests()
         {
             var databaseNamespace = DriverTestConfiguration.DatabaseNamespace;
-            __database = DriverTestConfiguration.Client.GetDatabase(databaseNamespace.DatabaseName);
+            __database = DriverTestConfiguration.Linq2Client.GetDatabase(databaseNamespace.DatabaseName);
```
JamesKovacs, line 43: Shouldn't this now be tested against LINQ3? Same with all the following tests. If our default is LINQ3, our tests should run against LINQ3. Thoughts?

JamesKovacs (changes requested):
Questions inline.

PR #900, opened 2022-10-13
CSHARP-4362: TestEmptySortedSet test failing on net6.

rstam:


src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
```diff
@@ -36,7 +36,8 @@ static TypeNameDiscriminator()
             {
                 typeof(object).GetTypeInfo().Assembly, // mscorlib
                 typeof(Queue<>).GetTypeInfo().Assembly, // System
-                typeof(HashSet<>).GetTypeInfo().Assembly // System.Core
+                typeof(HashSet<>).GetTypeInfo().Assembly, // System.Core
+                typeof(SortedSet<>).GetTypeInfo().Assembly
```
DmitryLukyanov, line 40: can you please add the same comment as in above lines? To better understand expected value

DmitryLukyanov (approved):
Minor comment and then LGTM

src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
rstam: I can't, because the location depends on which TF you are running on.

The above comments are actually wrong. They were only correct for .NET Framework.

DmitryLukyanov: at least we can say something like: `.net 6: System.Collection` and update it when we will work on it later? I will leave this up to you

DmitryLukyanov (approved):
LGTM

BorisDog (approved):
LGTM

DmitryLukyanov (approved):
LGTM

src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
```diff
@@ -27,17 +27,25 @@ namespace MongoDB.Bson.Serialization
     public static class TypeNameDiscriminator
     {
         // private static fields
-        private static Assembly[] __wellKnownAssemblies;
+        private static HashSet<Assembly> __wellKnownAssemblies = new HashSet<Assembly>();
```
BorisDog, line 30: can be made readonly?

BorisDog (approved):
LGTM (1 comment for your consideration)

src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
rstam: Will do.

It's misleading of course... while the field is readonly, the hash set is not!

PR #904, opened 2022-10-14
CSHARP-4365: Fix Sandcastle warning regarding Microsoft.Extensions.Logging.Abstraction assembly.

rstam:


BorisDog (approved):
LGTM (assuming it just works)

build.cake
```diff
@@ -118,6 +118,7 @@ Task("BuildArtifacts")
                     fileNames.Add("DnsClient.dll");
                     fileNames.Add("MongoDB.Libmongocrypt.dll");
                     fileNames.Add("SharpCompress.dll");
+                    fileNames.Add("Microsoft.Extensions.Logging.Abstractions.dll");
```
DmitryLukyanov, line 121: I thought about the same solution, but what I don't understand why it started to happen for microsoft library? Maybe it doesn't matter at this point..

rstam: Isn't this a rather new dependency added for the logging work Boris did?

DmitryLukyanov (approved):
LGTM. NOTE: I don't understand where we need this change and where not. Previously I thought that we need this change for 3rd party libs that may be not fully correct with some "definitions". But I would expect microsoft library should satisfy any "assumptions" about nugets, so it's a mystery for me

rstam:
> LGTM. NOTE: I don't understand where we need this change and where not. Previously I thought that we need this change for 3rd party libs that may be not fully correct with some "definitions". But I would expect microsoft library should satisfy any "assumptions" about nugets, so it's a mystery for me

It's not part of any target framework. It's a Nuget package.

build.cake
DmitryLukyanov: yes, but it's not the only new dependency, we have other ones like `ZstdSharp.Port` or `Microsoft.SourceLink.GitHub`, so it's unclear what is special in nugets from this list

rstam: Maybe it's because nothing from those dependencies appear in any public APIs.

DmitryLukyanov:
> > LGTM. NOTE: I don't understand where we need this change and where not. Previously I thought that we need this change for 3rd party libs that may be not fully correct with some "definitions". But I would expect microsoft library should satisfy any "assumptions" about nugets, so it's a mystery for me
> 
> It's not part of any target framework. It's a Nuget package.

can you clarify? I didn't say about TFs

build.cake
DmitryLukyanov: That's good point, maybe you're right!

rstam:
> can you clarify? I didn't say about TFs

I think Sandcastle knows about all the public types in net472. Anything else has to come from a DLL in this list.

build.cake
DmitryLukyanov: but same story with `SharpCompress.dll` :)

PR #907, opened 2022-10-18
CSHARP-4261: Verify that sample code works when custom serializer implements IRepresentationConfigurable.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4261Tests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public string Name { get; set; }
+        }
+
+        [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+        public class AsObjectIdAttribute : BsonSerializerAttribute
+        {
+            public AsObjectIdAttribute() : base(typeof(ObjectIdSerializer)) { }
+
+            private class ObjectIdSerializer : SerializerBase<string>, IRepresentationConfigurable
```
rstam, line 72: I modified their sample code to implement `IRepresentationConfigurable` and confirmed these tests pass now.

I'm going to explain this to the user in the JIRA ticket.

rstam:
There might be a way to make this work *without* requiring the custom serializer to implement `IRepresentationConfigurable` but that might have problems of its own. Let's discuss when @JamesKovacs gets back.

JamesKovacs (commented):
Looks reasonable. I don't think we need to include this test case in our suite as it is simply demonstrating a technique rather than reproducing/fixing a bug.

rstam:
Works as designed.

We've decided not to push repro to master.

PR #909, opened 2022-10-18
CSHARP-3144: Verify that supplied repro works with LINQ3.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
```diff
@@ -0,0 +1,110 @@
...
+        public void Where_with_Contains_should_work()
+        {
+            var collection = CreateCollection();
+
+            var queryable = collection
+                .AsQueryable()
+                .Where(x => x.Items.Select(e => e.GoodId).Contains(2));
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { $expr : { $in : [2, '$Items.GoodId'] } } }");
```
rstam, line 36: Please verify that you agree this translation is correct.

Is there a way to translate this more simply in filter language (without using `$expr`)?

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
```diff
@@ -0,0 +1,110 @@
...
+        public void Suggested_workaround_should_work()
+        {
+            var collection = CreateCollection();
+
+            var queryable = collection
+                .AsQueryable()
+                .Where(x => x.Items.Select(e => e.GoodId).Any(e => e == 2));
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { $expr : { $anyElementTrue : { $map : { input : '$Items.GoodId', as : 'e', in : { $eq : ['$$e', 2] } } } } } }");
```
rstam, line 52: Please verify that you agree this translation is correct.

Is there a way to translate this more simply in filter language (without using `$expr`)?

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
JamesKovacs: This is one potential correct translation, but it does require `$expr`. Another possible translation that doesn’t require `$expr` is:

```
[{$match: {"Items.GoodId": 2}}]
```

Quick testing shows that both produce the same results.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
JamesKovacs: The same MQL will work here too because `[{$match: {"Items.GoodId": 2}}]` evaluates to true if any array element has `GoodId: 2`.

JamesKovacs (changes requested):
While the generated MQL produces the correct results, we could use the simplified MQL noted below. The main advantage of the filter language formulation is that since `$expr` isn't used then an index can satisfy the query. See [SERVER-32549](https://jira.mongodb.org/browse/SERVER-32549).

I would suggest investing a few hours to determine how much effort would be required to produce the filter language MQL and whether there would be edge cases that would break this MQL but not the `$expr` formulation.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
rstam: Latest commit implements this suggestion.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3144Tests.cs
rstam: Latest commit also handles this case!

JamesKovacs (approved):
LGTM

I was concerned that we could potentially break an existing query that renders using `$expr` given the narrow requirements in `SelectMethodToFilterFieldTranslator`. But when `SelectMethodToFilterFieldTranslator` doesn’t know how to handle something, it simply falls back to the `$expr` impl by throwing an `ExpressionNotSupportedException`. This allows us some nicely flexibility because we can gradually add more features (if needed) but still fall back to the existing `$expr` impl for all other scenarios.

PR #911, opened 2022-10-20
CSHARP-3136: Add full support for $dateToString in LINQ3.

rstam:


src/MongoDB.Driver/Linq/DateTimeExtensions.cs
```diff
@@ -279,6 +279,18 @@ public static DateTime Subtract(this DateTime @this, long value, DateTimeUnit un
...
         }
 
+        /// <summary>
+        /// Converts a DateTime value to a string.
+        /// </summary>
+        /// <param name="this">The DateTime value.</param>
+        /// <param name="format">The format string (optional, can be null).</param>
+        /// <param name="timezone">The timezone to use in the returned string (optional, can be null).</param>
+        /// <returns>The DateTime value converted to a string.</returns>
+        public static string ToString(this DateTime @this, string format, string timezone)
```
rstam, line 289: Only the `NullableDateTime` overload takes has an `onNull` parameter.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/DateTimeMethod.cs
```diff
@@ -92,6 +94,8 @@ static DateTimeMethod()
             __subtractWithTimeSpanAndTimezone = ReflectionInfo.Method((DateTime @this, TimeSpan value, string timezone) => @this.Subtract(value, timezone));
             __subtractWithUnit = ReflectionInfo.Method((DateTime @this, long value, DateTimeUnit unit) => @this.Subtract(value, unit));
             __subtractWithUnitAndTimezone = ReflectionInfo.Method((DateTime @this, long value, DateTimeUnit unit, string timezone) => @this.Subtract(value, unit, timezone));
+            __toStringWithFormat = ReflectionInfo.Method((DateTime @this, string format) => @this.ToString(format));
```
rstam, line 97: This overload is part of the .NET API.

We will use the standard existing overload but will require that the `format` string be a MongoDB format string.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToStringMethodToAggregationExpressionTranslator.cs
```diff
@@ -13,64 +13,110 @@
...
+        };
+
+        private static readonly MethodInfo[] __dateTimeToStringMethodsWithOnNull = new[]
+        {
+            NullableDateTimeMethod.ToStringWithFormatAndTimezoneAndOnNull,
+        };
+
         public static AggregationExpression Translate(TranslationContext context, MethodCallExpression expression)
         {
-            if (TryTranslateInstanceToStringWithNoArguments(context, expression, out var translation) ||
```
rstam, line 27: Instead of the `Try` pattern we just determine which method to call and only call the method if fully expect it to succeed except in the case of unsupported expressions.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToStringMethodToAggregationExpressionTranslator.cs
```diff
@@ -13,64 +13,110 @@
...
-            if (method.DeclaringType != typeof(DateTime) || method.IsStatic || method.ReturnType != typeof(string) || method.Name != "ToString" || arguments.Count != 1 || arguments[0].Type != typeof(string))
-            {
-                translation = null;
-                return false;
-            }
+        private static AggregationExpression TranslateDateTimeToStringMethod(TranslationContext context, MethodCallExpression expression, MethodInfo method, Expression[] arguments)
+        {
 
-            var dateTimeExpression = expression.Object;
+            var dateTimeExpression = method.IsStatic ? arguments[0] : expression.Object;
```
rstam, line 76: Note that the `DateTime` value is `arguments[0]` for an extension method and `expression.Object` for an instance method.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToStringMethodToAggregationExpressionTranslator.cs
```diff
@@ -13,64 +13,110 @@
...
-        }
+            var dateAst = dateTimeTranslation.Ast;
 
-        private static bool TryTranslateInstanceToStringWithNoArguments(TranslationContext context, MethodCallExpression expression, out AggregationExpression translation)
-        {
-            var method = expression.Method;
-            var arguments = expression.Arguments;
-            translation = null;
+            AstExpression formatAst = null;
+            var formatExpression = method.IsStatic ? arguments[1] : arguments[0];
```
rstam, line 81: Note that `format` is either argument `0` or `1` depending on whether the method is an extension method or not.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/DateTimeMethod.cs
JamesKovacs: Makes sense to use the existing overload but insist that the format string is something that MongoDB understands. It would be messy to parse and convert .NET DateTime format strings to MongoDB equivalent ones.

src/MongoDB.Driver/Linq/NullableDateTimeExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Converts a NullableDateTime value to a string.
+        /// </summary>
+        /// <param name="this">The NullableDateTime value.</param>
+        /// <param name="format">The format string (optional, can be null).</param>
+        /// <param name="timezone">The timezone to use in the returned string (optional, can be null).</param>
+        /// <param name="onNull">The string to return if the NullableDateTime value is null.</param>
+        /// <returns>The NullableDateTime value converted to a string.</returns>
+        public static string ToString(this DateTime? @this, string format, string timezone, string onNull)
```
JamesKovacs, line 34: Do we even want to expose `onNull`? A more natural and flexible way to write this in C# would be:

```
var query = coll.AsQueryable().Select(x => new { D2 = x.N.ToString("%Y-%M-%D", null, null) ?? x.D.ToString() });
```

which is translated to:

```
test.coll.Aggregate([{ "$project" : { "D2" : { "$ifNull" : [{ "$dateToString" : { "date" : "$N", "format" : "%Y-%M-%D" } }, { "$toString" : "$D" }] }, "_id" : 0 } }])
```

Note that `onNull` can support arbitrary expressions, not just strings. The existing impl using `??` already can support arbitrary expressions whereas this `ToString` overload cannot. Maybe we should remove this `ToString` overload? What do you think?

Regardless we should have `DateTime?` extension methods for `format` and `format`/`timezone` overloads.

JamesKovacs (changes requested):
I have some API questions below. Let me know what you think...

src/MongoDB.Driver/Linq/NullableDateTimeExtensions.cs
rstam: That's a neat trick using `??`. But it does generate much more verbose MQL.

I think it's still better to have actual support for the `onNull` parameter because it translates to smaller MQL (and therefore presumably more efficient).

I don't see why you can't use an arbitrary expression for the `onNull` argument value. If there isn't already a test for that I can modify the test to verify that an expression is allowed.

rstam: See the refactored tests that now include testing that the `onNull` argument can be an expression.

JamesKovacs: I was thinking about this incorrectly. I thought we would need a `ToString` overload with `Expression onNull`. But we can have an expression as long as it evaluates to string. And it wouldn't work if the expression didn't evaluate to a string.

JamesKovacs (approved):
LGTM

PR #912, opened 2022-10-21
CSHARP-4368: Support Convert to BsonValue in LINQ3.

rstam:


src/MongoDB.Bson/Serialization/PrimitiveSerializationProvider.cs
```diff
@@ -53,6 +54,7 @@ static PrimitiveSerializationProvider()
                 { typeof(Nullable<>), typeof(NullableSerializer<>) },
                 { typeof(Object), typeof(ObjectSerializer) },
                 { typeof(ObjectId), typeof(ObjectIdSerializer) },
+                { typeof(Regex), typeof(RegexSerializer) },
```
rstam, line 57: This is a minor bug found while testing support for convert to BsonValue.

It could be factored out to its own ticket if you want.

src/MongoDB.Bson/Serialization/Serializers/RegexSerializer.cs
```diff
@@ -90,6 +90,10 @@ public override Regex Deserialize(BsonDeserializationContext context, BsonDeseri
             var bsonType = reader.GetCurrentBsonType();
             switch (bsonType)
             {
+                case BsonType.Null:
+                    reader.ReadNull();
+                    return null;
+
```
rstam, line 96: Another small bug found while testing convert to BsonValue.

This serializer was not handling `null` properly.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -797,6 +797,36 @@ public static AstExpression Sum(AstExpression array)
             return new AstUnaryExpression(AstUnaryOperator.Sum, array);
         }
 
+        public static AstExpression ToBool(AstExpression arg)
```
rstam, line 800: A small number of new factory methods.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConvertExpressionToAggregationExpressionTranslator.cs
```diff
@@ -65,5 +72,42 @@ public static AggregationExpression Translate(TranslationContext context, UnaryE
...
+            }
+
+            var convertedOperandAst = Type.GetTypeCode(operandType) switch
+            {
+                TypeCode.Object when operandType == typeof(byte[]) => ErrorIfUnexpectedDataType(operandAst, expectedType: "binData"),
+                TypeCode.Boolean => AstExpression.ToBool(operandAst),
+                TypeCode.DateTime => AstExpression.ToDate(operandAst),
+                TypeCode.Decimal => AstExpression.ToDecimal(operandAst),
+                TypeCode.Object when operandType == typeof(Decimal128) => AstExpression.ToDecimal(operandAst),
+                TypeCode.Double => AstExpression.ToDouble(operandAst),
```
rstam, line 91: `Guid` is conspicuously absent.

We *could* support it, but for one the conversion from `Guid` to `BsonValue` is deprecated because it is ambiguous regarding representation.

What we could do is ignore the representation issue and simply return whatever `BsonBinaryData` is in the database (of unknown representation unless it's `Standard`).

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConvertExpressionToAggregationExpressionTranslator.cs
```diff
@@ -65,5 +72,42 @@ public static AggregationExpression Translate(TranslationContext context, UnaryE
...
+
+            return new AggregationExpression(expression, convertedOperandAst, BsonValueSerializer.Instance);
+
+            static AstExpression ErrorIfUnexpectedDataType(AstExpression operandAst, string expectedType)
+            {
+                // this expression is designed to fail server side if the data encountered is not null or of the expected type
+                return AstExpression.Cond(
+                    AstExpression.In(AstExpression.Type(operandAst), AstExpression.ComputedArray(new AstExpression[] { "null", expectedType })),
+                    operandAst,
+                    AstExpression.Convert(operandAst, expectedType)); // will fail server side
```
rstam, line 109: We could talk about this...

If you think the errors should occur client side that's possibly a reasonable alternative.

rstam: I'm having doubts about all this server side "type checking".

Maybe it's better to just let whatever is stored served side (unexpected or not) come out as a BsonValue.

That does mean you might get unexpected results, but only if the data is invalid.

For example, `(BsonValue)x.IntProperty` might result in a `BsonDouble` if the number stored in the database is a double instead of an int.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4368Tests.cs
```diff
@@ -29,159 +30,254 @@ namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
...
         {
-            var expectedResult = value == null ? BsonNull.Value : (BsonValue)value;
-            Convert_to_BsonValue_from_TValue_should_work<string>(value, expectedResult, valueAsJson, "{ $toString : '$V' }", x => (BsonValue)x.V);
+            var value = BsonSerializer.Deserialize<string>(valueAsJson);
+            var expectedResult = BsonSerializer.Deserialize<BsonValue>(valueAsJson);
+            Convert_to_BsonValue_from_TValue_should_work<string>(value, valueAsJson, expectedResult, x => (BsonValue)x.V);
         }
 
-        private void Convert_to_BsonValue_from_TValue_should_work<TValue>(TValue value, BsonValue expectedResult, string valueAsJson, string convertExpression, Expression<Func<Document<TValue>, BsonValue>> projection)
+        private void Convert_to_BsonValue_from_TValue_should_work<TValue>(TValue value, string valueAsJson, BsonValue expectedResult, Expression<Func<Document<TValue>, BsonValue>> projection)
```
JamesKovacs, line 280: While we often name tests using snake_case, we don't name our helper methods like this. I would suggest `AssertCastToBsonValue` or something similar instead.

JamesKovacs (changes requested):
Minor naming issue for a test helper method. Otherwise LGTM.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4368Tests.cs
rstam: This method is not a simple Assert.

This IS the test.

Normally this would be a `Theory`, but it's got the `TValue` generic parameter, so we need separate `Fact` to introduce the actual `TValue` type.

Let me think about this...

JamesKovacs: I'm not too concerned with the naming since it is in tests. If you feel strongly about the method name, I'm ok leaving it. However these new tests are failing on certain variants in this patch. I haven't looked into why this might be yet. I'll let you investigate before I approve.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4368Tests.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        public class TestCase
+        {
+            public Type ValueType { get; set; }
+            public string ValueAsJson { get; set; }
+            public LambdaExpression Projection { get; set; }
+        }
+
+        public static TestCase CreateTestCase<TValue>(
```
rstam, line 62: Using a factory method for test cases allows more compact syntax.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4368Tests.cs
```diff
@@ -0,0 +1,196 @@
...
+        public void Convert_to_BsonValue_from_TValue_should_work_invoker(Type valueType, int i, string valueAsJson, string projectionAsString)
+        {
+            GuidMode.Set(GuidRepresentationMode.V3);
+
+            var testMethodInfo = this.GetType().GetMethod(nameof(Convert_to_BsonValue_from_TValue_should_work));
+            var testMethod = testMethodInfo.MakeGenericMethod(valueType);
+            testMethod.Invoke(this, new object[] { i, valueAsJson, projectionAsString });
+        }
+
+        public void Convert_to_BsonValue_from_TValue_should_work<TValue>(int i, string valueAsJson, string projectionAsString)
```
rstam, line 136: This is the *real* test method.

The `Convert_to_BsonValue_from_TValue_should_work_invoker` method is a helper method to work around limitations xUnit has in invoking generic test methods.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4368Tests.cs
rstam: See the latest. There is now just one xUnit test method, which is an invoker for the *real* test method.

These two methods work together to work around limitations in xUnit not being able to invoke generic test methods.

JamesKovacs (approved):
LGTM

PR #914, opened 2022-10-26
CSHARP-4369: Translate Group/Where/Select correctly in LINQ3.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionToFilterTranslator.cs
```diff
@@ -48,10 +48,23 @@ public static AstFilter Translate(TranslationContext context, Expression express
             return filter;
         }
 
-        public static AstFilter TranslateLambda(TranslationContext context, LambdaExpression lambdaExpression, IBsonSerializer parameterSerializer)
+        public static AstFilter TranslateLambda(
+            TranslationContext context,
+            LambdaExpression lambdaExpression,
+            IBsonSerializer parameterSerializer,
+            bool asRoot = false)
```
rstam, line 55: Note: currently all callers happen to pass `true` as the value of the `asRoot` argument, but that might be coincidental.

I would still like the caller's to explicitly set `asRoot` to `true` when it applies.

tests/MongoDB.Driver.Tests/Samples/AggregationSample.cs
```diff
@@ -88,16 +89,38 @@ public async Task States_with_pops_over_20000_queryable_method()
...
+                {
+                    "{ $group : { _id : '$state', __agg0 : { $sum : '$pop' } } }",
+                    "{ $match : { __agg0 : { $gt : 20000 } } }",
+                    "{ $project : { State : '$_id', TotalPopulation : '$__agg0', _id : 0 } }"
+                }
+                :
+                new[]
+                {
+                    "{ $group : { _id : '$state', __agg0 : { $sum : '$pop' } } }",
+                    "{ $match : { $expr : { $gt : ['$__agg0', 20000] } } }",
```
rstam, line 118: This is the only part of the pipeline that is slightly different from LINQ2.

rstam (commented):
At first I thought this ticket might be a bug in the pipeline optimizer, but it turned out to be a bug in the filter translator which prevented the pipeline optimizer from recognizing the needed optimization.

JamesKovacs (approved):
LGTM

PR #915, opened 2022-10-26
CSHARP-4376: Add LINQ3 support for untyped CreateQuery.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -77,7 +77,9 @@ internal sealed class MongoQueryProvider<TDocument> : MongoQueryProvider
         // public methods
         public override IQueryable CreateQuery(Expression expression)
         {
-            throw new NotImplementedException();
+            var outputType = expression.Type.GetSequenceElementType();
+            var queryType = typeof(MongoQuery<,>).MakeGenericType(typeof(TDocument), outputType);
+            return (IQueryable)Activator.CreateInstance(queryType, new object[] { this, expression });
```
rstam, line 82: As part of my testing I replaced the generic version of `CreateQuery` with:

    public override IQueryable<TOutput> CreateQuery<TOutput>(Expression expression)
    {
        // return new MongoQuery<TDocument, TOutput>(this, expression);
        return (IQueryable<TOutput>)CreateQuery(expression);
    }

to force ALL calls to `CreateQuery` to go through the non-generic path and the full test suite passed.

I then reverted that because we don't want to use reflection when we don't have to.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/MongoQueryProviderTests.cs
```diff
@@ -13,14 +13,67 @@
...
+            var queryable = collection.AsQueryable();
+            var parameter = Expression.Parameter(typeof(C), "x");
+            var expression =
+                Expression.Call(
+                    QueryableMethod.Select.MakeGenericMethod(typeof(C), typeof(int)),
+                    Expression.Constant(queryable, typeof(IQueryable<C>)),
+                    Expression.Quote(
+                        Expression.Lambda<Func<C, int>>(
+                            Expression.Property(parameter, "X"),
+                            parameter)));
```
rstam, line 68: The hardest part of these tests was getting the expression right!

JamesKovacs (approved):
LGTM

PR #916, opened 2022-10-26
CSHARP-4289: Projected class should use correct serializers.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -26,51 +27,97 @@ internal static class MemberInitExpressionToAggregationExpressionTranslator
...
-                    var argumentExpression = constructorArguments[i];
-                    var fieldName = GetFieldName(constructorParameter);
-                    var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
-                    computedFields.Add(AstExpression.ComputedField(fieldName, argumentTanslation.Ast));
-                }
+                var argumentExpression = constructorArguments[i];
+                var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
+                computedFields.Add(AstExpression.ComputedField(memberMap.ElementName, argumentTanslation.Ast));
+
+                memberMap.SetSerializer(argumentTanslation.Serializer);
```
rstam, line 45: This is the key change... we configure a new class map on the fly and set the correct serializer on each property.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -26,51 +27,97 @@ internal static class MemberInitExpressionToAggregationExpressionTranslator
...
+            {
+                var memberAssignment = (MemberAssignment)binding;
+                var member = memberAssignment.Member;
+                var memberMap = FindMemberMap(expression, classMap, member.Name);
+
+                var valueExpression = memberAssignment.Expression;
+                var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression);
+                computedFields.Add(AstExpression.ComputedField(memberMap.ElementName, valueTranslation.Ast));
+
+                memberMap.SetSerializer(valueTranslation.Serializer);
```
rstam, line 58: Same as above but now for properties being set instead of constructor arguments.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -26,51 +27,97 @@ internal static class MemberInitExpressionToAggregationExpressionTranslator
...
+            var ast = AstExpression.ComputedDocument(computedFields);
+
+            classMap.Freeze();
+            var serializerType = typeof(BsonClassMapSerializer<>).MakeGenericType(expression.Type);
+            var serializer = (IBsonSerializer)Activator.CreateInstance(serializerType, classMap);
+
+            return new AggregationExpression(expression, ast, serializer);
+        }
+
+        private static BsonClassMap CreateClassMap(Type classType)
```
rstam, line 70: Everything from here to the end of the file is unfortunate.

Basically we're creating a new class map hierarchy all the way up to `object` so that we can configure serializers as needed.

Normally class maps are created quite lazily and some work isn't performed until `Freeze` is called, but for this change to work we need the class maps to be a little more fully initialized up front.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -26,51 +27,97 @@ internal static class MemberInitExpressionToAggregationExpressionTranslator
...
-                {
-                    var constructorParameter = constructorParameters[i];
-                    var argumentExpression = constructorArguments[i];
-                    var fieldName = GetFieldName(constructorParameter);
-                    var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
-                    computedFields.Add(AstExpression.ComputedField(fieldName, argumentTanslation.Ast));
-                }
+                var argumentExpression = constructorArguments[i];
+                var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
+                computedFields.Add(AstExpression.ComputedField(memberMap.ElementName, argumentTanslation.Ast));
```
rstam, line 43: We could use the `memberMap.MemberName` here if we wanted to, as long as we configured the class map appropriately.

I'm not sure which is better.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4289Tests.cs
```diff
@@ -0,0 +1,87 @@
...
+        [Fact]
+        public void Select_using_named_class_should_work()
+        {
+            var collection = CreateCollection();
+
+            var queryable = collection.AsQueryable()
+                .Select(x => new R(x.Id) { W = x.Id });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $project : { v : '$_id', w : '$_id', _id : 0 } }");
```
rstam, line 52: Notice the lower case `v` and `w`.

We could choose to use the property names instead (upper case) and configure the class map accordingly.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -26,51 +27,97 @@ internal static class MemberInitExpressionToAggregationExpressionTranslator
...
-                for (var i = 0; i < constructorParameters.Length; i++)
-                {
-                    var constructorParameter = constructorParameters[i];
-                    var argumentExpression = constructorArguments[i];
-                    var fieldName = GetFieldName(constructorParameter);
-                    var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
-                    computedFields.Add(AstExpression.ComputedField(fieldName, argumentTanslation.Ast));
-                }
+                var argumentExpression = constructorArguments[i];
+                var argumentTanslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression);
```
JamesKovacs, line 42: `argumentTanslation` => `argumentTranslation`

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
JamesKovacs: I can't think of a reason to use one over the other. Let's use `memberMap.ElementName` unless we can think of a reason why that wouldn't work.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
JamesKovacs: Makes sense based on our conversation about this ticket.

JamesKovacs (approved):
LGTM. Just one spelling error to fix.

PR #918, opened 2022-10-28
CSHARP-3965: Support OrderBy with an expression that does not resolve to a field.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3965Tests.cs
```diff
@@ -0,0 +1,461 @@
...
+            var queryable = collection
+                .AsQueryable()
+                .OrderBy(x => x.Y + 1);
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $project : { _sortedDocument : '$$ROOT', _key1 : { $add : ['$Y', 1] }, _id : 0 } }",
+                "{ $sort : { _key1 : 1 } }",
+                "{ $replaceRoot : { newRoot : '$_sortedDocument' } }");
```
rstam, line 39: As I write these comments it occurs to me that some minor changes to the generated MQL might be beneficial:

    { $project : { _document : "$$ROOT", _id : 0, _key1 : expr1, ..., _keyn : exprn } }
    { $sort : { _key1 : d1, ..., _keyn : dn } }
    { $replaceRoot : { newRoot : "$_document } }

Naming the temporary field `_document` is shorter than `_sortedDocument` and also more correct since the document isn't sorted (yet) in the `$project` stage.

Putting the keys at the end of the `$project` stage makes it slightly easier to append a new key. But... it's easy enough no matter what so let's focus on making the generated MQL as logical and readable as possible as opposed to ease of implementation.

JamesKovacs: I like renaming `_sortedDocument` to `_document` and it makes sense for the reasons stated. I'm ambivalent on where `_id: 0` should appear. Keeping the keys at the end of the `$project` makes senes to me and if it simplifies the implementation, then I'm good with making that change.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/OrderByMethodToPipelineTranslator.cs
```diff
@@ -63,20 +62,199 @@ public static AstPipeline Translate(TranslationContext context, MethodCallExpres
...
+        {
+            return pipeline.AddStages(pipeline.OutputSerializer, newSortStages);
+        }
+
+        private static AstPipeline CombineSortStages(AstPipeline pipeline, AstStage[] newSortStages)
         {
-            var fieldPath = keySelector.GetFieldPath(context, parameterSerializer);
-            switch (methodName)
+            var oldSortStages = FindOldSortStages(pipeline);
+
```
JamesKovacs, line 73: I was initially confused by the magic numbers `1` and `3` and why only these various combinations mattered. I'd suggest adding a comment such as "1 is a $sort stage and 3 is $project/$sort/$replaceRoot".

JamesKovacs (changes requested):
Looks great. Minor changes mostly to make the changes you suggested.

JamesKovacs (approved):
LGTM

PR #921, opened 2022-10-29
CSHARP-4244: Support object comparison at root level in LINQ3.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstFieldOperationFilter.cs
```diff
@@ -38,6 +38,11 @@ public AstFieldOperationFilter(AstFilterField field, AstFilterOperation operatio
                 // normally an ExpressionNotSupported should have been thrown before reaching here
                 throw new ArgumentException($"Field must be represented as a string for regex filter operations: {field.Path}", nameof(field));
             }
+
+            if (_field.Path == "@<current>")
+            {
+                throw new ExpressionNotSupportedException("Field path cannot be \"@<current>\" in AstFieldOperationFilter.");
+            }
```
rstam, line 45: This exception will end up triggering falling back to `$expr`.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4244Tests.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var collection = CreateCollection();
+            var person = new Person { Id = 2, Name = "Jane Doe" };
+
+            var queryable = collection
+                .AsQueryable()
+                .Where(p => p == person);
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { $expr : { $eq : ['$$ROOT', { _id : 2, Name : 'Jane Doe' }] } } }");
```
rstam, line 36: This *has* to use `$expr` and `$eq` because the following is NOT correct:

    { $match : { _id : 2, Name : "Jane Doe" } }

in the general case. Because if any default fields are not serialized this would inadvertently become a partial QBE query.

JamesKovacs (approved):
LGTM

PR #922, opened 2022-10-31
CSHARP-2471: Add support for trig functions in LINQ3.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstBinaryOperator.cs
```diff
@@ -20,7 +20,7 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
     internal enum AstBinaryOperator
     {
         ArrayElemAt,
-        ATan2,
+        Atan2,
```
rstam, line 23: Getting the casing in sync with .NET and MongoDB.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstUnaryOperator.cs
```diff
@@ -20,21 +20,22 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
     internal enum AstUnaryOperator
     {
         Abs,
-        ACos,
-        ACosh,
+        Acos,
+        Acosh,
```
rstam, line 24: Getting the casing in sync with .NET and MongoDB.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/TrigMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,122 @@
...
+    internal static class TrigMethodToAggregationExpressionTranslator
+    {
+        private static readonly MethodInfo[] __binaryTrigMethods;
+        private static readonly MethodInfo[] __unaryTrigMethods;
+
+        static TrigMethodToAggregationExpressionTranslator()
+        {
+            __binaryTrigMethods = new[]
+            {
+                MathMethod.Atan2
```
rstam, line 36: I know... there's only one, but no harm in mirroring the unary implementation and being future proof.

src/MongoDB.Driver/Linq/MongoDBMath.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Similar to the standard Math class but contains MongoDB specific methods.
+    /// </summary>
+    public static class MongoDBMath
+    {
+        /// <summary>
+        /// Converts degrees to radians.
+        /// </summary>
+        /// <param name="degrees">The degrees.</param>
+        /// <returns>The radians.</returns>
+        public static double DegreesToRadians(double degrees)
```
rstam, line 30: I couldn't find any methods in .NET to convert degrees to radians or vice versa.

I'm not thrilled with having to use `MongoDBMath.DegreesToRadians`, but there is no way to add a new static method like `Math.DegreesToRadians` because we don't own the `Math` class.

If you and think of a better new public API for these two methods let's discuss.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2471Tests.cs
```diff
@@ -0,0 +1,121 @@
...
+                "$asinh" => x => Math.Asinh(x.X),
+#endif
+                "$atan" => x => Math.Atan(x.X),
+#if NETCOREAPP3_1_OR_GREATER
+                "$atanh" => x => Math.Atanh(x.X),
+#endif
+                "$cos" => x => Math.Cos(x.X),
+                "$cosh" => x => Math.Cosh(x.X),
+                "$degreesToRadians" => x => MongoDBMath.DegreesToRadians(x.X),
+                "$radiansToDegrees" => x => MongoDBMath.RadiansToDegrees(x.X),
```
rstam, line 69: Here's the new public API for these two methods in action.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2471Tests.cs
```diff
@@ -0,0 +1,121 @@
...
+        [InlineData("$atanh", 0.0, 0.0)]
+#endif
+        [InlineData("$cos", 0.0, 1.0)]
+        [InlineData("$cosh", 0.0, 1.0)]
+        [InlineData("$degreesToRadians", 0.0, 0.0)]
+        [InlineData("$radiansToDegrees", 0.0, 0.0)]
+        [InlineData("$sin", 0.0, 0.0)]
+        [InlineData("$sinh", 0.0, 0.0)]
+        [InlineData("$tan", 0.0, 0.0)]
+        [InlineData("$tanh", 0.0, 0.0)]
```
rstam, line 47: Do you think we need more "interesting" test values?

There are lot of zeros in there.

Might be fine... it's not our job to test the server implementation of these methods, but more  interesting values might be better at ensuring we translated to the correct server operator (because a lot of trig functions return 0.0 when called with 0.0).

src/MongoDB.Driver/Linq/MongoDBMath.cs
JamesKovacs: Let's make them extension methods in the `MongoDB.Driver.Linq` namespace. Then instead of writing:

`x => MongoDBMath.DegreesToRadians(x.Value)`

you could instead write:

`x => x.Value.DegreesToRadians()`

We could even shorten these names to `ToRadians()` and `ToDegrees()` as the other unit is implied though we have no way to check. If you want to keep the more verbose `DegreesToRadians` to better match the server functions, I'm good with that too.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2471Tests.cs
JamesKovacs: I think it's fine. The important thing is the translation of C# method calls into the correct MQL, not ensuring that the server's impl of a function's precision remains consistent.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2471Tests.cs
JamesKovacs: See above. I think these should be `x.X.ToDegrees()` and `x.X.ToRadians()`, but up for discussion.

JamesKovacs (changes requested):
See my comments on using extension methods for the conversion syntax.

src/MongoDB.Driver/Linq/MongoDBMath.cs
rstam: I considered this approach, but I think it has major disadvantages.

The first is lack of symmetry with the rest of the trig functions which are static methods in `Math`. So static methods in `MongoDBMath` is the closest we can come to that.

Another is that an extension method pollutes the namespace of `double` (admittedly only if you have the right `using` statement). Worse, we would need multiple extension methods for multiple numeric types.

`MongoDBMath.DegreesToRadians(180)` works because there is an implicit conversion from `int` to `double`.

An extension method on `double` will not be invoked for `180.DegreesToRadians`.

JamesKovacs (approved):
Valid point about needing extension methods on each numeric type and not being able to rely on implicit conversion. I hadn't thought of that wrinkle.

Your reasoning (as always) is sound. LGTM.

PR #923, opened 2022-10-31
CSHARP-3197: Verify that Select/Select works in LINQ3.

rstam:
Don't know what transformations LINQ2 tried to do when there were two Selects in a row, but in LINQ3 if you write two Selects you get two $projects.


JamesKovacs (approved):
LGTM

PR #925, opened 2022-10-31
CSHARP-2134: Verify that All works in LINQ3.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2134Tests.cs
```diff
@@ -0,0 +1,64 @@
...
+        [InlineData(false, true)]
+        [InlineData(true, false)]
+        public void Projection_of_ArrayOfDocuments_dictionary_keys_and_values_should_work(bool includeNullId, bool expectedResult    )
+        {
+            var collection = CreateCollection(includeNullId);
+
+            var result = collection
+                .AsQueryable()
+                .All(x => x.Id != null);
+
```
JamesKovacs, line 35: We should assert the MQL as we do in other tests.

JamesKovacs (changes requested):
Missing assertion on the generated MQL.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2134Tests.cs
rstam: Easier said than done... :)

`All` is a terminal method that returns `bool`, not an `IQueryable` that we can translate into a pipeline that we can assert against.

I'll have to think about whether there is a way to test the MQL.

JamesKovacs: Oh right! It didn't immediately occur to me that `All()` is a terminal expression returning a `bool`. You could rewrite it as a `Select(predicate).All(x => x)` but this will alter the generated MQL. Given we are proving that `All` works in LINQ3, this test is sufficient as-is.

JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2134Tests.cs
rstam: I found a good way to assert against the generated stages.

Even though we haven't done this before, I would definitely do it going forward, so I'm happy to have found a decent way to do this!

JamesKovacs (approved):
Clever. LGTM.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -52,12 +53,14 @@ internal abstract class MongoQueryProvider : IMongoQueryProvider
...
         public abstract TResult Execute<TResult>(Expression expression);
         public abstract Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken);
+        public abstract BsonDocument[] GetMostRecentPipelineStages();
     }
 
     internal sealed class MongoQueryProvider<TDocument> : MongoQueryProvider
     {
         // private fields
         private readonly IMongoCollection<TDocument> _collection;
+        private ExecutableQuery<TDocument> _mostRecentExecutableQuery;
```
rstam, line 63: The query provider will get garbage collected when the queryable gets garbage collected, which typically happens right after the queryable is "executed".

While adding this field extends the lifetime of the most recent executable query by a little bit, it's usually for an extremely short time.

The only time it won't be immediately garbage collected is if someone hangs on to a queryable and "executes" it over and over. Each time it's executed `_mostRecentExecutableQuery` gets overwritten.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -100,13 +103,22 @@ public override object Execute(Expression expression)
         public override TResult Execute<TResult>(Expression expression)
         {
             var executableQuery = ExpressionToExecutableQueryTranslator.TranslateScalar<TDocument, TResult>(this, expression);
+            _mostRecentExecutableQuery = executableQuery;
```
rstam, line 106: The runtime overhead of an extra assignment statement is negligible.

The impact on garbage collection was discussed above.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -100,13 +103,22 @@ public override object Execute(Expression expression)
...
         }
 
         public override Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken)
         {
             var executableQuery = ExpressionToExecutableQueryTranslator.TranslateScalar<TDocument, TResult>(this, expression);
+            _mostRecentExecutableQuery = executableQuery;
             return executableQuery.ExecuteAsync(_session, cancellationToken);
         }
+
+        public override BsonDocument[] GetMostRecentPipelineStages()
```
rstam, line 117: I chose to return `BsonDocument[]` instead of `AstPipeline` because `AstPipeline` is `internal`.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
JamesKovacs: Makes sense. When I LGTM'd this earlier, I considered whether we would be potentially leaking references to instances and decided we weren't. You have to explicitly hang onto an instance for this brief caching to be a problem.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
JamesKovacs: Agreed.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
JamesKovacs: Makes perfect sense.

PR #927, opened 2022-11-01
CSHARP-2472: Get sample from ticket to pass.

rstam:


rstam:
I wanted you to look at this PR as is for a first pass.

I think we should create a separate ticket for the change to `Convert` to use the simpler `$toDouble` (for example) than the longer form `$convert`. Now we only use `$convert` when the shorter `$toDouble`-like form.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -240,26 +240,34 @@ public static AstExpression Constant(BsonValue value)
...
             {
-                "MongoDB.Bson.ObjectId" => "objectId",
-                "System.Boolean" => "bool",
-                "System.DateTime" => "date",
-                "System.Decimal" => "decimal",
-                "System.Double" => "double",
-                "System.Int32" => "int",
-                "System.Int64" => "long",
-                "System.String" => "string",
-                _ => throw new ArgumentException($"Invalid toType: {toType.FullName}.", nameof(toType))
```
rstam, line 259: Moving this transformation to the translator instead allows us to throw an `ExpressionNotSupportedException` instead which is much better.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -240,26 +240,34 @@ public static AstExpression Constant(BsonValue value)
...
-        public static AstExpression Convert(AstExpression input, Type toType, AstExpression onError = null, AstExpression onNull = null)
-        {
-            Ensure.IsNotNull(toType, nameof(toType));
-            var to = toType.FullName switch
+            Ensure.IsNotNull(input, nameof(input));
+            Ensure.IsNotNull(to, nameof(to));
+
+            if (to is AstConstantExpression toConstantExpression &&
+                (toConstantExpression.Value as BsonString)?.Value is string toValue &&
+                toValue != null &&
```
rstam, line 248: The `toValue = null` check is technically not necessary but it's clearer to check for it. It's kind of subtle why letting `null` flow through below happens to work (somewhat coincidentally).

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2472Tests.cs
```diff
@@ -0,0 +1,89 @@
...
+                    q => q.equipment_id == equipmentId
+                    && q.timestamp >= startDate
+                    && q.timestamp <= endDate
+                )
+                .GroupBy(g => g.timestamp)
+                .Select(p => new MyDTO
+                {
+                    timestamp = p.Key,
+                    sqrt_calc = (decimal)Math.Sqrt(
+                        p.Sum(x => (double)x.my_decimal_value)
```
rstam, line 46: It's important to cast to `(double)` BEFORE calling `Sum` because `decimal` is represented in the database as a string (by default), and fortunately `$toDouble` handles converting from `string`!

The original poster used:

    (double)p.Sum(x => x.my_decimal_value)

which doesn't work because `Sum` treats invalid values (like `string`) as zero.

rstam:
Ticket number in commit is wrong. 

Must remember to fix it!!!

JamesKovacs (approved):
LGTM

PR #929, opened 2022-11-02
CSHARP-3136: Skip certain tests on server version 3.6.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
```diff
@@ -71,6 +75,11 @@ public void DateTime_ToString_with_format_should_work()
         [InlineData("%H:%M:%S", "-04:00", "{ $project : { _v : { $dateToString : { date : '$D', format : '%H:%M:%S', timezone : '-04:00' } }, _id : 0 } }", new[] { "23:04:05", "23:04:05" })]
         public void DateTime_ToString_with_format_and_timezone_constants_should_work(string format, string timezone, string expectedProjectStage, string[] expectedResults)
         {
+            if (format == null)
+            {
+                RequireServer.Check().VersionGreaterThanOrEqualTo("4.0");
```
rstam, line 80: I didn't create a `Feature` for this because it's more of a limitation on an existing feature.

Sound OK?

(If we did create a `Feature` for this it would be called something like `DateToStringOperatorOptionalFormat`)

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
```diff
@@ -27,6 +29,8 @@ public class CSharp3136Tests : Linq3IntegrationTest
         [Fact]
         public void DateTime_ToString_with_no_arguments_should_work()
         {
+            RequireServer.Check().Supports(Feature.ToStringOperator);
```
DmitryLukyanov, line 32: it doesn't matter now, but maybe better to use `SkippableFact`?

DmitryLukyanov (approved):
Minor comment and then LGTM, assuming tests passed

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
rstam: Good point. Will do.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
JamesKovacs: Shouldn't this guard clause be `if (format != null)`?

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
```diff
@@ -151,6 +164,11 @@ public void NullableDateTime_ToString_with_no_arguments_should_work()
         [InlineData("%H:%M:%S", "-04:00", "xx", "{ $project : { _v : { $dateToString : { date : '$N', format : '%H:%M:%S', timezone : '-04:00', onNull : 'xx' } }, _id : 0 } }", new[] { "23:04:05", "xx" })]
         public void NullableDateTime_ToString_with_format_and_timezone_and_onNull_constants_should_work(string format, string timezone, string onNull, string expectedProjectStage, string[] expectedResults)
         {
+            if (format == null || onNull != null)
+            {
+                RequireServer.Check().VersionGreaterThanOrEqualTo("4.0");
+            }
```
JamesKovacs, line 170: Same here. `format != null` then perform the check for server 4.0+.

JamesKovacs (changes requested):
Verify server checks. I think two of them are backwards.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
rstam: I think this is correct.

Format is *required* in server 3.6, so a missing format requires server >= 4.0.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp3136Tests.cs
rstam: See above.

JamesKovacs (approved):
Oh, server made `format` optional in 4.0? Weird. I thought they added it in 4.0. My mistake for not reading the docs myself.

LGTM.

PR #934, opened 2022-11-06
CSHARP-4385: Consider adding Directory.Build.props.

DmitryLukyanov:


CSharpDriver.sln
```diff
@@ -1,9 +1,12 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30011.22
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31919.166
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{D2012971-32BB-4C5F-BFC4-30A9994AB152}"
+	ProjectSection(SolutionItems) = preProject
+		src\Directory.Build.props = src\Directory.Build.props
```
DmitryLukyanov, line 8: it's possible to add this file into solution directory, but given that configuration makes sense only for src folder, probably better to leave it when it's currently placed

src/MongoDB.Bson/MongoDB.Bson.csproj
```diff
@@ -1,68 +1,24 @@
...
-  <ItemGroup>
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0" />
-  </ItemGroup>
-
-  <PropertyGroup>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2" PrivateAssets="All" />
```
DmitryLukyanov, line 16: we can move this to directory.build.props. If we add this reference to the legacy project. Thoughts?

src/Directory.Build.props
```diff
@@ -0,0 +1,55 @@
+<Project>
```
BorisDog, line 1: Maybe let's do another one for tests?

src/MongoDB.Driver.Legacy/MongoDB.Driver.Legacy.csproj
```diff
@@ -1,55 +1,19 @@
...
-    </PackageReference>
-  </ItemGroup>
-
   <PropertyGroup>
-    <DefineConstants>TRACE</DefineConstants>
-  </PropertyGroup>
-
-  <PropertyGroup>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <RootNamespace>MongoDB.Driver</RootNamespace>
```
BorisDog, line 17: Is RootNamespace needed? It's present only in 2 Legacy and Core projects.

BorisDog (changes requested):
Nice optimization!

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/MongoDB.Driver.SmokeTests.Sdk.csproj
```diff
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net472;net5.0;net6.0</TargetFrameworks>
-    <LangVersion>9</LangVersion>
+    <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
```
BorisDog, line 3: I think tests props should not be shared with smoke tests.
Smoke tests intended to be "standalone" projects, and should not share anything with driver specific code.
We might want to consider props for all smoke tests if/when we expand them with more projects.

DmitryLukyanov: I think they can (should?) share some configuration like TF to ensure that all TFs we support works well in consumers application too. But I'm ok to defer this to further tickets, reverted

tests/TestsProps/Directory.Build.props
```diff
@@ -0,0 +1,52 @@
+<Project>
```
BorisDog, line 1: It looks like `Directory.Build.props` naming is only relevant when the props needs to be applied to all project in same directory.
Maybe is should be called `Tests.Build.props` and optionally the containing directory `BuildProps` (as it's already under tests dir)?

tests/TestsProps/Directory.Build.props
```diff
@@ -0,0 +1,52 @@
...
+  <PropertyGroup>
+    <Company>MongoDB Inc.</Company>
+    <Copyright>Copyright © 2010-present MongoDB Inc.</Copyright>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <DefineConstants>TRACE</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(IsWindows)'=='true'">
+    <DefineConstants>$(DefineConstants);WINDOWS</DefineConstants>
```
BorisDog, line 34: Any idea what $(DefineConstants) means and why it's not applied to TRACE?

DmitryLukyanov: in this section you can define different consts (based on condition) that then it will be visible in `#if` construction in tests

tests/TestsProps/Directory.Build.props
DmitryLukyanov: done

BorisDog (approved):
LGTM.
Nice improvement!

PR #938, opened 2022-11-07
CSHARP-4406: Unite logging "Cluster" and "Sdam" categories

BorisDog:
- Fix "Invalidating description" log message in DefaultServer

SDAM category was mistakenly split to Cluster and SDAM categories. Spec specifies single category for both.


JamesKovacs (approved):
LGTM

PR #940, opened 2022-11-07
CSHARP-2472: Skip integration test on older servers.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2472Tests.cs
```diff
@@ -16,16 +16,19 @@
...
 
 namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
 {
     public class CSharp2472Tests : Linq3IntegrationTest
     {
-        [Fact]
+        [SkippableFact]
         public void Projection_of_ArrayOfDocuments_dictionary_keys_and_values_should_work()
         {
+            RequireServer.Check().Supports(Feature.ToStringOperator);
```
rstam, line 31: I'm using the `ToStringOperator` feature as representative of the whole family of `ToXyz` operators.

JamesKovacs: Given that there is no use of `$toString` in this test (and even `$toDouble` is a bit hidden in the MQL), please include a comment explaining the feature check. Otherwise someone might remove this check in the future thinking it was added in error.

JamesKovacs (changes requested):
Add comment to clarify the server check.

JamesKovacs (approved):
LGTM

PR #941, opened 2022-11-08
TryRegisterClassMap implementation

Laurianti:
I've implemented the TryRegisterClassMap method which is an atomic version of the existing RegisterClassMap method.
This method is useful in highly concurrent scenarios as it avoids creating external locks before registering a class map.

Current workaround:

```c#
lock (_lock)
{
    if (!BsonClassMap.IsClassMapRegistered(typeof(T)))
        BsonClassMap.RegisterClassMap<T>(Map);
}
```

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonClassMap.cs
+++ b/src/MongoDB.Bson/Serialization/BsonClassMap.cs
@@ -383,9 +383,9 @@ namespace MongoDB.Bson.Serialization
         /// </summary>
         /// <typeparam name="TClass">The class.</typeparam>
         /// <returns>The class map.</returns>
-        public static BsonClassMap<TClass> TryRegisterClassMap<TClass>()
+        public static bool TryRegisterClassMap<TClass>(out BsonClassMap<TClass> classMap)
         {
-            return TryRegisterClassMap<TClass>(cm => { cm.AutoMap(); });
+            return TryRegisterClassMap(cm => { cm.AutoMap(); }, out classMap);
         }
 
         /// <summary>
@@ -406,12 +406,12 @@ namespace MongoDB.Bson.Serialization
         /// </summary>
         /// <typeparam name="TClass">The class.</typeparam>
         /// <param name="classMapInitializer">The class map initializer.</param>
-        /// <returns>The class map.</returns>
-        public static BsonClassMap<TClass> TryRegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer)
+        /// <param name="classMap">The registered class map.</param>
+        /// <returns>false if already registered</returns>
+        public static bool TryRegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer, out BsonClassMap<TClass> classMap)
         {
-            var classMap = new BsonClassMap<TClass>(classMapInitializer);
-            TryRegisterClassMap(classMap);
-            return classMap;
+            classMap = new BsonClassMap<TClass>(classMapInitializer);
+            return TryRegisterClassMap(classMap);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Laurianti:
Hi,
Unfortunately this is not always the case, sometimes this registration is used in external libraries and it's not executed at the application startup.
For example we are using MassTransit and the RegisterClassMap call is executed when the library services are resolved by the IoC container, well after the application startup.
Another example where the startup approach doesn't work is when the HostBuilder/Host pattern (which have become the standard when creating a new .NET application) is leveraged to create and dispose multiple instances of the application (under the same process) for automated testing.
I think that this static configuration class doesn't fit anymore with the way the .NET ecosystem evolved, and doesn't play nice with the HostBuilder and the IoC container.
The best thing would be having an instance based configuration, this way every Host and/or every IoC container instantiated in the same program could have it's own class mapping registration, but I suppose that this would require substantial work to achieve.
Given what I've said before, I think that in the modern .NET ecosystem, having to relay on locks for registering the class maps becomes much more common, that's why I propose to have an atomic version of the registration out of the box.

rstam:
Class map registration is normally done at application startup, but it doesn't have to be. The only requirement is that it only be done once (and that it be done *before* it is first needed).

I think it is better for the calling code to keep track of whether it has registered class maps yet or not. That is more efficient than repeatedly calling a tentative registration method that will do nothing every time it is called after the first time.

Your point about the static nature of the class map registry is a good one. See https://jira.mongodb.org/browse/CSHARP-3985 for a proposal to make the configuration registries no longer be static.

rstam:
One recommended pattern for doing serialization configuration only once is something like:

```
public static class ClassMapRegisterer
{
    private static int __classMapsAreRegistered;

    public static void RegisterClassMaps()
    {
        if (Interlocked.CompareExchange(ref __classMapsAreRegistered, 1, 0) == 0)
        {
            BsonClassMap.RegisterClassMap<C>();
        }
    }
}
```

This is much more efficient than repeatedly calling a method like `TryRegisterClassMap`.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -391,6 +402,19 @@ public static BsonClassMap<TClass> RegisterClassMap<TClass>(Action<BsonClassMap<
...
+        /// <summary>
+        /// Creates and attempts to register a class map.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapInitializer">The class map initializer.</param>
+        /// <param name="classMap">The registered class map.</param>
+        /// <returns>false if already registered</returns>
+        public static bool TryRegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer, out BsonClassMap<TClass> classMap)
+        {
+            classMap = new BsonClassMap<TClass>(classMapInitializer);
```
rstam, line 414: This creates a new class map whether it is needed or not. Depending on how many times this method is called it could be very wasteful.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +439,39 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+        public static bool TryRegisterClassMap(BsonClassMap classMap)
+        {
+            var isRegistered = false;
+
+            if (classMap == null)
+            {
+                throw new ArgumentNullException("classMap");
+            }
+
+            BsonSerializer.ConfigLock.EnterWriteLock();
```
rstam, line 456: Taking a write lock when it probably isn't needed is going to interfere with any other threads that are trying to read from the class map registry.

Depending on how often this method is called it could be a performance issue.

rstam:
Thanks for providing more information about scenarios in which conditional class map registration might prove helpful.

I have created a JIRA ticket to track this feature request:

https://jira.mongodb.org/browse/CSHARP-4424

rstam:
@Laurianti could you take a look at https://github.com/mongodb/mongo-csharp-driver/pull/979 and let us know if that change satisfies your need?

In particular, do you feel a need for the method to be named `TryRegisterClassMap`? My thought is that whether or not the class was already registered or is registered as a result of this call is irrelevant. The only thing that matters is that after this call the class *is* registered.

Laurianti:
Thank you for asking @rstam. I have reviewed the changes in your PR (#979) and I am happy to say that it satisfies my requirement. I appreciate your efforts in addressing this issue. I will proceed to close my request. Thank you again for your help.

PR #944, opened 2022-11-10
CSHARP-3731: Verify that reported issue works in LINQ3.

rstam:
I hesitate to push this to master. Specially because it pollutes the global convention registry (but in a safe way).

Mainly it's just proving that CSHARP-3731 is no longer an issue.


JamesKovacs (approved):
LGTM

rstam:
Closed without merging to master.

PR #945, opened 2022-11-11
CSHARP-4343: Enable AWS ECS tests.

DmitryLukyanov:


evergreen/evergreen.yml
```diff
@@ -499,23 +498,57 @@ functions:
...
+        script: |
+          ${PREPARE_SHELL}
+          if [ "${skip_ECS_auth_test}" = "true" ]; then
+             echo "This platform does not support the ECS auth test, skipping..."
+             exit 0
+          fi
+          cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
+          . ./activate_venv.sh
+          echo "Project Directory: $PROJECT_DIRECTORY"
+          # SRC_DIRECTORY is workaround since EG_TOOLS expects "src" folder as a root
```
DmitryLukyanov, line 539: I've tried a number of attempts to move this logic on EG tools but with no luck. So here, I copy "mongo-csharp-driver" folder with driver sources into "src" and the content (`run-mongodb-aws-ecs-test.sh`) of "evergreen" folder to ".evergreen". All this structure is expected from EG_TOOLS scripts

evergreen/evergreen.yml
```diff
@@ -1509,6 +1548,7 @@ axes:
         display_name: "Windows 64-bit"
         variables:
           OS: "windows-64"
+          skip_ECS_auth_test: true
```
DmitryLukyanov, line 1551: ECS works only on linux

evergreen/evergreen.yml
```diff
@@ -536,7 +569,7 @@ functions:
         working_dir: mongo-csharp-driver
         script: |
           ${PREPARE_SHELL}
-          evergreen/run-mongodb-aws-test.sh
+          OS="${OS}" ASSERT_NO_URI_CREDS=true evergreen/run-mongodb-aws-test.sh
```
JamesKovacs, line 572: `${OS}` shouldn't have any spaces in it and thus doesn't need to be quoted. If we do quote it, we should quote it consistently on line 595 below.

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
```diff
@@ -131,6 +131,14 @@ async Task Find(IMongoClient client)
                 cursor.ToList();
             }
         }
+
+        [SkippableFact]
```
JamesKovacs, line 135: If this is a `SkippableFact`, shouldn't we have some `RequireEnvironment.Check()` or similar statement to start?

JamesKovacs (changes requested):
Overall looking good. Minor feedback. See below.

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
DmitryLukyanov: it was skippable, but not now. Yes, it should be changed

evergreen/evergreen.yml
DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
```diff
@@ -131,6 +131,14 @@ async Task Find(IMongoClient client)
                 cursor.ToList();
             }
         }
+
+        [Fact]
+        public void Ecs_should_fill_AWS_CONTAINER_CREDENTIALS_RELATIVE_URI()
+        {
+            var isEcs = Environment.GetEnvironmentVariable("AWS_ECS_TEST") != null;
+            var awsContainerRelativeUri = Environment.GetEnvironmentVariable("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI");
+            (awsContainerRelativeUri != null).Should().Be(isEcs);
```
DmitryLukyanov, line 140: this test was wrong, fixed

JamesKovacs (approved):
LGTM

PR #947, opened 2022-11-14
CSHARP-3740: Add native support for AWS IAM Roles for service accounts, EKS in particular.

DmitryLukyanov:
Asked a question about possible thead safe issue: https://github.com/aws/aws-sdk-net/discussions/2487

### DIFF
--- a/evergreen/evergreen.yml
+++ b/evergreen/evergreen.yml
@@ -432,7 +432,12 @@ functions:
               "iam_auth_assume_role_name" : "${iam_auth_assume_role_name}",
               "iam_auth_ec2_instance_account" : "${iam_auth_ec2_instance_account}",
               "iam_auth_ec2_instance_secret_access_key" : "${iam_auth_ec2_instance_secret_access_key}",
-              "iam_auth_ec2_instance_profile" : "${iam_auth_ec2_instance_profile}"
+              "iam_auth_ec2_instance_profile" : "${iam_auth_ec2_instance_profile}",
+              "iam_auth_assume_web_role_name": "${iam_auth_assume_web_role_name}",
+              "iam_web_identity_issuer": "${iam_web_identity_issuer}",
+              "iam_web_identity_rsa_key": "${iam_web_identity_rsa_key}",
+              "iam_web_identity_jwks_uri": "${iam_web_identity_jwks_uri}",
+              "iam_web_identity_token_file": "${iam_web_identity_token_file}"
           }
           EOF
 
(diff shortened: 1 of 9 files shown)


evergreen/evergreen.yml
```diff
@@ -551,6 +556,76 @@ functions:
...
+      params:
+        working_dir: mongo-csharp-driver
+        script: |
+          ${PREPARE_SHELL}
+          if [ "${skip_web_identity_auth_test}" = "true" ]; then
+             echo "This platform does not support the web identity auth test, skipping..."
+             exit 0
+          fi
+          PROJECT_DIRECTORY=${PROJECT_DIRECTORY} OS=$OS  ASSERT_NO_URI_CREDS=true evergreen/run-mongodb-aws-test.sh
+
```
JamesKovacs, line 628: Why are we running `run-mongodb-aws-test.sh` twice, once without `AWS_ROLE_SESSION_NAME` set and once with it set? If we are testing two different scenarios, we should probably separate this into two separate functions. Otherwise if one fails, it will not be immediately obvious which one. We would have to check whether `AWS_ROLE_SESSION_NAME` is set or not to understand which test case failed. Thoughts?

evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -9,7 +9,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 #
 # Environment variables used as output:
 #       AWS_TESTS_ENABLED       Allows runnings AWS tests
-#       AWS_ECS_TEST            Allows runnings EVS tests
+#       AWS_ECS_ENABLED         Allows runnings ECS tests
```
JamesKovacs, line 12: Allows runnings => Allows running

src/MongoDB.Driver.Core/Core/Authentication/External/AwsAuthenticationCredentialsProvider.cs
```diff
@@ -56,183 +60,28 @@ public AwsCredentials(string accessKeyId, SecureString secretAccessKey, string s
...
-        private readonly IEnvironmentVariableProvider _environmentVariableProvider;
+        public void Clear() => FallbackCredentialsFactory.Reset();
 
-        public AwsAuthenticationCredentialsProvider(IHttpClientWrapper httpClientWrapper, IEnvironmentVariableProvider environmentVariableProvider)
+        public AwsCredentials CreateCredentialsFromExternalSource(CancellationToken cancellationToken)
         {
-            _awsHttpClientHelper = new AwsHttpClientHelper(httpClientWrapper);
-            _environmentVariableProvider = Ensure.IsNotNull(environmentVariableProvider, nameof(environmentVariableProvider));
+            var creds = FallbackCredentialsFactory.GetCredentials();
+            var immutableCredentias = creds.GetCredentials();
```
JamesKovacs, line 70: `immutableCredentias` => `immutableCredentials`

src/MongoDB.Driver.Core/Core/Authentication/External/AwsAuthenticationCredentialsProvider.cs
```diff
@@ -56,183 +60,28 @@ public AwsCredentials(string accessKeyId, SecureString secretAccessKey, string s
...
-                result = null;
-                if (DateTime.TryParse(value, out var dateTime))
-                {
-                    result = dateTime;
-                    return true;
-                }
-                return false;
-            }
+            var creds = FallbackCredentialsFactory.GetCredentials();
+            var immutableCredentias = await creds.GetCredentialsAsync().ConfigureAwait(false);
```
JamesKovacs, line 77: `immutableCredentias` => `immutableCredentials`

src/MongoDB.Driver.Core/Core/Authentication/External/AwsAuthenticationCredentialsProvider.cs
```diff
@@ -14,38 +14,42 @@
 */
 
 using System;
-using System.Net.Http;
 using System.Security;
 using System.Threading;
 using System.Threading.Tasks;
+using Amazon.Runtime;
```
JamesKovacs, line 20: Shouldn't switching to use the AWS SDK be done in the context of https://jira.mongodb.org/browse/CSHARP-4390?

JamesKovacs: If the AWS SDK is responsible for caching, do we need any of this caching logic that was introduced in https://jira.mongodb.org/browse/CSHARP-4273?

JamesKovacs (changes requested):
Mostly questions with a few minor corrections.

src/MongoDB.Driver.Core/Core/Authentication/External/AwsAuthenticationCredentialsProvider.cs
JamesKovacs: Good point about Azure KMS.

While I'm not enthusiastic about splitting the CSHARP-4390 work between these two tickets, I understand your reasoning. It doesn't make sense to implement something in this ticket to only revert it in the next. Accepted.

JamesKovacs (approved):
LGTM

JamesKovacs (approved):
LGTM

PR #948, opened 2022-11-14
CSHARP-4424: Add a BsonClassMap method that registers a class map only if it has not already been registered.

rstam:


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -378,6 +378,17 @@ public static BsonClassMap<TClass> RegisterClassMap<TClass>()
             return RegisterClassMap<TClass>(cm => { cm.AutoMap(); });
         }
 
+        /// <summary>
+        /// Creates and attempts to register a class map.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMap">The registered class map.</param>
+        /// <returns>false if already registered</returns>
+        public static bool TryRegisterClassMap<TClass>(out BsonClassMap<TClass> classMap)
```
rstam, line 387: This method is from the original PR.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -391,6 +402,71 @@ public static BsonClassMap<TClass> RegisterClassMap<TClass>(Action<BsonClassMap<
             return classMap;
         }
 
+        /// <summary>
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapInitializer">The class map initializer.</param>
+        public static void RegisterClassMapIfNotAlreadyRegistered<TClass>(Action<BsonClassMap<TClass>> classMapInitializer)
```
rstam, line 410: This is my proposed alternative, assuming we want to support something like this at all.

This method is a slight variation of the `LookupClassMap` method.

Design goals:

* don't create a new class map unless it's needed
* don't take a write lock unless it's needed

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -391,6 +402,71 @@ public static BsonClassMap<TClass> RegisterClassMap<TClass>(Action<BsonClassMap<
...
+        }
+
+        /// <summary>
+        /// Creates and attempts to register a class map.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapInitializer">The class map initializer.</param>
+        /// <param name="classMap">The registered class map.</param>
+        /// <returns>false if already registered</returns>
+        public static bool TryRegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer, out BsonClassMap<TClass> classMap)
```
rstam, line 464: This method is from the original PR.

Main concern: it creates a new class map *every* time it is called whether it is needed or not.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +491,39 @@ public static void RegisterClassMap(BsonClassMap classMap)
             }
         }
 
+        /// <summary>
+        /// Attempts to register a class map.
+        /// </summary>
+        /// <param name="classMap">The class map.</param>
+        /// <returns>false if already registered</returns>
+        public static bool TryRegisterClassMap(BsonClassMap classMap)
```
rstam, line 499: This method is from the original PR.

Main concern: it takes a write lock every time, even when the class map is already registered (which is true every single time this method is called after the first time).

rstam:
See https://jira.mongodb.org/browse/CSHARP-4424

JamesKovacs (commented):
Overall seems like a reasonable request. There is debate over whether the method should be named `RegisterClassMapIfNotRegistered` or `TryRegisterClassMap`. I favour the first name as neither the `bool` return value nor the `out BsonClassMap` parameter has value and will likely be more confusing than helpful.

We can close this PR and create a CSHARP ticket to track this work.

rstam:
Closing this PR.

Track https://jira.mongodb.org/browse/CSHARP-4424 for further work.

PR #950, opened 2022-11-14
CSHARP-3237: Test that ToList works in LINQ3.

rstam:
I don't think this repro needs to be pushed to master.


JamesKovacs (approved):
LGTM

rstam:
Closed without merging to master.

PR #953, opened 2022-11-16
CSHARP-2113: Support LINQ queries against interfaces.

rstam:


src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -23,11 +23,24 @@ namespace MongoDB.Bson.Serialization.Serializers
...
+            classMap.AutoMap();
+            classMap.Freeze();
+            return new BsonClassMapSerializer<TInterface>(classMap);
+        }
+        #endregion
+
         // private fields
         private readonly Type _interfaceType;
         private readonly IDiscriminatorConvention _discriminatorConvention;
+        private readonly IBsonSerializer<TInterface> _interfaceSerializer;
```
rstam, line 43: In order to implement `TryGetMemberSerializationInfo` we need some serializer to delegate the question to.

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -110,5 +124,17 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
                 _objectSerializer.Serialize(context, args, value);
             }
         }
+
+        /// <inheritdoc/>
+        public bool TryGetMemberSerializationInfo(string memberName, out BsonSerializationInfo serializationInfo)
+        {
+            if (_interfaceSerializer is IBsonDocumentSerializer documentSerializer)
```
rstam, line 131: The `if` is a little redundant because we created the serializer so we presumably already know that it implements `IBsonDocumentSerializer`.

rstam (commented):
For some reason I don't feel 100% good about this one, but I can't explain why.

I think it's because I feel there are hidden gotchas around support for interfaces that we aren't seeing here.

Even so, I don't think this change would make anything worse.

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -23,11 +23,24 @@ namespace MongoDB.Bson.Serialization.Serializers
...
     {
+        #region static
+        private static IBsonSerializer<TInterface> CreateInterfaceSerializer()
+        {
+            var classMapDefinition = typeof(BsonClassMap<>);
+            var classMapType = classMapDefinition.MakeGenericType(typeof(TInterface));
+            var classMap = (BsonClassMap)Activator.CreateInstance(classMapType);
+            classMap.AutoMap();
+            classMap.Freeze();
+            return new BsonClassMapSerializer<TInterface>(classMap);
```
rstam, line 36: The key here was to not call `BsonClassMap.LookupClassMap` because that tries to register the class map and throws an error if the type is an interface.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
```diff
@@ -0,0 +1,93 @@
...
+        public interface IMakeTheTestFail
+        {
+            int I { get; set; }
+        }
+
+        public class FailingType : IMakeTheTestFail
+        {
+            public IMakeTheTestFail Inner { get; set; }
+            public int I { get; set; }
+        }
```
JamesKovacs, line 85: `IMakeTheTestFail` and `FailingType` no longer make sense because those tests pass with the changes to `DiscriminatedInterfaceSerializer`.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            var collection = GetCollection<FailingType>();
+
+            var queryable = collection
+                .AsQueryable()
+                .Where(x => x.Inner.I == 1);
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { 'Inner.I' : 1 } }");
+        }
```
JamesKovacs, line 74: All four tests should be integration tests. We will need a type discriminator on the subdocument to allow deserializing.

JamesKovacs (changes requested):
I understand your hesitancy and the feeling that there might be some deeper issues around interfaces that need addressing. But I concur that these changes shouldn't break any currently working use cases and allow some use cases that currently do not work.

Let's proceed with making these integration tests and ensuring that results can be deserialized correctly.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
rstam: Will do.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
rstam: Those are the names from the ticket.

I'll think about how to refactor this with new names that still tests the same thing.

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
rstam: No longer redundant because the user might supply a custom serializer.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
rstam: Done.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp2113Tests.cs
rstam: Done.

rstam (commented):
All tests are now integration tests.

JamesKovacs (approved):
LGTM

PR #956, opened 2022-11-18
Atlas search POC

BorisDog:
Based on @dgolub  [project](https://github.com/mongodb-labs/mongo-csharp-search)

Sample usage:
`coll.Aggregate().Search(SearchBuilders<Person>.Search.Text(x => x.Title, "cat")).ToList()`


src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -83,16 +83,18 @@ public static T IsEqualTo<T>(T value, T comparand, string paramName)
         }
 
         /// <summary>
-        /// Ensures that the value of a parameter is greater than or equal to zero.
+        /// Ensures that the value of a parameter is greater than or equal to a comparand.
```
dgolub, line 86:
```suggestion
        /// Ensures that the value of a parameter is greater than a comparand.
```

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -83,16 +83,18 @@ public static T IsEqualTo<T>(T value, T comparand, string paramName)
...
         /// <param name="paramName">The name of the parameter.</param>
         /// <returns>The value of the parameter.</returns>
-        public static int IsGreaterThanOrEqualToZero(int value, string paramName)
+        public static T IsGreaterThan<T>(T value, T comparand, string paramName) where T : IComparable<T>
         {
-            if (value < 0)
+            if (value.CompareTo(comparand) <= 0)
             {
-                var message = string.Format("Value is not greater than or equal to 0: {0}.", value);
+                var message = $"Value is not greater than or equal to {comparand}: {value}.";
```
dgolub, line 97:
```suggestion
                var message = $"Value is not greater than {comparand}: {value}.";
```

src/MongoDB.Driver/Search/AggregateFluentBuilder.cs
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Extension methods for <see cref="IAggregateFluent{TResult}"/>.
+    /// </summary>
+    public static class AggregateFluentBuilder
```
dgolub, line 22: The reason why I created a separate `AggregateFluentBuilder` class with extension methods rather than modifying `AggregateFluent` was that it was necessary in order to have Atlas Search in a separate assembly.  Now that we're integrating it into the driver, we might want to reconsider this decision.

src/MongoDB.Driver/Search/FuzzyOptions.cs
```diff
@@ -0,0 +1,65 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
```
dgolub, line 18: Should the namespace here be `MongoDB.Driver.Search`, given that it's in a subdirectory?

src/MongoDB.Driver/Search/ScoreDefinitionBuilder.cs
```diff
@@ -0,0 +1,132 @@
...
+            var document = new BsonDocument()
+            {
+                { "path", _path.Render(documentSerializer, serializerRegistry) },
+                { "undefined", _undefined, _undefined != 0 }
+            };
+
+            return new("boost", document);
+        }
+    }
+    internal sealed class ConstantScoreDefinition<TDocument> : ScoreDefinition<TDocument>
```
dgolub, line 118:
```suggestion
    }
    
    internal sealed class ConstantScoreDefinition<TDocument> : ScoreDefinition<TDocument>
```

src/MongoDB.Driver/Search/SearchBuilders.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// A static helper class containing various builders pertaining to Atlas Search.
+    /// </summary>
+    /// <typeparam name="TDocument">The type of the document.</typeparam>
+    public static class SearchBuilders<TDocument>
```
dgolub, line 21: The reason why I introduced a new `SearchBuilders` class rather than extending the existing `Builders` class was that it isn't possible to add static extension methods to a class.  This seems like another decision that would be worth reconsidering now that we're no longer forced into it.  I imagine that it would be a better experience for customers not to have to keep track of type different classes.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <returns>A text search definition.</returns>
+        public SearchDefinition<TDocument> Text<TField>(
+            QueryDefinition query,
+            Expression<Func<TDocument, TField>> path,
+            FuzzyOptions fuzzy = null,
+            ScoreDefinition<TDocument> score = null) =>
+            Text(query, new ExpressionFieldDefinition<TDocument>(path), fuzzy, score);
+    }
+
+    internal sealed class TextSearchDefinition<TDocument> : SearchDefinition<TDocument>
```
dgolub, line 63: We're restricting it to just `Text` for now and not adding support for any of the other operators?

dgolub (commented):
Integration looks good!  I left a few minor comments and suggestions.

JamesKovacs (changes requested):
Overall looking good. I like the overall direction. @dgolub raises some valid points that we should address. (Thank you, @dgolub!)

A few broad comments...
- We need more testing.
- We should determine if we can integration test given that `$search` only works on Atlas. Good question for the Java team since they implemented `$search` already.
- Support for additional operators beyond `Text`.
- Support for LINQ.
- Investigate whether it is possible to provide a friendlier error message than `MongoServerError: $search not enabled! Enable Search by setting serverParameter mongotHost to a valid "host:port" string`.

With respect to LINQ, we can probably limit it to chaining off of `AsQueryable()`. While it would be nice to support in `where` comprehensions, it adds complexity because `$search` must be the first operator in the pipeline.

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -104,79 +106,62 @@ public static int IsGreaterThanOrEqualToZero(int value, string paramName)
...
-        {
-            if (value < TimeSpan.Zero)
-            {
-                var message = string.Format("Value is not greater than or equal to zero: {0}.", TimeSpanParser.ToString(value));
-                throw new ArgumentOutOfRangeException(paramName, message);
-            }
-            return value;
-        }
+        public static long IsGreaterThanOrEqualToZero(long value, string paramName) =>
+            IsGreaterThanOrEqualTo(value, 0, paramName);
```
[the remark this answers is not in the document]
rstam: Same might apply to some of the other refactorings.

Do we really want one Ensure method calling another? Or should each Ensure method be complete by itself?

rstam (commented):
I think the overall approach integrates with how we support existing pipeline stages.

I've added a few comments here and there on the POC code.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1309,6 +1309,57 @@ public static class PipelineStageDefinitionBuilder
...
+        /// Flag that specifies whether to perform a full document lookup on the backend database
+        /// or return only stored source fields directly from Atlas Search.
+        /// </param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
+            SearchDefinition<TInput> query,
+            HighlightOptions<TInput> highlight = null,
+            string indexName = null,
+            SearchCountOptions count = null,
+            bool returnStoredSource = false)
```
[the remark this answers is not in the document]
dgolub: The JSON fields in `$search` are called `highlight` and `count`, but I have no objection to this change if you feel that it will make it easier for customers to understand what the parameters do.

src/MongoDB.Driver/Search/FuzzyOptions.cs
```diff
@@ -0,0 +1,65 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Search
```
rstam, line 18: I know there's talk about moving all these classes up one level in the namespace, but maybe we should leave them here.

Being in their own namespace could help in avoiding future name collisions.

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -298,6 +283,24 @@ public static string IsNotNullOrEmpty(string value, string paramName)
...
+        /// <param name="value">The value of the parameter.</param>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <param name="paramName">The name of the parameter.</param>
+        /// <returns>The value of the parameter.</returns>
+        public static T? IsNullOrBetween<T>(T? value, T min, T max, string paramName) where T : struct, IComparable<T>
+        {
+            if (value != null)
+            {
+                Ensure.IsBetween(value.Value, min, max, paramName);
```
rstam, line 299: The error message will be wrong when value is not null but not between.

The error message won't mention that null is valid.

This is probably a reason for each Ensure method to be self contained.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: I realized that the server uses the shorter names.

It's a conflict between our naming conventions and the server's.

I made this comment because normally a parameter names `count` would be of type `int`.

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
BorisDog: That's a good catch.
I still think that we should and can easily reuse the code. 
We got similar behaviour in the old `IsNull*` methods. I suggest to fix them to in a separate ticket.

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,29 @@ namespace MongoDB.Driver
     /// <typeparam name="TDocument">The type of the document.</typeparam>
     public static class Builders<TDocument>
     {
-        private static FilterDefinitionBuilder<TDocument> __filter = new FilterDefinitionBuilder<TDocument>();
-        private static IndexKeysDefinitionBuilder<TDocument> __index = new IndexKeysDefinitionBuilder<TDocument>();
-        private static ProjectionDefinitionBuilder<TDocument> __projection = new ProjectionDefinitionBuilder<TDocument>();
-        private static SortDefinitionBuilder<TDocument> __sort = new SortDefinitionBuilder<TDocument>();
-        private static UpdateDefinitionBuilder<TDocument> __update = new UpdateDefinitionBuilder<TDocument>();
```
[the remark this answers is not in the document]
rstam: My counter argument is that most of our classes already have private fields using the `_field` or `__staticField` naming convention.

In many cases (such as when validating in setters) an explicit backing field is required.

Why not just standardize on explicit backing fields and that way we can always know by glancing at the top of a class what state the class maintains.

Otherwise we have to tediously search the source file to see if there are any autoproperties we didn't know about.

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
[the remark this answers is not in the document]
rstam: Of course! But it changes the error messages.

src/MongoDB.Driver/Search/FuzzyOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Options for fuzzy search.
+    /// </summary>
+    public sealed class FuzzyOptions
+    {
+        private int? _maxEdits;
+        private int? _prefixLength;
+        private int? _maxExpansions;
```
rstam, line 27: This is an example of a class where we have to use explicit backing fields.

I would like us to standardize on this approach even in cases where autoproperties could be used.

src/MongoDB.Driver/Search/HighlightOptionsBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <param name="maxNumPassages">
+        /// The number of high-scoring passages to return per document in the highlighting results
+        /// for each field.
+        /// </param>
+        /// <returns>Highlighting options.</returns>
+        public HighlightOptions<TDocument> Options(
+            PathDefinition<TDocument> path,
+            int? maxCharsToExamine = null,
+            int? maxNumPassages = null)
+            => new()
```
rstam, line 43: You don't think `=>` should be indented one more tab stop?

Better yet use a method body... :)

src/MongoDB.Driver/Search/HighlightOptionsBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
```
rstam, line 44: You don't think `{` should be indented on tab stop further than `=> new()`?

src/MongoDB.Driver/Search/FuzzyOptions.cs
BorisDog: Yes there always will be cases where non trivial setter is required.
That should not deter us from using autoproperties for simpler classes. And especially for cases where set is private or init only.

src/MongoDB.Driver/Builders.cs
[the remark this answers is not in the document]
BorisDog: I think the key here is finding a good balance between following the existing style and evolving our code base to use new language features and mainstream standards. We have already adopted some mainstream conventions like string interpolation, `nameof`, local methods and other, despite having most of the code written differently. I hope we continue to do so.
Autoproperties are a very good example of a "mainstream" and very loved feature by C# community (which continues to evolve), and I believe it is perceived as sort of "default" in C# dev process.

Like with fields, all properties (auto implemented or not) through the source file, they will always reside in same section, like fields. Given that, and the fact the major part of code is cleaned up, the "state" of the class appears more clear to me. Autoproperties indicate that this is just a simple data holders, no custom properties, no special relation between property and field that needs to be inspected. 

For more complicated classes,  where inspecting class state/functionality becomes is more complicated, and I find using VS built-in class/document explorers very useful.

BorisDog:
Follow up work in [CSHARP-4440](https://jira.mongodb.org/browse/CSHARP-4440).

PR #961, opened 2022-11-22
CSHARP-4255: Automatically create Queryable Encryption keys.

DmitryLukyanov:


src/MongoDB.Driver/Encryption/ClientEncryption.cs
```diff
@@ -78,6 +78,64 @@ public ClientEncryption(ClientEncryptionOptions clientEncryptionOptions)
...
 
+        /// <summary>
+        /// Create encrypted collection.
+        /// </summary>
+        /// <param name="collectionNamespace">The collection namespace.</param>
+        /// <param name="createCollectionOptions">The create collection options.</param>
+        /// <param name="kmsProvider">The kms provider.</param>
+        /// <param name="dataKeyOptions">The datakey options.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public (IMongoCollection<TCollection> Collection, BsonDocument EncryptedFields) CreateEncryptedCollection<TCollection>(CollectionNamespace collectionNamespace, CreateCollectionOptions createCollectionOptions, string kmsProvider, DataKeyOptions dataKeyOptions, CancellationToken cancellationToken = default)
```
DmitryLukyanov, line 89: I'm going to discuss whether we can add a different return type than this, so might be not the final form

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -84,6 +83,69 @@ public ClientEncryptionProseTests(ITestOutputHelper testOutputHelper)
...
+                        case 3: // Case 3: Invalid ``keyId``
+                            {
+                                var effectiveEncryptedFields = encryptedFields.DeepClone();
+                                effectiveEncryptedFields["fields"].AsBsonArray[0].AsBsonDocument["keyId"] = true;
+                                var collection = CreateEncryptedCollection(client, clientEncryption, __collCollectionNamespace, encryptedFields, kmsProvider, async).Collection;
+
+                                var exception = Record.Exception(() => Insert(collection, async, new BsonDocument("ssn", "123-45-6789")));
+                                exception.Should().BeOfType<MongoBulkWriteException<BsonDocument>>().Which.Message.Should().Contain("Document failed validation");
+                            }
+                            break;
```
DmitryLukyanov, line 142: The positive scenario is not in the spec, but I'm going to discuss adding it, so one more case can be added here

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
```diff
@@ -76,6 +76,40 @@ public static BsonDocument GetEffectiveEncryptedFields(CollectionNamespace colle
...
+        public static IEnumerable<BsonDocument> IterateEmptyKeyIds(CollectionNamespace collectionNamespace, BsonDocument encryptedFields)
+        {
+            if (!EncryptedCollectionHelper.TryGetEffectiveEncryptedFields(collectionNamespace, encryptedFields, encryptedFieldsMap: null, out var storedEncryptedFields))
+            {
+                throw new InvalidOperationException("There are no encrypted fields defined for the collection.");
+            }
+
+            if (storedEncryptedFields.TryGetValue("fields", out var fields) && fields is BsonArray fieldsArray)
+            {
+                foreach (var field in fieldsArray)
```
BorisDog, line 88: `fieldsArray.OfType<BsonDocument>()` ?
or even
```
fieldsArray
    .OfType< BsonDocument>()
    .Where(b => b.TryGetElement("keyId", out var keyId) && keyId.Value == BsonNull.Value)
```

src/MongoDB.Driver.Core/Core/Encryption/EncryptedCollectionHelper.cs
```diff
@@ -76,6 +76,40 @@ public static BsonDocument GetEffectiveEncryptedFields(CollectionNamespace colle
...
+                    }
+                    else
+                    {
+                        // If `F` is not a document element, skip it.
+                        continue;
+                    }
+                }
+            }
+
+            yield break;
```
BorisDog, line 105: Not needed?

BorisDog (approved):
LGTM

src/MongoDB.Driver/Encryption/ClientEncryption.cs
DmitryLukyanov: I've discussed this question with spec author and he's confirmed that we can define public API in the consistent way with previous driver's API methods. So, I consider 2 changes:
1. Do not return created collection from the helper [method](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/IMongoDatabase.cs#L141). Instead do the same logic as we do with a regular CreateCollection method ie return nothing and request retrieving collection via GetCollection. 
2. Instead returning EncryptedFields, we can simply modify input EncryptedFields in the provided options.

I'm not too confident about these changes (as well as about the initial requirement), but I would probably want to avoid returning tuple in public API, thoughts?

rstam: I think Tuples are not recommended in public APIs.

Agreed that there is no need to return the `IMongoCollection<TDocument>` .

Do we need to return anything?

Also, shouldn't the type parameter be `TDocument` (not `TCollection`). But if we don't return the `IMongoCollection<TDocument>` I don't think this type parameter is needed any more anyway.

DmitryLukyanov: >Do we need to return anything?

yeah, I also think it can be simply `void` method

src/MongoDB.Driver/Encryption/ClientEncryption.cs
```diff
@@ -78,6 +78,64 @@ public ClientEncryption(ClientEncryptionOptions clientEncryptionOptions)
...
+        {
+            var effectiveEncryptedFields = createCollectionOptions?.EncryptedFields?.DeepClone()?.AsBsonDocument;
+
+            foreach (var fieldDocument in EncryptedCollectionHelper.IterateEmptyKeyIds(collectionNamespace, effectiveEncryptedFields))
+            {
+                var dataKey = CreateDataKey(kmsProvider, dataKeyOptions, cancellationToken);
+                EncryptedCollectionHelper.ModifyEndryptedFields(fieldDocument, dataKey);
```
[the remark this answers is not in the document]
DmitryLukyanov: true

PR #962, opened 2022-11-22
CSHARP-2727: Skip test using $toString when testing against server 3.6.

rstam:


JamesKovacs (approved):
LGTM

FYI - I found another use of `$toString` that isn’t guarded by a server check in our LINQ3 `CSharp3524Tests.cs` and wondered why it wasn’t failing. We only assert the MQL but never run it against a server. Probably not worth going back and turning it into an integration test now. All tests where we run `$toString` against a server now have the necessary `RequireServer.Check()` in place.

PR #970, opened 2022-11-24
CSHARP-4427: Negative array indexes should throw ExpressionNotSupportedException.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/GetItemMethodToFilterFieldTranslator.cs
```diff
@@ -43,6 +43,15 @@ public static AstFilterField Translate(TranslationContext context, MethodCallExp
                 if (indexExpression.Type == typeof(int))
                 {
                     var index = indexExpression.GetConstantValue<int>(containingExpression: expression);
+                    if (index < 0)
```
rstam, line 46: This check was missing.

tests/MongoDB.Driver.Tests/UpdateDefinitionBuilderTests.cs
```diff
@@ -243,13 +243,13 @@ public void Inc_Typed()
         public void Incorrect_index_should_throw_expected_exception_with_set()
         {
             var subject = CreateSubject<Person>();
-            string expectedErrorMessage = "Array indexes must be greater than or equal to -1.";
+            string expectedErrorMessage = "because negative indexes are not valid";
```
rstam, line 246: LINQ3 throws `ExpressionNotSupportedException` instead of `IndexOutOfRangeException` and the error message is slightly different.

tests/MongoDB.Driver.Tests/UpdateDefinitionBuilderTests.cs
```diff
@@ -670,9 +670,14 @@ private void Assert<TDocument>(UpdateDefinition<TDocument> update, string expect
 
         private void AssertThrow<TDocument, TException>(UpdateDefinition<TDocument> update, string errorMessage) where TException : Exception
         {
-            var exception = Record.Exception(() => { Render(update); });
+            AssertThrow<TDocument, TException>(update, errorMessage, LinqProvider.V2);
```
rstam, line 673: Having two overloads of `AssertThrow` is a temporary condition because in this PR all the other tests in this file are still using LINQ2.

JamesKovacs (approved):
LGTM. I like that the exception message provides clear guidance on what they should use instead of -1.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/ArraySerializerHelper.cs
```diff
@@ -38,5 +39,24 @@ public static IBsonSerializer GetItemSerializer(IBsonSerializer serializer)
                 throw new InvalidOperationException($"{serializer.GetType().FullName} must implement IBsonArraySerializer to be used with LINQ.");
             }
         }
+
+        public static IBsonSerializer GetItemSerializer(Expression expression, IBsonSerializer serializer)
```
rstam, line 43: Same as the above method but with `expression` parameter so we can throw `ExpressionNotSupportedException`  instead.

Maybe ALL callers of the original method should be switched to call the new method?

JamesKovacs (approved):
LGTM

PR #972, opened 2022-11-24
CSHARP-4429: Support resultSerializer = null in MongoQueryable.AppendStage

BorisDog:


tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4234Tests.cs
```diff
@@ -16,19 +16,21 @@
...
-        [Fact]
-        public void AppendStage_should_work()
+        [Theory]
+        [ParameterAttributeData]
+        public void AppendStage_should_work([Values(false, true)] bool resultSerializerExists)
         {
             var collection = CreateProductsCollection();
             var textStage = "{ $match : { $text : { $search : 'apples' } } }";
-            var resultSerializer = collection.DocumentSerializer;
+            var resultSerializer = resultSerializerExists ? collection.DocumentSerializer : null;
```
rstam, line 33: I think we often use a name like `useResultSerializer` for a test method parameter that says whether to use a result serializer or not, but the naming is pretty flexible.

rstam (approved):
LGTM

PR #974, opened 2022-11-25
CSHARP-4417: ConnectionId should use longs for LocalValue and ServerValue

JamesKovacs:
ConnectionId should use longs for LocalValue and ServerValue to match 64-bit integer connectionId returned from the server.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/ConnectionId.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/ConnectionId.cs
@@ -28,8 +28,8 @@ namespace MongoDB.Driver.Core.Connections
     {
         // fields
         private readonly ServerId _serverId;
-        private readonly int _localValue;
-        private readonly int? _serverValue;
+        private readonly long _localValue;
+        private readonly long? _serverValue;
         private readonly int _hashCode;
 
         // constructors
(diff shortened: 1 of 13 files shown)


src/MongoDB.Driver.Core/Core/Connections/ConnectionId.cs
```diff
@@ -81,7 +81,7 @@ public ServerId ServerId
         /// <value>
         /// The local value.
         /// </value>
-        public int LocalValue
+        public long LocalValue
```
rstam, line 84: Is this a breaking change?

Maybe we should add a new property, `LongLocalvalue` and leave this one alone.

This would be similar to `Enumerable.LongCount` being added while leaving `Enumerable.Count` alone.

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Connections/ConnectionId.cs
JamesKovacs: Yes, this would be a (hopefully minor) breaking change that we could put in 2.19.0. As much as I don't care for `LocalValue` (int) and `LongLocalValue` (long), it is a good suggestion for making this an additive change rather than a breaking one. Good suggestion.

JamesKovacs: Yes, this would be a (hopefully minor) breaking change that we could put in 2.19.0. As much as I don't care for `LocalValue` (int) and `LongLocalValue` (long), it is a good suggestion for making this an additive change rather than a breaking one. Good suggestion.

rstam (approved):
LGTM

PR #976, opened 2022-11-25
CSHARP-4413: Equality operator of MongoClientSettings is bugged

BorisDog:


src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -1109,7 +1109,7 @@ public override bool Equals(object obj)
                 _useTls == rhs._useTls &&
                 _waitQueueSize == rhs._waitQueueSize &&
                 _waitQueueTimeout == rhs._waitQueueTimeout &&
-                _writeConcern == rhs._writeConcern &&
+                object.Equals(_writeConcern, rhs._writeConcern) &&
```
DmitryLukyanov, line 1112: maybe better implementing `operator ==`?

BorisDog: I agree, I thought this might be BC, but looks like it's minor. Thoughts @DmitryLukyanov ?

DmitryLukyanov: yeah, it's a BC, but I also think it's minor and I don't see how it can affect users, so I think it's better to do it

rstam: I think as a general rule `Equals` should be implemented in terms of `Equals`.

Sometimes we rely on `==` being equivalent to `Equals` but to me it's clearer if `Equals` is implemented using `Equals`.

BorisDog: @DmitryLukyanov we address this in [CSHARP-4432](https://jira.mongodb.org/browse/CSHARP-4432)

DmitryLukyanov (approved):
LGTM

PR #979, opened 2022-11-28
CSHARP-4424: Added RegisterClassMapIfNotRegistered method to BsonClassMap.

rstam:


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,58 @@ public static void RegisterClassMap(BsonClassMap classMap)
             }
         }
 
+        /// <summary>
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapInitializer">The class map initializer.</param>
+        public static void RegisterClassMapIfNotRegistered<TClass>(Action<BsonClassMap<TClass>> classMapInitializer)
```
rstam, line 423: This method is a clone of LookupClassMap modified slightly.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,58 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+
+            BsonSerializer.ConfigLock.EnterReadLock();
+            try
+            {
+                BsonClassMap classMap;
+                if (__classMaps.TryGetValue(typeof(TClass), out classMap))
+                {
+                    if (classMap.IsFrozen)
+                    {
+                        return;
```
rstam, line 438: On every call after the first we will exit here because the class map is already registered.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,58 @@ public static void RegisterClassMap(BsonClassMap classMap)
             }
         }
 
+        /// <summary>
```
BorisDog, line 418: We probably should provide additional overload that receives `BsonClassMap` for convenience and parity with `RegisterClassMap(BsonClassMap classMap)`.

Also this overload seems to be the original request.
Also Github search shows that `RegisterClassMap<TClass>()` is quite popular.

rstam: As I commented on the original PR an overload that takes a `BsonClassMap` is a recipe for performance problems because the caller will probably be creating a new class map for every call but usually the class map will simply be ignored.

Given that do you still think such an overload is warranted?

I'm surprised anyone calls `RegisterClassMap<TClass>()`. It's what happens automatically even if you don't call it.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: I'm not totally opposed to the `TryRegisterClassMap` approach, it just seems pointless.

Of what use is it to the caller to know whether the class was already registered or not? The only thing that matters is that one war or the other it got registered.

rstam: I've asked the user on the original PR whether knowing whether the class was already registered or not is of any use (i.e. the `bool` return value of a `Try` variant).

BorisDog: I agree, return value does not seem very useful in our case, it's very similar to [IServiceCollection.TryAdd](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryadd?view=dotnet-plat-ext-7.0#microsoft-extensions-dependencyinjection-extensions-servicecollectiondescriptorextensions-tryadd(microsoft-extensions-dependencyinjection-iservicecollection-microsoft-extensions-dependencyinjection-servicedescriptor)).
I was just referring to the naming.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+            }
+        }
+
+        /// <summary>
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMap">The class map.</param>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>(BsonClassMap<TClass> classMap)
```
rstam, line 441: This is the only overload that (potentially) requires a new classMap to be instantiated on every call.

I don't think an application should use this overload, but if they do it's up to them to try and use a single instance of the classMap if they call this method more than once.

rstam (commented):
Implemented a suite of `TryRegisterClassMap` (instead of a single `RegisterClassMapIfNotRegistered` method) overloads at Boris's request. He felt strongly that this naming would be more familiar to users.

He's right about that. The `bool` return value seems rather pointless, but it's part of the `Try` naming convention so it is what it is.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+        /// <summary>
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>()
+        {
+            return TryRegisterClassMap(ClassMapFactory);
+
+            static BsonClassMap<TClass> ClassMapFactory()
```
BorisDog, line 427: Nice use of static local method. We should use them more where appropriate.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
             }
         }
 
+        /// <summary>
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>()
```
BorisDog, line 423: Wondering if we should use this opportunity to reuse `static ClassMapFactory` in `RegisterClassMap<T>()` to skip lambda creation.
Minor, as this is not on the hot path.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+            if (classMapInitializer == null)
+            {
+                throw new ArgumentNullException("classMapInitializer");
+            }
+
+            return TryRegisterClassMap(ClassMapFactory);
+
+            BsonClassMap<TClass> ClassMapFactory()
+            {
+                var classMap = new BsonClassMap<TClass>();
```
BorisDog, line 473: Minor: should we use `new BsonClassMap<TClass>(classMapInitializer)` for consistency with 
`RegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer)` implementation?

Then there is no strict need for the local method, and the code can be shortened to  `return TryRegisterClassMap(() => new BsonClassMap<TClass>(classMapInitializer));`

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapInitializer">The class map initializer (only called if the class map is not already registered).</param>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>(Action<BsonClassMap<TClass>> classMapInitializer)
+        {
+            if (classMapInitializer == null)
+            {
+                throw new ArgumentNullException("classMapInitializer");
```
BorisDog, line 466: `nameof(classMapInitializer)`

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMap">The class map.</param>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>(BsonClassMap<TClass> classMap)
+        {
+            if (classMap == null)
+            {
+                throw new ArgumentNullException("classMap");
```
BorisDog, line 445: nameof(classMap)

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -415,6 +415,114 @@ public static void RegisterClassMap(BsonClassMap classMap)
...
+        /// Registers a class map if it is not already registered.
+        /// </summary>
+        /// <typeparam name="TClass">The class.</typeparam>
+        /// <param name="classMapFactory">The class map factory (only called if the class map is not already registered).</param>
+        /// <returns>True if this call registered the class map, false if the class map was already registered.</returns>
+        public static bool TryRegisterClassMap<TClass>(Func<BsonClassMap<TClass>> classMapFactory)
+        {
+            if (classMapFactory == null)
+            {
+                throw new ArgumentNullException("classMapFactory");
```
BorisDog, line 489: nameof(classMapFactory)

BorisDog (approved):
LGTM (minor comments for consideration).

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: I wondered also. 

I usually prefer to leave existing code alone unless there's a strong reason.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Done

rstam: Note: wasn't done in the first place because I copy/pasted some code from another part of this file that was written before `nameof` existed.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Done on the first comment.

I wasn't sure if using `() =>` would result in an additional memory allocation for the lambda. Do you know?

If `() =>` doesn't have a performance penalty we can use it here and on line 448 also.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Done

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Done

BorisDog (approved):
LGTM!

JamesKovacs (approved):
LGTM

Laurianti:
Hi @rstam, just wanted to let you know that I checked out PR #979 and it completely satisfies my request. I've gone ahead and closed out my own PR #941. Thanks so much for your hard work on this! Appreciate it a lot.

PR #980, opened 2022-12-02
CSHARP-4182: Add specification tests for Range Index.

DmitryLukyanov:
I've just noticed that implemented test changes actually not only required and there are more spec changes in the scope of this ticket that will be added soon. So postpone reviewing this change for now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,7 +47,7 @@ Help
 src/packages
 
 # Nupkg artifacts
-*.nupkg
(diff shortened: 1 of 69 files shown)


JamesKovacs:
ACK. Put it up for PR review when all changes have been incorporated.

DmitryLukyanov:
Close in favor of https://github.com/mongodb/mongo-csharp-driver/pull/988

PR #981, opened 2022-12-03
CSHARP-4442: Hide aws sdk in release package.

DmitryLukyanov:
Without applied fix, a consumer will see:
image
and therefore will be able to use aws sdk public classes.
The applied change hides this internal (for the driver) classes from consumers. 
image
As a result, aws.sdk public classes will be unavailable for driver's consumers

### DIFF
--- a/tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
+++ b/tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
@@ -31,6 +31,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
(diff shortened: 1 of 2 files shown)


DmitryLukyanov:
I would also consider the same changes for other 3rd parties we have even though it might be considered as minor BC

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
```diff
@@ -31,6 +31,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.100.14" PrivateAssets="All" />
```
DmitryLukyanov, line 34: this is needed because aws.sdk classes became invisible for all consumers including our tests

DmitryLukyanov:
Apparently this change broke our tests in core TFs. We need investigating how it can be fixed as a separate effort, reverted for now

PR #986, opened 2022-12-10
CSHARP-4358: Update tests that only test against LINQ2 to test against LINQ3 instead of or in addition to LINQ2.

rstam:


rstam (commented):
With this PR all tests are now running either against LINQ3 or both LINQ2 and LINQ3.

Except for Find projections. There is some work to do there still.

JamesKovacs (approved):
LGTM

PR #987, opened 2022-12-12
CSHARP-4422: Custom xunit test cases discoverer for unified format tests

BorisDog:
- Single runner for all unified tests, replaces 15 unified runners.
- Navigation to unified spec test case (json file) directly via VS TestExplorer
- Introduced `CategoryAttribute(params string[] categories)` as a shortcut for [Trait("Category", "SomeCatergory")]


tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -0,0 +1,194 @@
...
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.Driver.TestHelpers;
+using MongoDB.Driver.Tests.UnifiedTestOperations;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MongoDB.Driver.Tests.Specifications
+{
+    public partial class UnifiedTestSpecRunner : LoggableTestClass
```
DmitryLukyanov, line 29: even though it may work, I would still prefer to have a separate class for a test runner. A super class that is responsible for everything (all tests) doesn't look like a good idea almost always.

DmitryLukyanov: Additionally we will probably have more often conflicts when we will implement spec tests even thought it will be about completely different areas

BorisDog: Take a look at it in the local test explorer, the tests organizations is quite straightforward.
Almost all tests runners differ only by the tests folder path (as the should), so there is no really a point of copy pasting all that code, and a folder basis.
So now adding a spec test folder requires one-liner test method, instead of runner + tests factory.

Could you think of a conflict example that becomes harder with a single runner?

DmitryLukyanov: >Take a look at it in the local test explorer, the tests organizations is quite straightforward.

I think, same we can say for runners structure before

>Almost all tests runners differ only by the tests folder path

If it's simple runner, the only logic we actually need is to create a new file. The rest of logic we currently have can be improved. We simply can create a base class that will contain all logic we need including testcase factory and even current Run method. In theory we can end-up with just creating a new file and configure path for simple cases. It's a question whether it's worth it, but it's possible

I really don't like idea about a super class responsible for everything. It gives just a mess since now we actually need to mix all logic between runners. Even now, you spread a runner methods and related helpers across the same .cs file and the situation will become only worse

>So now adding a spec test folder requires one-liner test method, instead of runner + tests factory.

I think it's a good that we need a new file for a runner to encapsulate all particular runner related logic in the same place. The rest of changes can be improved, moving everything into one class is only one option

>Could you think of a conflict example that becomes harder with a single runner?

yes, we all will need to work on the same file. So if at the same time two of us will work on let's say adding new methods here, it will be a merge conflict, it may or may nor be resolved automatically, but the key point we don't need this situation at all

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.Cmap.cs
```diff
@@ -14,17 +14,14 @@
...
 using MongoDB.Driver.Core.TestHelpers.Logging;
 using MongoDB.Driver.Tests.UnifiedTestOperations;
-using Xunit;
-using Xunit.Abstractions;
 
-namespace MongoDB.Driver.Tests.Specifications.command_logging_and_monitoring
+namespace MongoDB.Driver.Tests.Specifications
 {
-    public class CmapUnifiedTestRunner : LoggableTestClass
+    public partial class UnifiedTestSpecRunner : LoggableTestClass
```
DmitryLukyanov, line 24: I don't see why we need a partial class here

DmitryLukyanov (commented):
Initial comments

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestsDiscoverer.cs
```diff
@@ -0,0 +1,169 @@
...
+                    if (_testsToSkip?.Contains(testCase.Name) == true)
+                    {
+                        continue;
+                    }
+
+                    var description = testCase.Test["description"].AsString;
+                    var lineNumer = lines.FirstOrDefault(p => p.Line.Contains(description)).Index;
+                    testCase.Test.Add("_lineNumber", lineNumer);
+                    
+                    if (_generateAsyncVersion)
```
sanych-sun, line 135: Can we simply this code with?:

```
yield return new JsonDrivenTestCase($"{fileName}:{testCase.Name}", testCase.Shared, testCase.Test);

if (_generateAsyncVersion)
{
    var name = $"{fileName}:{testCase.Name}:async";
    var test = testCase.Test.DeepClone().AsBsonDocument.Add("async", true);
    yield return new JsonDrivenTestCase(name, testCase.Shared, test);
}
```

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestsDiscoverer.cs
```diff
@@ -0,0 +1,169 @@
...
+                }
+            }
+
+            protected override string GetTestCaseName(BsonDocument shared, BsonDocument test, int index) =>
+                GetTestName(test, index);
+        }
+    }
+
+    [XunitTestCaseDiscoverer("MongoDB.Driver.Tests.UnifiedTestOperations.UnifiedTestsDiscoverer", "MongoDB.Driver.Tests")]
+    public class UnifiedTestsAttribute : FactAttribute
```
sanych-sun, line 158: Do we really need to inherit from the FactAttribute?

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestsDiscoverer.cs
sanych-sun: Or consider removing of _generateAsyncVersion at all.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestsDiscoverer.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            var provider = reflectionTypeInfo.Type.GetField(skipTestsProvider, BindingFlags.NonPublic | BindingFlags.Static);
+            return provider.GetValue(null) as HashSet<string>;
+        }
+
+        private sealed class UnifiedTestCaseFactory : JsonDrivenTestCaseFactory
+        {
+            private static bool[] __asyncOptions = new[] { true, false };
+
```
sanych-sun, line 86: This is not used anymore.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.Sdam.cs
```diff
@@ -14,78 +14,15 @@
...
-using Xunit;
-using Xunit.Abstractions;
-using System.Linq;
 
-namespace MongoDB.Driver.Tests.Specifications.server_discovery_and_monitoring
+namespace MongoDB.Driver.Tests.Specifications
 {
-    [Trait("Category", "SDAM")]
-    public class ServerDiscoveryAndMonitoringUnifiedTestRunner : LoggableTestClass
+    public partial class UnifiedTestSpecRunner : LoggableTestClass
```
rstam, line 24: The intended use of `partial` classes is where part of the class is generated by a code generator (which might overwrite the file at any time) and user written code is in a second file that is not overwritten by the code generator.

If the argument for partial classes here is that a class is so big that you feel a need to spread it across multiple files, using partial classes doesn't address the fact that the class is too big.

I would not use `partial` here (or anywhere).

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -0,0 +1,220 @@
...
+    {
+        public UnifiedTestSpecRunner(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper, true)
+        {
+        }
+
+        [Category("SupportLoadBalancing")]
+        [UnifiedTests("change_streams.tests.unified")]
```
rstam, line 38: Consider renaming the attribute to make it clear that this is a "theory":

    [UnifiedTestsTheory("change_streams.tests.unified")]

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -0,0 +1,220 @@
...
+namespace MongoDB.Driver.Tests.Specifications
+{
+    public partial class UnifiedTestSpecRunner : LoggableTestClass
+    {
+        public UnifiedTestSpecRunner(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper, true)
+        {
+        }
+
+        [Category("SupportLoadBalancing")]
```
rstam, line 37: If `[Category]` is *only* useful in the unified specs space it should probably have a more precise name:

    [UnifiedTestsCategory("SupportLoadBalancing")]

If it is a more general concept that can be used aside from unified tests then the very general name should be OK.

rstam (changes requested):
Overall looks great. Just a few minor comments.

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.Sdam.cs
rstam: I won't withhold an LGTM to this PR based on the use of partial classes, but I'm not convinced it ever makes sense to use partial classes.

I'm curious what @JamesKovacs and @sanych-sun think about partial classes?

BorisDog: Refactored to regions.

rstam (approved):
LGTM

PR #988, opened 2022-12-12
CSHARP-4182: Support for Range Indexes.

DmitryLukyanov:


tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
             }
         }
 
+        [SkippableTheory]
+        [ParameterAttributeData]
+        public void RangeExplicitEncryptionTest(
+            [Range(1, 8)] int testCase,
+            // test case keys correspond to keys used in test configuration files
+            [Values("DoubleNoPrecision", "DoublePrecision", "Date", "Int", "Long")] string rangeSupportedType,
```
BorisDog, line 1883: nit:
1. just `rangeType` or `rangeDataType`
2. should "// test case keys..." comment be before/above the `testCase` argument?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+                    case "Long":
+                        {
+                            result.Should().BeOfType<BsonInt64>();
+                        }
+                        break;
+                    default: throw new Exception($"Unsupported rangeSupportedType {rangeSupportedType}.");
+                }
+            }
+
+            BsonValue GetValue<TValue>(TValue value, string rangeSupportedType) => rangeSupportedType switch
```
BorisDog, line 2146: Seems that `TValue` is not really used and will always be `int`.
So maybe make it non-generic and then you could skip the string parsing stage?
`"DoubleNoPrecision" => new BsonDouble(value)`
...

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+                    var encrypted30 = ExplicitEncrypt(clientEncryption, encryptOptions, value30, async);
+                    var encrypted200 = ExplicitEncrypt(clientEncryption, encryptOptions, value200, async);
+
+                    CreateCollection(encryptedClient, explicitEncryption, encryptedFields: encryptedFields);
+                    var encryptedCollection = GetCollection(encryptedClient, explicitEncryption);
+                    // bulk insert is not supported
+                    Insert(
+                        encryptedCollection,
+                        async,
+                        new BsonDocument { { encryptedKeyWithRangeSupportedType, encrypted0 }, { "_id", 0 } });
```
BorisDog, line 1921: You also could just use `new() { { encryptedKeyWithRangeSupportedType, encrypted0 }, { "_id", 0 } }`

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+
+            void RunTestCase(ClientEncryption clientEncryption, IMongoCollection<BsonDocument> encryptedCollection, int testCase)
+            {
+                switch (testCase)
+                {
+                    case 1: // can decrypt a payload
+                        {
+                            var insertPayload6 = ExplicitEncrypt(clientEncryption, encryptOptions, value6, async);
+                            var decryptedValue = ExplicitDecrypt(clientEncryption, insertPayload6, async);
+                            decryptedValue.Should().Be(BsonValue.Create(value6));
```
BorisDog, line 1976: Curious why do we need `BsonValue.Create` here?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+            void AssertResult(BsonValue result, string rangeSupportedType)
+            {
+                switch (rangeSupportedType)
+                {
+                    case "DoubleNoPrecision":
+                        {
+                            result.Should().BeOfType<BsonDouble>();
+                        }
+                        break;
+                    case "DoublePrecision":
```
BorisDog, line 2122: minor: You could combine DoubleNoPrecision and DoublePrecision cases.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,287 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+            void RunTestCase(ClientEncryption clientEncryption, IMongoCollection<BsonDocument> encryptedCollection, int testCase)
+            {
+                switch (testCase)
+                {
+                    case 1: // can decrypt a payload
+                        {
+                            var insertPayload6 = ExplicitEncrypt(clientEncryption, encryptOptions, value6, async);
+                            var decryptedValue = ExplicitDecrypt(clientEncryption, insertPayload6, async);
+                            decryptedValue.Should().Be(BsonValue.Create(value6));
+                            AssertResult(decryptedValue, rangeSupportedType);
```
BorisDog, line 1977: Also do we need `AssertResult`? 
`decryptedValue.Should().Be(BsonValue.Create(value6)); `would not validate the BsonValue type as well?

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -57,6 +57,7 @@ public class Feature
         private static readonly Feature __createIndexCommitQuorum = new Feature("CreateIndexCommitQuorum", WireVersion.Server44);
         private static readonly Feature __createIndexesCommand = new Feature("CreateIndexesCommand", WireVersion.Server26);
         private static readonly Feature __createIndexesUsingInsertOperations = new Feature("CreateIndexesUsingInsertOperations", WireVersion.Zero, WireVersion.Server42);
+        private static readonly Feature __csfleRangeExplicitAlgorithm = new Feature("CsfleRangeExplicitAlgorithm", WireVersion.Server62);
```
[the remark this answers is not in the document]
DmitryLukyanov: `Explicit` in this context means explicit encryption (ie via ClientEncryption). I agreed that it's better to remove this part. Will do it.

src/MongoDB.Driver/Encryption/ClientEncryption.cs
```diff
@@ -257,5 +299,20 @@ public void Dispose()
         /// <returns>The result.</returns>
         public Task<RewrapManyDataKeyResult> RewrapManyDataKeyAsync(FilterDefinition<BsonDocument> filter, RewrapManyDataKeyOptions options, CancellationToken cancellationToken = default) =>
             _libMongoCryptController.RewrapManyDataKeyAsync(filter, options, cancellationToken);
+
+        // private methods
+        private TEncryptedValue EnsureEncryptedData<TEncryptedValue>(BsonValue encryptedValue) where TEncryptedValue : BsonValue
```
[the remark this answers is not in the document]
DmitryLukyanov: If we (or libmongocrypt level) don't have a bug, this check is just in case, but I think it's a good check since it's easy to call not a right method since there are 2 almost the same inner methods with the same income arguments and the only difference in a return type

src/MongoDB.Driver/Encryption/ClientEncryption.cs
```diff
@@ -257,5 +299,20 @@ public void Dispose()
...
+        {
+            if (encryptedValue is TEncryptedValue convertedValue)
+            {
+                return convertedValue;
+            }
+            else
+            {
+                // should not be reached
+                var butMessage = encryptedValue == null ? " was null" : $"was {encryptedValue.GetType().Name}";
+                throw new InvalidOperationException($"The encrypted data must be {typeof(TEncryptedValue).Name}, but {butMessage}.");
```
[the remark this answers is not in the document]
DmitryLukyanov: yeah, good point, will do it

src/MongoDB.Driver/Encryption/EncryptOptions.cs
```diff
@@ -145,29 +231,33 @@ encryptionAlgorithm switch
...
             Ensure.That(!(_keyId.HasValue && _alternateKeyName != null), "Key Id and AlternateKeyName may not both be set.");
-            Ensure.That(!(_contentionFactor.HasValue && _algorithm != EncryptionAlgorithm.Indexed.ToString()), "ContentionFactor only applies for Indexed algorithm.");
-            Ensure.That(!(_queryType != null && _algorithm != EncryptionAlgorithm.Indexed.ToString()), "QueryType only applies for Indexed algorithm.");
+            Ensure.That(!(_contentionFactor.HasValue && (_algorithm != EncryptionAlgorithm.Indexed.ToString() && _algorithm != EncryptionAlgorithm.RangePreview.ToString())), "ContentionFactor only applies for Indexed or RangePreview algorithm.");
+            Ensure.That(!(_queryType != null && (_algorithm != EncryptionAlgorithm.Indexed.ToString() && _algorithm != EncryptionAlgorithm.RangePreview.ToString())), "QueryType only applies for Indexed or RangePreview algorithm.");
+            Ensure.That(!(_rangeOptions != null && _algorithm != EncryptionAlgorithm.RangePreview.ToString()), "RangeOptions only applies for RangePreview algorithm.");
```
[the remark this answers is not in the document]
DmitryLukyanov: I will double check it

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: >should "// test case keys..." comment be before/above the testCase argument?

this comment is relevant only to the second input argument, so I would say no, better to place it only on the line when it has meaning

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: i will check it

src/MongoDB.Driver/Encryption/EncryptOptions.cs
```diff
@@ -13,11 +13,77 @@
...
+    /// min, max, sparsity, and range must match the values set in the encryptedFields of the destination collection.
+    /// For double and decimal128, min/max/precision must all be set, or all be unset.
+    /// RangeOptions only applies when algorithm is "rangePreview".
+    /// </remarks>
+    public class RangeOptions
+    {
+        private readonly BsonValue _max;
+        private readonly BsonValue _min;
+        private readonly int? _precision;
+        private readonly long _sparsity;
```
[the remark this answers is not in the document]
DmitryLukyanov: >fields don't add any functionality

that's true, but it can be changed later. As we discussed previously, the reason why I prefer this approach: 
1. All code base is in this format already. 
2. This approach splits logic inside class on few simple and clear parts: 
 * Define fields for internal usage
 * Ctor
 * Define props for external usage.

I like this structure. The pain about few lines more in the file is smaller in my opinion comparing it with code consistency (for all cases) and a little bit more clear structure.

src/MongoDB.Driver/Encryption/EncryptOptions.cs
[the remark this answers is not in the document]
BorisDog: I mean the following case: `_rangeOptions == null` and _`algorithm == RangePreview`

src/MongoDB.Driver/Encryption/EncryptOptions.cs
```diff
@@ -13,11 +13,78 @@
...
+        {
+            _sparsity = sparsity;
+            _min = min.WithDefault(null);
+            _max = max.WithDefault(null);
+            _precision = precision.WithDefault(null);
+        }
+
+        // public properties
+        /// <summary>
+        /// Min range.
```
BorisDog, line 56: Minimum value

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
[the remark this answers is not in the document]
BorisDog: Also do we want to call it Preview, to differentiate with future non preview?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
BorisDog: I think this change is relevant in all other cases.

src/MongoDB.Driver/Encryption/EncryptOptions.cs
DmitryLukyanov: done

src/MongoDB.Driver/Encryption/EncryptOptions.cs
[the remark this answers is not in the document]
DmitryLukyanov: it looks like you're right, I will think about it

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: done

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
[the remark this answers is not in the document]
DmitryLukyanov: At this point I think we will use the same server version guard. Most-likely the only difference will be removing `Preview` suffix. I would defer any change here to the time when it will be clear what final shape we will end up here with.

src/MongoDB.Driver/Encryption/EncryptOptions.cs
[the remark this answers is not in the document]
DmitryLukyanov: discussed offline, there is a similar check mentioned in the documentation, but not this one. At this point, I think better to add arguments validation only for arguments combinations that are restricted in the spec

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,253 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+            }
+
+            BsonValue GetValue(int value, string rangeSupportedType) => rangeSupportedType switch
+            {
+                "DoubleNoPrecision" => new BsonDouble(value),
+                "DoublePrecision" => new BsonDouble(value),
+                "Date" => new BsonDateTime(millisecondsSinceEpoch: value),
+                "Int" => new BsonInt32(value),
+                "Long" => new BsonInt64(value),
+                _ => throw new Exception($"Unsupported rangeSupportedType {rangeSupportedType}.")
```
BorisDog, line 2119: nit: `$"Unsupported {nameof(rangeSupportedType)} {rangeSupportedType}."` or even better
`new ArgumentException($"Unsupported nameof(rangeSupportedType) {rangeSupportedType}.", nameof(rangeSupportedType))`

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2471,6 +2718,15 @@ private void DropView(CollectionNamespace viewNamespace)
             return encryptedValue;
         }
 
+        private BsonDocument ExplicitEncryptExpression(
+            ClientEncryption clientEncryption,
+            EncryptOptions encryptOptions,
+            BsonDocument expression,
+            bool async) =>
+            async
+                ? clientEncryption.EncryptExpressionAsync(expression, encryptOptions).GetAwaiter().GetResult()
```
BorisDog, line 2727: I would recommend to shift away from using `GetAwaiter().GetResult()` anti-pattern.
You can return `task` here, and make the whole test async.

BorisDog (changes requested):
Few minor comments.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: I'm ok with this approach, but don't really want to change unrelated tests in this file just to make them compiled after this change. So I would use an old approach for this file

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: I prefer the first one, since we throw in test in majority of cases only `Exception`, done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,253 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+                        {
+                            if (rangeType == "DoubleNoPrecision" || rangeType == "DoublePrecision")
+                            {
+                                throw new SkipException("Skip it based on spec requirement.");
+                            }
+
+                            var exception = Record.Exception(() =>
+                                ExplicitEncrypt(
+                                    clientEncryption,
+                                    encryptOptions.With(rangeOptions: new RangeOptions(sparsity: 1, min: BsonValue.Create(0), max: BsonValue.Create(200), precision: 2)),
```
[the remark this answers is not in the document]
DmitryLukyanov: no, it's needed. Since the input type is `Optional<BsonValue>`, otherwise, it's not compiled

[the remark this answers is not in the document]
BorisDog: I see, then let's use `(BsonValue)0`?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1875,6 +1875,267 @@ HttpClientWrapperWithModifiedRequest CreateHttpClientWrapperWithModifiedRequest(
...
+            BsonValue GetValue(int value, string rangeSupportedType) => rangeSupportedType switch
+            {
+                "DecimalNoPrecision" => new BsonDecimal128(value),
+                "DecimalPrecision" => new BsonDecimal128(value),
+                "DoubleNoPrecision" => new BsonDouble(value),
+                "DoublePrecision" => new BsonDouble(value),
+                "Date" => new BsonDateTime(millisecondsSinceEpoch: value),
+                "Int" => new BsonInt32(value),
+                "Long" => new BsonInt64(value),
+                _ => throw new Exception($"Unsupported {nameof(rangeSupportedType)} {rangeSupportedType}.")
```
BorisDog, line 2133: `ArgumentException` is used a lot in the driver in such cases already, also it's more suitable exception type her, while `Exception` should be used only when there is no better specific exception.
Up to you.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
BorisDog: Isn't this a new method used only by the new `RangeExplicitEncryptionTest` test?

DmitryLukyanov: Yes, but this test touches `Insert` that then touches everything else

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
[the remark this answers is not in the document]
DmitryLukyanov: I would prefer `BsonValue.Create`. If you have a factory method, I would use it. NOTE: no huge logic happens behind the scene, it will be almost the same logic as you suggest

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: it's interesting, I expected that we use it rare than it actually happen. We still use: `throw new Exception` more often.. but I'm ok to change it

DmitryLukyanov: done

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
DmitryLukyanov: discussed offline, changed

BorisDog (approved):
LGTM.

PR #989, opened 2022-12-12
CSHARP-4440: Incorporate MongoDB.Labs.Search library

BorisDog:


src/MongoDB.Driver/Search/SearchRange.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// Object that specifies range of scalar and DateTime values.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the range value.</typeparam>
+    public struct SearchRange<TValue> where TValue : struct, IComparable<TValue>
+    {
+        /// <summary>Empty range.</summary>
+        public static SearchRange<TValue> Empty { get; } = new(default, default, default, default);
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Range{TValue}"/> class.
```
dgolub, line 16:
```suggestion
        /// Initializes a new instance of the <see cref="SearchRange{TValue}"/> class.
```

src/MongoDB.Driver/Search/SearchDefinitions.cs
```diff
@@ -0,0 +1,379 @@
...
+            _query = Ensure.IsNotNull(query, nameof(query));
+            _tokenOrder = tokenOrder;
+            _fuzzy = fuzzy;
+        }
+
+        private protected override BsonDocument RenderOperator(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
+           new()
+           {
+                { "query", _query.Render() },
+                { "tokenOrder", _tokenOrder.ToCamelCase(), _tokenOrder == AutocompleteTokenOrder.Sequential },
```
dgolub, line 48:
```suggestion
                { "tokenOrder", _tokenOrder.ToCamelCase(), _tokenOrder != AutocompleteTokenOrder.Any },
```

This should be more future-proof in case they add other token order options.

src/MongoDB.Driver/Search/SearchDefinitions.cs
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        private protected override BsonDocument RenderOperator(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
+            new()
+            {
+                { _range.IsMinInclusive ? "gte" : "gt", () => ToBsonValue(_range.Min.Value), _range.Min != null },
+                { _range.IsMaxInclusive ? "lte" : "lt", () => ToBsonValue(_range.Max.Value), _range.Max != null },
+            };
+
+        private static BsonValue ToBsonValue(TField value) =>
```
dgolub, line 280: This is probably enforceable at compile time using the new generic math stuff that they added in the latest C# version, but I'm assuming that we can't use that for backward compatibility reasons?

dgolub (commented):
Left a couple of minor suggestions.  Otherwise, the changes look great.

src/MongoDB.Driver/Search/SearchRange.cs
BorisDog: Thanks.

src/MongoDB.Driver/Search/SearchDefinitions.cs
BorisDog: Thanks, good catch!

src/MongoDB.Driver/Search/SearchDefinitions.cs
BorisDog: Yeah backward compatibility, I wish we could use that.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,667 @@
...
+        /// </summary>
+        /// <typeparam name="TCoordinates">The type of the coordinates.</typeparam>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="geometry">
+        /// GeoJSON object specifying the MultiPolygon or Polygon to search within.
+        /// </param>
+        /// <param name="path">Indexed geo type field or fields to search.</param>
+        /// <param name="score">The score modifier.</param>
+        /// <returns>A geo within search definition.</returns>
+        public SearchDefinition<TDocument> GeoWithin<TCoordinates, TField>(
```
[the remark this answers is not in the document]
BorisDog: IIRC generic type inference fails.

src/MongoDB.Driver/Search/SpanDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+
+namespace MongoDB.Driver.Search
+{
+    /// <summary>
+    /// Base class for span clauses.
+    /// </summary>
+    /// <typeparam name="TDocument"></typeparam>
+    public abstract class SpanDefinition<TDocument>
```
rstam, line 24: Rename class to SearchSpanDefinition

src/MongoDB.Driver/Search/SpanDefinitionBuilder.cs
```diff
@@ -0,0 +1,198 @@
...
+using MongoDB.Driver;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Search
+{
+    /// <summary>
+    /// A builder for a span clause.
+    /// </summary>
+    /// <typeparam name="TDocument">The type of the document.</typeparam>
+    public sealed class SpanDefinitionBuilder<TDocument>
```
rstam, line 30: Rename class to SearchSpanDefinitionBuilder

src/MongoDB.Driver/Search/SpanDefinitionBuilder.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <param name="query">The string or string to search for.</param>
+        /// <param name="path">The indexed field or fields to search.</param>
+        /// <returns>A term span clause.</returns>
+        public SpanDefinition<TDocument> Term<TField>(
+            QueryDefinition query,
+            Expression<Func<TDocument, TField>> path) =>
+            Term(query, new ExpressionFieldDefinition<TDocument>(path));
+    }
+
+    internal sealed class FirstSpanDefinition<TDocument> : SpanDefinition<TDocument>
```
rstam, line 103: Rename class to FirstSearchSpanDefinition

Similarly for the rest of the classes in this file

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
     /// <typeparam name="TDocument">The type of the document.</typeparam>
     public static class Builders<TDocument>
     {
-        private static FilterDefinitionBuilder<TDocument> __filter = new FilterDefinitionBuilder<TDocument>();
-        private static IndexKeysDefinitionBuilder<TDocument> __index = new IndexKeysDefinitionBuilder<TDocument>();
-        private static ProjectionDefinitionBuilder<TDocument> __projection = new ProjectionDefinitionBuilder<TDocument>();
-        private static SortDefinitionBuilder<TDocument> __sort = new SortDefinitionBuilder<TDocument>();
-        private static UpdateDefinitionBuilder<TDocument> __update = new UpdateDefinitionBuilder<TDocument>();
+        /// <summary>Gets a <see cref="FilterDefinitionBuilder{TDocument}"/>.</summary>
```
rstam, line 26: I'm not a fan of these changes... as you now.

Also not a fan of random diffs that make reviewing changes more difficult.

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
...
+
+        /// <summary>Gets a <see cref="SortDefinitionBuilder{TDocument}"/>.</summary>
+        public static SortDefinitionBuilder<TDocument> Sort { get; } = new SortDefinitionBuilder<TDocument>();
+
+        /// <summary>Gets an <see cref="UpdateDefinitionBuilder{TDocument}"/>.</summary>
+        public static UpdateDefinitionBuilder<TDocument> Update { get; } = new UpdateDefinitionBuilder<TDocument>();
+
+        // Search builders
+        /// <summary>Gets a <see cref="SearchFacetBuilder{TDocument}"/>.</summary>
+        public static SearchFacetBuilder<TDocument> Facet { get; } = new SearchFacetBuilder<TDocument>();
```
rstam, line 43: Rename property to SearchFacet

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
...
 
-        /// <summary>
-        /// Gets a <see cref="FilterDefinitionBuilder{TDocument}"/>.
-        /// </summary>
-        public static FilterDefinitionBuilder<TDocument> Filter
-        {
-            get { return __filter; }
-        }
+        /// <summary>Gets a <see cref="PathDefinition{TDocument}"/>.</summary>
+        public static PathDefinitionBuilder<TDocument> Path { get; } = new PathDefinitionBuilder<TDocument>();
```
rstam, line 46: Rename property to SearchPath

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
...
 
-        /// <summary>
-        /// Gets an <see cref="IndexKeysDefinitionBuilder{TDocument}"/>.
-        /// </summary>
-        public static IndexKeysDefinitionBuilder<TDocument> IndexKeys
-        {
-            get { return __index; }
-        }
+        /// <summary>Gets a <see cref="ScoreDefinitionBuilder{TDocument}"/>.</summary>
+        public static ScoreDefinitionBuilder<TDocument> Score { get; } = new ScoreDefinitionBuilder<TDocument>();
```
rstam, line 49: Rename property to SearchScore

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
...
 
-        /// <summary>
-        /// Gets a <see cref="ProjectionDefinitionBuilder{TDocument}"/>.
-        /// </summary>
-        public static ProjectionDefinitionBuilder<TDocument> Projection
-        {
-            get { return __projection; }
-        }
+        /// <summary>Gets a <see cref="ScoreFunctionBuilder{TDocument}"/>.</summary>
+        public static ScoreFunctionBuilder<TDocument> ScoreFunction { get; } = new ScoreFunctionBuilder<TDocument>();
```
rstam, line 52: Rename property to SearchScoreFunction

src/MongoDB.Driver/Builders.cs
```diff
@@ -21,50 +23,38 @@ namespace MongoDB.Driver
...
 
-        /// <summary>
-        /// Gets an <see cref="UpdateDefinitionBuilder{TDocument}"/>.
-        /// </summary>
-        public static UpdateDefinitionBuilder<TDocument> Update
-        {
-            get { return __update; }
-        }
+        /// <summary> Gets a <see cref="SpanDefinitionBuilder{TDocument}"/>.</summary>
+        public static SpanDefinitionBuilder<TDocument> Span { get; } = new SpanDefinitionBuilder<TDocument>();
```
rstam, line 58: Rename property to SearchSpan

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1309,6 +1309,80 @@ public static class PipelineStageDefinitionBuilder
...
+        /// or return only stored source fields directly from Atlas Search.
+        /// </param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
+            SearchDefinition<TInput> searchDefinition,
+            HighlightOptions<TInput> highlight = null,
+            string indexName = null,
+            SearchCountOptions count = null,
+            bool returnStoredSource = false)
+        {
```
rstam, line 1331: Just wondering whether we have the best ordering of parameters here.

For example, if indexName is the most commonly used optional parameter, maybe it should be the first optional parameter.

I don't actually know what parameters are more likely to be used than others. I just want to make sure we get the optimal parameter ordering right the first time, because we can't change it later.

If we do change the order let's be sure to change it the same way everywhere.

rstam (commented):
This is my first pass at reviewing.

Overall this looks great.

I do think we need to be more careful about naming things in a way that doesn't risk name collisions now or in the future.

For example, you named a class `SearchCountOptions` to avoid colliding with the existing `CountOptions` class. This is good.

We should use the same naming convention for all our search related classes, both for consistent naming and to avoid future possible name collisions.

So for example, rename `FuzzyOptions` to `SearchFuzzyOptions` and `HighlightOptions` to `SearchHighlightOptions` and similarly for all our new classes.

I realize that all these new classes are in a `.Search` sub-namespace and thus are technically unambiguous in the face of name collisions, but only when prefixing them with full or partial namespaces. Let's just avoid potential collisions in the first place and adopt a consistent naming convention, even if it does make the type names slightly longer. Shorter is not always better when it creates ambiguity (either real ambiguity, as in a compile time error, or conceptual ambiguity such `QueryDefinition` vs `SearchQueryDefinition`, where the first doesn't sound like it has anything to do with Search and sounds more like it might apply to Find).

Once we've worked through these renaming suggestions I would like to review again in more detail.

Good work overall!

BorisDog (commented):
I am not a big fun of longer names, especially as Search namespace makes this a non issue. In our case actual names are less important, as mostly Builders are expected to be used. Also most of the classes are internal.
I don't think this change will have noticeable impact on our API, so made this change.

tests/MongoDB.Driver.Tests/Search/MongoQueryableTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        [Fact]
+        public void Search()
+        {
+            var query = _queryable
+                .Where(x => x.FirstName == "Alexandra")
+                .Search(Builders<Person>.Search.Text("Alex", x => x.FirstName))
+                .ToString();
+
+            query.Should().EndWith("Aggregate([{ \"$match\" : { \"fn\" : \"Alexandra\" } }, { \"$search\" : { \"text\" : { \"query\" : \"Alex\", \"path\" : \"fn\" } } }])");
+        }
```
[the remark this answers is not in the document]
rstam: Helper method:

        private IMongoQueryable<Person> CreateSubject()
        {
            var client = DriverTestConfiguration.Linq3Client;
            var database = client.GetDatabase(DriverTestConfiguration.DatabaseNamespace.DatabaseName);
            var collection = database.GetCollection<Person>(DriverTestConfiguration.CollectionNamespace.CollectionName);
            return collection.AsQueryable();
        }

rstam (changes requested):
Apologies that there are many comments.

Overall this looks great and these are just small comments to fine tune and polish the result.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,665 @@
...
+        /// <param name="path">Indexed geo type field or fields to search.</param>
+        /// <param name="relation">
+        /// Relation of the query shape geometry to the indexed field geometry.
+        /// </param>
+        /// <param name="score">The score modifier.</param>
+        /// <returns>A geo shape search definition.</returns>
+        public SearchDefinition<TDocument> GeoShape<TCoordinates>(
+            GeoJsonGeometry<TCoordinates> geometry,
+            SearchPathDefinition<TDocument> path,
+            GeoShapeRelation relation,
```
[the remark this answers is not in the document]
BorisDog: I've put the path first.
But I think `relation` should come after `geometry`, as we specify the main "query" object first and then all parameters.
I thinks it's also sort of a convention, some examples:
[NetTopologySuite](https://nettopologysuite.github.io/NetTopologySuite/api/NetTopologySuite.Geometries.Geometry.html#NetTopologySuite_Geometries_Geometry_Relate_NetTopologySuite_Geometries_Geometry_System_String_)
[ArgGic](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Geometry.GeometryEngine.Relate.html#Esri_ArcGISRuntime_Geometry_GeometryEngine_Relate_Esri_ArcGISRuntime_Geometry_Geometry_Esri_ArcGISRuntime_Geometry_Geometry_System_String_)

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,665 @@
...
+                new GeoWithinSearchDefinition<TDocument, TCoordinates>(geoWithinQuery, path, score);
+
+        /// <summary>
+        /// Creates a search definition that returns documents similar to the input documents.
+        /// </summary>
+        /// <param name="like">
+        /// One or more documents that Atlas Search uses to extract representative terms for.
+        /// </param>
+        /// <returns>A more like this search definition.</returns>
+        public SearchDefinition<TDocument> MoreLikeThis(IEnumerable<TDocument> like) =>
```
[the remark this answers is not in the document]
BorisDog: This is probably a good suggestion. I think we can think about this in a follow up ticket.

tests/MongoDB.Driver.Tests/Search/MongoQueryableTests.cs
```diff
@@ -0,0 +1,64 @@
...
+using FluentAssertions;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Search
+{
+    public class MongoQueryableTests
+    {
+        private readonly IMongoQueryable<Person> _queryable;
```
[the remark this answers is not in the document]
BorisDog: Thanks, I will consider this in the next round.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,377 @@
...
+                byte uint8Value => (BsonValue)uint8Value,
+                short int16Value => (BsonValue)int16Value,
+                ushort uint16Value => (BsonValue)uint16Value,
+                int int32Value => (BsonValue)int32Value,
+                uint uint32Value => (BsonValue)uint32Value,
+                long int64Value => (BsonValue)int64Value,
+                double doubleValue => (BsonValue)doubleValue,
+                DateTime dateTimeValue => (BsonValue)dateTimeValue,
+                _ => throw new InvalidCastException()
+            };
```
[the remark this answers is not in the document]
rstam: It's nice that you can reuse the same variable name on different cases!

I still think a few casts are called for to make clear that short integral values are converted to int:

        private static BsonValue ToBsonValue(TField value) =>
            value switch
            {
                sbyte v => (int)v,
                byte v => (int)v,
                short v => (int)v,
                ushort v => (int)v,
                int v => v,
                uint v => (int)v,
                long v => v,
                double v => v,
                DateTime v => v,
                _ => throw new InvalidCastException()
            };

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,383 @@
...
+    {
+        private readonly bool _allowAnalyzedField;
+        private readonly SearchQueryDefinition _query;
+
+        public RegexSearchDefinition(
+            SearchPathDefinition<TDocument> path,
+            SearchQueryDefinition query,
+            bool allowAnalyzedField,
+            SearchScoreDefinition<TDocument> score)
+            : base(OperatorType.Regex, path, score)
```
rstam, line 308: Indent call to : base one more tab.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,383 @@
...
+    {
+        private readonly SearchFuzzyOptions _fuzzy;
+        private readonly SearchQueryDefinition _query;
+
+        public TextSearchDefinition(
+            SearchPathDefinition<TDocument> path,
+            SearchQueryDefinition query,
+            SearchFuzzyOptions fuzzy,
+            SearchScoreDefinition<TDocument> score)
+            : base(OperatorType.Text, path, score)
```
rstam, line 346: Indent call to : base one more tab.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,383 @@
...
+    {
+        private readonly bool _allowAnalyzedField;
+        private readonly SearchQueryDefinition _query;
+
+        public WildcardSearchDefinition(
+            SearchPathDefinition<TDocument> path,
+            SearchQueryDefinition query,
+            bool allowAnalyzedField,
+            SearchScoreDefinition<TDocument> score)
+            : base(OperatorType.Wildcard, path, score)
```
rstam, line 370: Indent call to : base one more tab.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,666 @@
...
+        public SearchDefinition<TDocument> Autocomplete<TField>(
+            Expression<Func<TDocument, TField>> path,
+            SearchQueryDefinition query,
+            SearchAutocompleteTokenOrder tokenOrder = SearchAutocompleteTokenOrder.Any,
+            SearchFuzzyOptions fuzzy = null,
+            SearchScoreDefinition<TDocument> score = null) =>
+                Autocomplete(new ExpressionFieldDefinition<TDocument>(path), query, tokenOrder, fuzzy, score);
+
+        /// <summary>
+        /// Creates a search definition that combines two or more operators into a single query.
```
rstam, line 68: /// Creates a builder for a search definition that combines two or more operators into a single query.
 
or maybe:

        /// Creates a builder for a compound search definition.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: I still think the order should be:

    path, relation, geometry

for clarity and readability.

Consider these examples as illustrations of this order being more readable because the parameters are in the same order as the english language phrase describing the operation:

    SearchDefinitionBuilder<TDocument> search;

    // search for properties completely within a county
    // i.e. property is within county
    search.GeoShape(x => x.PropertyBoundary, GeoShapeRelation.Within, countyBoundary)

    // search for properties that are at least partially in a county
    // i.e. property intersects county
    search.GeoShape(x => x.PropertyBoundary, GeoShapeRelation.Intersects, countyBoundary)

    // search for properties that are not in a county
    // i.e. property is disjoint with county
    search.GeoShape(x => x.PropertyBoundary, GeoShapeRelation.Disjoint, countyBoundary)

    // search for the county a property is completely contained in
    // i.e. county contains property
    search.GeoShape(x => x.CountyBoundary, GeoShapeRelation.Contains, propertyBoundary);

    // search for all counties a property is even partially contained in
    // i.e. county intersects property
    search.GeoShape(x => x.CountyBoundary, GeoShapeRelation.Intersects, propertyBoundary);

@JamesKovacs you should probably look at this public API decision and chime in also.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,666 @@
...
+            where TCoordinates : GeoJsonCoordinates =>
+                GeoWithin(path, new GeoWithinGeometry<TCoordinates>(geometry), score);
+
+        /// <summary>
+        /// Creates a search definition that queries for geographic points within a given geo object.
+        /// </summary>
+        /// <typeparam name="TCoordinates">The type of the coordinates.</typeparam>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="path">Indexed geo type field or fields to search.</param>
+        /// <param name="area">Object that specifies the geo object to search within./// </param>
```
rstam, line 264: /// <param name="area">Object that specifies the area to search within./// </param>

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,666 @@
...
+            SearchScoreDefinition<TDocument> score = null)
+            where TCoordinates : GeoJsonCoordinates =>
+                GeoWithin(new ExpressionFieldDefinition<TDocument>(path), area, score);
+
+        /// <summary>
+        /// Creates a search definition that queries for geographic points within a given geo object.
+        /// </summary>
+        /// <typeparam name="TCoordinates">The type of the coordinates.</typeparam>
+        /// <param name="path">Indexed geo type field or fields to search.</param>
+        /// <param name="area">Object that specifies the geo object to search within./// </param>
```
rstam, line 279: /// <param name="area">Object that specifies the area to search within./// </param>

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,666 @@
...
+        /// Creates a search definition that queries for documents where a floating-point
+        /// field is in the specified range.
+        /// </summary>
+        /// <returns>A fluent range interface.</returns>
+        public SearchDefinition<TDocument> Range<TField>(
+            Expression<Func<TDocument, TField>> path,
+            SearchRange<TField> range,
+            SearchScoreDefinition<TDocument> score = null)
+            where TField : struct, IComparable<TField> =>
+            Range(new ExpressionFieldDefinition<TDocument>(path), range, score);
```
rstam, line 538: Indent one more tab stop

src/MongoDB.Driver/Search/SearchFacetBuilder.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// A list of date values that specify the boundaries for each bucket.
+        /// </param>
+        /// <param name="default">
+        /// The name of an additional bucket that counts documents returned from the operator that
+        /// do not fall within the specified boundaries.
+        /// </param>
+        /// <returns>A date search facet.</returns>
+        public SearchFacet<TDocument> Date(
+            SearchPathDefinition<TDocument> path,
+            string name,
```
rstam, line 47: I think this is the exception to putting the path first. `name` should go first.

We're building a facet, and the parameter to building a facet are `name` + `everything else`, so `name` should go first and `everything else` after that.

src/MongoDB.Driver/Search/SearchFacetBuilder.cs
```diff
@@ -0,0 +1,274 @@
...
+            _path = Ensure.IsNotNull(path, nameof(path));
+            _numBuckets = Ensure.IsNullOrBetween(numBuckets, 1, 1000, nameof(numBuckets));
+        }
+
+        public override BsonDocument Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
+            new()
+            {
+                { "type", "string" },
+                { "path", _path.Render(documentSerializer, serializerRegistry) },
+                { "numBuckets", _numBuckets, _numBuckets != null }
```
rstam, line 271: Does this even work? I would expect to see:

                { "numBuckets", () => _numBuckets.Value, _numBuckets != null }

src/MongoDB.Driver/Search/SearchFuzzyOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Gets or sets the number of characters at the beginning of each term in the result that
+        /// must exactly match.
+        /// </summary>
+        public int? PrefixLength
+        {
+            get => _prefixLength;
+            set => _prefixLength = Ensure.IsNullOrGreaterThanOrEqualToZero(value, nameof(value));
+        }
```
rstam, line 57: Missing blank line

src/MongoDB.Driver/Search/SearchHighlightOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets the document field to search.
+        /// </summary>
+        public SearchPathDefinition<TDocument> Path
+        {
+            get => _path;
+            set => _path = Ensure.IsNotNull(value, nameof(value));
+        }
```
rstam, line 73: Missing blank line.

src/MongoDB.Driver/Search/SearchMetaResult.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Generic;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace MongoDB.Driver.Search
+{
+    /// <summary>
+    /// A search count result set.
+    /// </summary>
+    public sealed class SearchMetaCountResult
```
rstam, line 25: This class is impossible to instantiate. Either add a constructor (and remove `private set`) or just make the class mutable (public `set`).

Yes it's a shame we can't use `init` because we are targetting some ancient TFs.

I think the claim that this class is OK as is just because the serializers can get around the non-instantiability of the class is not a convincing argument.

Same for the other classes in this file.

src/MongoDB.Driver/Search/SearchPathDefinition.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        /// <summary>
+        /// Performs an implicit conversion from an array of field names to 
+        /// <see cref="SearchPathDefinition{TDocument}"/>.
+        /// </summary>
+        /// <param name="fieldNames">The array of field names.</param>
+        /// <returns>
+        /// The result of the conversion.
+        /// </returns>
+        public static implicit operator SearchPathDefinition<TDocument>(string[] fieldNames) =>
```
rstam, line 88: Would there be any point in adding a `List<string>` overload for symmetry with the other methods?

src/MongoDB.Driver/Search/SearchPathDefinitionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+            new SingleSearchPathDefinition<TDocument>(field);
+
+        /// <summary>
+        /// Creates a search path for a single field.
+        /// </summary>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="field">The field definition.</param>
+        /// <returns>A single-field search path.</returns>
+        public SearchPathDefinition<TDocument> Single<TField>(Expression<Func<TDocument, TField>> field) =>
+            Single(new ExpressionFieldDefinition<TDocument>(field));
```
rstam, line 91: Missing blank line.

src/MongoDB.Driver/Search/SearchPathDefinitionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            _field = Ensure.IsNotNull(field, nameof(field));
+            _analyzerName = Ensure.IsNotNull(analyzerName, nameof(analyzerName));
+        }
+
+        public override BsonValue Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) => new BsonDocument()
+        {
+            {  "value", _field.Render(documentSerializer, serializerRegistry).FieldName },
+            {  "multi", _analyzerName }
+        };
```
rstam, line 119: This formatting makes the braces look like the body of a method. Consider this instead:

        public override BsonValue Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
            new BsonDocument()
            {
                {  "value", _field.Render(documentSerializer, serializerRegistry).FieldName },
                {  "multi", _analyzerName }
            };

src/MongoDB.Driver/Search/SearchPathDefinitionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            _field = Ensure.IsNotNull(field, nameof(field));
+        }
+
+        public override BsonValue Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry)
+        {
+            var renderedField = _field.Render(documentSerializer, serializerRegistry);
+            return new BsonString(renderedField.FieldName);
+        }
+    }
```
rstam, line 149: Missing blank line.

src/MongoDB.Driver/Search/SearchPathDefinitionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public WildcardSearchPathDefinition(string query)
+        {
+            _query = Ensure.IsNotNull(query, nameof(query));
+        }
+
+        public override BsonValue Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) => new BsonDocument()
+        {
+            { "wildcard", _query }
+        };
```
rstam, line 162: This formatting makes the braces look like a method body. Consider this instead:

        public override BsonValue Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
            new BsonDocument()
            {
                { "wildcard", _query }
            };

src/MongoDB.Driver/Search/SearchRange.cs
```diff
@@ -0,0 +1,134 @@
...
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Search
+{
+    /// <summary>
+    /// Object that specifies range of scalar and DateTime values.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the range value.</typeparam>
+    public struct SearchRange<TValue> where TValue : struct, IComparable<TValue>
+    {
```
rstam, line 26: We normally put the static members in a #region at the top of the class:

        #region static
        /// <summary>Empty range.</summary>
        public static SearchRange<TValue> Empty { get; } = new(default, default, default, default);
        #endregion

src/MongoDB.Driver/Search/SearchRange.cs
```diff
@@ -0,0 +1,134 @@
...
+        public TValue? Max { get; }
+
+        /// <summary>Gets the lower bound of the range.</summary>
+        public TValue? Min { get; }
+    }
+
+    /// <summary>
+    /// A builder for a SearchRange.
+    /// </summary>
+    public static class SearchRangeBuilder
```
rstam, line 65: It's a bit odd to see factory methods and extension methods in the same class.

Possibly consider moving the extension methods to `SearchRangeExtensions` class?

Not a big deal. Your choice.

src/MongoDB.Driver/Search/SearchRange.cs
```diff
@@ -0,0 +1,134 @@
...
+            => SearchRange<TValue>.Empty.Gt(value);
+
+        /// <summary>
+        /// Adds a greater than value to a search range.
+        /// </summary>
+        /// <param name="searchRange">Search range.</param>
+        /// <param name="value">The value.</param>
+        /// <returns>Search range.</returns>
+        public static SearchRange<TValue> Gt<TValue>(this SearchRange<TValue> searchRange, TValue value) where TValue : struct, IComparable<TValue>
+            => new(value, searchRange.Max, false, searchRange.IsMaxInclusive);
```
rstam, line 82: Consider adding some parameter names for clarity:

            => new(min: value, searchRange.Max, isMinInclusive: false, searchRange.IsMaxInclusive);

rstam: i.e. when I read `false` I don't know what it means without navigating to the method implementation, and `isMinInclusive: false` documents what `false` means in this case.

rstam: Same for other methods below.

src/MongoDB.Driver/Search/SearchRange.cs
```diff
@@ -0,0 +1,134 @@
...
+        public static SearchRange<TValue> Lt<TValue>(this SearchRange<TValue> searchRange, TValue value) where TValue : struct, IComparable<TValue>
+            => new(searchRange.Min, value, searchRange.IsMinInclusive, false);
+
+        /// <summary>
+        /// Creates a less than or equal search range.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>search range.</returns>
+        public static SearchRange<TValue> Lte<TValue>(TValue value) where TValue : struct, IComparable<TValue>
+            => SearchRange<TValue>.Empty.Lte(value);
```
rstam, line 124: Missing blank line.

src/MongoDB.Driver/Search/SearchScoreDefinitionBuilder.cs
```diff
@@ -0,0 +1,149 @@
...
+        private readonly double _value;
+
+        public BoostValueSearchScoreDefinition(double value)
+        {
+            _value = Ensure.IsGreaterThanZero(value, nameof(value));
+        }
+
+        public override BsonDocument Render(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
+            new("boost",  new BsonDocument("value", _value));
+    }
```
rstam, line 123: Missing blank line.

src/MongoDB.Driver/Search/SearchScoreFunctionBuilder.cs
```diff
@@ -0,0 +1,276 @@
...
+        /// The number of use to determine the distance from <paramref name="origin"/>.
+        /// </param>
+        /// <returns>A Guassian score function.</returns>
+        public SearchScoreFunction<TDocument> Gauss(
+            SearchPathDefinition<TDocument> path,
+            double origin,
+            double scale,
+            double decay = 0.5,
+            double offset = 0)
+            => new GaussSearchScoreFunction<TDocument>(path, origin, scale, decay, offset);
```
rstam, line 80: Indent one more tab stop

src/MongoDB.Driver/Search/SearchScoreFunctionBuilder.cs
```diff
@@ -0,0 +1,276 @@
...
+        /// The number of use to determine the distance from <paramref name="origin"/>.
+        /// </param>
+        /// <returns>A Guassian score function.</returns>
+        public SearchScoreFunction<TDocument> Gauss(
+            Expression<Func<TDocument, double>> path,
+            double origin,
+            double scale,
+            double decay = 0.5,
+            double offset = 0)
+            => Gauss(new ExpressionFieldDefinition<TDocument>(path), origin, scale, decay, offset);
```
rstam, line 106: Indent one more tab stop

src/MongoDB.Driver/Search/SearchSpanDefinitionBuilder.cs
```diff
@@ -0,0 +1,199 @@
...
+        public SearchSpanDefinition<TDocument> Subtract(
+            SearchSpanDefinition<TDocument> include,
+            SearchSpanDefinition<TDocument> exclude) =>
+                new SubtractSearchSpanDefinition<TDocument>(include, exclude);
+
+        /// <summary>
+        /// Creates a span clause that matches a single term.
+        /// </summary>
+        /// <param name="query">The string or strings to search for.</param>
+        /// <param name="path">The indexed field or fields to search.</param>
```
rstam, line 86: Probably better if doc comments are in the same order as the parameters.

src/MongoDB.Driver/AggregateFluentBase.cs
```diff
@@ -215,6 +216,26 @@ public virtual IAggregateFluent<TNewResult> ReplaceWith<TNewResult>(AggregateExp
             throw new NotImplementedException();
         }
 
+        /// <inheritdoc />
+        public virtual IAggregateFluent<TResult> Search(
+            SearchDefinition<TResult> query,
```
rstam, line 221: Rename `query` to `searchDefinition` to be consistent with all other `Search` methods.

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -83,16 +83,18 @@ public static T IsEqualTo<T>(T value, T comparand, string paramName)
         }
 
         /// <summary>
-        /// Ensures that the value of a parameter is greater than or equal to zero.
+        /// Ensures that the value of a parameter is greater than a comparand.
```
rstam, line 86: Carefully double check alphabetical order of methods?

Watch for: `HasValue`, `IsGreaterThan`, `IsGreaterThanOrEqualTo`.

tests/MongoDB.Driver.Tests/Search/MongoQueryableTests.cs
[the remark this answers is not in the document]
rstam: Don't forget to switch from constructor setup to `CreateSubject` helper method.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,908 @@
...
+                    new GeoJson2DGeographicCoordinates(-152.446289, 22.512557)));
+
+        private static readonly GeoWithinCircle<GeoJson2DGeographicCoordinates> __testCircle =
+            new GeoWithinCircle<GeoJson2DGeographicCoordinates>(
+                new GeoJsonPoint<GeoJson2DGeographicCoordinates>(
+                    new GeoJson2DGeographicCoordinates(-161.323242, 22.512557)),
+                7.5);
+
+        private static readonly GeoJsonPolygon<GeoJson2DGeographicCoordinates> __testPolygon =
+                            new GeoJsonPolygon<GeoJson2DGeographicCoordinates>(
```
rstam, line 44: Indentation is wrong on this line.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,908 @@
...
+                new GeoJsonPolygonCoordinates<GeoJson2DGeographicCoordinates>(
+                    new GeoJsonLinearRingCoordinates<GeoJson2DGeographicCoordinates>(
+                        new List<GeoJson2DGeographicCoordinates>()
+                        {
+                            new GeoJson2DGeographicCoordinates(-161.323242, 22.512557),
+                            new GeoJson2DGeographicCoordinates(-152.446289, 22.065278),
+                            new GeoJson2DGeographicCoordinates(-156.09375, 17.811456),
+                            new GeoJson2DGeographicCoordinates(-161.323242, 22.512557)
+                        })));
```
rstam, line 53: Missing blank line.

tests/MongoDB.Driver.Tests/Search/SearchPathDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        [Fact]
+        public void Single_Typed()
+        {
+            var subject = CreateSubject<Person>();
+
+            AssertRendered(subject.Single(x => x.FirstName), new BsonString("fn"));
+            AssertRendered(subject.Single("FirstName"), new BsonString("fn"));
+        }
```
rstam, line 113: Missing blank line

tests/MongoDB.Driver.Tests/PipelineDefinitionBuilderTests.cs
```diff
@@ -117,6 +118,140 @@ public void Merge_should_add_expected_stage()
...
 
+        [Fact]
+        public void Search_should_add_expected_stage()
+        {
+            var pipeline = new EmptyPipelineDefinition<BsonDocument>();
+            var builder = new SearchDefinitionBuilder<BsonDocument>();
+            var result = pipeline.Search(builder.Text("bar", "foo"));
+            var stages = RenderStages(result, BsonDocumentSerializer.Instance);
+            stages[0].Should().Be(
+                BsonDocument.Parse("{ $search: { text: { query: 'foo', path: 'bar' } } }"));
```
rstam, line 129: Simplify to:

            var stages = RenderStages(result, BsonDocumentSerializer.Instance);
            stages[0].Should().Be("{ $search: { text: { query: 'bar', path: 'foo' } } }");

No need for `BsonDocument.Parse`

rstam: Same for all other new tests immediately below.

tests/MongoDB.Driver.Tests/PipelineDefinitionBuilderTests.cs
```diff
@@ -117,6 +118,140 @@ public void Merge_should_add_expected_stage()
             stages[0].Should().Be("{ $merge : { into : { db : 'database', coll : 'collection' } } }");
         }
 
+        [Fact]
+        public void Search_should_add_expected_stage()
+        {
+            var pipeline = new EmptyPipelineDefinition<BsonDocument>();
+            var builder = new SearchDefinitionBuilder<BsonDocument>();
```
rstam, line 125: Add a blank line after Arrange section.

Same for other new tests below.

tests/MongoDB.Driver.Tests/PipelineDefinitionBuilderTests.cs
```diff
@@ -117,6 +118,140 @@ public void Merge_should_add_expected_stage()
...
+            var result = pipeline.Search(builder.Text("bar", "foo"));
```
rstam, line 126: Add a blank line after Act section.

Same for other new tests below.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,383 @@
...
+            : base(OperatorType.GeoWithin, path, score)
+        {
+            _area = Ensure.IsNotNull(area, nameof(area));
+        }
+
+        private protected override BsonDocument RenderOperator(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry) =>
+            new(_area.Render());
+    }
+
+    internal sealed class MoreLikeThisSearchDefinition<TDocument> : OperatorSearchDefinition<TDocument>
```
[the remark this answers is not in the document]
rstam: @JamesKovacs  let's discuss this one.

src/MongoDB.Driver/Search/SearchFacetBuilder.cs
rstam: Same for all methods in this file.

rstam (changes requested):
I can see the finish line!

tests/MongoDB.Driver.Tests/Search/MongoQueryableTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            query.ToString().Should().EndWith("Aggregate([{ \"$search\" : { \"text\" : { \"query\" : \"Alex\", \"path\" : \"fn\" } } }])");
+        }
+
+        [Fact]
+        public void SearchMeta()
+        {
+            var subject = CreateSubject();
+
+            var query = subject
+                .Where(x => x.FirstName == "Alexandra")
```
rstam, line 42: Remove the call to `Where`?

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: Do you need to edit your comment? There are multiple `polygon` parameters.

I see nothing wrong with adding more (possibly optional) parameters at the end.

The following looks fine to me:

    GeoShape(point, GeoShapeRelation.Within, polygon, padding, threshold)

rstam (changes requested):
Looks like we're only a few issues away from being done.

Besides the requested changes here we have the following design issues to discuss with @JamesKovacs 

* GeoShape parameter order
* MoreLike with partial documents

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: Note that in our own documentation the order is `path`, `relation`, `geometry`:

https://www.mongodb.com/docs/atlas/atlas-search/geoShape/

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: This is the example from our documentation:

    {
        "$search": {
           "index": <index name>, // optional, defaults to "default"
           "geoShape": {
              "path": "<field-to-search>",
              "relation": "contains | disjoint | intersects | within",
              "geometry": <GeoJSON-object>,
              "score": <score-options>
           }
    }

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
BorisDog: I think typed overload is convenient. `morelikethis` documentation gives example of executing `morelikethis` queries with objects returned by previous queries. Also rendered document does not have to be entire document, as null can be omitted.
With such usage pattern in mind typed overload is useful. 

I agree that `BsonDocument` overload would be useful, added.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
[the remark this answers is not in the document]
BorisDog: > it's about what Range needs.

Missed this nuance, that's a good point! Done.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
BorisDog: All our search operators are in form of `Operator(subject, queryObject, params)`, which happens to align with general convention for geospatial libs, of `Relate(Geomerty, relation, params)`. So with that in mind, the API should be:
`GeoShape(point, polygon, relation, padding, threshold)`. It's also more intuitive imho. 

If you think we should match the ordering of parameters to the official docs, we could do that (even that there are some problematic nuances with such approach). In this case we should match rest of the operators as well, and alternate the position of the `path` argument (it does not always come first).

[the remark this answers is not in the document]
rstam: I didn't mean to propose that we should slavishly follow whatever order the server documentation is in. The server documentation is not consistent with itself. We should be consistent with ourselves.

I just meant this particular example as showing that I'm not the only one who thinks of this operation using the same order as I would say them in: `path`, `relation`, `geometry`. As in "county contains property", instead of "county property contains" (which is not grammatical).

Doesn't sound like we're getting any closer to a consensus, since we each are coming at this from different directions. I think it should be "readable" (and to my mind more "intuitive"). You think it should be consistent with external libraries that have nothing to do with our driver (and to your different interpretation of what is "intuitive").

I'm just going to defer to @JamesKovacs  on this one. If James wants to go the same way you are suggesting, then we will go that way. I won't like it, but I will accept it.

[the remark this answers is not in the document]
BorisDog: Yes I understand the grammatical argument, but I don't think we should try to follow this approach at all costs, I see it more as "nice to have". Yes, those libraries have nothing to do with out driver, but I think their conventions in this area are sort of a good standard.

Reminder: NetTopologySuite is used by Microsoft EF core, CosmosDB and Npsql.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -0,0 +1,396 @@
...
+
+        private protected override BsonDocument RenderArguments(IBsonSerializer<TDocumentCollection> documentSerializer, IBsonSerializerRegistry serializerRegistry)
+        {
+            var likeArray = typeof(TDocumentLike) == typeof(BsonDocument) ?
+                new BsonArray(_like.OfType<BsonDocument>()) :
+                new BsonArray(_like.OfType<TDocumentCollection>().Select(e => e.ToBsonDocument(documentSerializer)));
+
+            return new("like", likeArray);
+        }
+    }
```
rstam, line 203: I like this approach in general, but I think it is too restrictive. I could see myself creating a smaller POCO to use for query by example calls to MoreLike, so `TDocument` and `BsonDocument` are not the only two valid choices for the like values. For example, I might declare a `PersonQBE` class that is a subset of `Person` and use that with `MoreByLike`. If I have to use `BsonDocument` I could get lots of things wrong.

While it is true that allowing arbitrary `TLike` classes could be misused, the same is already true by allowing `BsonDocument`.

Would you consider simplifying this class to:

    internal sealed class MoreLikeThisSearchDefinition<TDocument, TLike> : OperatorSearchDefinition<TDocument>
    {
        private readonly TLike[] _like;

        public MoreLikeThisSearchDefinition(IEnumerable<TLike> like)
            : base(OperatorType.MoreLikeThis)
        {
            _like = Ensure.IsNotNull(like, nameof(like)).ToArray();
        }

        private protected override BsonDocument RenderArguments(IBsonSerializer<TDocument> documentSerializer, IBsonSerializerRegistry serializerRegistry)
        {
            var likeSerializer = serializerRegistry.GetSerializer<TLike>();
            var likeArray = new BsonArray(_like.Select(e => e.ToBsonDocument(likeSerializer)));
            return new("like", likeArray);
        }
    }

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,686 @@
...
+
+        /// <summary>
+        /// Creates a search definition that returns documents similar to the input documents.
+        /// </summary>
+        /// <param name="like">
+        /// One or more documents that Atlas Search uses to extract representative terms for.
+        /// </param>
+        /// <returns>A more like this search definition.</returns>
+        public SearchDefinition<TDocument> MoreLikeThis(IEnumerable<TDocument> like) =>
+            new MoreLikeThisSearchDefinition<TDocument, TDocument>(like);
```
rstam, line 297: Consider the more flexible:

    public SearchDefinition<TDocument> MoreLikeThis<TLike>(IEnumerable<TLike> like) =>
        new MoreLikeThisSearchDefinition<TDocument, TLike>(like);

and

    public SearchDefinition<TDocument> MoreLikeThis<TLike>(params TLike[] like) =>
        MoreLikeThis((IEnumerable<TLike>)like);

which would make the `BsonDocument` overloads unnecessary and allow them to be removed.

rstam (changes requested):
Small changes requested. Let me know what you think of them.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
BorisDog: Good point and good suggestion, done.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
BorisDog: Done.

rstam (changes requested):
Minor comments.

Still pending are some questions for @JamesKovacs, *specially* the one about the order of the parameters for `GeoSearch`.

evergreen/run-atlas-search-test.sh
```diff
@@ -0,0 +1,17 @@
...
+
+############################################
+#            Main Program                  #
+############################################
+
+echo "Running Atlas Search driver tests"
+
+export ATLAS_SEARCH_TESTS_ENABLED=true
+
+powershell.exe .\\build.ps1 --target TestAtlasSearch
```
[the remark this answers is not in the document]
BorisDog: No reason. Copied this from other Atlas tests scripts. Changed.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -0,0 +1,668 @@
...
+        /// <summary>
+        /// Creates a search definition that returns documents similar to the input documents.
+        /// </summary>
+        /// <typeparam name="TLike">The type of the like documents.</typeparam>
+        /// <param name="like">
+        /// One or more documents that Atlas Search uses to extract representative terms for.
+        /// </param>
+        /// <returns>A more like this search definition.</returns>
+        public SearchDefinition<TDocument> MoreLikeThis<TLike>(params TLike[] like) =>
+            new MoreLikeThisSearchDefinition<TDocument, TLike>(like);
```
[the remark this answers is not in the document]
BorisDog: Good point, changed.

src/MongoDB.Driver/Search/SearchHighlight.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// <summary>
+    /// Represents a result of highlighting.
+    /// </summary>
+    public sealed class SearchHighlight
+    {
+        /// <summary>
+        /// Gets or sets the document field which returned a match.
+        /// </summary>
+        [BsonElement("path")]
+        public string Path { get; private set; }
```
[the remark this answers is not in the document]
BorisDog: We can use the `init` modifier (by [introducing ](https://btburnett.com/csharp/2020/12/11/csharp-9-records-and-init-only-setters-without-dotnet5.html) the internal `IsExternalInit`)

src/MongoDB.Driver/Search/SearchHighlightOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <param name="maxNumPassages">
+        /// The number of high-scoring passages to return per document in the highlighting results
+        /// for each field.
+        /// </param>
+        /// <returns>Highlighting options.</returns>
+        public SearchHighlightOptions(
+            Expression<Func<TDocument, object>> path,
+            int? maxCharsToExamine = null,
+            int? maxNumPassages = null) :
+            this(new ExpressionFieldDefinition<TDocument>(path), maxCharsToExamine, maxNumPassages)
```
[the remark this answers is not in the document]
BorisDog: Done.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
```diff
@@ -0,0 +1,519 @@
...
+
+namespace MongoDB.Driver.Tests.Search
+{
+    [Trait("Category", "AtlasSearch")]
+    public class AtlasSearchTests : LoggableTestClass
+    {
+        #region static
+
+        private static readonly GeoJsonPolygon<GeoJson2DGeographicCoordinates> __testPolygon =
+            new(new(new(new GeoJson2DGeographicCoordinates[]
```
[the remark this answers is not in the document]
BorisDog: Yes, this is more readable :) !
The idea is to skip the multi-line boilerplate code, which does not provide any useful into for regular tests maintenance. The only important bit of information are the coordinates themselves.
This way the whole tests class is simpler and more readable.

In the rare case the geo object needs to be edited, it's can be easily done with VS intellisense. Seems that mouse hovering or pressing F12 is a reasonable price to pay.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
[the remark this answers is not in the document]
JamesKovacs: I’m in favour of the `path, relation, geometry` suggestion as the `relation` acts like an operator.

```{
  field1: {$in: [ 1, 2, 3 ]},
  field2: {$gt: 42}
}```

`path` is `fieldX`. `relation` is the operator. `geometry` is equivalent to the operator argument. The following is an imagining of `path, geometry, relation` for `$in` and `$gt`.

```{
  field1: {arg: [ 1, 2, 3 ], op: "in"},
  field2: {arg: 42, op: "gt"}
}```

So I would argue that `path, relation, geometry` makes more sense syntactically.

File: src/MongoDB.Driver/Search/SearchMetaResult.cs (hunk 31 above)
Comment (JamesKovacs) [reply src/MongoDB.Driver/Search/SearchMetaResult.cs:25]: See my previous comment on `init`. If users can instantiate instances with init-only properties and adding `IsExternalInit` doesn't cause compat problems (I don't think it does), then I'm good going down this path.

**Review by JamesKovacs — CHANGES_REQUESTED** [review]
See my previous inline comments. To summarize:
1. `GeoShape` param order should be `path, relation, geometry`.
2. Using `init` for `SearchHighlight` and related classes sounds like a good solution as long as users of our driver can instantiate instances of those classes.

**Comment by rstam** [comment]
If using `init` on older TFs poses a risk we can just use `public` and be safe.

Sure we'd rather the class be immutable, but if the user mutates it, it can't hurt us.

**Comment by JamesKovacs** [comment]
Summary of my `init` investigations. TL;DR: We can't use `init`.

The `init` magic is done by the compiler. You can use it on earlier TFMs using the `IsExternalInit` trick however you must use a compiler compatible with `init`. And this is where we run into problems...

I defined a library `Lib` with an `init` property. TFM is `netstandard2.0` and `LangVersion` is 9.0. I then created a console app `App` and referenced `Lib`. App is `netcoreapp3.1` with `LangVersion` 8.0. I instantiated an instance of the init-only containing class from `Lib`. When I compiled, I got the following error:

```
error CS8400: Feature 'init-only setters' is not available in C# 8.0. Please use language version 9.0 or greater.
```

If you change the `LangVersion` of App to 9.0, then everything works. The problem is that if we use `init` in our driver we will require our users to upgrade their toolchain to C# 9 or later. If they’re still using `netcoreapp3.1` or earlier, they might not be able to upgrade their toolchain.

The safest option is to make the setters public on `SearchHighlight` and related so that users can instantiate their own instances for testing purposes.

File: src/MongoDB.Driver/Search/SearchHighlight.cs [hunk 62]
```diff
@@ -0,0 +1,105 @@
...
+    /// </summary>
+    public sealed class SearchHighlight
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SearchHighlight"/> class.
+        /// </summary>
+        /// <param name="path">document field which returned a match.</param>
+        /// <param name="score">Score assigned to this result.</param>
+        /// <param name="texts">Objects containing the matching text and the surrounding text.</param>
+        public SearchHighlight(string path, double score, SearchHighlightText[] texts)
```
BorisDog, line 32: Adopted the ctors approach. Thanks @rstam for making this work.

JamesKovacs (approved):
LGTM

PR #990, opened 2022-12-13
CSHARP-4445: Support 64-bit values for Skip and Limit/Take.

rstam:


src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -215,7 +215,7 @@ public interface IAggregateFluent<TResult> : IAsyncCursorSource<TResult>
         /// </summary>
         /// <param name="limit">The limit.</param>
         /// <returns>The fluent aggregate interface.</returns>
-        IAggregateFluent<TResult> Limit(int limit);
+        IAggregateFluent<TResult> Limit(long limit);
```
rstam, line 218: This is source compatible, but binary breaking.

I assume that's OK?

Same for similar changes below.

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -1066,6 +1066,27 @@ public static IMongoQueryable<TSource> Skip<TSource>(this IMongoQueryable<TSourc
...
+        /// remaining elements.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of source</typeparam>
+        /// <param name="source">An <see cref="IMongoQueryable{TSource}"/> to return elements from.</param>
+        /// <param name="count">The number of elements to skip before returning the remaining elements.</param>
+        /// <returns>
+        /// An <see cref="IMongoQueryable{TSource}"/> that contains elements that occur after the
+        /// specified index in the input sequence.
+        /// </returns>
+        public static IMongoQueryable<TSource> Skip<TSource>(this IMongoQueryable<TSource> source, long count)
```
rstam, line 1080: For the LINQ versions of Skip and Take we need new overloads because the standard Skip and Take are defined in .NET LINQ classes (not in our code).

JamesKovacs (approved):
LGTM.

Verified that on MongoDB 3.6, the following works:
```
test> db.foo.aggregate([{$limit:NumberLong("1")}])
[ { _id: ObjectId("6398bc6cfed4ffeedca75eba"), a: 1 } ]
```

And specifying a limit greater than `int.MaxValue` also works (even as far back as MongoDB 3.6) - which suggests that the server always treated the arguments as `int64`.

PR #1000, opened 2022-12-22
CSHARP-4449: Implement Find projections in LINQ3.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -140,8 +140,7 @@ static Expression RemovePossibleConvertToObject(Expression expression)
             IBsonSerializer<TSource> sourceSerializer,
             IBsonSerializerRegistry serializerRegistry)
         {
-            // TODO: implement using LINQ3 instead of falling back to LINQ2
-            return LinqProviderAdapter.V2.TranslateExpressionToFindProjection(expression, sourceSerializer, serializerRegistry);
+            return TranslateExpressionToProjection(expression, sourceSerializer, serializerRegistry, translationOptions: null);
```
rstam, line 143: Now that the server supports the full aggregation language in `Find` projections we can simply translate a `Find` projection the same way as a `$project` stage projection.

Note: this only works on server version 4.4 and higher. Users on older server versions can still use `Find` projections, with the limitation that they can't project computed fields (`Find` projections prior to server 4.4 are limited to including/excluding fields).

tests/MongoDB.Driver.Tests/FindFluentTests.cs
```diff
@@ -333,8 +335,12 @@ public void ToString_should_return_the_correct_string()
 
             var str = find.ToString();
 
+            var expectedProjection = linqProvider == LinqProvider.V2 ?
+                "{ \"FirstName\" : 1, \"LastName\" : 1, \"_id\" : 0 }" :
+                "{ \"_v\" : { \"$concat\" : [\"$FirstName\", \" \", \"$LastName\"] }, \"_id\" : 0 }";
```
rstam, line 340: LINQ2 and LINQ3 translate this differently.

LINQ2 handles the computation client side. LINQ3 handles the computation server side (as it should be).

JamesKovacs: Won't this test fail with LINQ3 on MongoDB 4.2 and earlier? Do we need a `RequireServer` guard clause?

tests/MongoDB.Driver.Tests/IFindFluentExtensionsTests.cs
```diff
@@ -228,15 +230,22 @@ public void Project_should_generate_the_correct_fields_when_a_string_is_used()
...
 
-        [Fact]
-        public void Project_should_generate_the_correct_fields_and_assign_the_correct_result_serializer()
+        [Theory]
+        [ParameterAttributeData]
+        public void Project_should_generate_the_correct_fields_and_assign_the_correct_result_serializer(
+            [Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
         {
-            var subject = CreateSubject()
+            var subject = CreateSubject(linqProvider)
```
JamesKovacs, line 238: Same issue here. I think we need `RequireServer` guard clause to prevent LINQ3 version from running on MongoDB 4.2 and earlier.

JamesKovacs (changes requested):
Based on your comments, we need to skip the LINQ3 tests on MongoDB 4.2 and earlier.

tests/MongoDB.Driver.Tests/FindFluentTests.cs
rstam: Done

tests/MongoDB.Driver.Tests/IFindFluentExtensionsTests.cs
rstam: Done

JamesKovacs (approved):
LGTM

PR #1005, opened 2022-12-30
CSHARP-4453: Support Bucket and BucketAuto stages in LINQ3.

rstam:


src/MongoDB.Driver/AggregateBucketAutoResultIdSerializer.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            return new AggregateBucketAutoResultIdSerializer<TValue>(valueSerializer);
+        }
+    }
+
+    /// <summary>
+    /// A serializer for AggregateBucketAutoResultId.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the values.</typeparam>
+    public class AggregateBucketAutoResultIdSerializer<TValue> : ClassSerializerBase<AggregateBucketAutoResultId<TValue>>, IBsonDocumentSerializer
```
rstam, line 45: I'm wondering now as I write these comments whether we could have just let this be class mapped?

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -71,7 +71,15 @@ public static class IAggregateFluentExtensions
...
         {
             Ensure.IsNotNull(aggregate, nameof(aggregate));
-            return aggregate.AppendStage(PipelineStageDefinitionBuilder.Bucket(groupBy, boundaries, output, options));
+            if (aggregate.Database.Client.Settings.LinqProvider == LinqProvider.V2)
+            {
+                return aggregate.AppendStage(PipelineStageDefinitionBuilder.Bucket(groupBy, boundaries, output, options));
+            }
+            else
+            {
+                var (bucketStage, projectStage) = PipelineStageDefinitionBuilder.BucketForLinq3(groupBy, boundaries, output, options);
```
rstam, line 80: Similar to what we had to do for `GroupForLinq3`.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -110,13 +118,47 @@ public static class IAggregateFluentExtensions
             this IAggregateFluent<TResult> aggregate,
             Expression<Func<TResult, TValue>> groupBy,
             int buckets,
-            Expression<Func<IGrouping<TValue, TResult>, TNewResult>> output,
+            Expression<Func<IGrouping<TValue, TResult>, TNewResult>> output, // the IGrouping has been wrong all along, only fixing it for LINQ3
```
rstam, line 121: LINQ2 depends on this being declared this way, even thought it is wrong.

I imagine it would be hard to fix LINQ2, and I have no idea how to even begin.

Better to just fix it in LINQ3.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -24,40 +24,49 @@
...
         #region static
         public static AstPipeline Optimize(AstPipeline pipeline)
         {
-            var optimizer = new AstGroupPipelineOptimizer();
+            var optimizer = new AstGroupingPipelineOptimizer();
             for (var i = 0; i < pipeline.Stages.Count; i++)
             {
                 var stage = pipeline.Stages[i];
-                if (stage is AstGroupStage groupStage)
+                if (IsGroupingStage(stage))
```
rstam, line 36: The gist of the changes in this PR are that there are 3 kinds of "grouping" stages:

1. Group
2. Bucket
3. BucketAuto

Any names using "grouping" refer to any of the 3. If a name uses just "group" it refers to the first.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupExpressionStageDefinitions.cs
```diff
@@ -1,189 +0,0 @@
...
-using MongoDB.Driver.Core.Misc;
-using MongoDB.Driver.Linq.Linq3Implementation.Ast.Optimizers;
-using MongoDB.Driver.Linq.Linq3Implementation.Misc;
-using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
-using MongoDB.Driver.Linq.Linq3Implementation.Translators;
-using MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToPipelineTranslators;
-
-namespace MongoDB.Driver.Linq.Linq3Implementation
-{
-    internal sealed class GroupExpressionStageDefinitions<TInput, TKey, TOutput>
```
rstam, line 33: This class has been replaced by the new `GroupingAndProjectStageDefinitions` class.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupExpressionStageDefinitions.cs
```diff
@@ -1,189 +0,0 @@
...
-                IBsonSerializer<TInput> inputSerializer,
-                IBsonSerializerRegistry serializerRegistry,
-                LinqProvider linqProvider)
-            {
-                if (linqProvider != LinqProvider.V3)
-                {
-                    throw new InvalidOperationException("GroupExpressionStageDefinitions can only be used with LINQ3.");
-                }
-
-                var expression = CreateExpression(inputSerializer);
```
rstam, line 83: I found a simpler way to get the `AstPipelineOptimizer` to do its thing without having to create an `Expression` (not to mention that there is no possible `Expression` for `Bucket` or `BucketAuto` because LINQ doesn't have those).

See the `RenderGroupingStage` method in `GroupingAndProjectStageDefinitions.cs`.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupExpressionStageDefinitions.cs
```diff
@@ -1,189 +0,0 @@
...
-
-                var selectExpression = Expression.Call(
-                    QueryableMethod.Select.MakeGenericMethod(typeof(IGrouping<TKey, TInput>), typeof(TOutput)),
-                    groupByExpression,
-                    _groupExpression);
-
-                return selectExpression;
-            }
-
-            private class PseudoQueryProvider : IMongoQueryProvider
```
rstam, line 116: A nice side effect of the new way of invoking the `AstPipelineOptimizer` is that we no longer need a `PseudoQueryProvider` or a `PseudoSource`.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupingAndProjectStageDefinitions.cs
```diff
@@ -0,0 +1,299 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Optimizers;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Stages;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
+using MongoDB.Driver.Linq.Linq3Implementation.Translators;
+using MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation
+{
+    internal static class GroupingAndProjectStageDefinitions
```
rstam, line 33: These are static factory methods. Static factory methods are nice because they can infer type parameters (constructor calls can't).

src/MongoDB.Driver/Linq/Linq3Implementation/GroupingAndProjectStageDefinitions.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Represents a pair of grouping and project stage definitions.
+    /// </summary>
+    /// <typeparam name="TInput">The input type.</typeparam>
+    /// <typeparam name="TGrouping">The grouping type.</typeparam>
+    /// <typeparam name="TOutput">The output type.</typeparam>
+    public sealed class GroupingAndProjectStageDefinitions<TInput, TGrouping, TOutput>
```
rstam, line 80: `TGrouping` is a type parameter because `BucketAuto` has a different `IGrouping` than the other 2.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupingAndProjectStageDefinitions.cs
```diff
@@ -0,0 +1,299 @@
...
+        private readonly Expression<Func<TGrouping, TOutput>> _projection;
+        private RenderedPipelineStageDefinition<TGrouping> _renderedGroupingStage;
+        private RenderedPipelineStageDefinition<TOutput> _renderedProjectStage;
+
+        public GroupingStageDefinition(Expression<Func<TGrouping, TOutput>> projection)
+        {
+            _projection = projection;
+        }
+
+        public override RenderedPipelineStageDefinition<TGrouping> Render(IBsonSerializer<TInput> inputSerializer, IBsonSerializerRegistry serializerRegistry, LinqProvider linqProvider)
```
rstam, line 133: When asked to `Render` the grouping stage we also render the project stage and we run the pair through the `AstPipelineOptimizer` so that accumulators can be moved from the project stage to the grouping stage.

src/MongoDB.Driver/Linq/Linq3Implementation/GroupingAndProjectStageDefinitions.cs
```diff
@@ -0,0 +1,299 @@
...
+        public void Deconstruct(
+            out PipelineStageDefinition<TInput, TGrouping> groupingStage,
+            out PipelineStageDefinition<TGrouping, TOutput> projectStage)
+        {
+            groupingStage = _groupingStage;
+            projectStage = _projectStage;
+        }
+    }
+
+    internal abstract class GroupingStageDefinition<TInput, TGrouping, TOutput> : PipelineStageDefinition<TInput, TGrouping>
```
rstam, line 122: There are 3 subclasses for each of the grouping stages:

1. `BucketStageDefinition`
2. `BucketAutoStageDefinition`
3. `GroupStageDefinition`

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/IGroupingSerializer.cs
```diff
@@ -107,5 +107,10 @@ public static IBsonSerializer Create(IBsonSerializer keySerializer, IBsonSeriali
             var serializerType = typeof(IGroupingSerializer<,>).MakeGenericType(keyType, elementType);
             return (IBsonSerializer)Activator.CreateInstance(serializerType, keySerializer, elementSerializer);
         }
+
+        public static IBsonSerializer<IGrouping<TKey, TElement>> Create<TKey, TElement>(IBsonSerializer<TKey> keySerializer, IBsonSerializer<TElement> elementSerializer)
+        {
+            return new IGroupingSerializer<TKey, TElement>(keySerializer, elementSerializer);
```
rstam, line 113: If the type parameters can be inferred we don't need to use reflection.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -880,15 +933,15 @@ public static class PipelineStageDefinitionBuilder
         /// <param name="translationOptions">The translation options.</param>
         /// <returns>The stage.</returns>
         /// <remarks>This method can only be used with LINQ3 but that can't be verified until Render is called.</remarks>
-        public static GroupForLinq3Result<TInput, TValue, TOutput> GroupForLinq3<TInput, TValue, TOutput>(
+        public static GroupingAndProjectStageDefinitions<TInput, IGrouping<TValue, TInput>, TOutput> GroupForLinq3<TInput, TValue, TOutput>(
```
rstam, line 936: `GroupingAndProjectStageDefinitions` replaces `GroupForLinq3Result`.

This could be considered a breaking change, but LINQ3 has not been the default up to now so maybe it's OK? Let's discuss.

We could choose to keep the `GroupForLinq3Result` return type and just do a conversion to it.

tests/MongoDB.Driver.Tests/AggregateFluentBucketAutoTests.cs
```diff
@@ -211,48 +213,95 @@ public void BucketAuto_typed_should_return_expected_result()
...
-            var result = subject.BucketAuto(
-                e => e.Year,
-                buckets,
-                g => new { _id = default(AggregateBucketAutoResultId<int?>), Years = g.Select(e => e.Year), Count = g.Count() });
+            if (linqProvider == LinqProvider.V2)
+            {
+                var result = subject.BucketAuto(
+                    e => e.Year,
+                    buckets,
+                    g => new { _id = default(AggregateBucketAutoResultId<int?>), Years = g.Select(e => e.Year), Count = g.Count() });
```
rstam, line 230: Notice the odd way `_id` is used here.

It is only present to add the `_id` property to the anonymous class. LINQ2 actually removes the `_id` from the translation!

tests/MongoDB.Driver.Tests/AggregateFluentBucketAutoTests.cs
```diff
@@ -211,48 +213,95 @@ public void BucketAuto_typed_should_return_expected_result()
...
+                var stage = result.Stages.Single();
+                var serializerRegistry = BsonSerializer.SerializerRegistry;
+                var exhibitSerializer = serializerRegistry.GetSerializer<Exhibit>();
+                var renderedStage = stage.Render(exhibitSerializer, serializerRegistry, linqProvider);
+                renderedStage.Document.Should().Be("{ $bucketAuto : { groupBy : \"$year\", buckets : 4, output : { Years : { $push : \"$year\" }, Count : { $sum : 1 } } } }");
+            }
+            else
+            {
+                var result = subject.BucketAutoForLinq3(
+                    e => (int?)e.Year,
```
rstam, line 241: The only reason we have to cast to `int?` is because the test data is missing the `Year` in one document. This could be considered an error in the test data, but at the same time this illustrates what you would have to do if a `struct` valued property might be missing in the database (not a problem with `class` because `null` is a valid value).

tests/MongoDB.Driver.Tests/AggregateFluentBucketAutoTests.cs
```diff
@@ -211,48 +213,95 @@ public void BucketAuto_typed_should_return_expected_result()
...
+                    buckets,
+                    g => new { Key = g.Key, Years = g.Select(e => e.Year), Count = g.Count() });
```
rstam, line 243: Notice the use of `Key = g.Key` instead of the `_id = default(AggregateBucketAutoResultId<int?>)` hack.

In LINQ3 the `IGrouping` is declared correctly and `g.Key` for `BucketAuto` is of type `AggregateBucketAutoResultId<TValue>`.

JamesKovacs (changes requested):
Minor comments below. We had a larger discussion around `XyzForLinq3`. Robert is going to prototype an idea for possibly removing the need for `XyzForLinq3` methods. I'll review again after those changes.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -147,10 +182,21 @@ static bool IsLastStageThatCanBeOptimized(AstStage stage)
...
             return mappings.ToArray();
+
+            static AstStage CreateNewGroupingStage(AstStage groupingStage, AccumulatorSet accumulators)
+            {
+                return groupingStage switch
+                {
+                    AstGroupStage groupStage => AstStage.Group(groupStage.Id, accumulators),
+                    AstBucketStage bucketStage => AstStage.Bucket(bucketStage.GroupBy, bucketStage.Boundaries, bucketStage.Default, accumulators),
+                    AstBucketAutoStage bucketAutoStage => AstStage.BucketAuto(bucketAutoStage.GroupBy, bucketAutoStage.Buckets, bucketAutoStage.Granularity, accumulators),
+                    _ => throw new Exception($"Unexpected groupintStage node type: {groupingStage.NodeType}.")
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/AggregateBucketAutoResultIdSerializer.cs
```diff
@@ -0,0 +1,103 @@
+﻿/* Copyright 2016-present MongoDB Inc.
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -103,7 +95,7 @@ public static class IAggregateFluentExtensions
         }
 
         /// <summary>
-        /// Appends a $bucketAuto stage to the pipeline (this overload can only be used with LINQ2).
+        /// Appends a $bucketAuto stage to the pipeline (this method can only be used with LINQ3).
```
JamesKovacs, line 98: This is an overload, not just a method.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -118,20 +110,20 @@ public static class IAggregateFluentExtensions
...
-                throw new InvalidOperationException("This overload of BucketAuto can only be used with LINQ2.");
+                throw new InvalidOperationException("This overload of BucketAuto can only be used with LINQ3.");
             }
 
             return aggregate.AppendStage(PipelineStageDefinitionBuilder.BucketAuto(groupBy, buckets, output, options));
         }
 
         /// <summary>
-        /// Appends a $bucketAuto stage to the pipeline (this method can only be used with LINQ3).
+        /// Appends a $bucketAuto stage to the pipeline (this overload can only be used with LINQ2).
```
JamesKovacs, line 126: This is a new method, not an overload.

JamesKovacs (changes requested):
Very nice refactoring! Minor feedback on some comments. Otherwise LGTM.

src/MongoDB.Driver/AggregateBucketAutoResultIdSerializer.cs
rstam: I've thought about it some more, and the answer is no.

We have to be able to pass in the correct value serializer. If we just let it be automapped the value serializer would be looked up in the registry and that might or might not be the correct one.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
rstam: Done

src/MongoDB.Driver/IAggregateFluentExtensions.cs
rstam: Done

JamesKovacs (approved):
LGTM

PR #1007, opened 2023-01-03
CSHARP-4465: Include correct type in exception in BsonTypeMapper.MapToDotNetValue .

rstam:


rstam:
@BorisDog the BsonTypeMapper class is basically obsolete. It was created in the early days before serializers were invented.

Not sure what you are using BsonTypeMapper for, but you might want to avoid it.

JamesKovacs (approved):
LGTM

rstam:
Closing this PR without merging. Turns out there was already an external PR with the same fix, but the JIRA ticket didn't mention it.

PR #1008, opened 2023-01-06
CSHARP-4475: Add an AllowedTypes filter to ObjectSerializer.

rstam:


src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
```diff
@@ -651,13 +651,23 @@ private void SerializeDiscriminator(BsonSerializationContext context, Type nomin
...
+                }
             }
-            else
+            catch (Exception ex)
             {
-                SerializeExtraElements(context, obj, memberMap);
+                var message = string.Format(
+                    "An error occurred while serializing the {0} {1} of class {2}: {3}", // terminating period provided by nested message
+                    memberMap.MemberName, (memberMap.MemberInfo is FieldInfo) ? "field" : "property", memberMap.ClassMap.ClassType.FullName, ex.Message);
+                throw new BsonSerializationException(message, ex);
```
rstam, line 670: This change helped me figure out why some tests were failing.

We already do the same thing on Deserialize, so it seems like we should do it here in Serialize also.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
```diff
@@ -59,7 +58,7 @@ public DictionarySerializerBase()
         /// </summary>
         /// <param name="dictionaryRepresentation">The dictionary representation.</param>
         public DictionarySerializerBase(DictionaryRepresentation dictionaryRepresentation)
-            : this(dictionaryRepresentation, new ObjectSerializer(), new ObjectSerializer())
+            : this(dictionaryRepresentation, BsonSerializer.LookupSerializer<object>(), BsonSerializer.LookupSerializer<object>())
```
rstam, line 61: Look it up in the registry instead of newing it up to pick up any specially configured ObjectSerializer.

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -81,7 +81,7 @@ public DiscriminatedInterfaceSerializer(IDiscriminatorConvention discriminatorCo
 
             _interfaceType = typeof(TInterface);
             _discriminatorConvention = discriminatorConvention;
-            _objectSerializer = new ObjectSerializer(_discriminatorConvention);
+            _objectSerializer = ((ObjectSerializer)BsonSerializer.LookupSerializer<object>()).WithDiscriminatorConvention(_discriminatorConvention);
```
rstam, line 84: We need to lookup the ObjectSerializer in the registry and then reconfigure the discriminatorConvention.

WithDiscriminator is a new method.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -264,12 +315,27 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
...
         // private methods
         private object DeserializeDiscriminatedValue(BsonDeserializationContext context, BsonDeserializationArgs args)
         {
             var bsonReader = context.Reader;
 
             var actualType = _discriminatorConvention.GetActualType(bsonReader, typeof(object));
+            if (!_allowedTypes(actualType))
+            {
+                throw new BsonSerializationException($"Type {actualType.FullName} is not configured as an allowed type for this instance of ObjectSerializer.");
+            }
```
rstam, line 337: This is the key change that ensures only allowed types are deserialized.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -333,6 +399,11 @@ private object DeserializeDiscriminatedValue(BsonDeserializationContext context,
 
         private void SerializeDiscriminatedValue(BsonSerializationContext context, BsonSerializationArgs args, object value, Type actualType)
         {
+            if (!_allowedTypes(actualType))
+            {
+                throw new BsonSerializationException($"Type {actualType.FullName} is not configured as an allowed type for this instance of ObjectSerializer.");
+            }
+
```
rstam, line 406: While the most important check is in Deserialize I think Serialize should do the same thing. Why let you serialize something that can't be deserialized?

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -361,5 +432,93 @@ private void SerializeDiscriminatedValue(BsonSerializationContext context, BsonS
                 }
             }
         }
+
+        // nested types
+        private static class DefaultFrameworkAllowedTypes
+        {
+            private readonly static HashSet<Type> __allowedTypes = new HashSet<Type>
```
rstam, line 439: This list is pretty complete, but I'm sure someone will report missing types and we will add to this list over time.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -361,5 +432,93 @@ private void SerializeDiscriminatedValue(BsonSerializationContext context, BsonS
...
+                return type.IsConstructedGenericType ? IsAllowedGenericType(type) : IsAllowedType(type);
+
+                static bool IsAllowedType(Type type) =>
+                    __allowedTypes.Contains(type) ||
+                    type.IsArray && AllowedTypes(type.GetElementType()) ||
+                    type.IsEnum;
+
+                static bool IsAllowedGenericType(Type type) =>
+                    __allowedGenericTypes.Contains(type.GetGenericTypeDefinition()) &&
+                    type.GetGenericArguments().All(a => AllowedTypes(a));
```
rstam, line 520: A generic type is allowed if the generic type definition is allowed and all the type arguments are also allowed.

tests/MongoDB.Bson.TestHelpers/TestObjectSerializerRegisterer.cs
```diff
@@ -0,0 +1,58 @@
...
+                    // note: because xUnit has no way of letting us run some initialization code before any tests are run
+                    // and because xUnit runs tests in random order it's very likely some other test has had the side effect
+                    // of already registering a standard ObjectSerializer, so we have to use reflection below to replace
+                    // any existing ObjectSerializer with one configured with AllowedTypes for testing
+
+                    var testObjectSerializer = new ObjectSerializer(TestAllowedTypes);
+                    var serializerRegistry = BsonSerializer.SerializerRegistry;
+                    var cacheInfo = typeof(BsonSerializerRegistry).GetField("_cache", BindingFlags.NonPublic | BindingFlags.Instance);
+                    var cache = (ConcurrentDictionary<Type, IBsonSerializer>)cacheInfo.GetValue(serializerRegistry);
+                    cache.AddOrUpdate(typeof(object), testObjectSerializer, (objectType, existingObjectSerializer) => testObjectSerializer); // might replace existing ObjectSerializer!
```
rstam, line 44: I wouldn't normally write code like this, but it's the only way to get our tests to pass given that we can't reliably register a custom ObjectSerializer early enough.

tests/MongoDB.Bson.Tests/Jira/CSharp1559Tests.cs
```diff
@@ -26,6 +26,11 @@ namespace MongoDB.Bson.Tests.Jira
 {
     public class CSharp1559Tests
     {
+        static CSharp1559Tests()
+        {
+            TestObjectSerializerRegisterer.EnsureTestObjectSerializerIsRegistered();
+        }
```
rstam, line 32: This static constructor here and similarly in many classes below ensures that the correct ObjectSerializer is registered before running these tests.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
```diff
@@ -36,6 +38,32 @@ public class C
...
+
+            BsonClassMap.RegisterClassMap<T>(cm =>
+            {
+                cm.MapProperty(t => t.D).SetSerializer(dictionarySerializer);
+                cm.MapProperty(t => t.ID).SetSerializer(idictionarySerializer);
+                cm.MapProperty(t => t.IROD).SetSerializer(ireadOnlyDictionarySerializer);
+                cm.MapProperty(t => t.ROD).SetSerializer(readOnlyDictionarySerializer);
+                cm.MapProperty(t => t.SD).SetSerializer(sortedDictionarySerializer);
+                cm.MapProperty(t => t.SL).SetSerializer(sortedListSerializer);
+            });
```
rstam, line 64: The problem here is that while line 43 ensures that the correct ObjectSerializer is registered, line 43 doesn't help with any Dictionary (and similar) serializers that might already be registered and that have cached the wrong ObjectSerializer (as the key and value serializers) before line 43 was executed.

The workaround is to manually configure a class map for `T` that has all the correct serializers regardless of what may or may not already be in the registry.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DictionarySerializerTests.cs
```diff
@@ -53,6 +55,30 @@ public override int GetHashCode()
...
+            BsonClassMap.RegisterClassMap<T>(cm =>
+            {
+                cm.MapProperty(t => t.HT).SetSerializer(hashTableSerializer);
+                cm.MapProperty(t => t.ID).SetSerializer(iDictionarySerializer);
+                cm.MapProperty(t => t.LD).SetSerializer(listDictionarySerializer);
+                cm.MapProperty(t => t.OD).SetSerializer(orderedDictionarySerializer);
+                cm.MapProperty(t => t.SL).SetSerializer(sortedListDictionarySerializer);
+            });
+        }
+
```
rstam, line 81: Similar issue as above.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -294,7 +299,7 @@ public void constructor_with_discriminator_convention_should_initialize_instance
         [Fact]
         public void constructor_with_discriminator_convention_should_throw_when_discriminator_convention_is_null()
         {
-            var exception = Record.Exception(() => new ObjectSerializer(null));
```
rstam, line 297: With the addition of a new constructor `new ObjectSerializer(null)` has become ambiguous.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -294,7 +299,7 @@ public void constructor_with_discriminator_convention_should_initialize_instance
...
+            var exception = Record.Exception(() => new ObjectSerializer(discriminatorConvention: null));
```
rstam, line 302: Using `discriminatorConvention: null` resolves the ambiguity.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -24,10 +26,36 @@ namespace MongoDB.Bson.Serialization.Serializers
     /// </summary>
     public class ObjectSerializer : ClassSerializerBase<object>
     {
+        #region static
         // private static fields
+        private static readonly Func<Type, bool> __allAllowedTypes = t => true;
         private static readonly ObjectSerializer __instance = new ObjectSerializer();
+        private static readonly Func<Type, bool> __noAllowedTypes = t => false;
```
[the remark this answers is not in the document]
rstam: I struggled a bit with this name.

`AllowedTypes` is a suffix here.

This is a filter that allows `No` types (as opposed to `All` types).

That's how I came up with `NoAllowedTypes`.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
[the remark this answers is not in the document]
rstam: Good optimization. Done.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -361,5 +432,93 @@ private void SerializeDiscriminatedValue(BsonSerializationContext context, BsonS
...
+                typeof(System.Collections.Generic.LinkedList<>),
+                typeof(System.Collections.Generic.List<>),
+                typeof(System.Collections.Generic.Queue<>),
+                typeof(System.Collections.Generic.SortedDictionary<,>),
+                typeof(System.Collections.Generic.SortedList<,>),
+                typeof(System.Collections.Generic.SortedSet<>),
+                typeof(System.Collections.Generic.Stack<>),
+                typeof(System.Collections.ObjectModel.Collection<>),
+                typeof(System.Collections.ObjectModel.ReadOnlyCollection<>),
+                typeof(System.Collections.ObjectModel.ReadOnlyDictionary<,>),
```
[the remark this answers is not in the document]
rstam: I can't remember why I omitted KeyValuePair. Don't see any harm in adding it.

I got my list by scanning through the types loaded into Reflector. PriorityQueue is not there. Admittedly I was looking at net472 DLLs. Maybe PriorityQueue is not in net472, or maybe it's in a DLL I didn't have loaded into Reflector.

I'll try to think of a better list to check against.

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -127,6 +117,40 @@ public void RegisterSerializationProvider(IBsonSerializationProvider serializati
...
+            {
+                throw new ArgumentNullException("type");
+            }
+            if (serializer == null)
+            {
+                throw new ArgumentNullException("serializer");
+            }
+            EnsureRegisteringASerializerForThisTypeIsAllowed(type);
+
+            if (_cache.TryAdd(type, serializer))
```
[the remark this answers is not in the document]
rstam: I considered using `GetOrAdd` but it doesn't quite fit our needs in this case.

We need to know whether the value was added or not (i.e. whether it already existed) in order to check for error and provide the proper error message when attempting to register a serializer that is already registered.

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -127,6 +117,40 @@ public void RegisterSerializationProvider(IBsonSerializationProvider serializati
...
+            {
+                return true;
+            }
+            else
+            {
+                var existingSerializer = _cache[type];
+                if (!existingSerializer.Equals(serializer))
+                {
+                    var message = string.Format("There is already a different serializer registered for type {0}.", BsonUtils.GetFriendlyTypeName(type));
```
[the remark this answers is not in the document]
rstam: Sure, I like string interpolation a lot.

This was just copy/pasted from existing code and adapted.

I will change it to string interpolation here in this new code, but I'm reluctant to change existing code.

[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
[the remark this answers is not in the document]
rstam: Do you want to help me come up with more types to add to this list?

I'm going to push a new commit addressing your other comments, but leave this pending for now.

tests/MongoDB.Bson.TestHelpers/RegisterObjectSerializerFixture.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Bson.TestHelpers
+{
+    public class RegisterObjectSerializerFixture
+    {
```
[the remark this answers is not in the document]
rstam: sealed is good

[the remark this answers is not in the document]
rstam: Done

tests/MongoDB.Bson.Tests/Jira/CSharp1559Tests.cs
rstam: Turns out that if you don't actually need to use the fixture in your tests you can just omit this constructor.

We're only using the collection fixture for its side effect of registering the object serializer.

I'm going to remove all these constructors.

tests/MongoDB.Bson.Tests/RegisterObjectSerializerCollection.cs
```diff
@@ -0,0 +1,25 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using MongoDB.Bson.TestHelpers;
+using Xunit;
+
+namespace MongoDB.Bson.Tests
+{
+    [CollectionDefinition("RegisterObjectSerializer")]
```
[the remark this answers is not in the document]
rstam: Done

tests/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
rstam: Originally this was done here because this was the only test affected by this.

But it does make sense to do it `RegisterObjectSerializerFixture`.

rstam: There is a problem though... we can't move the class map registration to the fixture.

I'm going to leave this alone for now. Think about it a bit more and let me know if you want to pursue it further.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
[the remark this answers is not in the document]
BorisDog: I can see just those two: 
`ObservableCollection`,
`ReadOnlyObservableCollection`

tests/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
BorisDog: Maybe we could just register serializers for `Dictionary<>`, `IDictionary<>` like we do with `ObjectSerializer` and skip the mapping?

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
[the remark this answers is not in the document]
BorisDog: Sorry, I missed that `GetOrAdd` return value would not allow us to indicate if it was already inserted.

src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
JamesKovacs: SGTM

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
JamesKovacs: Agreed.

JamesKovacs (approved):
LGTM

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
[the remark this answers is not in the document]
BorisDog: Reminder, do we want to the mentioned types above as well?

tests/MongoDB.Bson.Tests/Serialization/Serializers/DictionaryGenericSerializerTests.cs
BorisDog: Reminder.

BorisDog (changes requested):

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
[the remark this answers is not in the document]
rstam: The full list I added in this commit is:

    ListDictionary
    OrderedDictionary
    Version
    KeyValuePair<,>
    KeyedCollection<,>
    ObservableCollection<>
    ReadOnlyObservableCollection<>
    System.Tuple<,,,,,,,> // one missing Tuple class

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -165,6 +217,20 @@ public override object Deserialize(BsonDeserializationContext context, BsonDeser
             }
         }
 
+        /// <inheritdoc/>
+        public override bool Equals(object obj) =>
+            obj is ObjectSerializer other &&
```
rstam, line 222: For non-sealed classes `obj is ObjectSerializer` is not a good enough test because `obj` might be a subclass of `ObjectSerializer` in which case we want to return `false`.

Because `ObjectSerializer` is `sealed` we know that if obj is an `ObjectSerializer` it can't be a subclass of `ObjectSerializer`.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -165,6 +217,20 @@ public override object Deserialize(BsonDeserializationContext context, BsonDeser
...
+            _allowedTypes.Equals(other._allowedTypes) &&
+            _discriminatorConvention.Equals(other._discriminatorConvention) &&
+            _guidRepresentation == other._guidRepresentation;
+
+        /// <inheritdoc/>
+        public override int GetHashCode() =>
+            new Hasher()
+            .Hash(_discriminatorConvention)
```
rstam, line 230: Deliberately left `_alowedTypes` out of the computation of the hash code because we don't know anything about the behavior of `GetHashCode` for delegates (and it's not worth finding out).

Technically `public override int GetHashCode() => 0` would be correct also if we don't anticipate serializers being put into a hash table (and even if they are `0` would just be a small performance hit).

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -14,20 +14,49 @@
...
 using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Shared;
 
 namespace MongoDB.Bson.Serialization.Serializers
 {
     /// <summary>
     /// Represents a serializer for objects.
     /// </summary>
-    public class ObjectSerializer : ClassSerializerBase<object>
+    public sealed class ObjectSerializer : ClassSerializerBase<object>
```
rstam, line 28: Changed to `sealed` in part to simplify the implementation of `Equals`.

Any concern that this could be considered a breaking change?

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -165,6 +217,20 @@ public override object Deserialize(BsonDeserializationContext context, BsonDeser
             }
         }
 
+        /// <inheritdoc/>
```
rstam, line 220: There is another ticket to implement `Equals` for *all* serializers.

In this PR I'm only implementing `Equals` for `ObjectSerializer` because it is required for the tests of `TryRegisterSerializer` to pass.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -165,6 +217,20 @@ public override object Deserialize(BsonDeserializationContext context, BsonDeser
...
+        public override bool Equals(object obj) =>
+            obj is ObjectSerializer other &&
+            _allowedTypes.Equals(other._allowedTypes) &&
+            _discriminatorConvention.Equals(other._discriminatorConvention) &&
+            _guidRepresentation == other._guidRepresentation;
```
rstam, line 225: I firmly believe `Equals` should be implemented by calling `Equals` on fields in order to preserve the semantics of `Equals` (without being forced to analyze whether `==` is semantically equivalent to `Equals` or not on a case by case basis).

But... in the case of `enum` fields calling `Equals` would involve boxing. It would still be correct, but there would be a small performance hit.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
BorisDog: I agree with your remark that `GetHashCode() => 0` would work, as this class is not designed to be used for lookup in hash tables.
Would not `=> base.GetHashCode()` be sufficient then?
The default implementation calculates the hash based on reference, so I would recommend to offload this to .net framework.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
JamesKovacs: It will break anyone with a `IBsonSerializer<T>` deriving from `ObjectSerializer`. A quick google for `mongodb objectserializer` returns our API docs and a few SO questions, but no one recommending deriving a new serializer from `ObjectSerializer`. It's probably not common, but we can't know for certain.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
JamesKovacs: Agreed.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
BorisDog: We can avoid this by checking `other.GetType()`

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
JamesKovacs: To avoid the `sealed` breaking change, would the following not be sufficient?
```
obj is ObjectSerializer other &&
GetType() == obj.GetType() &&
```

JamesKovacs (changes requested):
While I suspect that no one has subclassed `ObjectSerializer` there is no way to know for sure. See my suggested change to `Equals` avoids the need for this breaking change.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
rstam: Removing sealed to avoid possible though unlikely breaking change.

JamesKovacs (approved):
LGTM

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
BorisDog: > If Equals returns true for x and y then x.GetHashCode() MUST EQUAL y.GetHashCode(). Calling base.GetHashCode() violates that.

I think this only applies to cases where `Hashcode` is actually used, which is lookups. But I agree that it's a "bon ton" not to violate this in any case.

I think 0 is a good option, to state that "we don't care", otherwise using `Hasher` leads to wrong impression that this class is designed and optimized to be used for lookups.

BorisDog (approved):
LGTM.
Comment re. `GetHashcode` for your consideration.

PR #1014, opened 2023-01-30
CSHARP-4499: Fix ConvertExpressionToFilterFieldTranslator when target type is object.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/ConvertExpressionToFilterFieldTranslator.cs
```diff
@@ -135,6 +145,12 @@ private static AstFilterField TranslateConvertEnumToUnderlyingType(AstFilterFiel
             return AstFilter.Field(field.Path, targetSerializer);
         }
 
+        private static AstFilterField TranslateConvertToBaseType(AstFilterField field, Type targetType)
+        {
+            var targetSerializer = BsonSerializer.LookupSerializer(targetType);
```
rstam, line 150: I'm a little concerned that the target serializer in the registry might not be compatible with the field serializer, but we have no was of asking a serializer "give me a compatible serializer for one of your base classes".

The same problem already exists in `TranslateConvertToDerivedType` in the other direction.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/ConvertExpressionToFilterFieldTranslator.cs
```diff
@@ -135,6 +145,15 @@ private static AstFilterField TranslateConvertEnumToUnderlyingType(AstFilterFiel
             return AstFilter.Field(field.Path, targetSerializer);
         }
 
+        private static AstFilterField TranslateConvertToBaseType(AstFilterField field, Type targetType)
+        {
+            var baseType = targetType;
+            var derivedType = field.Serializer.ValueType;
+            var derivedTypeSerializer = field.Serializer;
+            var targetSerializer = DowncastingSerializer.Create(baseType, derivedType, derivedTypeSerializer);
+            return AstFilter.Field(field.Path, targetSerializer);
```
rstam, line 154: The idea here is that we need a serializer for TBase that enforces that actual values are of type TDerived (by downcasting) when Serialize is called.

If the field is only being used to figure out the name this change is moot, but for any cases where the field is also used to serialize constant values this enforces that the values are of the proper type for the field.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4499Tests.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class CSharp4499Tests : Linq3IntegrationTest
+    {
+        [Theory]
+        [ParameterAttributeData]
+        public void ExpressionFieldDefinition_should_work([Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
+        {
+            var collection = CreateCollection(linqProvider);
+            var fieldDefinition = new ExpressionFieldDefinition<ConcreteClass, object>(x => x.InternalId);
+            //var fieldDefinition = new ExpressionFieldDefinition<ConcreteClass>((ConcreteClass x) => x.InternalId);
+            //var fieldDefinition = new ExpressionFieldDefinition<ConcreteClass, ObjectId>(x => x.InternalId);
```
rstam, line 32: These were the two workarounds that didn't use `object`.

I should remove them before pushing to master.

JamesKovacs (approved):
LGTM. Don't forget to remove the commented workarounds in the test code.

PR #1015, opened 2023-01-30
CSHARP-4506: Use EncryptionFieldsMap with CreateEncryptedCollection.

DmitryLukyanov:


DmitryLukyanov:
It might be that this change is not needed since the spec currently considers revering this requirement. Let's put this PR in draft until the actual spec wording will be changed. If it will be reverted, the only required logic in this PR will be skipping fle2-Range tests on macos.

DmitryLukyanov:
This requirement was cancelled in this PR https://github.com/mongodb/specifications/pull/1376

PR #1016, opened 2023-02-01
CSHARP-4514: createEncryptedCollection should not accept keyAltNames.

DmitryLukyanov:


DmitryLukyanov:
There is no c# ticket for this change yet, I will update this description as soon as driver ticket will be moved to implementing

PR #1017, opened 2023-02-01
CSHARP-4509: Handle expression based field selection when using object instead of TField.

rstam:


rstam:
Based on csharp4499.

Only review last two commits (which I probably should have squashed into one, but at first I thought rebasing on csharp4499 alone would fix the issue).

JamesKovacs (approved):
LGTM

PR #1018, opened 2023-02-02
Implement IDisposable in MongoClient

semyon-p:
https://jira.mongodb.org/browse/CSHARP-4513

My suggestion is to implement IDisposable in MongoClient, it'll allow controlling connections in such kind of way.

using (var client = new MongoClient("mongodb://localhost:27017"))
{
    var db = client.GetDatabase("config");
    var collection = db.GetCollection<Element>("test");
    var x = collection.Aggregate().FirstOrDefault();
    Console.WriteLine($"{x.Name} : {x.Value}");
}

### DIFF
--- a/src/MongoDB.Driver/MongoClient.cs
+++ b/src/MongoDB.Driver/MongoClient.cs
@@ -35,7 +35,7 @@ using MongoDB.Driver.Linq;
 namespace MongoDB.Driver
 {
     /// <inheritdoc/>
-    public class MongoClient : MongoClientBase
+    public class MongoClient : MongoClientBase, IDisposable
     {
         #region static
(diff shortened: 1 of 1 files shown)


JamesKovacs:
Thank you for submitting this pull request. Unfortunately this fix does not work in the general case because {{Cluster}} can be shared between multiple {{MongoClient}} instances based on shared {{ClusterKey}}. It is not possible to simply dispose the {{Cluster}} when the {{MongoClient}} is disposed. The engineering work required to implement a disposable {{MongoClient}} is captured in [CSHARP-3431](https://jira.mongodb.org/browse/CSHARP-3431).

PR #1019, opened 2023-02-06
CSHARP-4480: Add a test that drivers emit a CommandSucceededEvent when ok=1 and a writeConcernError is returned.

DmitryLukyanov:


BorisDog (approved):
LGTM, assuming EG is green :)

PR #1021, opened 2023-02-06
CSHARP-4517: Values being compared must be of the same type.

rstam:


JamesKovacs (approved):
LGTM

JamesKovacs (commented):
There are a lot of LINQ2 test failures in the patch build - which doesn't make immediate sense to me as the code changes are in LINQ3. Please investigate these failures before merging.

JamesKovacs (changes requested):
Changes look good but the patch build is showing a lot of LINQ2 failures, which don't make sense to me as the changes were isolated to LINQ3 code. Please investigate before merging.

rstam:
The tests that are failing are not LINQ2 test but rather Linq2ImplementationTestsOnLinq3 tests.

I'm going to have to rework this fix.

I'll re-request your review when it's ready.

rstam:
Looks like it was sufficient to relax the requirement that operands be of the *same* type to just require that they be of *compatible* types, where compatible means one of:

* one operand is null
* both operands are the same type
* one operand can be implicitly converted to the other

JamesKovacs (approved):
LGTM

PR #1022, opened 2023-02-06
CSHARP-4505: Repro.

rstam:


JamesKovacs (commented):
Agreed. Good to verify, but no need to push to our main branch. You can close this PR without merging.

rstam:
Closed without merging since it is simply proving that something already works.

PR #1024, opened 2023-02-08
CSHARP-4496: Add decimal support to ObjectSerializer.

JamesKovacs:
Added support for `ObjectSerializer` to both read and write `Decimal128` data type.


rstam (approved):
LGTM

PR #1026, opened 2023-02-14
CSHARP-4524: Handle projections using constructors for classes that use public fields instead of public properties.

rstam:


JamesKovacs (approved):
I found an edge case where `new SpawnData(x.StartDate, x.SpawnPeriod) { ExtraInfo = 42 }` doesn't work because `MemberInitExpressionToAggregationExpressionTranslator` behaves differently than `NewExpressionToAggregationExpressionTranslator`. Split into https://jira.mongodb.org/browse/CSHARP-4550.

LGTM. We can fix the MemberInit edge case in the other ticket.

PR #1028, opened 2023-02-17
CSHARP-4054: One-to-many LINQ query using Join fails with InvalidOperationException.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/JoinMethodToPipelineTranslator.cs
```diff
@@ -38,10 +38,23 @@ public static AstPipeline Translate(TranslationContext context, MethodCallExpres
...
+                if (pipeline.OutputSerializer is IWrappedValueSerializer pipelineOutputWrappedSerializer)
+                {
+                    outerAst = AstExpression.GetField(root, pipelineOutputWrappedSerializer.FieldName);
+                    outerSerializer = pipelineOutputWrappedSerializer.ValueSerializer;
+                }
+                else
+                {
+                    outerAst = root;
+                    outerSerializer = pipeline.OutputSerializer;
+                }
```
rstam, line 54: This new code is to avoid double wrapping.

In this case the output of the pipeline (from line 40) coming into the Join was already wrapped. So we want the stage being added in line 56 to be:

    { $project : { _v : '$_v', _id : 0 } }

instead of 

    { $project : { _v : '$$ROOT', _id : 0 } }

JamesKovacs (approved):
LGTM

PR #1032, opened 2023-02-23
CSHARP-4542: LINQ3 support for projecting to a ValueTuple.

rstam:


rstam:
No driver changes.

Just verifying that CSHARP-4524 fixes the issue. The difference between `Tuple` and `ValueTuple` is that `Tuple` has *properties* called `Item1` etc and `ValueTuple` has *fields` called `Item1` etc.

Note; the driver is using `BsonClassMapSerializer` for `ValueTuple`, we probably should provide custom serializers for all the `ValueTuple` types for better performance. If you agree I can create a ticket for that.

JamesKovacs (approved):
LGTM

> Note; the driver is using `BsonClassMapSerializer` for `ValueTuple`, we probably should provide custom serializers for all the `ValueTuple` types for better performance. If you agree I can create a ticket for that.

Agreed. Let's create a separate ticket to create customer serializers for ValueTuples.

rstam:
Closing without merging because the new tests merely verify that CSHARP-4524 fixed the issue.

PR #1039, opened 2023-03-13
CSHARP-4564: Implement ValueSerializers.

rstam:


src/MongoDB.Bson/Serialization/Serializers/ValueTupleSerializers.cs
```diff
@@ -0,0 +1,1393 @@
...
+            T1 item1 = default;
+            switch (reader.GetCurrentBsonType())
+            {
+                case BsonType.Array:
+                    reader.ReadStartArray();
+                    item1 = _lazyItem1Serializer.Value.Deserialize(context);
+                    reader.ReadEndArray();
+                    break;
+
+                case BsonType.Document:
```
rstam, line 91: ValueTuples serialized before this serializer was implemented will have been serialized as documents using the `BsonClassMapSerializer`.

We could make the representation configurable. If you want to do that I suggest we do that as a separate ticket.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ValueTupleSerializersTests.cs
```diff
@@ -0,0 +1,807 @@
...
+        [Fact]
+        public void Deserialize_should_deserialize_document_value()
+        {
+            var json = "{ \"_id\" : 1, \"T\" : { Item1 : false } }";
+            var expectedValue = CreateValue();
+
+            var value = BsonSerializer.Deserialize<C>(json);
+
+            Assert.Equal(expectedValue, value);
+        }
```
rstam, line 85: There are no tests for serializing or deserializing `null` because `ValueTuple` is a `struct` (as opposed to `Tuple` which is a `class`).

rstam:
A good way to review this is to compare the existing TupleSerializer to the new ValueTupleSerializer and the same for the tests.

Since ValueTupleSerializer is a clone of TupleSerializer I've made the smallest changes possible (so no formatting improvements).

src/MongoDB.Bson/Serialization/Serializers/ValueTupleSerializers.cs
```diff
@@ -0,0 +1,1393 @@
...
+            _lazyItem1Serializer = new Lazy<IBsonSerializer<T1>>(() => item1Serializer);
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ValueTupleSerializer{T1}" /> class.
+        /// </summary>
+        /// <param name="serializerRegistry">The serializer registry.</param>
+        public ValueTupleSerializer(IBsonSerializerRegistry serializerRegistry)
+        {
+            if (serializerRegistry == null) { throw new ArgumentNullException("serializerRegistry"); }
```
BorisDog, line 57: `nameof`

src/MongoDB.Bson/Serialization/Serializers/ValueTupleSerializers.cs
```diff
@@ -0,0 +1,1393 @@
...
+
+            _lazyItem1Serializer = new Lazy<IBsonSerializer<T1>>(() => serializerRegistry.GetSerializer<T1>());
+        }
+
+        // public properties
+        /// <summary>
+        /// Gets the Item1 serializer.
+        /// </summary>
+        public IBsonSerializer<T1> Item1Serializer
```
BorisDog, line 66: consider shorter form public `IBsonSerializer<T1> Item1Serializer => _lazyItem1Serializer.Value` everywhere.

BorisDog (changes requested):
Few minor comments

src/MongoDB.Bson/Serialization/Serializers/ValueTupleSerializers.cs
rstam: I mentioned in a comment that this file was cloned from `TupleSerializers.cs` and that I deliberately avoided making any formatting changes to make it easy to diff the two files.

Do you still think we should make these changes? And if yes, do I make the same changes to `TupleSerializers.cs`?

BorisDog: I don't see any real-world use case where we would need to diff the two files. But either way is fine. For your consideration.

BorisDog (approved):
LGTM.

rstam (commented):
Applied the requested formatting changes.

BorisDog (approved):
LGTM

PR #1045, opened 2023-03-22
CSHARP-4576: Atlas search: support score in Compound operator

BorisDog:


tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
```diff
@@ -80,12 +80,20 @@ public void Autocomplete()
...
+            var searchDefinition = Builders.Search.Compound(Builders.SearchScore.Constant(score))
                 .Must(Builders.Search.Text(x => x.Body, "life"), Builders.Search.Text(x => x.Body, "liberty"))
                 .MustNot(Builders.Search.Text(x => x.Body, "property"))
-                .Must(Builders.Search.Text(x => x.Body, "pursuit of happiness")));
+                .Must(Builders.Search.Text(x => x.Body, "pursuit of happiness"));
 
+            var projectionDefinition = Builders.Projection
+                .Include(x => x.Body)
+                .Include(x => x.Title)
+                .MetaSearchScore("score");
```
JamesKovacs, line 92: We should include strongly-typed expressions as well as string-based field names.

JamesKovacs (changes requested):
Looks good. We should include `MetaSearchScore(x => x.Score)` rather than forcing users to specify the field name as a string.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
BorisDog: I agree, that's also applies to `MetaSearchHighlights` and `MetaTextScore`.
I think it's better to address this in separate ticket.

JamesKovacs: Fair enough. Create a separate ticket and it can be implemented post-2.19.1.

JamesKovacs (approved):
LGTM

PR #1049, opened 2023-03-28
CSHARP-4551: Avoid using PID when generating ObjectIds.

DmitryLukyanov:


src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
```diff
@@ -44,26 +38,7 @@ public class AscendingGuidGenerator : IIdGenerator
...
-
-            __machineProcessId = new byte[5]
-            {
-                machineHash[0],
-                machineHash[1],
-                machineHash[2],
-                (byte)(processId >> 8),
-                (byte)(processId)
-            };
+            __machineProcessId = BitConverter.GetBytes(ObjectId.CalculateRandomValue());
```
rstam, line 41: Doesn't this create 8 bytes?

rstam: Calculate RandomValue returns a 5-byte value. (OK 8 bytes but 3 bytes are zeros).

Make sure you are using the right 5 bytes.

Probably better to select the right 5 bytes here than later.

rstam (changes requested):

src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
DmitryLukyanov: Yes, but we still use only 5 in the below steps. Yes, we can do it now too, but we also need to do it later in any case (since GenerateId is public). Do you think it's worth creating a new array just for this simple issue?

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -421,7 +425,8 @@ private static ObjectId Create(int timestamp, long random, int increment)
         [MethodImpl(MethodImplOptions.NoInlining)]
         private static int GetCurrentProcessId()
         {
-            return Process.GetCurrentProcess().Id;
+            using var process = Process.GetCurrentProcess();
```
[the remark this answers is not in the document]
rstam: I looked for his comment and the better way is `Environment.ProcessId`.

But it's not available on any of our target frameworks.

src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
rstam: I think it's worth it to get the CORRECT 5 bytes now, and then we know we're good.

I'd also rename the field `__random`. (Or if you prefer `__random5Bytes`)

So here's the code I would suggest for the static constructor:

    // static constructor
    static AscendingGuidGenerator()
    {
        var random = ObjectId.CalculateRandomValue();
        var random8Bytes = BitConverter.GetBytes(random);
        __random = new byte[5];
        Array.Copy(random8Bytes, __random, 5); // the 5 bytes we need are the first 5 bytes assuming little-endian
    }

I don't know if we still support big-endian machines. If so we need to get the last 5 bytes instead on a big-endian machine.

Allocating an extra array doesn't matter. This is a static constructor and only executes once. Since we only need 5 bytes our array should be 5 bytes.

src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
```diff
@@ -129,32 +104,5 @@ public bool IsEmpty(object id)
         {
```
rstam, line 104: I would modify the doc comment for the `machineProcessId` parameter to read:

    /// <param name="machineProcessId">A 5 byte array with 5 random bytes.
    /// Originally the first 3 bytes represented a machine id and the next 2 bytes
    /// represented the process id but that is no longer true. </param>

I would add an intermediate variable to the implementation for clarity:

    public object GenerateId(
        long tickCount,
        byte[] machineProcessId,
        int increment)
    {
        if (machineProcessId == null) { throw new ArgumentNullException(nameof(machineProcessId)); }
        if (machineProcessId.Length != 5) { throw new ArgumentException($"{nameof(machineProcessId)} argument must be exactly 5 bytes", nameof(machineProcessId)); }
        var random5Bytes = machineProcessId; // changing the parameter name could be considered a breaking change
        var a = (int)(tickCount >> 32);
        var b = (short)(tickCount >> 16);
        var c = (short)(tickCount);
        var d = new byte[8];
        Array.Copy(random5Bytes, d, 5);
        d[5] = (byte)(increment >> 16);
        d[6] = (byte)(increment >> 8);
        d[7] = (byte)(increment);
        return new Guid(a, b, c, d);
    }

src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
rstam: Good  point about `GenerateId` being public. We aren't very good about validating arguments in this method but we should thrown an `ArgumentException` if the array passed in is *not* exactly 5 bytes long.

rstam (changes requested):

src/MongoDB.Bson/Serialization/IdGenerators/AscendingGuidGenerator.cs
DmitryLukyanov: done

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -385,17 +384,22 @@ public static void Unpack(byte[] bytes, out int timestamp, out int machine, out
             increment = (bytes[9] << 16) + (bytes[10] << 8) + bytes[11];
         }
 
-        // private static methods
-        private static long CalculateRandomValue()
+        // internal static methods
+        internal static long CalculateRandomValue()
         {
+#if NET472
```
rstam, line 390: Should this be `NET472_OR_GREATER`?

rstam (changes requested):
Looking good.

One small question I forgot to ask last time around.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
DmitryLukyanov: Checked that the same collision behavior happens with .net frameworks higher than 472 like 48 too. Done

rstam (approved):
LGTM assuming we don't need to consider big-endian machines.

DmitryLukyanov:
> LGTM assuming we don't need to consider big-endian machines.

I think, we can add big-endian support later if we have request for it

PR #1051, opened 2023-03-30
CSHARP-4483: connectionId returned in heartbeats may be int32, double, or int64.

JamesKovacs:
Although CSHARP-4417 changes the client and server connectionIds to use `int64`, the server description update code path was not tested. The update code path is where the server response is used update the `ConnectionId.LongServerValue`. `HelloResult.ConnectionIdServerValue` still parsed `connectionId` in the response as an `int32`. This has been fixed and additional tests added.

Although `HelloResult` is public, it is in Core. We renamed this class during the offensive language removal initiative and no one complained. Thus changing the data type of `HelloResult.ConnectionIdServerValue` from `int?` to `long?` shouldn't be problematic.

### DIFF
--- a/tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
+++ b/tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
@@ -128,7 +128,9 @@ namespace MongoDB.Driver.Core.Connections
 
         [Theory]
         [ParameterAttributeData]
-        public void InitializeConnection_should_acquire_connectionId_from_hello_response([Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId, [Values(false, true)] bool async)
+        public void InitializeConnection_should_acquire_connectionId_from_hello_response(
+            [Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId,
+            [Values(false, true)] bool async)
         {
             var formattedServerConnectionId = $"{serverConnectionId}" + (serverConnectionId is double ? ".0" : "");
             var helloResponse = MessageHelper.BuildCommandResponse(RawBsonDocumentHelper.FromJson($"{{ ok : 1, connectionId : {formattedServerConnectionId} }}"));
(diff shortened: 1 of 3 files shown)


tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
```diff
@@ -128,9 +128,10 @@ public void ConnectionInitializerContext_should_throw_when_arguments_are_null()
 
         [Theory]
         [ParameterAttributeData]
-        public void InitializeConnection_should_acquire_connectionId_from_hello_response([Values(false, true)] bool async)
+        public void InitializeConnection_should_acquire_connectionId_from_hello_response([Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId, [Values(false, true)] bool async)
```
rstam, line 131: Consider reformatting on multiple lines:

    [Theory]
    [ParameterAttributeData]
    public void InitializeConnection_should_acquire_connectionId_from_hello_response(
        [Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId,

tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
```diff
@@ -140,15 +141,15 @@ public void InitializeConnection_should_acquire_connectionId_from_hello_response
...
             var sentMessages = connection.GetSentMessages();
             sentMessages.Should().HaveCount(1);
-            result.ConnectionId.LongServerValue.Should().Be(1);
+            result.ConnectionId.LongServerValue.ShouldBeEquivalentTo(serverConnectionId);
         }
 
         [Theory]
         [ParameterAttributeData]
-        public void InitializeConnection_should_acquire_connectionId_from_legacy_hello_response([Values(false, true)] bool async)
+        public void InitializeConnection_should_acquire_connectionId_from_legacy_hello_response([Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId, [Values(false, true)] bool async)
```
rstam, line 149: Consider reformatting on multiple lines:

    [Theory]
    [ParameterAttributeData]
    public void InitializeConnection_should_acquire_connectionId_from_legacy_hello_response(
        [Values(1, int.MaxValue, (long)int.MaxValue + 1, long.MaxValue, 1d, (double)int.MaxValue+1, (double)int.MaxValue*4)] object serverConnectionId,

rstam (changes requested):
Two minor changes requested.

I'm also willing to LGTM as is.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
JamesKovacs: Done.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ConnectionInitializerTests.cs
JamesKovacs: Done.

PR #1052, opened 2023-03-30
CSHARP-4586: Map constructor in MemberInit and simplify $project for simple field inclusion.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstSimplifier.cs
```diff
@@ -322,6 +322,20 @@ public override AstNode VisitMatchStage(AstMatchStage node)
             return node;
         }
 
+        public override AstNode VisitProjectStageSetFieldSpecification(AstProjectStageSetFieldSpecification node)
```
rstam, line 325: I can't believe we weren't doing this simplification. Apparently neither was LINQ2 and that's why we didn't notice it (because all the LINQ2 tests we copied were not relying on this simplification).

This change did result in a modest number of tests having to be updated. Take a look and see if you are OK with how the tests changed.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -60,6 +64,7 @@ public static AggregationExpression Translate(TranslationContext context, Member
 
             var ast = AstExpression.ComputedDocument(computedFields);
 
+            classMap.MapConstructor(constructorInfo, memberNames);
```
rstam, line 67: You pointed out some days back that New and MemberInit were not implemented exactly the same.

This was one of the differences.

Sometimes the constructor would have been mapped automatically when automapping, but the user's test class is "partially" immutable and we don't handle that well in automapping. That's a separate issue as well.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4586Tests.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            public Model(string id)
+            {
+                Id = id;
+            }
+
+            public string Id { get; private set; }
+        }
+
+        private class View1
```
rstam, line 187: This class maps well because it is immutable.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationTests/Jira/CSharp4586Tests.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            public View1(string id)
+            {
+                Id = id;
+            }
+
+            public string Id { get; }
+        }
+
+        private class View2
```
rstam, line 197: This class does not map well because we don't think it is immutable, even though it kind of is.

The `init` property throws our mapping off. We think it is a writable property and that therefore the class is not immutable.

A somewhat unfortunate decision we made many years ago was that we do NOT automap constructors UNLESS the class is immutable. This decision was made in the name of backward compatibility, though I'm not sure how breaking mapping constructors is (for the first few years we didn't support constructors at all and when we did start supporting them we made the decision to only automap them for immutable classes).

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstSimplifier.cs
rstam: It's possible that we might only want to perform this transformation for Find projections? (as opposed to for `$project` stages in an aggregation pipeline).

It's not strictly necessary in aggregation pipelines. It's also no longer strictly necessary in Find projections when using newer servers, but it is mandatory when using older servers that don't support aggregation style projections with Find.

What do you think?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
rstam: Note that this call to `MapConstructor` is a no-op if the class map already has a creator mapped for this constructor.

That's why this code already worked with immutable classes (because AutoMap automatically maps constructors for immutable classes) and didn't work with the user's "semi-immutable" class which we didn't consider to be immutable  because it has an `init` property setter (and therefore AutoMap ignored the constructor during mapping).

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstSimplifier.cs
JamesKovacs: Seems like a good optimization.

rstam:
I missed that one failing test because it was in another project!

JamesKovacs (approved):
LGTM

PR #1064, opened 2023-04-13
CSHARP-4595: Support GridFS with Stable API.

JamesKovacs:
`GridFSBucket.Find` was failing because we always included `noCursorTimeout` even when not specified. `noCursorTimeout` is not supported with Stable API. That change was simple (removing the defaulting to false), but we had no integration tests around GridFS. Additionally while we test Stable API, we don't test it in strict mode nor do we have a mechanism to do so. I added `Feature.StableApi` so that I could write a strict mode Stable API test. Feedback appreciated especially if you can think of a better way to test it.


DmitryLukyanov (approved):
LGTM

PR #1065, opened 2023-04-13
CSHARP-4609: Replace DowncastingSerializer with more permissive ConvertIfPossibleSerializer when rendering values associated with a FieldDefnition.

rstam:


src/MongoDB.Bson/Serialization/Serializers/DowncastingSerializer.cs
```diff
@@ -17,6 +17,27 @@
 
 namespace MongoDB.Bson.Serialization.Serializers
 {
+    /// <summary>
+    /// An interface implemented by DowncastingSerializer.
+    /// </summary>
+    public interface IDowncastingSerializer
```
rstam, line 23: This interface makes it possible to work with a DowncastingSerializer<TBase, TDerived> without using reflection (because the TBase and TDerived generic type arguments can be anything at runtime).

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -38,6 +38,13 @@ public static IBsonSerializer GetSerializerForValueType(IBsonSerializer fieldSer
             // these will normally be equal unless we've removed some Convert(s) that the compiler put in
             if (fieldType == valueType)
             {
+                if (fieldSerializer is IDowncastingSerializer downcastingSerializer)
+                {
+                    var derivedSerializer = downcastingSerializer.DerivedSerializer;
+                    var derivedType = derivedSerializer.ValueType;
+                    return ConvertIfPossibleSerializer.Create(valueType, derivedType, derivedSerializer, serializerRegistry);
+                }
+
```
rstam, line 47: This will have NO impact on LINQ2, because when using LINQ2 we will never encounter an instance a DowncastingSerializer. Only LINQ3 uses DowncastingSerializer.

Essentially this change substitutes a stricter DowncastingSerializer with a more permissive ConvertIfPossibleSerializer, but *only* in the context of working with `FieldDefinition` (which should prevent destabilizing LINQ3 in any way).

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -177,6 +181,20 @@ private static IBsonSerializer WithStringRepresentation(IBsonSerializer serializ
...
+            public static IBsonSerializer Create(
+                Type valueType,
+                Type fieldType,
+                IBsonSerializer fieldSerializer,
+                IBsonSerializerRegistry serializerRegistry)
+            {
+                var convertIfPossibleSerializerType = typeof(ConvertIfPossibleSerializer<,>).MakeGenericType(valueType, fieldType);
+                var convertIfPossibleSerializerConstructor = convertIfPossibleSerializerType.GetTypeInfo().GetConstructor(new[] { fieldSerializer.GetType(), typeof(IBsonSerializerRegistry) });
+                return (IBsonSerializer)convertIfPossibleSerializerConstructor.Invoke(new object[] { fieldSerializer, serializerRegistry });
+            }
```
rstam, line 195: This helper method simply encapsulates in one place the reflection base code to create an instance of a ConvertIfPossibleSerializer, because now there are two place we might create one.

JamesKovacs (approved):
LGTM

PR #1067, opened 2023-04-17
CSHARP-4492: Drivers should not create the ECC collection in v2 of queryable encryption.

DmitryLukyanov:


src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
```diff
@@ -82,6 +81,7 @@ CreateCollectionOperation CreateInnerCollectionOperation(string collectionName)
         private readonly MessageEncoderSettings _messageEncoderSettings;
         private bool? _noPadding;
         private BsonDocument _storageEngine;
+
```
JamesKovacs, line 84: Extraneous newline.

src/MongoDB.Driver/Encryption/ClientEncryption.cs
```diff
@@ -119,6 +120,7 @@ public CreateEncryptedCollectionResult CreateEncryptedCollection(IMongoDatabase
             Ensure.IsNotNull(collectionName, nameof(collectionName));
             Ensure.IsNotNull(createCollectionOptions, nameof(createCollectionOptions));
             Ensure.IsNotNull(kmsProvider, nameof(kmsProvider));
+            EnsureFeatureSupported(database.Client.Cluster, Feature.Csfle2QEv2, cancellationToken);
```
JamesKovacs, line 123: Won't this prevent FLEv1 from being used with newer drivers?

JamesKovacs (changes requested):
Two issues...
- Question about whether this change prevents new drivers from using FLEv1 features.
- Minor formatting issue.

src/MongoDB.Driver.Core/Core/Operations/CreateCollectionOperation.cs
DmitryLukyanov: done

src/MongoDB.Driver/Encryption/ClientEncryption.cs
DmitryLukyanov: Inside this method there is a validation on encryptedFields value which is part of CSFLE2. So, FLE1 should be not affected. NOTE: I'm not sure off the top of my head whether it's even possible to be in this mode with latest libmongocrypt bingins

tests/MongoDB.Driver.Core.TestHelpers/JsonDrivenTests/CommandStartedEventAsserter.cs
```diff
@@ -172,6 +172,42 @@ private void AssertCommandAspect(BsonDocument actualCommand, string name, BsonVa
...
+                        if (commandName == "create") // create encrypted collection
+                        {
+                            var encryptedFieldsDocument = actualCommand[name].AsBsonDocument;
+                            foreach (var expectedItem in expectedValue.AsBsonDocument)
+                            {
+                                switch (expectedItem.Name)
+                                {
+                                    case "escCollection":
+                                    case "ecocCollection" :
+                                    case "eccCollection":
```
JamesKovacs, line 185: `eccCollection` was removed. We shouldn't need this case anymore.

JamesKovacs (changes requested):
Question about the asserter and then we're good.

tests/MongoDB.Driver.Core.TestHelpers/JsonDrivenTests/CommandStartedEventAsserter.cs
DmitryLukyanov: in theory yes, but json tests assert that this value is not filled

JamesKovacs: Ah ok. LGTM then.

JamesKovacs (approved):
LGTM

PR #1070, opened 2023-04-24
CSHARP-4626: Fix aws tests on mac.

DmitryLukyanov:


evergreen/evergreen.yml
```diff
@@ -459,7 +459,11 @@ functions:
         working_dir: mongo-csharp-driver
         script: |
           cat <<'EOF' > "${PROJECT_DIRECTORY}/prepare_mongodb_aws.sh"
-            alias urlencode='python -c "import sys, urllib as ul; sys.stdout.write(ul.quote_plus(sys.argv[1]))"'
+            if [ "macos-1100" = "$OS" ]; then
+              alias urlencode='python -c "import sys, urllib.parse; sys.stdout.write(urllib.parse.quote(sys.argv[1]))"'
+            else
+              alias urlencode='python -c "import sys, urllib as ul; sys.stdout.write(ul.quote_plus(sys.argv[1]))"'
+            fi
```
JamesKovacs, line 466: Looks like this is a difference between Python2 and Python3. I suspect that macos-1100 and later use python3 as the default python whereas our other OS variants use python2 as the default. I would suggest always using python3.

https://stackoverflow.com/questions/31827012/python-importing-urllib-quote

JamesKovacs (changes requested):
Switch to use python3 on all variants.

PR #1071, opened 2023-04-26
CSHARP-4629: Use max wire version from a connection level instead of server.

DmitryLukyanov:


JamesKovacs (approved):
LGTM. Please change the ticket title and commit description to provide clearer details about the fix.

PR #1076, opened 2023-05-05
Added a test that verifies CSHARP-4637

alkampfergit:
I've added a test that verifies bug I've opened in jira [CSHARP-4637](https://jira.mongodb.org/projects/CSHARP/issues/CSHARP-4637). Actually the bug was introduced in commit 9bbca930bc062c36f806a6d17dbf3a40d97ff831 when base serialization for decimal whas changed to Decimal128 actually breaking serialization as showed in the test.

Also I think that tests for serialization / deserialization are not entirely correct, you should verify that the rehydrated object in memory is the same of the original object.


rstam:
Thank you for providing this test.

While your test mentions CSHARP-4637 I believe this was fixed by CSHARP-4648.

PR #1077, opened 2023-05-06
CSHARP-4496: ObjectSerializer should deserialize BsonDecimal128 as decimal.

rstam:


tests/MongoDB.Bson.Tests/Jira/CSharp4496Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using FluentAssertions;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Serializers;
+using Xunit;
+
+namespace MongoDB.Bson.Tests.Jira
+{
+    public class CSharp4496Tests
```
rstam, line 24: Class and file name were wrong.

tests/MongoDB.Bson.Tests/Jira/CSharp4496Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var subject = new ObjectSerializer();
+            var json = "NumberDecimal('1.5')";
+            using var reader = new JsonReader(json);
+            var context = BsonDeserializationContext.CreateRoot(reader);
+
+            var result = subject.Deserialize(context);
+
+            var value = result.Should().BeOfType<decimal>().Subject;
+            value.Should().Be(1.5M);
```
rstam, line 37: The previous test passed in a way that hid the bug.

tests/MongoDB.Bson.Tests/Jira/CSharp4496Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        [Fact]
+        public void ObjectSerializer_should_serialize_decimal_as_BsonDecimal128()
+        {
+            var subject = new ObjectSerializer();
+            var value = 1.5M;
+
+            var result = value.ToJson(serializer: subject);
+
+            result.Should().Be("NumberDecimal(\"1.5\")");
```
rstam, line 48: This test for the serialize direction was missing.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -217,7 +217,8 @@ public override object Deserialize(BsonDeserializationContext context, BsonDeser
                     return bsonDateTime.ToUniversalTime();
 
                 case BsonType.Decimal128:
-                    return bsonReader.ReadDecimal128();
+                    var decimal128 = bsonReader.ReadDecimal128();
+                    return Decimal128.ToDecimal(decimal128);
```
JamesKovacs, line 221: As discussed in Slack, due to the differences in range between `Decimal128` and `decimal`, we should return a `Decimal128` as a `Decimal128`, but the `ObjectSerializer` shouldn't serialize a `decimal` as a `Decimal128`. By default, the `ObjectSerializer` will serialize a `decimal` as `{ _t: "System.Decimal", _v: "3.1415" }`. This will allow `decimal` values to roundtrip while still allowing `Decimal128` values to be read using their full range.

Precision and Range of decimal types:
`decimal` ±1.0 x 10^-28 to ±7.9228 x 10^28 with 28-29 digits precision.
`Decimal128` ±10^-6143 to ±10^6144 with 34 digits precision.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types
https://www.mongodb.com/developer/products/mongodb/bson-data-types-decimal128/

JamesKovacs (changes requested):
We should remove the automatic `decimal`=>`Decimal128` serialization, but keep the `Decimal128` deserialization. This will allow `decimal` values to roundtrip while still being able to read `Decimal128` values from the database without truncation or loss of precision.

PR #1081, opened 2023-05-10
CSHARP-4638: AWS Lambda metadata detection logic is too permissive

sanych-sun:


rstam (approved):
LGTM

I really don't know anything about AWS Lambda but I agree that your code change implements the change the ticket requires

PR #1084, opened 2023-05-12
CSHARP-4652: TagSet.Equals should ignore tags order

BorisDog:


src/MongoDB.Driver.Core/TagSet.cs
```diff
@@ -46,7 +44,7 @@ public TagSet()
         /// <param name="tags">The tags.</param>
         public TagSet(IEnumerable<Tag> tags)
         {
-            _tags = Ensure.IsNotNull(tags, nameof(tags)).ToList();
+            _tags = Ensure.IsNotNull(tags, nameof(tags)).OrderBy(t => t.Name).ToArray();
```
JamesKovacs, line 47: Tag names aren't guaranteed to be unique. For example a node could have a `TagSet` of `{ usage: "operational", usage: "reporting" }`. This will not match `{ usage: "reporting", usage: "operational" }`, but it should.

Add `ThenBy(t => t.Value)` as well as a test for this edge case.

JamesKovacs (changes requested):
Found an edge case that needs addressing.

JamesKovacs (approved):
LGTM

PR #1089, opened 2023-05-17
CSHARP-4550: NewExpression and MemberInitExpression behaviour differs

sanych-sun:
Please note this PR is to be followed by another, that will actually remove all code related to new expression translation, by calling init expression translator.


rstam (changes requested):
Several refactoring suggestions discussed offline.

rstam (approved):
LGTM

PR #1095, opened 2023-05-25
CSHARP-4504: Do not perform server selection to determine sessions support

BorisDog:


src/MongoDB.Driver/Encryption/MongocryptdFactory.cs
```diff
@@ -74,18 +75,26 @@ public static void EnsureThatExtraOptionsAreValid(IReadOnlyDictionary<string, ob
     internal class MongocryptdFactory
     {
         private readonly bool? _bypassQueryAnalysis;
+        private readonly IEventSubscriber _eventSubscriber;
         private readonly IReadOnlyDictionary<string, object> _extraOptions;
 
-        public MongocryptdFactory(IReadOnlyDictionary<string, object> extraOptions, bool? bypassQueryAnalysis)
+        public MongocryptdFactory(IReadOnlyDictionary<string, object> extraOptions, bool? bypassQueryAnalysis, IEventSubscriber eventSubscriber = null)
```
BorisDog, line 81: This change is for testing, were access the events is needed.

src/MongoDB.Driver.Core/Core/Bindings/CoreSession.cs
```diff
@@ -521,11 +521,6 @@ private void EnsureTransactionsAreSupported()
 
             var connectedDataBearingServers = _cluster.Description.Servers.Where(s => s.State == ServerState.Connected && s.IsDataBearing).ToList();
 
-            if (connectedDataBearingServers.Count == 0)
```
BorisDog, line 524: Now that we don't necessary have selected server, switch to "best effort" validation. Compatible with [spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#starttransaction).

JamesKovacs (approved):
LGTM

PR #1102, opened 2023-05-30
CSHARP-4647: Remove "Public Technical Preview" from Queryable Encryption Equality API

BorisDog:
Update libmongocrypt to 1.8.0

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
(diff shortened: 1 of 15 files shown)


nuget.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
```
BorisDog, line 3: Will be reverted.

JamesKovacs (approved):
LGTM

PR #1113, opened 2023-06-20
Introduction of MongoUrlTypeConverter

Corniel:
There are multiple reasons why you would prefer to have types like `MongoUrl` being decorated with a custom `TypeConverter`, but for me the most pressing one is, that it allows typed configuration to use properties to be of the the type `MongoUrl`.


Corniel:
> @Corniel thank you for your PR. Please submit a Jira ticket describing the use case for this functionality, it would be helpful to have a sample of such typed configuration. [How to submit a ticket](https://github.com/mongodb/mongo-csharp-driver/blob/master/CONTRIBUTING.md)

See: [CSHARP-4689](https://jira.mongodb.org/browse/CSHARP-4689)

Example added to PR.

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.ComponentModel;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+
+namespace MongoDB.Driver.Convertion;
```
sanych-sun, line 23: We do not use file scoped namespaces. Please convert to traditional namespace declaration.

tests/MongoDB.Driver.Tests/MongoUrlTests.cs
```diff
@@ -457,6 +459,30 @@ public void TestSrvMaxHosts()
...
+        [Fact]
+        public void Decorated_with_TypeConverter()
+        {
+            var converter = TypeDescriptor.GetConverter(typeof(MongoUrl));
+
+            converter.Should().BeOfType<MongoUrlTypeConverter>();
+        }
+
+        [Fact]
+        public void Converts_from_string()
```
sanych-sun, line 471: Please use the following name instead:

Convert_from_string_should_work

tests/MongoDB.Driver.Tests/MongoUrlTests.cs
```diff
@@ -457,6 +459,30 @@ public void TestSrvMaxHosts()
...
+        {
+            var converter = TypeDescriptor.GetConverter(typeof(MongoUrl));
+
+            converter.ConvertFrom("mongodb://localhost").Should().Be(new MongoUrl("mongodb://localhost"));
+        }
+
+        [Fact]
+        public void Converts_to_string()
```
sanych-sun, line 479: Please use the following name instead:

Convert_to_string_should_work

tests/MongoDB.Driver.Tests/MongoUrlTests.cs
```diff
@@ -457,6 +459,30 @@ public void TestSrvMaxHosts()
...
+
+            converter.Should().BeOfType<MongoUrlTypeConverter>();
+        }
+
+        [Fact]
+        public void Converts_from_string()
+        {
+            var converter = TypeDescriptor.GetConverter(typeof(MongoUrl));
+
+            converter.ConvertFrom("mongodb://localhost").Should().Be(new MongoUrl("mongodb://localhost"));
```
sanych-sun, line 475: I would suggest to split this into 2 separate statements:

`var result = converter.ConvertFrom("mongodb://localhost");`
and
`result.Should().Be(new MongoUrl("mongodb://localhost"));`

To follow the arrange-act-assert paradigm.

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
sanych-sun, line 23: I suppose MongoUrlTypeConverter could be in the same namespace as MongoUrl class.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/MongoUrlTests.cs
```diff
@@ -457,6 +459,30 @@ public void TestSrvMaxHosts()
             subject.SrvMaxHosts.Should().Be(2);
         }
 
+        [Fact]
+        public void Decorated_with_TypeConverter()
```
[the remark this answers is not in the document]
Corniel: I disagree. You want to tell the story that `MongoUrl` is decorated. The other two tests are about the TypeConvert's behavior.

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
Corniel: That could be. I prefer to have converts in their own namespace.

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+
+/// <summary>Implements a <see cref="TypeConverter"/> for converting <see cref="MongoUrl"/>.</summary>
+public sealed class MongoUrlTypeConverter : TypeConverter
+{
+    /// <inheritdoc />
+    [Pure]
```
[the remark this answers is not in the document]
Corniel: See: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.contracts.pureattribute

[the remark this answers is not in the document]
sanych-sun: Yes, but is this attribute is being used by any tools?

https://github.com/dotnet/docs/issues/17640

Documentation even says " Code contracts aren't supported in .NET 5+ (including .NET Core versions)."
https://github.com/dotnet/docs/blob/main/docs/framework/debug-trace-profile/code-contracts.md

So what the reason for us to have this attribute?

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
sanych-sun: It makes sense if assembly contains or will eventually have lots of converters. Not sure if we will have many of them. On other hand having whole namespace for single class looks weird. Also having related classes sitting together sounds as a good idea for me.

src/MongoDB.Driver/Convertion/MongoUrlTypeConverter.cs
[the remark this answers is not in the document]
Corniel: That can be the case. Both .NET's analyzers and Sonar report on results not being consumed of pure methods. But you could argue that this whole library is not decorated a such, and therefor to drop it. That is fine with me. (That being said, I decorate all my libraries with `[Pure]` attributes when applicable.)

src/MongoDB.Driver/MongoUrlTypeConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.ComponentModel;
+using System.Diagnostics.Contracts;
```
sanych-sun, line 18: This using is not in use anymore. Please remove it.

sanych-sun (changes requested):
Looks good and almost ready to be merged.

sanych-sun (approved):

PR #1116, opened 2023-06-23
CSHARP-4688: Calling queryable.First when queryable is of type IMongoQueryable<T> doesn't add { $limit : 1 } stage to pipeline.

rstam:


rstam:
This is an initial POC for just Any and First (and async versions).

If you approve of the approach we can flesh it out for the rest of the affected methods.

src/MongoDB.Driver.Core/IAsyncCursorSource.cs
```diff
@@ -57,6 +57,11 @@ public static class IAsyncCursorSourceExtensions
         /// <returns>True if the cursor contains any documents.</returns>
         public static bool Any<TDocument>(this IAsyncCursorSource<TDocument> source, CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (source is IQueryable<TDocument> queryable && cancellationToken == CancellationToken.None)
+            {
+                return Queryable.Any(queryable);
```
rstam, line 62: In this case we can forward directly to `Queryable.Any` because `Queryable` is a type known to Core.

src/MongoDB.Driver.Core/IAsyncCursorSource.cs
```diff
@@ -72,6 +77,11 @@ public static bool Any<TDocument>(this IAsyncCursorSource<TDocument> source, Can
         /// <returns>A Task whose result is true if the cursor contains any documents.</returns>
         public static async Task<bool> AnyAsync<TDocument>(this IAsyncCursorSource<TDocument> source, CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (source is IMongoQueryableForwarder<TDocument> queryableForwarder)
+            {
+                return await queryableForwarder.AnyAsync(cancellationToken).ConfigureAwait(false);
```
rstam, line 82: In this case we cannot forward directly to `MongoQueryable.AnyAsync` becase this file is in Core and `MongoQueryable` is in High.

So we need an intermediate interface known to both Core and High that Core can use to forward this `AnyAsync` method to `MongoQueryable`.

Note that this intermediate interface is `internal` so we are not touching the public API.

Further note that once Core and High are merged we will no longer need this intermediate interface and will instead be able to forward directly to `MongoQueryable.AnyAsync`.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQuery.cs
```diff
@@ -62,13 +62,13 @@ public MongoQuery(MongoQueryProvider<TDocument> provider, Expression expression)
         public override IAsyncCursor<TOutput> Execute()
         {
             var executableQuery = ExpressionToExecutableQueryTranslator.Translate<TDocument, TOutput>(_provider, _expression);
-            return executableQuery.Execute(_provider.Session, CancellationToken.None);
```
rstam, line 65: These changes are to ensure that all executions go through the provider which is in charge of capturing the resulting stages.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQuery.cs
```diff
@@ -115,5 +115,8 @@ public override string ToString()
                 return ex.ToString();
             }
         }
+
+        Task<bool> IMongoQueryableForwarder<TOutput>.AnyAsync(CancellationToken cancellationToken) => MongoQueryable.AnyAsync(this, cancellationToken);
+        Task<TOutput> IMongoQueryableForwarder<TOutput>.FirstAsync(CancellationToken cancellationToken) =>  MongoQueryable.FirstAsync(this, cancellationToken);
```
rstam, line 120: These are the almost trivial statements forwarding calls to `MongoQueryable`.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -115,13 +115,33 @@ public override object Execute(Expression expression)
         public override TResult Execute<TResult>(Expression expression)
         {
             var executableQuery = ExpressionToExecutableQueryTranslator.TranslateScalar<TDocument, TResult>(this, expression);
+            return Execute(executableQuery);
+        }
+
+        public TResult Execute<TResult>(ExecutableQuery<TDocument, TResult> executableQuery)
+        {
+            return Execute(executableQuery, CancellationToken.None);
```
rstam, line 123: For now we default to `CancellationToken.None`.

I envision a future where a provider could be configured with a `CancellationToken` (similar to how `_options` are configurable). If that comes to pass the default would be `_cancellationToken` instead.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Linq3IntegrationTest.cs
```diff
@@ -57,6 +58,14 @@ protected void CreateCollection<TDocument>(IMongoCollection<TDocument> collectio
...
         }
 
+        protected (BsonDocument[] Stages, TResult Result) ExecuteQueryCapturingStages<TInput, TResult>(IMongoQueryable<TInput> queryable, Func<IMongoQueryable<TInput>, TResult> executor)
+        {
+            var provider = (MongoQueryProvider)queryable.Provider;
+            var result = executor(queryable);
+            var stages = provider.GetMostRecentPipelineStages();
+            return (stages, result);
+        }
+
```
rstam, line 68: New helper method to make it easier to execute a query and capture the stages.

Centralizing that in this method also makes it easier to adjust to future changes in the API to capture stages (a separate PR that is also currently under discussion).

src/MongoDB.Driver.Core/IAsyncCursorSource.cs
rstam: Note: to handle a cancellationToken would require a change to the public API (for example, adding a new overload of `MongoQueryable.Any` that takes a `CancellationToken`).

JamesKovacs (approved):
LGTM

sanych-sun (changes requested):
Minor changes requested

sanych-sun (approved):
LGTM

PR #1118, opened 2023-06-27
CSHARP-4691: Support GetType comparison in LINQ3.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/GetTypeComparisonExpressionToFilterTranslator.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (discriminatorValue.IsBsonArray)
+            {
+                var discriminatorValues = discriminatorValue.AsBsonArray;
+                var filters = new AstFilter[discriminatorValues.Count + 1];
+                filters[0] = AstFilter.Size(discriminatorField, discriminatorValues.Count); // don't match subclasses
+                for (var i = 0; i < discriminatorValues.Count; i++)
+                {
+                    var discriminatorItemField = discriminatorField.SubField(i.ToString(), BsonValueSerializer.Instance);
+                    filters[i + 1] = AstFilter.Eq(discriminatorItemField, discriminatorValues[i]);
+                }
```
rstam, line 60: This weird way of checking the `_t` value is to work around how MQL handles array queries.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/TypeIsExpressionToFilterTranslator.cs
```diff
@@ -29,14 +31,20 @@ public static AstFilter Translate(TranslationContext context, TypeBinaryExpressi
...
+
+                var discriminatorConvention = field.Serializer is ObjectSerializer objectSerializer ?
+                    objectSerializer.DiscriminatorConvention :
+                    BsonSerializer.LookupDiscriminatorConvention(actualType);
                 var discriminatorField = field.SubField(discriminatorConvention.ElementName, BsonValueSerializer.Instance);
-                var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                var discriminatorValue = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                if (discriminatorValue is BsonArray array)
+                {
+                    discriminatorValue = array.Last();
```
rstam, line 44: We need just the last value so that we match against subclasses also (as per `is` semantics).

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,366 @@
...
+    public class CSharp4691Tests : Linq3IntegrationTest
+    {
+        static CSharp4691Tests()
+        {
+            BsonClassMap.RegisterClassMap<Activity1>(cm =>
+            {
+                cm.AutoMap();
+                Func<Type, bool> allowedTypes = t => t == typeof(string) || t == typeof(MyActivityObject) || t == typeof(MyActivityObjectDerived);
+                var discriminatorConvention = new HierarchicalDiscriminatorConvention("_t");
+                var objectSerializer = new ObjectSerializer(discriminatorConvention, allowedTypes);
```
rstam, line 38: We need a custom object serializer for the `Object` property for two reasons:

1. We have to configure `AllowedTypes`
2. We have to configure the discriminator convention

If we don't configure the discriminator convention then the generated MQL will be wrong because it will be using the wrong discriminator values.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,366 @@
...
+                cm.MapMember(x => x.Object).SetSerializer(objectSerializer);
+            });
+        }
+
+        [Theory]
+        [ParameterAttributeData]
+        public void Find_with_object_property_GetType_comparison_to_base_class_should_work(
+            [Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
+        {
+            Implementation<MyActivityObject, int>();
```
rstam, line 48: This is a bit weird, but I wanted to use a generic method as similar as possible to the one in the JIRA ticket.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,366 @@
...
+                        x.Object.GetType() == typeof(TActivityObject) &&
+                        ((TActivityObject)x.Object).Id.Equals(activityId)
+                    );
+
+                var filter = TranslateFindFilter(collection, find);
+                var results = find.ToList();
+
+                if (linqProvider == LinqProvider.V2)
+                {
+                    // LINQ2 filter and results are both wrong
```
rstam, line 69: But we're not going to fix any of the LINQ2 bugs discovered in this test.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,366 @@
...
+                where TId : notnull, IEquatable<TId>, new()
+            {
+                var collection = GetCollection2(linqProvider);
+                var activityId = 1;
+
+                var find = collection
+                    .Find(x =>
+                        x.Object != null &&
+                        x.Object.GetType() == typeof(TActivityObject) &&
+                        ((TActivityObject)(object)x.Object).Id.Equals(activityId)
```
rstam, line 213: Double cast (with intermediate `object`) required to compile without errors.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,366 @@
...
+            public ActivityObject<int> Object { get; set; }
+        }
+
+        private abstract class ActivityObject<TId>
+        {
+            public abstract TId Id { get; set; }
+        }
+
+        [BsonDiscriminator(RootClass = true)]
+        [BsonKnownTypes(typeof(MyActivityObjectDerived))]
```
rstam, line 356: Without these attributes the polymorphic serialization wouldn't work.

rstam:
Added a second commit supporting GetType equals and is in aggregation expressions as well as filters.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/GetTypeComparisonExpressionToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,69 @@
...
+                rightExpression is ConstantExpression;
+        }
+
+        public static AggregationExpression Translate(TranslationContext context, BinaryExpression expression)
+        {
+            var leftExpression = expression.Left;
+            var rightExpression = expression.Right;
+
+            if (
+                leftExpression is MethodCallExpression methodCallExpression &&
```
sanych-sun, line 46: It looks as almost exact copy of lines 34-37. Could we separate this condition into internal  method with signature like 
`bool CanConvertInternal(BinaryExpression expression, out ..., out ...)`

And call this new method from both places. So we will be sure that conditions are processed consistently.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
sanych-sun: Does it mean you will remove/mark as Ignored all this tests before merge?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -0,0 +1,622 @@
...
+                where TActivityObject : ActivityObject<TId>
+                where TId : notnull, IEquatable<TId>, new()
+            {
+                var collection = GetCollection1(linqProvider);
+                var activityId = 1;
+
+                var find = collection
+                    .Find(x =>
+                        x.Object != null &&
+                        x.Object is TActivityObject &&
```
sanych-sun, line 136: Do we support "is" pattern matching? Like:
`x.Object is TActivityObject activity && activity.Id == activityId`

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
rstam: No need to. This setup doesn't alter any global state.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/GetTypeComparisonExpressionToAggregationExpressionTranslator.cs
rstam: See what you think of the refactoring.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
rstam: This type of pattern matching is not supported in expression trees (unfortunately).

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/GetTypeComparisonExpressionToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,78 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators
+{
+    internal static class GetTypeComparisonExpressionToAggregationExpressionTranslator
+    {
+        // public static methods
+        public static bool CanTranslate(BinaryExpression expression)
+        {
+            return CanTranslate(expression, out _, out _);
```
rstam, line 31: It kind of bothers me when I see out parameters being discarded. It makes me feel like they shouldn't have been there in the first place. But the only way to avoid it here would be to have two separate implementations of this method, one with out parameters and one without out parameters. But that defeats the whole purpose of having only a single implementation of the logic.

So throwing away the out parameters seems like the lesser evil here...

sanych-sun (approved):
LGTM

PR #1129, opened 2023-07-04
CSHARP-4666: Support simple projections to a field in Find on older servers.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -139,7 +141,39 @@ static Expression RemovePossibleConvertToObject(Expression expression)
             Expression<Func<TSource, TProjection>> expression,
             IBsonSerializer<TSource> sourceSerializer,
             IBsonSerializerRegistry serializerRegistry)
-            => TranslateExpressionToProjectionInternal(expression, sourceSerializer, new AstFindProjectionSimplifier());
+        {
+            var renderedProjection = TranslateExpressionToProjectionInternal(expression, sourceSerializer, new AstFindProjectionSimplifier());
+
+            // { _v : '$field', _id : 0 } => { field : 1, _id : 0 } or { _id : 1 } for compatibility with older servers
```
rstam, line 147: By implementing this transformation here it ONLY affects `Find`

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4538Tests.cs
```diff
@@ -35,7 +35,7 @@ public void Project_with_nullable_value_should_work()
                 .Project(x => x.D);
 
             var projection = TranslateFindProjection(collection, find);
-            projection.Should().Be("{ _v : '$D', _id : 0  }");
+            projection.Should().Be("{ D : 1, _id : 0 }");
```
rstam, line 38: Only two tests were affected since we limited the transformation to `Find` only.

JamesKovacs: Line 30 above should no longer be necessary.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -139,7 +141,39 @@ static Expression RemovePossibleConvertToObject(Expression expression)
...
+                var elements = projectionDocument.Elements.ToArray();
+                if (elements.Length == 2)
+                {
+                    if (elements[0].Name == wrappedValueSerializer.FieldName &&
+                        elements[0].Value is BsonString wrappedValueExpression &&
+                        elements[1].Name == "_id" &&
+                        elements[1].Value == 0)
+                    {
+                        var fieldPath = wrappedValueExpression.AsString;
+                        if (Regex.IsMatch(fieldPath, @"^\$[^.$]+"))
```
JamesKovacs, line 160: What is the regex protecting against? If dollars or dots are present in the field name, `$getField` will be used and the outer predicates will evaluate to false. It is probably worth a guard clause that `fieldPath.StartsWith("$")`.

JamesKovacs (changes requested):
Overall looks reasonable. Minor changes requested.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
rstam: `$x.y` is a valid field reference, but for example it references the `2` in `{ x : { y : 2 } }`

JamesKovacs: `{ "x.y": 1 }` is a valid `find` projection even for 4.2 and earlier as far as I recall. Plus the regex `^\$[^.$]+` matches `$x.y` because the `$x` part matches. So I'm still not understanding what the regex is protecting against.

JamesKovacs (changes requested):
Still not clear to me what the regex is protecting against. See comment.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
rstam: Yes, it is a valid projection as far as the server is concerned.

But the wrapped values serializer only works with fields one level deep. 

So it only works with `$x`, not with `$x.y`.

rstam: Good point about the regex needing to be fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -139,7 +140,50 @@ static Expression RemovePossibleConvertToObject(Expression expression)
             Expression<Func<TSource, TProjection>> expression,
             IBsonSerializer<TSource> sourceSerializer,
             IBsonSerializerRegistry serializerRegistry)
-            => TranslateExpressionToProjectionInternal(expression, sourceSerializer, new AstFindProjectionSimplifier());
+        {
```
rstam, line 143: Alex objected to this change offline because it does an odd transformation in an odd place in an odd domain.

It is true that it would be nice if the MQL were backward compatible in the first place, but at the point we're generating this MQL we don't know yet whether it's for a Find or not, and it would be rather invasive to try to pass that information down.

It is true that this is an odd place to be doing a transformation. We've already done all the work, and at the very last minute we pull a switcheroo. We discussed doing this in the AstFindSimplifier, but that's not quite possible because the transformation requires a serializer adjustment, and AstFindSimplifier doesn't have access to or return a serializer.

It is true that it's in an odd domain. By that I mean that we usually do our MQL generation (and transformations like simplifications) in the AST domain, and here we're working very late with MQL that has already been rendered to a BsonDocument.

I spent all day today trying to get some other approach working. I wish I hadn't.

I think in spite of the valid objections to this somewhat hacky approach we should nevertheless go with it anyway. Here are the reasons:

1. It is crystal clear that this only applies to Find
2. The changes are in this ONE place only without need to any other refactoring
3. Once server 4.4 is the minimum supported version we can remove this ONE block of code with no architectural side effects

@sanych-sun and @JamesKovacs what do you think?

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -139,7 +140,50 @@ static Expression RemovePossibleConvertToObject(Expression expression)
...
+            {
+                var projectionDocument = renderedProjection.Document;
+                var projectionElements = projectionDocument.Elements.ToArray();
+                if (projectionElements.Length == 2)
+                {
+                    if (projectionElements[0].Name == wrappedValueSerializer.FieldName &&
+                        projectionElements[0].Value is BsonString wrappedValueExpression &&
+                        projectionElements[1].Name == "_id" &&
+                        projectionElements[1].Value == 0 &&
+                        IsFieldNamePath(wrappedValueExpression.AsString, out var fieldName))
```
rstam, line 157: I got rid of the regular expression and wrote a helper method instead.

https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
sanych-sun: I still do not like the fact that we:
1) Doing this patchwork after whole logic is done.
2) We remaster BSON document itself bypassing all existing abstractions and classes designed to compose a projection.

However I do not want to be a show-stopper if James is OK with this approach.

JamesKovacs: Discussed in Slack. While the solution is not ideal, it is the best option available for the reasons stated above.

JamesKovacs (approved):
LGTM

sanych-sun (approved):

rstam:
Closing in favor of alternate approach in https://github.com/mongodb/mongo-csharp-driver/pull/1131

PR #1133, opened 2023-07-07
CSHARP-4626: Fix urlencode on MacOS variants.

JamesKovacs:
Using `python` worked until some OS variants started to map `python` to `python2` but others `python3`. The standard way to work around this is to use `python3` where you require `python3`, but some variants (cough cough Windows) don't map `python3`.

I tried the alternate approach of doing `python3` detection as many of our OCSP and other tests do. Unfortunately these aliases are evaluated before `drivers-evergreen-tools` is sourced and thus our `python3` detection scripts aren't available yet.

I then tried switching to `jq -sRjr @uri` for `urlencode`, which worked great except some MacOS machines don't have `jq` installed and the EG user doesn't have sufficient privileges to run `brew install jq`.

This lead me to take the same approach as `mongo-python-driver` to ensure that `python3` is used for these build scripts - which is to include the `python3` path as an environment variable configured for each OS variant.

Complete run of all variants can be found here:
https://spruce.mongodb.com/version/64a759231e2d1741743513aa/tasks

### DIFF
--- a/evergreen/evergreen.yml
+++ b/evergreen/evergreen.yml
@@ -459,7 +459,7 @@ functions:
         working_dir: mongo-csharp-driver
         script: |
           cat <<'EOF' > "${PROJECT_DIRECTORY}/prepare_mongodb_aws.sh"
-            alias urlencode='python -c "import sys, urllib as ul; sys.stdout.write(ul.quote_plus(sys.argv[1]))"'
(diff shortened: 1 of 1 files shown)


BorisDog (approved):
LGTM, looks like just using `python3` in the cleanest approach.

PR #1136, opened 2023-07-18
CSHARP-4360: Add log messages to SDAM spec

BorisDog:


src/MongoDB.Driver.Core/Core/Logging/StructuredLogTemplateProvidersSdam.cs
```diff
@@ -35,18 +36,18 @@ private static void AddSdamTemplates()
...
-                (e, _) => GetParams(e.ConnectionId, "Heartbeat succeeded"));
+                SdamCommonParams(Awaited, DurationMS, Reply),
+                (e, o) => GetParams(e.ConnectionId, "Server heartbeat succeeded", e.Awaited, (long)e.Duration.TotalMilliseconds, DocumentToString(e.Reply, o)));
 
             AddTemplateProvider<ServerHeartbeatFailedEvent>(
                 LogLevel.Debug,
-                SdamCommonParams(),
-                (e, _) => GetParams(e.ConnectionId, "Heartbeat failed"));
+                SdamCommonParams(Awaited, DurationMS, Failure),
+                (e, o) => GetParams(e.ConnectionId, "Server heartbeat failed", e.Awaited, (long)e.Timestamp.Millisecond, FormatCommandException(e.Exception, o)));
```
JamesKovacs, line 50: `e.Timestamp` returns a `DateTime`. `DateTime.Millisecond` returns the millisecond portion of the `DateTime`, which is a number between 0 and 999. I don't think this is what you're looking for.

JamesKovacs (changes requested):
Looks good. Minor comment on logging the wrong component of a `DateTime`. See inline for details.

src/MongoDB.Driver.Core/Core/Logging/StructuredLogTemplateProvidersSdam.cs
BorisDog: Good catch, thanks!
I had marked this as TODO, and forgot. Added `Duration` field.

JamesKovacs (approved):
LGTM

PR #1138, opened 2023-07-19
Ensure find projection is supported on servers prior to 4.4.

rstam:


rstam:
@JamesKovacs see what you think of this idea. If you think it has merit we can create a JIRA ticket for it.

This idea grew out of a concern about how some find projections are failing silently on servers prior to 4.4. This PR turns the silent failures into exceptions.

I think the idea is fairly safe. On servers 4.4 and newer this change is a no-op. While there is a risk that we will throw an exception on some projection that is actually supported prior to 4.4, if/when we find out that we missed a valid case we can just tweak the code accordingly.

src/MongoDB.Driver.Core/Core/Operations/FindOperation.cs
```diff
@@ -640,5 +645,56 @@ private ReadCommandOperation<BsonDocument> CreateOperation(RetryableReadContext
...
+                   (bsonInt32Value.Value == 0 || bsonInt32Value.Value == 1))
+                {
+                    return true;
+                }
+
+                if (value is BsonInt64 bsonInt64Value &&
+                   (bsonInt64Value.Value == 0L || bsonInt64Value.Value == 1L))
+                {
+                    return true;
+                }
```
JamesKovacs, line 682: Projection definitions are truthy. So `{ X : 1.0 }` is a valid projection as is `{ X : { $numberDecimal: "3.1415" } }`. I think the first check should be:

```
if (value is BsonBoolean || value.IsNumeric())
{
    return true;
}
```

While the driver code may not produce these unusual numeric projections, we have an implicit case from `BsonDocument`/`string` to `ProjectionDefinition<T>` and thus we need to support them.

JamesKovacs (changes requested):
This is a reasonable guard clause to implement. We do need to make it more permissive due to projection definitions being treated as truthy. See below.

rstam:
I was aware of the "truthy" variants but didn't think we had to support them. If this concerns you we can try to be more detailed in our checks and look for things that "look" like aggregation expressions.

My concern is that's it's far worse to accidentally fail silently because the logic failed to identify an invalid projection than it is to not support the occasional non-canonical projection.

Remember also that it's all moot with 4.4+ because we don't even perform these checks there.

rstam:
This PR has been superseded by https://github.com/mongodb/mongo-csharp-driver/pull/1147.

PR #1141, opened 2023-07-25
Enable PerformanceCounters on net6.0-windows

jairbubbles:
While migrating an application to .NET6+ I noticed PerformanceCounter related code was not available. This is an attempt to put it back.

As a side effect, targetting `net6.0-windows` is enabling some code analyzers issues. I'm leaving the code at is for now. Let me know if you have interest to go forward with the changes.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Events/Diagnostics/PerformanceCounterEventSubscriber.cs
+++ b/src/MongoDB.Driver.Core/Core/Events/Diagnostics/PerformanceCounterEventSubscriber.cs
@@ -13,7 +13,7 @@
 * limitations under the License.
 */
 
-#if NET472
+#if NET472 || WINDOWS
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


JamesKovacs:
Thank you for contributing this PR. Unfortunately it is not possible to use compile-time directives to differentiate between Windows and non-Windows operating systems at runtime. When you compile the assemblies and build the NuGet package on Windows, it would include code that only works on `net6.0-windows`. When you ran the driver on a non-Windows OS, it would call APIs that do not exist and would fail.

In order to do runtime detection, you would have to use code similar to the following:

```
if (OperatingSystemHelper.CurrentOperatingSystem == OperatingSystemPlatform.Windows)
```

where `OperatingSystemHelper` is an internal helper class in the driver.

Additionally we currently ship `net472` (full .NET Framework on Windows), `netstandard2.0` (AKA .NET Core 2.X), and `netstandard2.1` (AKA .NET Core 3.X) targets in our NuGet package. `netstandard2.1` is compatible with .NET 5+ as well, but it is not possible to access .NET 6.0 specific features like those in the `net6.0-windows` TFM. We plan to introduce newer TFMs in an upcoming release to leverage new features, but this hasn't been done yet.

I have closed this PR, but I encourage you to file a [CSHARP ticket](https://jira.mongodb.org/browse/CSHARP) to request this feature. We would be happy to provide you guidance in that ticket as to when the `net6.0` TFM will be added to the driver so that this feature could be implemented in a new PR.

PR #1142, opened 2023-07-26
CSHARP-4731: Support assigning List<T> value to IList<T> property in Select.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -57,7 +58,8 @@ public static AggregationExpression Translate(TranslationContext context, Member
                     var constructorArgumentSerializer = constructorArgumentTranslation.Serializer ?? BsonSerializer.LookupSerializer(constructorArgumentType);
                     var memberMap = EnsureMemberMap(expression, classMap, creatorMapParameter);
                     EnsureDefaultValue(memberMap);
-                    memberMap.SetSerializer(constructorArgumentSerializer);
+                    var memberSerializer = CoerceSourceSerializerToMemberSerializer(memberMap, constructorArgumentSerializer);
+                    memberMap.SetSerializer(memberSerializer);
```
rstam, line 62: An alternate approach might be to translate this as if there was a `Convert` from `List<T>` to `IList<T>` in the `Expression`.

I'm a little surprised the compiler didn't put one in.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
shiya-kohn: The compiler doesn't put in `Convert` for things that are already the target type, convert is for casts and implicit conversions (such as structs to nullable)

PR #1144, opened 2023-07-31
CSHARP-4723: Linq V2 Projection when upgrading from 2.19.1 to 2.19.2

sanych-sun:


src/MongoDB.Driver/IFindFluentExtensions.cs
```diff
@@ -57,7 +57,7 @@ public static class IFindFluentExtensions
             Ensure.IsNotNull(find, nameof(find));
             Ensure.IsNotNull(projection, nameof(projection));
 
-            return find.Project<TNewProjection>(new ExpressionProjectionDefinition<TDocument, TNewProjection>(projection, null));
+            return find.Project<TNewProjection>(new FindExpressionProjectionDefinition<TDocument, TNewProjection>(projection));
```
rstam, line 60: This is the change I was expecting.

In *this* case we *know* the projection will be used with a `Find`.

src/MongoDB.Driver/ProjectionDefinitionBuilder.cs
```diff
@@ -408,7 +408,7 @@ public ProjectionDefinition<TSource> Exclude(Expression<Func<TSource, object>> f
         /// </returns>
         public ProjectionDefinition<TSource, TProjection> Expression<TProjection>(Expression<Func<TSource, TProjection>> expression)
         {
-            return new ExpressionProjectionDefinition<TSource, TProjection>(expression, null);
+            return new FindExpressionProjectionDefinition<TSource, TProjection>(expression);
```
rstam, line 411: This doesn't seem right.

In this case we *don't* know whether the projection will be used with `Find` until it is actually used.

rstam (changes requested):

rstam (approved):
LGTM

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -141,7 +141,17 @@ static Expression RemovePossibleConvertToObject(Expression expression)
             Expression<Func<TSource, TProjection>> expression,
             IBsonSerializer<TSource> sourceSerializer,
             IBsonSerializerRegistry serializerRegistry)
-            => TranslateExpressionToProjection(expression, sourceSerializer, ProjectionHelper.CreateFindProjection, new AstFindProjectionSimplifier());
+        {
+            if (typeof(TSource) == typeof(TProjection)
```
rstam, line 145: Comparing types is redundant because they will always be equal for `x => x`.

This PR seems to overlap with the discussions we are having on Slack.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
```diff
@@ -141,7 +141,17 @@ static Expression RemovePossibleConvertToObject(Expression expression)
...
+                && expression.Parameters.Count == 1
+                && expression.Body == expression.Parameters.First())
```
rstam, line 147: `[0]` is often more efficient than `First()`.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
rstam: Note: this check should probably be in `TranslateExpressionToProjection` instead of here, as per discussions in Slack.

rstam (changes requested):

rstam:
Let's focus on the PR for CSHARP-4681 first. This PR might no longer be necessary once we merge that PR.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
sanych-sun: This code was done in scope of another PR.

src/MongoDB.Driver/Linq/Linq3Implementation/LinqProviderAdapterV3.cs
sanych-sun: This code was done in scope of another PR.

rstam (approved):
LGTM

PR #1146, opened 2023-07-31
CSHARP-4681: InvalidCastException when rendering projections

sanych-sun:


rstam (changes requested):
We are discussing some questions off line.

rstam (approved):
LGTM

PR #1156, opened 2023-08-09
CSHARP-4749: Investigation.

rstam:
I don't think we want to push anything to master here. I just want you to look at my repro and also read my analysis in the CSHARP-4749 JIRA ticket to see if you agree.


rstam:
Investigation concluded for now.

PR #1161, opened 2023-08-16
Proof of concept for supporting compound primary keys in LINQ with the EF Core Provider.

rstam:


src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -24,10 +26,24 @@ namespace MongoDB.Bson.Serialization
...
     {
+        #region static
+        /// <summary>
+        /// Creates anew instance of the BsonSerializationinfo class with an element path instead of an element name.
+        /// </summary>
+        /// <param name="elementPath">The element path.</param>
+        /// <param name="serializer">The serializer.</param>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <returns>A BsonSerializationInfo.</returns>
+        public static BsonSerializationInfo CreateWithPath(IEnumerable<string> elementPath, IBsonSerializer serializer, Type nominalType)
```
rstam, line 37: Adding a new public constructor was a breaking change so I added a new static factory method instead.

Note: a new public constructor resulted in compile time ambiguity when called with `null` as the first argument.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -37,19 +53,32 @@ public class BsonSerializationInfo
...
 
         // public properties
         /// <summary>
-        /// Gets or sets the dotted element name.
+        /// Gets the element name.
         /// </summary>
         public string ElementName
         {
-            get { return _elementName; }
+            get { return _elementPath?.Single(); }
```
rstam, line 73: One advantage of using `Single` here is that whenever LINQ2 calls this property it will throw an exception if the member is grouped under a nested path. That's far better than silently doing the wrong thing.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -92,20 +121,21 @@ public object DeserializeValue(BsonValue value)
         /// </summary>
         /// <param name="newSerializationInfo">The new info.</param>
         /// <returns>A new BsonSerializationInfo.</returns>
+        [Obsolete("This method is no longer relevant because field names are now allowed to contain dots.")]
```
rstam, line 124: We never call this method.

Should we delete it? Technically that's a breaking change, but the method no longer works reliably anyway since the server now allows `.` in element names.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/MemberSerializationInfo.cs
```diff
@@ -21,18 +22,18 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Misc
     internal class MemberSerializationInfo
     {
         // private fields
-        private readonly string _elementName;
+        private readonly IReadOnlyList<string> _elementPath;
         private readonly IBsonSerializer _serializer;
 
         // constructors
-        public MemberSerializationInfo(string elementName, IBsonSerializer serializer)
+        public MemberSerializationInfo(IReadOnlyList<string> elementPath, IBsonSerializer serializer)
```
rstam, line 29: This is an internal class so this is not a publicly breaking change.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberExpressionToAggregationExpressionTranslator.cs
```diff
@@ -77,8 +77,12 @@ public static AggregationExpression Translate(TranslationContext context, Member
...
 
             var serializationInfo = DocumentSerializerHelper.GetMemberSerializationInfo(containerTranslation.Serializer, member.Name);
-            var ast = AstExpression.GetField(containerTranslation.Ast, serializationInfo.ElementName);
-            return new AggregationExpression(expression, ast, serializationInfo.Serializer);
+            var subField = containerTranslation.Ast;
+            foreach (var subFieldName in serializationInfo.ElementPath)
+            {
+                subField = AstExpression.GetField(subField, subFieldName);
+            }
+            return new AggregationExpression(expression, subField, serializationInfo.Serializer);
```
rstam, line 85: This is a relatively simple change. We just need to call `GetField` more than once when there is a path.

If the path list has count == 1 this code is equivalent to the previous code.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/MemberExpressionToFilterFieldTranslator.cs
```diff
@@ -89,9 +89,13 @@ public static AstFilterField Translate(TranslationContext context, MemberExpress
...
             {
-                var subFieldName = memberSerializationInfo.ElementName;
                 var subFieldSerializer = memberSerializationInfo.Serializer;
-                return field.SubField(subFieldName, subFieldSerializer);
+                var subField = field;
+                foreach (var subFieldName in memberSerializationInfo.ElementPath)
+                {
+                    subField = subField.SubField(subFieldName, subFieldSerializer);
+                }
+                return subField;
```
rstam, line 98: Same as above but this time for filters.

I'm surprised I only found TWO places that needed to be changed. Hope I didn't miss any.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
```diff
@@ -0,0 +1,241 @@
...
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharpxxxTests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Deserialize_should_ungroup_members()
+        {
+            var serializedOrderItem = "{ _id : { OrderId : 1, ProductId : 2 }, Quantity : 3 }";
```
rstam, line 34: Note that `OrderId` and `ProductId` are grouped under the `_id` element.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
```diff
@@ -0,0 +1,241 @@
...
+
+            if (linqProvider == LinqProvider.V2)
+            {
+                var exception = Record.Exception(() => Translate(collection, queryable));
+                exception.Should().BeOfType<InvalidOperationException>(); // LINQ2 doesn't understand grouped members
+            }
+            else
+            {
+                var stages = Translate(collection, queryable);
+                AssertStages(stages, "{ $project : { _v : '$_id.OrderId', _id : 0 } }");
```
rstam, line 71: Note that the use of `$_id.OrderId` instead of `$OrderId`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        [BsonSerializer(typeof(OrderItemsSerializer))]
+        private class OrderItem
+        {
+            public int OrderId { get; set; }
+            public int ProductId { get; set; }
+            public int Quantity { get; set; }
+        }
+
+        private class OrderItemsSerializer : ClassSerializerBase<OrderItem>, IBsonDocumentSerializer
```
rstam, line 162: This is for testing.

We would implement similar code in the `EntitySerializer`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            // public methods
+            public bool IsGroupedMember(string name, out string groupElementName)
```
rstam, line 165: Oops this method is obsolete. Ignore it. It was when I was experimenting with a new interface instead.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -24,10 +26,24 @@ namespace MongoDB.Bson.Serialization
...
+        {
+            return new BsonSerializationInfo(elementPath.ToList(), serializer, nominalType);
+        }
+        #endregion
+
         // private fields
-        private string _elementName;
-        private IBsonSerializer _serializer;
-        private Type _nominalType;
+        private readonly IReadOnlyList<string> _elementPath;
```
sanych-sun, line 44: I would suggest to have both: `_elementPath` and `_elementName`. We can use _elementName for situation when there is a single element path. And switch to the _elementPath in that rare case when there are multiple. It could save some allocations.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
sanych-sun: Agree with the fact that we want to throw here. But I think we have to throw something more readable rather that just "Sequence contains more then one element".

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/MemberExpressionToFilterFieldTranslator.cs
sanych-sun: What about:
SelectMethodToFilterFieldTranslator line 47

sanych-sun (changes requested):

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -37,19 +53,32 @@ public class BsonSerializationInfo
...
         public string ElementName
         {
-            get { return _elementName; }
+            get { return _elementPath?.Single(); }
+        }
+
+        /// <summary>
+        /// Gets element path.
+        /// </summary>
+        public IReadOnlyList<string> ElementPath
```
JamesKovacs, line 79: Given this is a collection of paths, shouldn't this be `ElementPaths`?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            var collection = GetCollection<OrderItem>("test", linqProvider);
+            CreateCollection(
+                collection,
+                new OrderItem { OrderId = 1, ProductId = 2, Quantity = 3 },
+                new OrderItem { OrderId = 4, ProductId = 5, Quantity = 6 });
+            return collection;
+        }
+
+        [BsonSerializer(typeof(OrderItemsSerializer))]
```
JamesKovacs, line 154: So it is the responsibility of `OrderItemsSerializer` to configure the composite `_id`. If we agree on this path forward, it would be useful to allow specification of multiple `[BsonId(Order=N)]` attributes as well as `BsonClassMap<T>` configuration.

JamesKovacs (changes requested):
Straightforward and reasonable approach. Minor comments inline.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
rstam: I guess a related question is: do we need a variable length path at all?

Is it sufficient to only support a single level of nesting?

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
JamesKovacs: While I agree that the inclusion of `.` in element names breaks this method, that is not a mainline use case. This method works for the common case of element names without `.`. We can make the breaking change in 3.0.0, but not in a 2.X release IMHO.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
JamesKovacs: OK.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharpxxxxTests.cs
JamesKovacs: I agree that the most common use case is for EF Core, but it would be surprising for EF Core to write data that the driver cannot interoperate with.

JamesKovacs (commented):
See comments inline.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
rstam: I'm not proposing we make any changes to this method.

I'm merely pointing out that it's already broken, and that we don't use it.

Although I do wish we could just delete it.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
rstam: I've pushed a commit that has mutually exclusive `_elementName` and `_elementPath` to reduce allocations.

Note the changes in the users of this class that now have to check whether `ElementPath` is `null` or not and only use `ElementName` when there is no `ElementPath`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ToFilterFieldTranslators/MemberExpressionToFilterFieldTranslator.cs
rstam: I think we can ignore this one for now.

If it ever gets used when a path is involved the call to `ElementName` will throw an exception and we can deal with it then.

rstam:
Moving to a new PR now that we have a JIRA ticket number:

https://github.com/mongodb/mongo-csharp-driver/pull/1161

PR #1163, opened 2023-08-17
CSHARP-4757: Modify BsonSerializationInfo to handle properties that are grouped in a nested document.

rstam:


sanych-sun (approved):
LGTM + there might be more code we need to improve with the ElementPath property knowledge, but I suppose it could be done on-demand.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -43,13 +60,35 @@ public BsonSerializationInfo(string elementName, IBsonSerializer serializer, Typ
...
         public string ElementName
         {
-            get { return _elementName; }
+            get
+            {
+                if (_elementPath != null)
+                {
+                    throw new InvalidOperationException("When ElementPath is not null you must use it instead.");
+                }
+                return _elementName;
```
JamesKovacs, line 82: `return _elementPath[0];`

This eliminates the need for the `_elementName` field.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/MemberSerializationInfo.cs
```diff
@@ -22,6 +24,7 @@ internal class MemberSerializationInfo
     {
         // private fields
         private readonly string _elementName;
+        private readonly IReadOnlyList<string> _elementPath;
```
JamesKovacs, line 27: Here we could always use `_elementPath` and eliminate the branching logic.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberExpressionToAggregationExpressionTranslator.cs
```diff
@@ -77,7 +77,19 @@ public static AggregationExpression Translate(TranslationContext context, Member
...
+                ast = AstExpression.GetField(containerTranslation.Ast, serializationInfo.ElementName);
+            }
+            else
+            {
+                ast = containerTranslation.Ast;
+                foreach (var subFieldName in serializationInfo.ElementPath)
+                {
+                    ast = AstExpression.GetField(ast, subFieldName);
+                }
+            }
```
JamesKovacs, line 92: This code block could be simplified to iterate over `ElementPath`, which for the common case would only contain a single string.

JamesKovacs (changes requested):
Let me know what you think about only storing `_elementPath` and eliminating the branch logic based on whether `ElementPath` is null.

src/MongoDB.Bson/Serialization/BsonSerializationInfo.cs
```diff
@@ -92,20 +131,21 @@ public object DeserializeValue(BsonValue value)
         /// </summary>
         /// <param name="newSerializationInfo">The new info.</param>
         /// <returns>A new BsonSerializationInfo.</returns>
+        [Obsolete("This method is no longer relevant because field names are now allowed to contain dots.")]
```
rstam, line 135: I would like to argue in favor of removing this method even though it is public.

The method makes little sense and probably doesn't even work as intended. If we don't remove it then it seems like we would have to commit to fixing it.

sanych-sun (approved):
LGTM

JamesKovacs (approved):
LGTM. Discussed via Slack and Zoom. Maintaining both `_elementPath` and `_elementName` avoids an extra allocation of an array as the mainline use case only has a single `_elementName`.

PR #1165, opened 2023-08-18
CSHARP-4763: Canonical repro of client side projection workarounds.

rstam:


rstam:
I think we should push this branch to master even though it's only a repro so that we can link to the code from the ticket.

JamesKovacs (commented):
I agree that this is the crux of the problem. The one disadvantage of `AsEnumerable()` is that it retrieves the entire document, not just the needed fields. Although you could project into an anonymous type first followed by `AsEnumerable()` and the client-side `Select`, this is asking users to do a lot of extra work.

I see two potential paths forward:
1. Implement client-side projections similar to LINQ2.
2. Implement `ClientSelect` (needs a better name) that automatically projects the required fields and performs the remainder of the work client-side.

So the question is whether we want an auto-magic API like LINQ2 or whether we want an explicit API (like `ClientSelect`)?

shiya-kohn:
Given how many client-side projections we have in our codebase (v2), and our desire to upgrade to v3,  I really hope the team chooses to restore support for client-side projections. They should be optimized to project only the known accessed keys, obviously `x => LocalMethod(x)` won't be optimizable, and I don't expect support for aggregation expressions, but `Find` or even simple `Where().Select().ToList()` where no complex aggregations are needed should work.

rstam:
Closing for now until we have a PR for an actual implementation.

PR #1167, opened 2023-08-24
CSHARP-4756: MongoDB.Bson.BsonSerializationException (when serializing List of MongoDBRef)

sanych-sun:


rstam (approved):
Had a hard time at first wrapping my head around the changes but eventually I think I convinced myself that the changes are good as is.

LGTM

PR #1174, opened 2023-09-12
CSHARP-4018: Require hello_command + OP_MSG when loadBalanced=true

adelinowona:
Ensures that if load balancing is enabled, both 'hello' and OP_MSG are used for the connection handshake.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Authentication/DefaultAuthenticator.cs
+++ b/src/MongoDB.Driver.Core/Core/Authentication/DefaultAuthenticator.cs
@@ -86,7 +86,7 @@ namespace MongoDB.Driver.Core.Authentication
             if (!description.HelloResult.HasSaslSupportedMechs
                 && Feature.ScramSha256Authentication.IsSupported(description.MaxWireVersion))
             {
-                var command = CustomizeInitialHelloCommand(HelloHelper.CreateCommand(_serverApi));
+                var command = CustomizeInitialHelloCommand(HelloHelper.CreateCommand(_serverApi, loadBalanced: connection.Settings.LoadBalanced));
                 var helloProtocol = HelloHelper.CreateProtocol(command, _serverApi);
                 var helloResult = HelloHelper.GetResult(connection, helloProtocol, cancellationToken);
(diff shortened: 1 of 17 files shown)


src/MongoDB.Driver.Core/Core/Authentication/DefaultAuthenticator.cs
```diff
@@ -86,7 +86,7 @@ public void Authenticate(IConnection connection, ConnectionDescription descripti
             if (!description.HelloResult.HasSaslSupportedMechs
                 && Feature.ScramSha256Authentication.IsSupported(description.MaxWireVersion))
             {
-                var command = CustomizeInitialHelloCommand(HelloHelper.CreateCommand(_serverApi));
+                var command = CustomizeInitialHelloCommand(HelloHelper.CreateCommand(_serverApi, loadBalanced: connection.Settings?.LoadBalanced ?? false));
```
JamesKovacs, line 89: `BinaryConnection` and related `IConnection` classes assert that `Settings` cannot be null. So no need to use the null conditional operator. Same for the usages in server monitor, RTT monitor, and connection factory.

src/MongoDB.Driver.Core/Core/Connections/HelloHelper.cs
```diff
@@ -50,7 +50,7 @@ internal static BsonDocument CreateCommand(ServerApi serverApi, bool helloOk = f
                 (topologyVersion != null && maxAwaitTime.HasValue),
                 $"Both {nameof(topologyVersion)} and {nameof(maxAwaitTime)} must be filled or null.");
 
-            var helloCommandName = helloOk || serverApi != null ? "hello" : OppressiveLanguageConstants.LegacyHelloCommandName;
+            var helloCommandName = loadBalanced || helloOk || serverApi != null ? "hello" : OppressiveLanguageConstants.LegacyHelloCommandName;
```
JamesKovacs, line 53: I would suggest `helloOk || loadBalanced || serverApi != null` as the ordering.

JamesKovacs (changes requested):
Minor changes requested.

JamesKovacs (approved):
LGTM

PR #1175, opened 2023-09-12
CSHARP-4786: Added shim to uuidhelpers.js so that it works in both the old mongo shell as well as the new mongosh shell.

JamesKovacs:
Based on [question in the community forums](https://www.mongodb.com/community/forums/t/c-decode-and-encode-guids-in-mongosh/244163). It was easy enough to shim the existing script so that it works in both old and new shells.


adelinowona (approved):

adelinowona:
LGTM

PR #1176, opened 2023-09-14
CSHARP-4566: Support widening numeric and nullable-numeric conversions.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs
```diff
@@ -48,12 +48,6 @@ public static AggregationExpression Translate(TranslationContext context, Binary
                 throw new ExpressionNotSupportedException(expression, because: "operand types are not compatible with each other");
             }
 
-            if (IsArithmeticExpression(expression))
-            {
-                leftExpression = ConvertHelper.RemoveWideningConvert(leftExpression);
-                rightExpression = ConvertHelper.RemoveWideningConvert(rightExpression);
-            }
-
```
rstam, line 56: We now have code to translate these `Convert`s and no longer want to remove them.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4596Tests.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Linq.Expressions;
+using FluentAssertions;
+using MongoDB.Driver.Linq;
+using MongoDB.Driver.Tests.Linq.Linq3Implementation;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3ImplementationTests.Jira
+{
+    public class CSharp4596Tests : Linq3IntegrationTest
```
rstam, line 26: Note that this PR fixes CSharp4596 also. That ticket was also related to numeric conversions, which this PR addresses.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4596Tests.cs
```diff
@@ -0,0 +1,151 @@
...
+    {
+        [Fact]
+        public void Test_widening_expressions()
```
rstam, line 29: This tests can be removed. It's just experimenting to make sure all these conversions are compiled to a single `ConvertExpression`.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationWithLinq2Tests/MongoQueryableIntComparedToDoubleWithStringRepresentationTests.cs
```diff
@@ -43,18 +44,28 @@ public class C
         }
 
         [Theory]
-        [InlineData(1.5, "{ \"I\" : \"1.5\" }")]
```
rstam, line 46: I would argue that this translation was questionable. What we actually want to do is pass the value to the underlying serializer, and in this case `1.5` can't be converted to an `Int32` (without loss) to pass to the underlying `Int32Serializer`.

JamesKovacs: Agreed.

JamesKovacs (commented):
Overall looks great. (So many test cases!) :) Just a question about widening converts and then I'll LGTM this.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/ConvertHelper.cs
```diff
@@ -71,11 +77,14 @@ internal static class ConvertHelper
             (typeof(ulong), typeof(float)),
             (typeof(ulong), typeof(double)),
             (typeof(ulong), typeof(decimal)),
-            (typeof(float), typeof(double)),
-            (typeof(float), typeof(decimal)),
-            (typeof(double), typeof(decimal))
```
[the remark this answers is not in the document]
rstam: Because I wasn't sure they actually are widening converts.

Maybe float -> double should stay.

Possibly the other two as well. Let's think about it...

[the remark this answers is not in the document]
rstam: The C# language does not have implicit conversions from `float` or `double` to `decimal`.

I think the reason it does not is that these conversions are not guaranteed to be lossless. Widening conversions are supposed to be lossless.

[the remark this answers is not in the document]
JamesKovacs: `float` to `double` is definitely a lossless widening convert. We should add that back in. Agreed that `float` or `double` to `decimal` in not guaranteed to be lossless and you need a cast to assign a `decimal` to a `float`/`double`. Users probably already have those casts in their LINQ expressions to make the C# compiler happy. I'm ok with removing those. But let's keep `float` to `double`.

JamesKovacs (changes requested):
See my comment about adding the `float` to `double` widening convert back in.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/ConvertHelper.cs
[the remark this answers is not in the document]
JamesKovacs: Oh, my bad. You didn't remove `float` to `double`. The line just got modified. LGTM.

JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationWithLinq2Tests/MongoQueryableIntComparedToDoubleWithStringRepresentationTests.cs
```diff
@@ -43,18 +44,28 @@ public class C
...
             var subject = __collection.AsQueryable();
 
             var queryable = subject.Where(x => x.I == value);
 
-            AssertFilter(queryable, expectedFilter);
+            if (expectedFilter == "throws")
+            {
+                var exception = Record.Exception(() => AssertFilter(queryable, expectedFilter));
+                exception.Should().BeOfType<ExpressionNotSupportedException>();
+                exception.Message.Should().Contain("The constant 1.5 could not be converted to Int32 in order to be serialized using the proper serializer");
```
sanych-sun, line 59: Minor: `1.5` should not be hardcoded here, but built based on parameter value.

sanych-sun (approved):
LGTM + minor comment

rstam:
Note: even thought the status of this PR is "Merged" it was later reverted when some remaining issues with it were discovered.

PR #1177, opened 2023-09-15
CSHARP-4722: Add durations to connection pool events

adelinowona:


BorisDog (changes requested):
Few minor comments

src/MongoDB.Driver.Core/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -180,11 +180,12 @@ public AcquireConnectionHelper(ExclusiveConnectionPool pool)
 
             public IConnectionHandle AcquireConnection(CancellationToken cancellationToken)
             {
+                Stopwatch stopwatch = new();
```
sanych-sun, line 183: Minor: we could save a little on allocations by using the following technic:
capture `Stopwatch.GetTimestamp()` before the measured code
 and compare it to the timestamp after and convert diffrence from ticks to ms.

adelinowona: In this particular case, I think it may be overkill since the previous version of the code was also allocating a new stopwatch under the cover so I am not really adding anything.  However, I like the savings on allocations (even though they may be very small) from this technique but I would advocate that if we do that here, then we might as well do it in other places where we use a stopwatch to save on allocations there too and maintain uniformity in the code. This may be good for quick wins ticket for the future though. Let me know what you think.

sanych-sun (approved):
LGTM

BorisDog (approved):
LGTM

PR #1179, opened 2023-09-21
CSHARP-4789: MongoUrlBuilder builds urls with invalid query string fo…

adelinowona:
…rmat

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
@@ -123,7 +123,7 @@ namespace MongoDB.Driver.Core.Configuration
 
         internal ConnectionString(string connectionString, bool isInternalRepresentation, IDnsResolver dnsResolver)
         {
-            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString));
+            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString)).Replace(';', '&');
             _isInternalRepresentation = isInternalRepresentation;
 
             _allOptions = new NameValueCollection(StringComparer.OrdinalIgnoreCase);
--- a/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
@@ -123,7 +123,7 @@ namespace MongoDB.Driver.Core.Configuration
 
         internal ConnectionString(string connectionString, bool isInternalRepresentation, IDnsResolver dnsResolver)
         {
-            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString)).Replace(';', '&');
+            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString));
             _isInternalRepresentation = isInternalRepresentation;
 
             _allOptions = new NameValueCollection(StringComparer.OrdinalIgnoreCase);
--- a/tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
+++ b/tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
@@ -118,7 +118,7 @@ namespace MongoDB.Driver.Tests
             }
 #pragma warning restore 618
 
-            var connectionString = "mongodb://username:password@host/database?" + string.Join(";", new[] {
+            var connectionString = "mongodb://username:password@host/database?" + string.Join("&", new[] {
                 "authMechanism=GSSAPI",
                 "authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true",
                 "authSource=db",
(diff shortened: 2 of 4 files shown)


src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs
```diff
@@ -123,7 +123,7 @@ public ConnectionString(string connectionString) : this(connectionString, false,
 
         internal ConnectionString(string connectionString, bool isInternalRepresentation, IDnsResolver dnsResolver)
         {
-            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString));
+            _originalConnectionString = Ensure.IsNotNull(connectionString, nameof(connectionString)).Replace(';', '&');
```
rstam, line 126: I don't know about this... if the user chose to use `;` instead of `&` who are we to replace it?

But maybe it makes sense to normalize their non-standard choice.

In that case I worry about the `Replace` being too broad. Are we sure there aren't sometimes instances of `;` that should *not* be replaced?

For example, what if the password contains `;`?

adelinowona: Agreed, we shouldn't normalize the user's non-standard choice.

tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
```diff
@@ -222,17 +222,17 @@ public void TestAll()
...
                 {
                     var defaultGuidRepresentation = BsonDefaults.GuidRepresentation;
                     if (builder.GuidRepresentation == defaultGuidRepresentation)
                     {
-                        expectedConnectionString = expectedConnectionString.Replace("uuidRepresentation=pythonLegacy;", "");
+                        connectionString = connectionString.Replace("uuidRepresentation=pythonLegacy;", "");
                     }
                 }
 #pragma warning restore 618
+                var expectedConnectionString = connectionString.Replace(';', '&');
```
JamesKovacs, line 235: Rather than bulk replacing `;` with `&`, we should update the tests to use `&` as that is now our default representation.

tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
```diff
@@ -316,15 +316,16 @@ public void TestComputedWaitQueueSize_UsingMultiple()
         {
 #pragma warning disable 618
             var built = new MongoUrlBuilder { Server = _localhost, MaxConnectionPoolSize = 123, WaitQueueMultiple = 2.0 };
-            var connectionString = "mongodb://localhost/?maxPoolSize=123;waitQueueMultiple=2";
+            var connectionString = "mongodb://localhost/?maxPoolSize=123&waitQueueMultiple=2";
+            var expectedConnectionString = connectionString.Replace(';', '&');
```
JamesKovacs, line 320: Since `connectionString` contains no `;`, `Replace` is a NOP and not necessary.

tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
```diff
@@ -376,7 +377,7 @@ public void TestCompressors(CompressorType[] compressors, string compressionProp
             if (!string.IsNullOrWhiteSpace(compressionProperty))
             {
                 var @params = compressionProperty.Split(';');
```
JamesKovacs, line 379: Rather than splitting on `;`, we should use `&` in our tests and split on those when needed.

tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
```diff
@@ -969,10 +970,11 @@ public void TestMaxStaleness(int? ms, string formatString, string[] values)
             if (readPreference != null) { built.ReadPreference = readPreference; }
 
             var canonicalConnectionString = string.Format(formatString, values[0]);
+            var expectedConnectionString = canonicalConnectionString.Replace(';', '&');
```
JamesKovacs, line 973: Use `&` in tests.

tests/MongoDB.Driver.Tests/MongoUrlBuilderTests.cs
```diff
@@ -1070,6 +1072,7 @@ public void TestReadPreference_SecondaryWithOneTagSet()
             var readPreference = new ReadPreference(ReadPreferenceMode.Secondary, tagSets);
             var built = new MongoUrlBuilder { Server = _localhost, ReadPreference = readPreference };
             var connectionString = "mongodb://localhost/?readPreference=secondary;readPreferenceTags=dc:ny,rack:1";
+            var expectedConnectionString = connectionString.Replace(';', '&');
```
JamesKovacs, line 1075: Same here. Use `&` in tests.

tests/MongoDB.Driver.Tests/MongoUrlTests.cs
```diff
@@ -310,17 +310,18 @@ public void TestAll()
...
                 {
                     var defaultGuidRepresentation = BsonDefaults.GuidRepresentation;
                     if (url.GuidRepresentation == defaultGuidRepresentation)
                     {
-                        expectedConnectionString = expectedConnectionString.Replace("uuidRepresentation=pythonLegacy;", "");
+                        connectionString = connectionString.Replace("uuidRepresentation=pythonLegacy;", "");
                     }
                 }
 #pragma warning restore 618
+                var expectedConnectionString = connectionString.Replace(';', '&');
```
JamesKovacs, line 324: And here.

JamesKovacs (changes requested):
Overall looks good. Just need to switch our tests to use `&` by default rather than `;` and thus eliminate the need to call `Replace`.

JamesKovacs (approved):
LGTM

PR #1181, opened 2023-09-25
Fix null reference exception causing failed tests

adelinowona:
The commit for CSHARP-4018 is causing tests to fail in relation to changes that may have introduced null reference hazards.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
@@ -129,7 +129,7 @@ namespace MongoDB.Driver.Core.Servers
                     }
                     else
                     {
-                        var helloCommand = HelloHelper.CreateCommand(_serverApi, helloOk, loadBalanced: _roundTripTimeConnection.Settings?.LoadBalanced ?? false);
+                        var helloCommand = HelloHelper.CreateCommand(_serverApi, helloOk, loadBalanced: _roundTripTimeConnection.Settings.LoadBalanced);
                         var helloProtocol = HelloHelper.CreateProtocol(helloCommand, _serverApi);
 
                         var stopwatch = Stopwatch.StartNew();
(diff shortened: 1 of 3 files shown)


adelinowona:
Using the null-coalescing operator in both the RoundTripMonitor and ServerMonitor code seemed to have fixed the null reference issue. However, there are other places where I make use of the pattern of checking for the LoadBalanced setting in the connection settings but I am not too sure if I should also make use of the null-coalescing operator in those cases just to err on the side of caution. Those uses can be found in `DefaultAuthenticator.Authenticate(Async)` and `CommandWireProtocol.CreateSupportedWireProtocol`. I can't guarantee that in these cases the connection settings will never be null.

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -129,7 +129,7 @@ private void MonitorServer()
                     }
                     else
                     {
-                        var helloCommand = HelloHelper.CreateCommand(_serverApi, helloOk, loadBalanced: _roundTripTimeConnection.Settings.LoadBalanced);
+                        var helloCommand = HelloHelper.CreateCommand(_serverApi, helloOk, loadBalanced: _roundTripTimeConnection.Settings?.LoadBalanced ?? false);
```
BorisDog, line 132: Why is it possible not for `Settings` to be null? (might be a small breaking change for users).

sanych-sun (approved):
LGTM

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
adelinowona: It can be null but we would like to know if the connection used is loadbalanced or not in order to use hello command and OP_MSG so I have to make sure if Settings is null we use a default `false` value.

adelinowona:
There doesn't seem to be a code path in the driver that allows for connection settings to be null in a connection object that is in use. It seems the root cause might be that the setup for `CSharp3302Tests ` didn't set the connection settings for the mock connection object used in tests which resulted in the null reference.

JamesKovacs (approved):
LGTM

NOTE: The failing variants are single-test failures of known flaky tests.

BorisDog (approved):
LGTM

rstam (approved):
Testing with mock objects can be very fragile since typically you only set up the mock JUST ENOUGH for the code under test to pass, and then when that code changes the mock setup is no longer adequate.

LGTM

damieng (approved):

PR #1182, opened 2023-09-25
CSHARP-4792: Verify that delimiting slash bewteen hosts and options is optional.

JamesKovacs:
I sync'd `$/specifications/connection-string/tests` to the spec repo. The changes are minimal and all tests pass. The reason for the large diff is a formatting change in the JSON files. The YAML files (from which the JSON files were generated) were not significantly altered.


rstam (approved):
LGTM

PR #1184, opened 2023-09-27
CSHARP-4714: Reset EC2 instance profile at end of test if necessary.

JamesKovacs:
This is an additional cleanup step if the EC2 instance profile was changed during variant setup for MONGODB-AWS auth testing. This should improve the stability of the EG environment and avoid failures on future runs by any driver when run on the same host.


sanych-sun (approved):
LGTM

PR #1185, opened 2023-09-27
EF-26: Pass EF Provider data to Client Meta Data Handshake

sanych-sun:


src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -87,6 +92,11 @@ bool IsLegacyLoaded()
             {
                 return Type.GetType("MongoDB.Driver.MongoServer, MongoDB.Driver.Legacy") != null;
             }
+
+            bool IsEFLoaded()
+            {
+                return Type.GetType("MongoDB.EntityFrameworkCore.Storage.MongoDatabaseWrapper, MongoDB.EntityFrameworkCore") != null;
```
damieng, line 98: Maybe we could choose a different class name - I had plans to rename that one at some point as I think it's not great.

`MongoQueryContext` in `MongoDB.EntityFrameworkCore.Query` perhaps?

rstam: Agreed that we should pick a type that is *very* stable for this test.

rstam (approved):
LGTM

damieng (approved):

PR #1188, opened 2023-09-29
CSHARP-4799: Pipeline stage with BsonDocumentSearchDefinition can be rendered only once

BorisDog:


src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -76,7 +76,7 @@ public BsonDocumentSearchDefinition(BsonDocument document)
 
         /// <inheritdoc />
         public override BsonDocument Render(SearchDefinitionRenderContext<TDocument> renderContext) =>
-            Document;
+`            (BsonDocument)Document.DeepClone();
```
adelinowona, line 79: Typo `'` on this line :)

tests/MongoDB.Driver.Tests/Search/SearchDefinitionTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Search;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Search
+{
+    public class SearchDefinitionTests
+    {
+        [Theory]
+        [InlineData(null)]
```
adelinowona, line 27: Just curious, what is the null test case testing?

tests/MongoDB.Driver.Tests/Search/SearchDefinitionTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var document = new BsonDocument("operator", "parameter");
+            var searchDefinition = (BsonDocumentSearchDefinition<BsonDocument>)document;
+
+            var pipeline = (new EmptyPipelineDefinition<BsonDocument>()).Search(searchDefinition, searchOptions);
+
+            var renderedPipeline = pipeline.Render(BsonSerializer.SerializerRegistry.GetSerializer<BsonDocument>(), BsonSerializer.SerializerRegistry);
+            AssertStages();
+
+            renderedPipeline = pipeline.Render(BsonSerializer.SerializerRegistry.GetSerializer<BsonDocument>(), BsonSerializer.SerializerRegistry);
```
adelinowona, line 62: Is the returned BsonDocument somehow altered in between the two `Render` calls? From my understanding it seems we want it so that subsequent calls to render will continue to render the same result even if the returned BsonDocument is modified. I don't explicitly see it unless it is not required?

adelinowona (changes requested):

tests/MongoDB.Driver.Tests/Search/SearchDefinitionTests.cs
sanych-sun, line 27: Should we test empty document (`"{}"`) as well?

sanych-sun (approved):
LGTM + minor comment

src/MongoDB.Driver/Search/SearchDefinition.cs
BorisDog: Thanks!

tests/MongoDB.Driver.Tests/Search/SearchDefinitionTests.cs
BorisDog: It tests that `null` is returned in the following case:
`var x = (SearchDefinition<BsonDocument>)null;`

BorisDog: I don't think so, the only distinction is between `null` and not `null`.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionTests.cs
BorisDog: `SearchDefinition` stage alters the underlying `BsonDocument` returned by `BsonDocumentSearchDefinition`.
So we need to test that pipeline containing the `Search` stage can be rendered more than once, so we are testing this through `pipeline.Render`.

You are correct in your observation that we are not testing the rendering with modification of `renderedPipeline`, but that's not a goal of this test case.

adelinowona: Oh ok, thanks for the info.

adelinowona (approved):

PR #1189, opened 2023-10-04
CSHARP-4803: Refactor StringExpressionToRegexFilterTranslator.TryTranslate to not use an exception for flow control when possible.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
```diff
@@ -230,33 +230,52 @@ public static bool CanTranslate(Expression expression)
...
+
         public static bool TryTranslate(TranslationContext context, Expression expression, out AstFilter filter)
         {
-            try
-            {
-                filter = Translate(context, expression);
-                return true;
-            }
-            catch (ExpressionNotSupportedException)
+            if (CanTranslate(expression))
```
rstam, line 244: It was not possible to eliminate the try/catch entirely, but by calling `CanTranslate` first we can dramatically reduce the number of scenarios where an exception might still be thrown.

An example of a scenario where an exception might still be thrown is when we discover many levels down in the call stack from here that a string field is not actually stored in the database as a string. By the time we find that out we are deep into a chain of calls, and an exception is still the easiest way to back out.

But for most common scenarios an exception is no longer thrown.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
```diff
@@ -608,7 +622,7 @@ bool IsImpossibleMatch(Modifiers modifiers, string value)
             }
         }
 
-        private static AstFilter TranslateStringComparison(TranslationContext context, Expression expression, Expression leftExpression, AstComparisonFilterOperator comparisonOperator, Expression rightExpression)
+        private static AstFilter TranslateStringEqualityComparison(TranslationContext context, Expression expression, Expression leftExpression, AstComparisonFilterOperator comparisonOperator, Expression rightExpression)
```
rstam, line 625: Slightly better name, to emphasize that only equality (include not equals) comparisons are translated here.

sanych-sun (approved):
LGTM

PR #1191, opened 2023-10-05
Allow the DEFAULT auth mechanism to be specified explicitly

cknaap:
MongoDBCompass can copy a connectionstring, and for the DEFAULT auth mechanism it will include `authMechanism=DEFAULT` literaly. This PR makes the `ToAuthenticator` method accept that.

Before this, it would only use the default if the authMechanism was not specified at all, and including it in the connectionstring would lead to an exception.

### DIFF
--- a/specifications/auth/tests/connection-string.json
+++ b/specifications/auth/tests/connection-string.json
@@ -12,6 +12,18 @@
         "mechanism_properties": null
       }
     },
+      {
+          "description": "should use the default source and mechanism, when DEFAULT is specified",
+          "uri": "mongodb://user:password@localhost?authMechanism=DEFAULT",
+          "valid": true,
+          "credential": {
+              "username": "user",
+              "password": "password",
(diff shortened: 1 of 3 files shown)


JamesKovacs:
Hi, @cknaap,

Thank you for submitting this PR. I confirmed that MongoDB Compass is including the connection string option `authMechanism=DEFAULT` when you select username/password as the authentication mechanism. However this is invalid according to the Auth specification. See the [Auth Related Options section](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#auth-related-options) for allowed values. The correct behaviour is to simply omit `authMechanism` if the default mechanism rules are to be followed. I've filed [COMPASS-7308](https://jira.mongodb.org/browse/COMPASS-7308) to have this issue resolved in Compass.

Thank you for bringing this to our attention.

Sincerely,
James

PR #1195, opened 2023-10-06
CSHARP-4566: Fix some tests.

rstam:


tests/MongoDB.Driver.Tests/FilterDefinitionBuilderIntComparedToDoubleWithStringRepresentationTests.cs
```diff
@@ -43,7 +43,7 @@ public void Eq_with_field_name_should_render_correctly(double value, string expe
         }
 
         [Theory]
-        [InlineData(1.5, "{ I : \"1.5\" }")]
+        [InlineData(1.0, "{ I : \"1\" }")]
```
rstam, line 46: I believe this test is confirming how things *should* work.

What bothers me is that the test just before this didn't have to be fixed. It's because when using a field name `"I"` the LINQ machinery isn't used at all, so the changes for CSHARP-4566 don't come into play as they do when using a lambda expression `x => x.I` to sepcifify the field.

rstam: We will have to decide whether to address the `"I"` case or not before release.

sanych-sun (approved):
LGTM

rstam:
Closed because there are unresolved issues. I will create a new PR when further changes are made.

PR #1200, opened 2023-10-16
CSHARP-4809: If a string serializer doesn't implement IHasRepresentationSerializer don't assume string values are actually represented as strings.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
```diff
@@ -519,8 +519,11 @@ private static (AstFilterField, Modifiers) TranslateField(TranslationContext con
             {
                 var field = ExpressionToFilterFieldTranslator.Translate(context, fieldExpression);
 
-                if (field.Serializer is IRepresentationConfigurable representationConfigurable &&
```
rstam, line 522: At the time this code was written the `IHasRepresentationSerializer` interface didn't exist.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
```diff
@@ -519,8 +519,11 @@ private static (AstFilterField, Modifiers) TranslateField(TranslationContext con
...
-                    representationConfigurable.Representation != BsonType.String)
+                if (field.Serializer is not IHasRepresentationSerializer hasRepresentationSerializer)
```
rstam, line 522: If the serializer does not implement `IHasRepresentationSerializer` then we are unable to safely determine that the configured string serializer actually stores string values as strings in the database.

BorisDog (approved):
LGTM

JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4809Tests.cs
```diff
@@ -0,0 +1,99 @@
...
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp4809Tests : Linq3IntegrationTest
+    {
+        [Theory]
+        [ParameterAttributeData]
+        public void Filter_by_Id_with_custom_serializer_should_work(
```
sanych-sun, line 32: I can see only "negative" case, the case when string was represented as non-string. Should we have positive test(s) as well?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4809Tests.cs
```diff
@@ -0,0 +1,99 @@
...
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.Driver.Linq;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp4809Tests : Linq3IntegrationTest
```
sanych-sun, line 28: Can we try to use more meaningful name for test class? Like a `StringExpressionToRegexFilterTranslatorTests`

sanych-sun (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
rstam: I don't think this is a breaking change. It is a bug fix.

We used erroneously generate regex queries against values that weren't strings in the database.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4809Tests.cs
rstam: https://mongodb.slack.com/archives/GQ27C9N7J/p1697495123127329?thread_ts=1697471201.169569&cid=GQ27C9N7J

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4809Tests.cs
rstam: There are plenty of tests for the normal case elsewhere.

I can add one here also if you want.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/StringExpressionToRegexFilterTranslator.cs
sanych-sun: If someone implemented their own serializer for string for any reasons. It was not throwing before (but simply assume it should be string out there), but it will throw after the next release.

rstam: OK I see your point. I think we discussed this and decided that simply assuming the representation was a string was not safe.

@JamesKovacs does this need to be reconsidered?

The core question is: if we have no evidence that some string serializer uses string representation or not what should we assume? That it does? Or throw because we can't determine whether it does?

rstam: Note: if you look at the test in this PR the test still works even though an exception is thrown. This test specifically uses a user writer custom serializer that does not implement `IHasRepresentationSerializer` (the situation you were worried about).

That's because the exception simply means that this string expression can't be translated using regular expressions. The exception will be caught and the expression will then be translated in some other way that doesn't use regular expressions (and therefore doesn't require knowing whether the database representation is a string).

PR #1201, opened 2023-10-17
CSHARP-4808: "Updating stale electionId and setVersion" even when electionId/setVersion match

BorisDog:


src/MongoDB.Driver.Core/Core/Clusters/MultiServerCluster.cs
```diff
@@ -707,7 +707,9 @@ public override bool Equals(object obj)
             {
                 var other = obj as ElectionInfo;
 
-                return other != null && _setVersion == other.SetVersion && other.ElectionId == _electionId;
+                return other != null &&
+                    _setVersion == other.SetVersion &&
+                    (_electionId == null && other.ElectionId == null || other.ElectionId?.Equals(_electionId) == true);
```
rstam, line 712: I think you can simplify this line to:

    object.Equals(_electionId, other.ElectionId)

rstam (changes requested):
One minor change suggested.

src/MongoDB.Driver.Core/Core/Clusters/MultiServerCluster.cs
BorisDog: Good suggestion, thanks!

rstam (approved):
LGTM

PR #1202, opened 2023-10-18
CSHARP-4813: Verify more carefully that property named Count is a collection property.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Misc/DocumentSerializerHelper.cs
```diff
@@ -22,7 +23,7 @@ internal static class DocumentSerializerHelper
     {
         public static MemberSerializationInfo GetMemberSerializationInfo(IBsonSerializer serializer, string memberName)
         {
-            if (!(serializer is IBsonDocumentSerializer documentSerializer))
+            if (!IsDocumentSerializer(serializer, out var documentSerializer))
```
rstam, line 26: At least one serializer (`DictionarySerializer`) implemented `IBsonDocumentSerializer` when it shouldn't have.

However, I tried removing that implementation and it broke LINQ2.

The solution was to replace `serializer is IBsonDocumentSerializer` with a call to a helper method that filters out the mistaken implementations. Note that only the LINQ3 implementation uses this helper method.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/SerializationHelper.cs
```diff
@@ -25,6 +27,35 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Misc
...
     internal static class SerializationHelper
     {
+        public static void EnsureRepresentationIsArray(Expression expression, IBsonSerializer serializer)
+        {
+            if (!HasArrayRepresentation(serializer))
+            {
+                throw new ExpressionNotSupportedException(expression, because: "the expression is not represented as an array in the database");
+            }
+
+            // TODO: replace this heuristic with a more accurate determination
```
rstam, line 37: We really should require serializers to implement `IHasRepresentationSerializer` so that we could check if the representation is an array or not.

For now we approximate it.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableProperty.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
+{
+    internal static class EnumerableProperty
+    {
+        // public methods
+        public static bool IsCountProperty(MemberExpression expression)
```
rstam, line 28: This prevents false hits on `Count` properties that aren't from a class that represents a collection.

JamesKovacs (approved):
LGTM

rstam:
So here are the key insights for the third commit:

The following two questions are not the same:

* does serializer represent values as a document
* does serializer represent members as fields in a document

and:

`serializer is IBsonDocumentSerializer` doesn't tell us anything (because some classes implemented `IBsonDocumentSerializer` when they shouldn't have)

So all code that used `is IBsonDocumentSerializer` now uses the new helper methods `IsRepresentedAsDocument` and `AreMembersRepresentedAsFields` instead (LINQ3 only, LINQ2 is unchanged).

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableProperty.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            // Count is not actually a property defined by IEnumerable but rather defined by several sub-interfaces
+            return
+                expression.Expression != null &&
+                expression.Member is PropertyInfo propertyInfo &&
+                propertyInfo.Name == "Count" &&
+                propertyInfo.PropertyType == typeof(int) &&
+                propertyInfo.GetGetMethod().GetParameters().Length == 0 &&
```
rstam, line 36: Interesting fact that's easy to miss: properties can have parameters too so we need to check that this Count property has 0 parameters.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/DocumentSerializerHelper.cs
```diff
@@ -14,17 +14,67 @@
...
+            }
+
+            if (serializer is IDowncastingSerializer downcastingSerializer)
+            {
+                return AreMembersRepresentedAsFields(downcastingSerializer.DerivedSerializer, out documentSerializer);
+            }
+
+            if (serializer is IImpliedImplementationInterfaceSerializer impliedImplementationSerializer)
+            {
+                return AreMembersRepresentedAsFields(impliedImplementationSerializer.ImplementationSerializer, out documentSerializer); ;
```
JamesKovacs, line 39: Double semi-colon at the end of line.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableProperty.cs
JamesKovacs: Good point. I'd forgotten that fact.

JamesKovacs (approved):
LGTM. Minor issue with double semi-colon needs fixing.

PR #1203, opened 2023-10-18
CSHARP-4716: Log informational message client-side based on detected environment (DocumentDB or CosmosDB)

BorisDog:


src/MongoDB.Driver.Core/Core/Clusters/ClusterFactory.cs
```diff
@@ -102,5 +107,33 @@ private LoadBalancedCluster CreateLoadBalancedCluster(ClusterSettings setting)
...
+                }
+            }
+
+            bool IsCosmosDB(string host) =>
+                IsExternalEnvironment(host, "CosmosDB", "https://www.mongodb.com/supportability/cosmosdb", ".cosmos.azure.com");
+
+            bool IsDocumentDB(string host) =>
+                IsExternalEnvironment(host, "DocumentDB", "https://www.mongodb.com/supportability/documentdb", ".docdb.amazonaws.com", ".docdb-elastic.amazonaws.com");
+
+            bool IsExternalEnvironment(string host, string environment, string documentationUrl, params string[] suffixes)
```
adelinowona, line 127: `IsExternalEnvironment()` seems to always return false.

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterFactoryTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        [Theory]
+        [InlineData("mongodb://a.mongo.cosmos.azure.com.tld:19555")]
+        [InlineData("mongodb://a.docdb-elastic.amazonaws.com.t")]
+        [InlineData("mongodb+srv://a.example.com")]
+        [InlineData("mongodb+srv://a.mongodb.net/")]
+        [InlineData("mongodb+srv://a.mongo.cosmos.azure.com.tld/")]
+        [InlineData("mongodb+srv://a.docdb-elastic.amazonaws.com.tld/")]
+        public void ClusterFactory_should_not_log_if_no_external_environment_is_detected(string connectionString)
```
adelinowona, line 75: Should we add a test with multiple external hosts to ensure that only one log message is emitted?

adelinowona (changes requested):

src/MongoDB.Driver.Core/Core/Clusters/ClusterFactory.cs
BorisDog: Good catch! Done, thanks!

tests/MongoDB.Driver.Core.Tests/Core/Clusters/ClusterFactoryTests.cs
BorisDog: Added.

adelinowona (approved):
LGTM

src/MongoDB.Driver.Core/Core/Clusters/ClusterFactory.cs
```diff
@@ -102,5 +107,35 @@ private LoadBalancedCluster CreateLoadBalancedCluster(ClusterSettings setting)
         {
             return new LoadBalancedCluster(setting, _serverFactory, _eventSubscriber, _loggerFactory);
         }
+
+        private void ProcessClusterEnvironment(ClusterSettings settings)
+        {
+            foreach (var (host, _) in  settings.EndPoints.Select(EndPointHelper.GetHostAndPort))
+            {
+                if (IsCosmosDB(host) || IsDocumentDB(host))
```
JamesKovacs, line 115: Cross-cloud clusters are possible in Atlas. You can deploy a cluster with some nodes in Azure and others AWS. We should log informational messages for both in such a case.

As well, `IsCosmosDB` and `IsDocumentDB` call `IsExternalEnvironment`, which has side effects. This is unexpected. I would not expect calling `IsCosmosDB` in a tight loop to chew up log space. There's nothing wrong with the logic, but naming should be adjusted to reflect what the method is actually doing. Something like `LogIfCosmosDB`. I don't particularly care for that name and would love a better suggestion, but hopefully you get the idea.

JamesKovacs (changes requested):

src/MongoDB.Driver.Core/Core/Clusters/ClusterFactory.cs
BorisDog: Good point regarding the naming. Done.

JamesKovacs (approved):
LGTM

PR #1204, opened 2023-10-18
CSHARP-4670: Implement Drivers Performance Benchmarking Spec

adelinowona:
To make the review process more manageable and efficient, I've divided the PR into two batches for review. Each batch focuses on specific sections to ensure a detailed yet digestible review process.

Batch 1: Overall methodology, benchmarking approach and code structure changes.
- all files **not** in the BSON, Multi-Doc, ParallelBench and Single-Doc folders

Batch 2: Performance improvements and benchmarks accuracy
- all files in the these folders: BSON, Multi-Doc, ParallelBench and Single-Doc


BorisDog (changes requested):
Partial review.
Looks really good overall. Mainly minor and styling comments.

CSharpDriver.sln
```diff
@@ -64,6 +64,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SmokeTests", "SmokeTests",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MongoDB.TestHelpers", "tests\MongoDB.TestHelpers\MongoDB.TestHelpers.csproj", "{DF888021-744F-4A8B-9324-831DEFC48AB8}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "benchmarks", "benchmarks\benchmarks.csproj", "{6C1F8B13-7CAD-44BB-B26E-CDCF9785A1A8}"
```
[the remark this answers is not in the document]
adelinowona: I think this is a good suggestion, it also provides a place to keep future benchmark projects.

benchmarks/download-data.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
adelinowona: Agreed, it helps clean up the root directory the benchmark project

benchmarks/BenchmarkResult.cs
```diff
@@ -0,0 +1,31 @@
...
+
+            foreach (int percentile in new[]{10, 25, 50, 75, 80, 90, 95, 98, 99})
+            {
+                _percentiles[percentile] = benchmarkReport.ResultStatistics.Percentiles.Percentile(percentile);
+            }
+
+            Score = datasetSize / (_percentiles[50] / 1000);
+        }
+
+        public double GetPercentile(int percentile)
```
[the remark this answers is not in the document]
adelinowona: The spec says to collect those percentiles and report them so that's why I have it here. But after inspecting most of the other driver's benchmarks, no one is reporting the percentiles yet. So I think we can remove this functionality for now and introduce it when needed or asked for.

benchmarks/download-data.sh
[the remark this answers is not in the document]
sanych-sun: I suppose it could be in the `benchmarks` folder. Having the file in `evergreen` folder might be confusing, because the script is intended to be used locally as well.

sanych-sun (changes requested):

benchmarks/BSON/DeepBsonDecodingBenchmark.cs
```diff
@@ -0,0 +1,38 @@
...
+            _reader = new BsonBinaryReader(_stream);
+            _context = BsonDeserializationContext.CreateRoot(_reader);
+        }
+
+        [Benchmark]
+        public void DeepBsonDecoding()
+        {
+            for (int i = 0; i < 10000; i++)
+            {
+                BsonDocumentSerializer.Instance.Deserialize(_context);
```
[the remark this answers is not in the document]
adelinowona: discussed in slack

benchmarks/BSON/DeepBsonEncodingBenchmark.cs
```diff
@@ -0,0 +1,39 @@
...
+            _context = BsonSerializationContext.CreateRoot(_writer);
+            _document = ReadExtendedJson("../../../../../../../data/extended_bson/deep_bson.json");
+        }
+
+        [Benchmark]
+        public void DeepBsonEncoding()
+        {
+            for (int i = 0; i < 10000; i++)
+            {
+                BsonDocumentSerializer.Instance.Serialize(_context, _document);
```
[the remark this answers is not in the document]
adelinowona: discussed in slack

benchmarks/BSON/DeepBsonDecodingBenchmark.cs
```diff
@@ -0,0 +1,38 @@
...
+        public void Setup()
+        {
+            _bytes = ReadExtendedJsonToBytes("../../../../../../../data/extended_bson/deep_bson.json");
+            _stream = new MemoryStream(_bytes);
+            _reader = new BsonBinaryReader(_stream);
+            _context = BsonDeserializationContext.CreateRoot(_reader);
+        }
+
+        [Benchmark]
+        public void DeepBsonDecoding()
```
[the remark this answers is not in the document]
adelinowona: done

benchmarks/CustomExporters/EvergreenExporter.cs
```diff
@@ -0,0 +1,107 @@
+using System.IO;
+using System.Linq;
+using MongoDB.Bson.IO;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Exporters;
+using System.Collections.Generic;
+using static benchmarks.BenchmarkExtensions;
+
+namespace benchmarks.CustomExporters
```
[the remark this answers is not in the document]
adelinowona: done

benchmarks/Multi-Doc/FindManyBenchmark.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using static benchmarks.BenchmarkExtensions;
+
+namespace benchmarks.Multi_Doc
```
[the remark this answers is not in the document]
adelinowona: done

benchmarks/BSON/DeepBsonDecodingBenchmark.cs
```diff
@@ -0,0 +1,38 @@
+using System.IO;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using BenchmarkDotNet.Attributes;
+using MongoDB.Bson.Serialization.Serializers;
+using static benchmarks.BenchmarkExtensions;
```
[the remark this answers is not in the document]
adelinowona: done. Renamed to MongoDB.Driver.Benchmarks instead

benchmarks/BenchmarkResult.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            Name = benchmarkReport.BenchmarkCase.Descriptor.WorkloadMethodDisplayInfo;
+
+            _percentiles = new Dictionary<int, double>();
+
+            foreach (int percentile in new[]{10, 25, 50, 75, 80, 90, 95, 98, 99})
+            {
+                _percentiles[percentile] = benchmarkReport.ResultStatistics.Percentiles.Percentile(percentile);
+            }
+
```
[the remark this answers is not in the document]
adelinowona: will do when we decide to add percentiles to the benchmark report in evergreen

benchmarks/BenchmarkResult.cs
```diff
@@ -0,0 +1,31 @@
+using BenchmarkDotNet.Reports;
+using System.Collections.Generic;
+
+namespace benchmarks
+{
+    public class BenchmarkResult
```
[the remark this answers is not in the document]
adelinowona: done

benchmarks/BenchmarkExtensions.cs
```diff
@@ -0,0 +1,90 @@
+using System.IO;
+using MongoDB.Bson;
+using System.Collections.Generic;
+
+namespace benchmarks
+{
+    public static class BenchmarkExtensions
```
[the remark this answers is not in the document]
adelinowona: Updated benchmark suite to use categories for filtering!

benchmarks/BenchmarkExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        public static IReadOnlyList<string> SingleBenchmarks { get; } =
```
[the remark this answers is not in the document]
adelinowona: not doing since this code is now removed

benchmarks/BSON/DeepBsonDecodingBenchmark.cs
```diff
@@ -0,0 +1,38 @@
+using System.IO;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using BenchmarkDotNet.Attributes;
+using MongoDB.Bson.Serialization.Serializers;
```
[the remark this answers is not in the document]
adelinowona: done

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/FindManyBenchmark.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        private void PopulateCollection()
+        {
+            var documents = new List<BsonDocument>();
+            for (int i = 0; i < 10000; i++)
+            {
+                var documentCopy = _tweetDocument.DeepClone().AsBsonDocument;
+                documents.Add(documentCopy);
+            }
+            _collection.InsertMany(documents);
```
BorisDog, line 64: Consider:

```
var documents = Enumerable
                .Range(0, 10000)
                .Select(_ => _tweetDocument.DeepClone().AsBsonDocument);
 _collection.InsertMany(documents);
```

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/GridFSDownloadBenchmark.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Benchmark]
+        public void GridFsDownload()
+        {
+            _gridFsBucket.DownloadAsBytes(_fileId);
+        }
+
+        [GlobalCleanup]
+        public void Teardown()
+        {
+            _client.DropDatabase("perftest");
```
BorisDog, line 52: I think we need to port `DisposableMongoClient` from tests, and use it here to dispose the cluster.

benchmarks/MongoDB.Driver.Benchmarks/Exporters/EvergreenExporter.cs
```diff
@@ -0,0 +1,130 @@
...
+                benchmarkResults.Add(new BenchmarkResult(report, benchmarkName, GetDatasetSize(benchmarkName)));
+            }
+
+            var resultsPath = Path.Combine(summary.ResultsDirectoryPath, _outputFile);
+            if (File.Exists(resultsPath))
+            {
+                File.Delete(resultsPath);
+            }
+
+            using (var jsonWriter = new JsonWriter(File.CreateText(resultsPath), new JsonWriterSettings { Indent = true }))
```
BorisDog, line 64: I think `File.CreateText` just overwrites the existing file, no need to delete.

benchmarks/MongoDB.Driver.Benchmarks/Exporters/EvergreenExporter.cs
```diff
@@ -0,0 +1,130 @@
...
+            jsonWriter.WriteString("name", "megabytes_per_second");
+            jsonWriter.WriteDouble("value", score);
+            jsonWriter.WriteEndDocument();
+            jsonWriter.WriteEndArray();
+
+            jsonWriter.WriteEndDocument();
+        }
+
+        private static void WriteIndividualBenchmarkToResults(JsonWriter jsonWriter, BenchmarkResult benchmarkResult)
+        {
```
BorisDog, line 114: Maybe the previous `WriteCompositeScoreToResults` method can be reused here?

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileImportBenchmark.cs
```diff
@@ -0,0 +1,80 @@
...
+        private IMongoCollection<BsonDocument> _collection;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            string mongoUri = Environment.GetEnvironmentVariable("MONGODB_URI");
+            _client = mongoUri != null ? new MongoClient(mongoUri) : new MongoClient();
+            _client.DropDatabase("perftest");
+            _database = _client.GetDatabase("perftest");
+        }
```
BorisDog, line 43: Looks like this setup can be extracted to 
```
_database = MongoConfiguration.CreateDatabase();
_client = _database.Client;
```

benchmarks/MongoDB.Driver.Benchmarks/MongoDB.Driver.Benchmarks.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Optimize>true</Optimize>
+    <Configuration>Release</Configuration>
+    <NoWarn>NU1701</NoWarn>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.13.8" />
+    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.13.8" Condition="'$(OS)' == 'Windows_NT'"/>
```
BorisDog, line 26: Out of curiosity, why is this package needed on Windows?

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/FindManyBenchmark.cs
```diff
@@ -0,0 +1,67 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
```
BorisDog, line 20: Remove/Sort usings everywhere.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            string extendedJson = File.ReadAllText(resourcePath);
+            var document = BsonDocument.Parse(extendedJson);
+            return document.ToBson();
+        }
+
+        public static int GetDatasetSize(string benchmarkName) =>
+            benchmarkName switch
+            {
+                "GridFSMultiFileUploadBenchmark" or "GridFSMultiFileDownloadBenchmark" => 262144000,
```
BorisDog, line 39: Let use `nameof`:
`nameof(InsertOneLargeBenchmark) or nameof(InsertManyLargeBenchmark) => 1111`

benchmarks/MongoDB.Driver.Benchmarks/Exporters/EvergreenExporter.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        public IEnumerable<string> ExportToFiles(Summary summary, ILogger consoleLogger)
+        {
+            var benchmarkResults = new List<BenchmarkResult>();
+            foreach (var report in summary.Reports)
+            {
+                string benchmarkName;
+                if (report.BenchmarkCase.Descriptor.HasCategory(DriverBenchmarkCategory.BsonBench))
```
BorisDog, line 47: Looks like this code is shared with LocalExporter and can be extracted to 
`var benchmarkName = BenchmarkHelper.GetBenchmarkName(BenchmarkCase)`
or event simpler:
`var  benchmarResult = BenchmarkHelper.GetBenchmarkResult(report);`
and then `GetDatasetSize` can stay private.

benchmarks/MongoDB.Driver.Benchmarks/Exporters/LocalExporter.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Linq;
+using BenchmarkDotNet.Loggers;
+using BenchmarkDotNet.Reports;
+using BenchmarkDotNet.Exporters;
+using System.Collections.Generic;
+using static MongoDB.Benchmarks.BenchmarkHelper;
+
+namespace MongoDB.Benchmarks.Exporters
+{
+    public class LocalExporter : IExporter
```
BorisDog, line 26: sealed

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/FindManyBenchmark.cs
```diff
@@ -0,0 +1,67 @@
...
+            _tweetDocument = ReadExtendedJson("../../../../../../../data/single_and_multi_document/tweet.json");
+            _collection = _client.GetDatabase("perftest").GetCollection<BsonDocument>("corpus");
+            PopulateCollection();
+        }
+
+        [Benchmark]
+        public void FindManyAndEmptyCursor()
+        {
+            _collection.Find(Builders<BsonDocument>.Filter.Empty).ToList();
+        }
```
BorisDog, line 48: Do we need simple verifications of the correctness of the results? For  example count verification in this case.
Also to avoid any possible compiler optimization where unused results can be discarded (I don't think it applies here but not 100% sure).

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/GridFSMultiFileDownloadBenchmark.cs
```diff
@@ -0,0 +1,98 @@
...
+        [IterationSetup]
+        public void BeforeTask()
+        {
+            ClearDirectory();
+        }
+
+        [Benchmark]
+        public void GridFsMultiDownload()
+        {
+            Task[] tasks = new Task[50];
```
BorisDog, line 54: We need to understand the threading model prescribed by spec here.
Multiple tasks execution is not guarantied to be fully parallelized, and depends on various factors and various across different machines, setups, load.
So we might need to use threads directly or other solutions to for deterministic behavior.

BorisDog (changes requested):

benchmarks/MongoDB.Driver.Benchmarks/MongoDB.Driver.Benchmarks.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <PackageReference Include="NDesk.Options" Version="0.2.1" />
```
sanych-sun, line 27: Is this project still alive? Could we use [System.CommandLine](https://www.nuget.org/packages/System.CommandLine) instead? It's still in beta, but at least this is Microsoft supported one.

sanych-sun (changes requested):

benchmarks/MongoDB.Driver.Benchmarks/MongoDB.Driver.Benchmarks.csproj
adelinowona: I think it was auto added when I created the benchmark project. With BenchmarkDotNet you can add diagnosers for your benchmarks such as GC and memory allocations, JIT etc. Some of the diagnosers are only available on windows so I think the package is auto added so that one can add some diagnosers that will only run if they ran the benchmarks on a windows machine. We are not using any of those windows-only diagnosers so I can remove this for now.

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/FindManyBenchmark.cs
adelinowona: I tested correctness as I was writing the benchmarks so the benchmarks as is should be correct. Having correctness checks in some benchmarks seem very trivial while in others it may require doing something a bit more involved which may start having us diverge from performance work to correctness work. Most of the other driver benchmarks I have seen don't do correctness checks. I assume they probably did test for correctness as well while writing the benchmarks but I'll ask just out of curiosity.

adelinowona: Regarding the possible compiler optimization, I looked at the generated IL code and I don't think the compiler optimizations will do anything since almost all of the operations in benchmarks are performed on a member variable of a class which means compiler will be probably think that the variable is modified in some way by the calls thus it remains cautious. I also played around with this lab to confirm: [SharpLab](https://sharplab.io/)

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            string extendedJson = File.ReadAllText(DataFolderPath + resourcePath);
+            var document = BsonDocument.Parse(extendedJson);
+            return document.ToBson();
+        }
+
+        public static double CalculateCompositeScore(IEnumerable<BenchmarkResult> benchmarkResults, string benchmarkCategory)
+        {
+            Func<BenchmarkResult, bool> predicate;
+            if (benchmarkCategory == DriverBenchmarkCategory.DriverBench)
```
BorisDog, line 49: Please double check if this switch is needed.

BorisDog (changes requested):
Great refactoring, few more comments

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
```diff
@@ -0,0 +1,102 @@
...
+                predicate = benchmark => categoriesToMatch.Any(s => benchmark.Categories.Contains(s)); // select any benchmarks part of the read or write categories
+            }
+            else
+            {
+                predicate = benchmark => benchmark.Categories.Contains(benchmarkCategory);
+            }
+
+            var identifiedBenchmarksScores = benchmarkResults
+                .Where(predicate)
+                .Select(benchmark => benchmark.Score).ToList();
```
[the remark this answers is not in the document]
adelinowona: I am calling ToList here to avoid possible multiple enumeration since I am calling Any and Average on the enumerable. This helps in avoiding the costs of doing multiple enumerations if the objects being enumerated are large thus saving some performance.

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileExportBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+        public void MultiFileExport()
+        {
+            ThreadingUtilities.ExecuteOnNewThreads(16, threadNumber =>
+            {
+                var numFilesToExport = threadNumber == 15 ? 10 : 6;
+                var startingFileNumber = threadNumber * 6;
+                for (int i = 0; i < numFilesToExport; i++)
+                {
+                    var filepath = $"{DataFolderPath}parallel/tmpLDJSON/ldjson{(startingFileNumber+i):D3}.txt";
+                    var documents = _collection.Find(Builders<BsonDocument>.Filter.Empty).Skip((startingFileNumber+i) * 5000).Limit(5000).ToList();
```
sanych-sun, line 68: Let's discuss this code in offline.

sanych-sun (commented):
Generally looks good + some comments.

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileExportBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+
+                    using (StreamWriter streamWriter = File.CreateText(filepath))
```
BorisDog, line 70: minor: `using (var streamWriter...)`

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileExportBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+        [GlobalCleanup]
+        public void Teardown()
+        {
+            _tmpDirectory.Delete(true);
+            _client.Dispose();
+        }
+
+        private void ClearDirectory()
+        {
+            foreach (var file in _tmpDirectory.EnumerateFiles())
```
BorisDog, line 90: Consider dropping the whole dir.

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileExportBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        private void PopulateCollection()
+        {
+            for (int i = 0; i < 100; i++)
+            {
+                var resourcePath = $"{DataFolderPath}parallel/ldjson_multi/ldjson{i:D3}.txt";
+                var documents = new List<BsonDocument>(5000);
+                documents.AddRange(File.ReadLines(resourcePath).Select(BsonDocument.Parse));
```
BorisDog, line 102: minor:  consider
`var documents = File.ReadLines(resourcePath).Select(BsonDocument.Parse).ToArray();`

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileImportBenchmark.cs
```diff
@@ -0,0 +1,77 @@
...
+            ThreadingUtilities.ExecuteOnNewThreads(16, threadNumber =>
+            {
+                var numFilesToImport = threadNumber == 15 ? 10 : 6;
+                var startingFileNumber = threadNumber * 6;
+                for (int i = 0; i < numFilesToImport; i++)
+                {
+                    var resourcePath = $"{DataFolderPath}parallel/ldjson_multi/ldjson{(startingFileNumber+i):D3}.txt";
+                    var documents = new List<BsonDocument>(5000);
+                    documents.AddRange(File.ReadLines(resourcePath).Select(BsonDocument.Parse));
+                    _collection.InsertMany(documents);
```
BorisDog, line 66: Consider:
var documents = File.ReadLines(resourcePath).Select(BsonDocument.Parse).ToArray();

Also this is code appears in other places, consider moving to `BenchmarkHelper`

BorisDog (changes requested):
Looks great, mainly stylistic comments.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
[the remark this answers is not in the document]
rstam: `ToArray` implementation creates the equivalent of a `List` first and then once that list is populated and it knows how many elements are in it it then creates an array and copies the elements to it, so it requires an extra memory allocation and some copying.

Other than that `ToArray` is fine.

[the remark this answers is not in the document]
rstam: Also, a `List<T>` can be cast to an `IReadOnlyList<T>` if you want to hide the rest of the `List<T>` methods and properties.

`IReadOnlyList<T>` is like an array `T[]` but is immutable.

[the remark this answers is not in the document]
BorisDog: Optimizations considerations are probably neglectable in most cases here, but `ToArray` is probably more optimal in most of our use cases:

- If the underlying enumerable is `ICollection` (not here) creating temporary list is skipped.
- Copying:
 `ToList` will have [N, 2N) bytes copied(16 + 32 + , ... N^(floor(log(N ))
 `ToArray` will have N bytes copied.
- Allocation: You are correct that `ToArray` will allocate additional copy, but it would be gen0 object and quickly collected. But `ToList` will always (almost always) allocate more memory (up to N additional bytes) that will survive much longer. After first gen0 collection:
`ToList` [N, 2N) allocated bytes
`ToArray` N allocated bytes

So adopting `ToArray` instead of `ToList` as a rule of thumb will allow us to benefit from having less copying and less long surviving memory, as well as having simpler data type to work with.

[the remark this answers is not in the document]
rstam: I don't see how `ToArray` will result in less copying. It will always allocate roughly the same amount of memory as `ToList` for the temporary list and then an additional allocation for the final array. (With the exception of classes that implement `ToArray` more efficiently, but as you say that's not the case here).

I agree that if the result is to be long lived the array version will prevent only the required memory to hold the array and no more from being garbage collected.

But in this case the result of `ToList` is short lived anyway.

I don't mind encouraging the use of `ToArray`, but the argument should be that you want the result to be an *array* and not a performance or memory use argument.

benchmarks/MongoDB.Driver.Benchmarks/MongoDB.Driver.Benchmarks.csproj
adelinowona: Discussed on slack

benchmarks/MongoDB.Driver.Benchmarks/MultiDoc/GridFSUploadBenchmark.cs
```diff
@@ -0,0 +1,62 @@
...
+    [IterationCount(100)]
+    [BenchmarkCategory(DriverBenchmarkCategory.MultiBench, DriverBenchmarkCategory.WriteBench, DriverBenchmarkCategory.DriverBench)]
+    public class GridFsUploadBenchmark
+    {
+        private DisposableMongoClient _client;
+        private byte[] _fileBytes;
+        private GridFSBucket _gridFsBucket;
+
+        [Params(52428800)]
+        public int BenchmarkDataSetSize { get; set; }
```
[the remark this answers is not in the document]
adelinowona: This stores the dataset size for the benchmark and it is accessed in the BenchmarkResult constructor. I created this in an effort to keep all benchmark related data in their own class so I don't have to externally get that info from somewhere else.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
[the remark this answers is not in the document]
BorisDog: Less copying:
`ToList` -> Each underlying array will be copied to the new one when list grows, so we'll end up with [N, 2N) bytes copied.
`ToArray` -> When growing the underlying array creator structure, additional array is added for new items, while preserving old array. Exactly N bytes will be copied.

As mentioned at the begging of this thread my motivation to encourage usage of `Array` is to use simpler types where possible and opt-in into more complex types only when additional specific functionality is required.
This performance analysis is to address the concern of additional memory being used in `ToArray`,  while theoretically `ToArray` appears to be more optimal.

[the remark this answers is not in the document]
rstam: > ToArray -> When growing the underlying array creator structure, additional array is added for new items, while preserving old array. Exactly N bytes will be copied.

That's not the implementation I saw. Perhaps they have improved the implementation in more recent versions of .NET. That would be great.

I also like the idea of using the simplest data type possible, but an array is not quite that. It's mutable. The simplest data type possible would be `IReadOnlyList<T>`, assuming all you want to do is consume the results.

But like I said earlier, I'm fine with using `ToArray`.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            string extendedJson = File.ReadAllText(DataFolderPath + resourcePath);
+            var document = BsonDocument.Parse(extendedJson);
+            return document.ToBson();
+        }
+
+        public static double CalculateCompositeScore(IEnumerable<BenchmarkResult> benchmarkResults, string benchmarkCategory)
+        {
+            var identifiedBenchmarksScores = benchmarkResults
+                .Where(benchmark => benchmark.Categories.Contains(benchmarkCategory))
```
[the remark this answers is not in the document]
adelinowona: Yeah probably doesn't really matter here but it doesn't hurt to have more efficient code if it is just a simple change.

benchmarks/MongoDB.Driver.Benchmarks/Exporters/EvergreenExporter.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+        }
+
+        public IEnumerable<string> ExportToFiles(Summary summary, ILogger consoleLogger)
+        {
+            var benchmarkResults = summary.Reports.Select(report => new BenchmarkResult(report)).ToArray();
+
+            var resultsPath = Path.Combine(summary.ResultsDirectoryPath, _outputFile);
+
+            using (var jsonWriter = new JsonWriter(File.CreateText(resultsPath), new JsonWriterSettings { Indent = true }))
```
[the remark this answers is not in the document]
adelinowona: Seems like `JsonWriter` doesn't dispose the `StreamWriter`, the dispose of `JsonWriter` calls its `Close` method which in turn calls the `Flush` method on the `StreamWriter` but no disposing is done.

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/MultiFileImportBenchmark.cs
adelinowona: I think it only appears in two places, I think it is isolated enough to not move it `BenchmarkHelper`

benchmarks/MongoDB.Driver.Benchmarks/ParallelBench/GridFSMultiFileDownloadBenchmark.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            ClearDirectory();
+        }
+
+        [Benchmark]
+        public void GridFsMultiDownload()
+        {
+            ThreadingUtilities.ExecuteOnNewThreads(16, threadNumber =>
+            {
+                var numFilesToDownload = threadNumber == 15 ? 5 : 3;
```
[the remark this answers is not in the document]
adelinowona: I was using 16 threads as I wanted to use the optimal thread count for the evergreen host on which the benchmarks are run which has 16 logical CPUs. Alex also already suggested I use ConcurrentQueue since my current set up might be hard to read. So I will explore using using it as that is better than having this funky work distribution.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+using MongoDB.Driver;
+using MongoDB.Driver.TestHelpers;
+
+namespace MongoDB.Benchmarks
+{
+    public static class BenchmarkHelper
+    {
+        public const string DataFolderPath = "../../../../../../../data/";
+
+        public static void AddFilesToQueue(ConcurrentQueue<(string, int)> filesQueue, string directoryPath, string fileNamePrefix, int fileCount)
```
adelinowona, line 31: Added this function as the piece of code encapsulated here would have been reused in the parallel benchmarks to fill the concurrentQueue.

benchmarks/MongoDB.Driver.Benchmarks/scripts/compare-results.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+
+# Copyright 2021 MongoDB, Inc.
```
BorisDog, line 3: 2010

benchmarks/MongoDB.Driver.Benchmarks/scripts/download-data.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# Copyright 2021 MongoDB, Inc.
```
BorisDog, line 3: 2010

BorisDog (approved):
Great job, LGTM!
Minor optional comments.

sanych-sun (approved):
LGTM

PR #1206, opened 2023-10-19
CSHARP-4783: API docs generating with docfx

sanych-sun:
Here is an example of generated site: https://sanych-sun.github.io/mongo-csharp-driver/api/index.html


.gitignore
```diff
@@ -68,3 +68,5 @@ artifacts
 packages
 /tools
 /evergreen/krb5.conf.empty
+/build
```
JamesKovacs, line 71: Other tasks build into the `artifacts` directory. We should use that rather than adding another directory.

docfx_project/api/index.md
```diff
@@ -0,0 +1 @@
+# MongoDB C# Driver API Reference
```
JamesKovacs, line 1: MongoDB .NET/C# Driver API Reference

evergreen/apidocs-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
JamesKovacs, line 1: Most of our other scripts use `verb-noun.sh`. Consider changing this to `build-apidocs.sh`.

evergreen/evergreen.yml
```diff
@@ -1305,7 +1305,7 @@ tasks:
 
     - name: atlas-search-index-helpers-test
       commands:
-        - func: run-atlas-search-index-helpers-test
```
JamesKovacs, line 1308: Unrelated change even if it is fixing whitespace.

JamesKovacs (changes requested):
This looks great. Minor feedback on naming. As well since this commit replaces Sandcastle with DocFX, we should remove the `./Docs` folder as part of this commit. I realize that this will remove both the Sandcastle-related files as well as the reference docs. The reference docs aren't related, but we stopped updating them awhile ago.

.gitignore
sanych-sun: Ups, I've discovered that test results file being output into the ./build directory and did not check if other tasks output to the same folder. Will update here and some another tasks (packages creation script also output to the ./build folder now, will update it to use ./artifacts folder).

evergreen/apidocs-build.sh
sanych-sun: I probably would like to review this naming convention. If we use opposite convention: `noun-verb.sh` this let us group together related scripts. For example:
apidocs-build
apidocs-validate
packages-pack
packages-push
packages-test
tests-run

sanych-sun: I've reviewed other drivers/repos: all of them are using `verb-noun.sh` notation. I'll update scripts to follow the same naming conventions. Thanks.

docfx_project/.gitignore
```diff
@@ -0,0 +1,2 @@
+.manifest
```
sanych-sun, line 1: As we removed "old" `Docs` folder, should we rename this folder to `Docs` instead of `docfx_project`?

evergreen/evergreen.yml
```diff
@@ -47,7 +47,7 @@ functions:
            fi
```
sanych-sun, line 47: I've applied proper naming to all recently added scripts/tasks. Also I'll create a separate green-build-friday task to implement CI step to call docfx.

docfx_project/.gitignore
JamesKovacs: SGTM

JamesKovacs (changes requested):
Cake includes a [DocFX add-in](https://cakebuild.net/dsl/docfx/). We should keep the `ApiDocs` task and use that add-in. If we decide to replace Cake entirely with shell scripts eventually, we can do it then. But for now we should keep the Cake tasks for consistency.

PR #1209, opened 2023-10-20
CSHARP-4633: Avoiding Errors if Change Stream Events Exceed 16MB

BorisDog:


tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            const int dataLength = 10 * 1024 * 1024;
+
+            var collectionName = DriverTestConfiguration.CollectionNamespace.CollectionName;
+            using var client = DriverTestConfiguration.CreateDisposableClient();
+            var db = client.GetDatabase(DriverTestConfiguration.DatabaseNamespace.DatabaseName);
+            db.DropCollection(collectionName);
+            db.CreateCollection(collectionName, new() { ChangeStreamPreAndPostImagesOptions = new ChangeStreamPreAndPostImagesOptions() { Enabled = true } });
+            var collection = db.GetCollection<BsonDocument>(collectionName);
+            var stageDocument = BsonDocument.Parse("{ \"$changeStreamSplitLargeEvent\": {} }");
```
rstam, line 50: Should we add builder support for this new stage (similar to `IAggregateFluet.ChangeStream` etc...).

rstam (changes requested):
One minor question before I LGTM.

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
BorisDog: Good question, I will look into that

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            var changeStreamCursor = collection.Watch(pipeline, new() { FullDocumentBeforeChange = ChangeStreamFullDocumentBeforeChangeOption.Required });
+
+            var update = Builders<BsonDocument>.Update.Set("value", new string('z', dataLength));
+            collection.UpdateOne(Builders<BsonDocument>.Filter.Empty, update);
+
+            var events = changeStreamCursor.ToEnumerable().Take(2).ToArray();
+            events[0].SplitEvent.Fragment.Should().Be(1);
+            events[0].SplitEvent.Of.Should().Be(2);
+            events[1].SplitEvent.Fragment.Should().Be(2);
+            events[1].SplitEvent.Of.Should().Be(2);
```
rstam, line 64: Why are we seeing only 2 fragments instead of 3?

Can we expand this integration test to show how to consume all the information from all the fragments?

rstam (changes requested):

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
rstam: @JamesKovacs I'm starting to think the API in this PR is not sufficient for an application to handle change stream events that have been fragmented.

Consider: when an application receives fragment 1 it can't really handle it because it is missing the information contained in the following fragments. And the following fragments are meaningless by themselves and only have meaning when all fragments are considered together.

These fragmented events started out as ONE event on the server and only became multiple documents on the wire due to wire limitations on document size. Logically they continue to represent a single event, and the application really should combine all fragments back into a single logical event before it handles it.

I think at a minimum we should write a sample application that shows how the API could be used to successfully handle fragmented events. That exercise will either prove that the current API is sufficient or will expose what functionality is missing.

BorisDog: @JamesKovacs I suggest to consider and brainstorm the additional API for fragments combination in a separate ticket.
This PR can just expose the `SplitEvent` field as spec prescribes, and stay aligned with other drivers. This "fragmented" API is needed regardless of the decision to introduce additional "combining" API.
This will also allow us to gather users feedback for possible fragment combination from other drivers, before committing to certain "combination" API.

For the sample application, we can just add it to our documentation (docs ticket) and/or extend the current [server sample](https://www.mongodb.com/docs/manual/reference/operator/aggregation/changeStreamSplitLargeEvent/#example)

rstam: I leave it as an open question whether we want to refactor this unit test to illustrate how a user might write code to reassemble the fragments.

I would think that rewriting this test as above would still be fully in the spirit of the spec, and it would be an example we could link to if users ask us how to reassemble the complete event from the fragments.

rstam: Given the above POC that there is a decent way for users to reassemble the complete event from the fragments I am willing to LGTM this PR essentially as is (there are a few requested changes pending).

rstam (changes requested):

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
JamesKovacs: I looked into reassembling the fragments in our drivers, but that poses its own problems. For example if you reassemble a change stream event, it might exceed 16MB and thus cannot be stored in MongoDB any longer. As well if the server crashes after sending fragment 1 but before sending subsequent fragments, you could potentially block until the read times out. Each fragment also has its own resume token, which allows you to resume after any fragment.

Via Slack, @rstam asked me what guarantees the server makes about the splitEvent. Here's what I know...

The server impl will pack the 1st through Nth fields that fit within 16MB into the first fragment. It will then pack the N+1 through N+M fields into the next 16MB. Lather, rinse, repeat until you’re out of fields.

I don’t think they guarantee that the fields will be packed linearly, but that’s the current impl. They could in theory switch to a best-fit algorithm in the future.

Also annoyingly “required” fields are no longer required. A required field will be in one of the fragments, but it won’t be repeated in the others.

The only things that are guaranteed:
- you will receive the fragments in order.
- there will be no other change stream events interspersed with the fragments.
- you can resume a fragment. (e.g. If you received fragment 1 and 2 out of a 3 fragment event, you can resume after fragment 2 and you will be guaranteed to get fragment 3 next.)

@rstam's impl of fragment merging would be useful to include in our tests and/or documentation as one possible way to do it. I thought of implementing a `MergedBsonDocument` that would be backed by multiple `BsonDocument` instances to avoid the copying overhead, but the large fields would be reference copies anyways. So the current impl is probably sufficient.

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -292,6 +293,11 @@ public class Feature
         /// </summary>
         public static Feature ChangeStreamPrePostImages => __changeStreamPrePostImages;
 
+        /// <summary>
+        /// Gets the change stream splitEvent feature.
+        /// </summary>
+        public static Feature ChangeStreamSplitEvent => __changeStreamSplitEvent;
```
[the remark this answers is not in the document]
BorisDog: Done.

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
BorisDog: Thanks for the additional clarification @JamesKovacs.
I love the idea of `MergedBsonDocument`, we could also think of additional elegant ideas that won't require copying the large data chunks.

In the prose test, I think we should have the `splitEvents` objects validation as described by spec.
We could add events combination sample to this test by copying the BackingDocument, but I don't see a value for it to reside in our testing suite, as this does not really test any functionality, but more of an example.
Also I am hesitant to provide users guidance to copy over large `BackingDocument`s, without proper documentation.

I would suggest putting an effort towards working with docs team and documenting `splitEvent` API, with C# samples demonstrating both ways to consume the split events (merging and not merging), describing pros and cons of each approach.

tests/MongoDB.Driver.Core.Tests/ChangeStreamSplitEventSerializerTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using MongoDB.Bson.Serialization;
+using Xunit;
+
+namespace MongoDB.Driver
+{
+    public class ChangeStreamSplitEventSerializerTests
+    {
+        [Theory]
+        [InlineData("{ fragment : 1, of: 2 }", 1, 2)]
+        [InlineData("{ fragment : 1, of: 2 }", 1, 2)]
```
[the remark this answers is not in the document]
BorisDog: Good catch, changed, thanks.

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -51,6 +51,7 @@ public class Feature
         private static readonly Feature __changeStreamForDatabase = new Feature("ChangeStreamForDatabase", WireVersion.Server40);
         private static readonly Feature __changeStreamPostBatchResumeToken = new Feature("ChangeStreamPostBatchResumeToken", WireVersion.Server40);
         private static readonly Feature __changeStreamPrePostImages = new Feature("ChangeStreamPrePostImages", WireVersion.Server60);
+        private static readonly Feature __changeStreamSplitEventStage = new Feature("ChangeStreamSplitEvent", WireVersion.Server70);
```
rstam, line 54: Change the value of the `name` argument to the `Feature` constructor to `"ChangeStreamSplitEventStage"` to match the feature name.

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
rstam: What did you decide?

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
rstam: We can leave the test as is.

> I love the idea of MergedBsonDocument, we could also think of additional elegant ideas that won't require copying the large data chunks.

I don't like this idea. And it's moot at this point anyway since we aren't going to provide any API to help users consume multiple fragments.

Also, the POC I pasted above does NOT involve copying large data chunks. It merely copies references to the large data from additional fragments to the primary underlying base document (so yes, a few bytes will be allocated to add a new element to the underlying base document, but the large data is NOT copied).

rstam (changes requested):

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
BorisDog: Done.

tests/MongoDB.Driver.Tests/Specifications/change-streams/prose-tests/ChangeStreamProseTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            var collectionName = DriverTestConfiguration.CollectionNamespace.CollectionName;
+            using var client = DriverTestConfiguration.CreateDisposableClient();
+            var db = client.GetDatabase(DriverTestConfiguration.DatabaseNamespace.DatabaseName);
+            db.DropCollection(collectionName);
+            db.CreateCollection(collectionName, new() { ChangeStreamPreAndPostImagesOptions = new ChangeStreamPreAndPostImagesOptions() { Enabled = true } });
+            var collection = db.GetCollection<BsonDocument>(collectionName);
+
+            collection.InsertOne(new BsonDocument("value", new string('q', dataLength)));
+
+            var pipeline = new EmptyPipelineDefinition<ChangeStreamDocument<BsonDocument>>().ChangeStreamSplitLargeEvent();
```
BorisDog, line 53: Added a builder for `ChangeStreamSplitLargeEvent` cc @rstam.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -150,6 +150,18 @@ public static class IAggregateFluentExtensions
             return aggregate.AppendStage(PipelineStageDefinitionBuilder.BucketAutoForLinq2(groupBy, buckets, output, options));
         }
 
+        /// <summary>
+        ///Appends a $changeStreamSplitLargeEvent stage to the pipeline.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result.</typeparam>
+        /// <param name="aggregate">The aggregate.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        public static IAggregateFluent<ChangeStreamDocument<TResult>> ChangeStreamSplitLargeEvent<TResult>(this IAggregateFluent<ChangeStreamDocument<TResult>> aggregate)
```
rstam, line 159: Shouldn't this be defined in `IAggregateFluent.cs` and implemented in `AggregateFluent.cs` the same as the similar `ChangeStream` method?

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -150,6 +150,18 @@ public static class IAggregateFluentExtensions
             return aggregate.AppendStage(PipelineStageDefinitionBuilder.BucketAutoForLinq2(groupBy, buckets, output, options));
         }
 
+        /// <summary>
+        ///Appends a $changeStreamSplitLargeEvent stage to the pipeline.
```
rstam, line 154: Missing blank between `///` and `Appends`.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -300,6 +300,18 @@ public static class PipelineDefinitionBuilder
             return pipeline.AppendStage(PipelineStageDefinitionBuilder.ChangeStream<TIntermediate>(options));
         }
 
+        /// <summary>
+        /// Appends the $changeStreamSplitLargeEvent stage.
```
rstam, line 304: Slight rewording suggested: `a` instead of `the`:

    /// Appends a $changeStreamSplitLargeEvent stage.

BorisDog: Done.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
BorisDog: Done.

src/MongoDB.Driver/IAggregateFluentExtensions.cs
BorisDog: The intention is to enable adding `ChangeStreamSplitLargeEvent` stage to `IAggregateFluent<ChangeStreamDocument<TResult>>`  and not to `IAggregateFluent<TResult>`.

rstam: Oh I see. That's a great ides!

rstam (approved):
LGTM

PR #1211, opened 2023-10-27
CSHARP-4821: LINQ3 Where with enum filter followed by Select with cast throws System.ArgumentException.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConvertExpressionToAggregationExpressionTranslator.cs
```diff
@@ -88,21 +88,21 @@ public static AggregationExpression Translate(TranslationContext context, UnaryE
...
+                        case "System.Decimal": to = "decimal"; serializer = DecimalSerializer.Decimal128Instance; break; // not the default representation
+                        case "System.Double": to = "double"; serializer = DoubleSerializer.Instance; break;
+                        case "System.Int32": to = "int"; serializer = Int32Serializer.Instance; break;
+                        case "System.Int64": to = "long"; serializer = Int64Serializer.Instance; break;
+                        case "System.String": to = "string"; serializer = StringSerializer.Instance; break;
+                        default: throw new ExpressionNotSupportedException(expression, because: $"conversion to {expressionType} is not supported");
+                    }
 
                     ast = AstExpression.Convert(ast, to);
-                    serializer = context.KnownSerializersRegistry.GetSerializer(expression);
```
rstam, line 105: There is some bug in the known serializers registry that is returning the wrong serializer here.

In this particular case the set of types being converted to is small and consists of only primitive .NET types and we can map directly to the required serializers without using the known serializers registry.

sanych-sun: What if someone registered their custom serializer for the type? Should we respect it?

rstam: I don't think so. The result of the conversion on the server is a known type, and we want to deserialize it using the standard serializer for that type, regardless of any custom serializers the user might have registered.

By the way, anyone who registers a custom serializer for `Int32` (or any other primitive type) is shooting themselves in the foot big time.

sanych-sun (approved):
LGTM

PR #1214, opened 2023-11-15
CSHARP-4843: Suppress any exceptions in the code checking if legacy driver was loaded

sanych-sun:


src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -88,14 +88,18 @@ internal static BsonDocument CreateDriverDocument(string driverVersion)
...
-            }
-
-            bool IsEFLoaded()
-            {
-                return Type.GetType("MongoDB.EntityFrameworkCore.Query.MongoQueryContext, MongoDB.EntityFrameworkCore") != null;
+                try
+                {
+                    return Type.GetType(typeName) != null;
+                }
+                // ReSharper disable once EmptyGeneralCatchClause
```
BorisDog, line 97: Would have been great if there was another way to disable the warning.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -88,14 +88,18 @@ internal static BsonDocument CreateDriverDocument(string driverVersion)
...
+                catch (Exception)
```
rstam, line 98: why not just `catch`?

    catch
    {
        retrun false
    }

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
sanych-sun: Removed

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -88,14 +88,18 @@ internal static BsonDocument CreateDriverDocument(string driverVersion)
...
             {
-                return Type.GetType("MongoDB.Driver.MongoServer, MongoDB.Driver.Legacy") != null;
-            }
-
-            bool IsEFLoaded()
-            {
-                return Type.GetType("MongoDB.EntityFrameworkCore.Query.MongoQueryContext, MongoDB.EntityFrameworkCore") != null;
+                try
+                {
+                    return Type.GetType(typeName) != null;
```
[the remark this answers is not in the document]
sanych-sun: Documentation does say that, but as it was reported for some condition it could throw. Also as the results of mine investigation there is a way to end up with exception here. For example if you subscribed to `AppDomain.AssemblyResolve` event and throw in the handler - the exception would be re-throwed out of `Type.GetType`. So we still have to suppress anything here, as we really do not need the type, we just trying to check if it is available.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
sanych-sun: Done.

BorisDog (approved):
LGTM

PR #1218, opened 2023-11-22
CSHARP-4854: Suppress vulnerability warnings for netcoreapp2.1 TF in tests

BorisDog:


tests/BuildProps/Tests.Build.props
```diff
@@ -47,4 +47,11 @@
     <PackageReference Include="xunit" Version="2.4.2" />
     <PackageReference Include="JunitXml.TestLogger" Version="2.1.81" />
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="Microsoft.NETCore.App">
+      <NoWarn>NU1903,NU1902</NoWarn> <!--Temporary disable vulnerability warning to allow netcoreapp2.1 TF-->
```
adelinowona, line 53: Why don't we suppress all the warning codes?

adelinowona (approved):
LGTM

tests/BuildProps/Tests.Build.props
BorisDog: Only those turned out to be problematic so far. We can suppress additional ones if needed.

PR #1222, opened 2023-11-24
EF-73: Add more EF Core metadata to Client Metadata Handshake

sanych-sun:


src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -372,19 +373,38 @@ internal static BsonDocument RemoveOptionalFieldsUntilDocumentIsLessThan512Bytes
...
             {
                 // ignore any exceptions here.
+                type = null;
                 return false;
             }
         }
+
+        private static string GetAssemblyVersion(Assembly assembly)
+        {
+            var versionAttribute = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
```
sanych-sun, line 393: Try to read the `AssemblyInformationalVersionAttribute` insted of `AssemblyFileVersionAttribute`, because it contains version suffixes (like a -dev or -preview.1), when `AssemblyFileVersionAttribute` does not.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -372,19 +373,38 @@ internal static BsonDocument RemoveOptionalFieldsUntilDocumentIsLessThan512Bytes
...
+            if (versionAttribute == null)
```
sanych-sun, line 394: Fallback to the "previous" logic incase `AssemblyInformationalVersionAttribute` is missed (this is kind of overkill here, but want to make sure we have some version in any case)

BorisDog (changes requested):

BorisDog (approved):
LGTM

PR #1223, opened 2023-11-28
CSHARP-4706: Support for $out to Time-series collections

adelinowona:


tests/MongoDB.Driver.Tests/Jira/CSharp3397Tests.cs
```diff
@@ -52,6 +55,46 @@ public void Aggregate_out_to_collection_should_work()
 
             results.Single().Count.Should().Be(1);
         }
+
+        [Fact]
+        public void Aggregate_out_to_time_series_collection_should_work()
```
BorisDog, line 60: I am not sure we need such test for time-series here. Probably `Execute_to_time_series_collection_should_work` is sufficient. cc @JamesKovacs

src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -303,6 +303,24 @@ public interface IAggregateFluent<TResult> : IAsyncCursorSource<TResult>
         /// <returns>A cursor.</returns>
         IAsyncCursor<TResult> Out(string collectionName, CancellationToken cancellationToken = default(CancellationToken));
 
+        /// <summary>
+        /// Appends an out stage to the pipeline and executes it, and then returns a cursor to read the contents of the output collection.
+        /// </summary>
+        /// <param name="outputCollection">The output collection.</param>
+        /// <param name="timeSeriesOptions">The time series options</param>
```
BorisDog, line 310: Period at the end (here and elsewhere).

tests/MongoDB.Driver.Core.Tests/Core/Operations/AggregateToCollectionOperationTests.cs
```diff
@@ -607,6 +607,36 @@ public void CreateCommand_should_return_expected_result_when_MaxTime_is_set(long
             result.Should().HaveCount(1);
         }
 
+        [Theory]
```
BorisDog, line 610: Lets also add tests in `PipelineDefinitionBuilderTests` and in `PipelineStageDefinitionBuilderTests` (no need to cover all Out cases, just for time-series.).

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
```diff
@@ -330,7 +330,7 @@ private IReadOnlyList<BsonDocument> SimplifyOutStageIfOutputDatabaseIsSameAsInpu
         {
             var lastStage = pipeline.Last();
             var lastStageName = lastStage.GetElement(0).Name;
-            if (lastStageName == "$out" && lastStage["$out"] is BsonDocument outDocument)
+            if (lastStageName == "$out" && lastStage["$out"] is BsonDocument outDocument && !outDocument.Contains("timeseries"))
```
BorisDog, line 333: Do we have this tested directly or indirectly somewhere?

BorisDog (changes requested):
Looks great, some comments about testing.

tests/MongoDB.Driver.Tests/Jira/CSharp3397Tests.cs
adelinowona: I added this test here since server added support of $out to time series on secondaries. The original tests in CSharp3397Tests.cs test for support of $merge and $out executing on secondaries (https://jira.mongodb.org/browse/CSHARP-3397) so I thought it would be appropriate to add my test here. Also I should probably rename the test to `Aggregate_out_to_time_series_collection_on_secondary_should_work`

src/MongoDB.Driver.Core/Core/Operations/AggregateToCollectionOperation.cs
adelinowona: This is tested indirectly through the `Aggregate_out_to_time_series_collection_on_secondary_should_work` test and the AggregateToCollection time series tests. Without this, the simplification of the out stage will just remove the timeseries options resulting in the time series tests failing.

tests/MongoDB.Driver.Tests/PipelineDefinitionBuilderTests.cs
```diff
@@ -141,6 +141,20 @@ public void Merge_should_add_expected_stage()
             stages[0].Should().Be("{ $merge : { into : { db : 'database', coll : 'collection' } } }");
         }
 
+        [Fact]
+        public void Out_with_time_series_options_should_add_expected_stage()
+        {
+            var client = DriverTestConfiguration.Client;
```
BorisDog, line 147: Looks like we just need the collection namespace for rendering.
Maybe we can just mock the collection, and then the test would not be dependent on server?

tests/MongoDB.Driver.Tests/PipelineStageDefinitionBuilderTests.cs
```diff
@@ -391,6 +391,18 @@ public void Merge_with_WhenNotMatched_should_return_the_expected_result(MergeSta
...
+        [Fact]
+        public void Out_with_time_series_options_should_return_expected_result()
+        {
+            var client = DriverTestConfiguration.Client;
+            var outputCollection = client.GetDatabase("database").GetCollection<BsonDocument>("collection");
+            var timeSeriesOptions = new TimeSeriesOptions("time", "symbol");
+
+            var result = PipelineStageDefinitionBuilder.Out(outputCollection, timeSeriesOptions);
+
+            RenderStage(result).Document.Should().Be("{ $out: { db: 'database', coll: 'collection', timeseries: { timeField: 'time', metaField: 'symbol' } } }");
```
BorisDog, line 403: Same comment about mocking collection.

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -40,7 +40,9 @@ public class Feature
         private static readonly Feature __aggregateLet = new Feature("AggregateLet", WireVersion.Server36);
         private static readonly Feature __aggregateMerge = new Feature("AggregateMerge", WireVersion.Server42);
         private static readonly Feature __aggregateOut = new Feature("AggregateOut", WireVersion.Server26);
+        private static readonly Feature __aggregateOutTimeSeries = new Feature("AggregateOutTimeSeries", WireVersion.Server70);
         private static readonly Feature __aggregateOutOnSecondary = new Feature("AggregateOutOnSecondary", WireVersion.Server50);
+        private static readonly Feature __aggregateOutTimeSeriesOnSecondary = new Feature("AggregateOutTimeSeriesOnSecondary", WireVersion.Server70);
```
BorisDog, line 45: Not sure that we need the secondary feature.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -970,6 +970,25 @@ public static class PipelineDefinitionBuilder
...
+        /// <param name="outputCollection">The output collection.</param>
+        /// <param name="timeSeriesOptions">The time-series options</param>
+        /// <returns>A new pipeline with an additional stage.</returns>
+        /// <exception cref="System.NotSupportedException"></exception>
+        public static PipelineDefinition<TInput, TOutput> Out<TInput, TOutput>(
+            this PipelineDefinition<TInput, TOutput> pipeline,
+            IMongoCollection<TOutput> outputCollection,
+            TimeSeriesOptions timeSeriesOptions)
+        {
+            Ensure.IsNotNull(pipeline, nameof(pipeline));
```
BorisDog, line 988: We should either merge overloads (add `timeSeriesOptions = null` parameter) or forbid null here.
I personally would prefer having less overloads for simplicity.

BorisDog (changes requested):

src/MongoDB.Driver.Core/Core/Misc/Feature.cs
```diff
@@ -40,6 +40,7 @@ public class Feature
         private static readonly Feature __aggregateLet = new Feature("AggregateLet", WireVersion.Server36);
         private static readonly Feature __aggregateMerge = new Feature("AggregateMerge", WireVersion.Server42);
         private static readonly Feature __aggregateOut = new Feature("AggregateOut", WireVersion.Server26);
+        private static readonly Feature __aggregateOutTimeSeries = new Feature("AggregateOutTimeSeries", WireVersion.Server70);
```
BorisDog, line 43: Alphabetical order (field and property).

BorisDog (changes requested):

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
JamesKovacs: I see that we are already doing this elsewhere in our Fluent Aggregate API. Sigh. I guess we can use `TimeSeriesOptions timeSeriesOptions = null` here to reduce the number of overloads.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1244,17 +1244,35 @@ public static class PipelineStageDefinitionBuilder
...
             return new BsonDocumentPipelineStageDefinition<TInput, TInput>(new BsonDocument("$out", outDocument));
         }
 
+        /// <summary>
+        /// Creates a $out stage.
+        /// </summary>
+        /// <typeparam name="TInput">The type of the input documents.</typeparam>
+        /// <param name="outputCollection">The output collection.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Out<TInput>(
```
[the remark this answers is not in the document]
JamesKovacs: It is a minor breaking change (and I don't like it for the reasons stated above), but we are already doing it elsewhere in our public API. We can do it here too to reduce the number of overloads. We will release this in a new version, not a patch build. So users should recompile their apps and not notice the breakage as the C# compiler will do the right thing.

JamesKovacs (changes requested):
While I don't like optional parameters (see comment below for why), we use them elsewhere. I'm fine with using an optional parameter for `timeSeriesOptions` and not adding more overloads. Fix that and some minor comment issues and you're good to go.

BorisDog (approved):
LGTM!

PR #1224, opened 2023-11-28
CSHARP-4858: Add IsMissing and IsNullOrMissing MongoDBFunctions.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class CSharp4858Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Select_IsMissing_should_work()
+        {
+            var collection = GetCollection();
+
+            var queryable = collection.AsQueryable()
+                .Select(x => MongoDBFunctions.IsMissing(x.S));
```
rstam, line 32: Damien thinks `MongoDBFunctions` should not be a static class and that `IsMissing` should be an extension method instead.

To the user the difference would be that they would write

    MongoDB.Functions.IsMissing(x.S)

Note the period between `MongoDB` and `Functions`.

@JamesKovacs let's discuss when you get back.

rstam: Damien would also like to see the LINQ provider support

    EF.Functions.IsMissing(x.S)

even though the C# driver knows nothing about the type `DbFunctions` (that's the type of the `EF.Functions` static property) or anything about any extension methods defined on `DbFunctions`.

The question from the perspective of the C# driver is whether we want to start writing EF Core provider specific code (so far we have avoided doing so).

Ideally we would avoid having EF Core specific code in the C# driver. From a pragmatic viewpoint we might be willing to do some of that.

rstam: Besides the question of whether this class should be static or not we should also think hard about naming.

We currently have a few places custom LINQ related methods live:

* Various extensions methods on existing .NET classes (e.g. DateTime, String, NullableDateTime)
* MongoDBMath
* LinqExtensions
* MongoEnumerable (similar to .NET Enumerable)
* MongoQueryable (similar to .NET Queryable)

MongoDBFunctions is named similarly to `EF.Functions`, but I don't think we need to follow EF Core naming conventions in the C# driver. Maybe a simpler name would be possible.

Hopefully we don't end up with too many new places for custom LINQ methods to live.

BorisDog: Additional point to consider:
As we are planning to support trimming in [CSHARP-4840](https://jira.mongodb.org/browse/CSHARP-4840), it would be great to ensure that at least all new code supports trimming.
In this PR the reflection is an immediate suspect for trimming compatibility issues. If it can be avoided that would be preferable.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/MongoDBFunctionsMethod.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        // private static fields
+        private static readonly MethodInfo __isMissing;
+        private static readonly MethodInfo __isNullOrMissing;
+
+        // static constructor
+        static MongoDBFunctionsMethod()
+        {
+            __isMissing = ReflectionInfo.Method((object field) => MongoDBFunctions.IsMissing(field));
+            __isNullOrMissing = ReflectionInfo.Method((object field) => MongoDBFunctions.IsNullOrMissing(field));
```
rstam, line 30: Don't these two static references to `IsMissing` and `IsNullOrMissing` avoid any trimming concerns?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
BorisDog: Usually reflection is problematic, as the trimming process does not know of dynamically loaded types, and might trim them.
In this case we just need to make sure that there are no issues around usage of `MongoDBFunctionsMethod` in the trimmed mode.
Testing this in not trivial, as our driver can't even load in trimmed mode, so if you choose to research this issue, probably the best way to do so is via sample app.
Or just avoid reflection if possible.

rstam: Thanks for checking that.

JamesKovacs: What is the problem with using `MongoDbFunctions.IsNullOrMissing` in EF? It is a MongoDB-specific extension that other providers won't understand.

While we could write it as an extension method, we end up with `field.IsNullOrMissing()` - which while possible would suggest we are calling a method on a potentially null value. Not a huge deal as it is an extension method. No different than writing `someString.IsNullOrEmpty()`, which is simply defined as:

```
public class StringExtensions
{
    public static bool IsNullOrEmpty(this string str)
    {
        return String.IsNullOrEmpty(str);
    }
}
```

Totally valid C# but still quirky as it looks like you're calling a method on a potentially `null` value but not throwing.

Is there an existing `EF.Functions.IsMissing`? Why wouldn't we simply use `MongoDbFunctions.IsMissing` directly?

The larger question of whether we should merge the various extensions is a 3.0 driver question. It would be nice to simply `using MongoDB.Driver.Extentions` (or whatever we call it) and have them all available. But that sounds like a task for another ticket targeted at the 3.0 driver.

JamesKovacs (approved):
LGTM. See my inline comment for my discussion points.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
sanych-sun: I would vote for implementing this as extensions methods ([empty class](https://github.com/dotnet/efcore/blob/e84e5add7a1046c7248515884fcd321cc03678ef/src/EFCore/DbFunctions.cs) that exists only to be extended). Because if we would ever came to the point when our LINQ provider would be extendable from outside it would be nice extension point to inject 3rd-party functions. And they would be available same way as built-in.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class CSharp4858Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Select_IsMissing_should_work()
+        {
+            var collection = GetCollection();
+
+            var queryable = collection.AsQueryable()
+                .Select(x => MongoDBLinq.Functions.IsMissing(x.S));
```
rstam, line 32: I first thought of using `MongoDB.Functions` (patterned after `EF.Functions`) but we can't have a class named `MongoDB` because it conflicts with the root of our namespaces which is also `MongoDB`.

I suppose we could consider `MDB.Functions` but we normally avoid abbreviations.

rstam:
The latest commit explores Alex's suggestion to define `IsMissing` and `IsNullOrMissing` as extension methods against a dummy `MongoDBFunctions` class. This idea comes from how EF Core does something similar.

This is just a prototype of such an approach. I need you to review the approach and the naming even more than reviewing the code itself. If we use this approach we are committing to it for future methods.

rstam:
The latest commit adds an Exists method to test for the existence of a field.

Using `Exists` is more convenient than `!IsMissing`.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
damieng: Two comments:

**1. Trimming** - Given these are going to be static methods (either extension or regular) I don't think they're going to cause a problem for trimming. While reflection has limitations I haven't seen anything that says the compiler is incapable of identifying references inside expression trees that the compiler builds during compilation.

**2. Naming** - Let's go with whatever works best for the LINQ provider for now. EF doesn't *require* them to be in `EF.Functions` so let's see if it causes confusion or not. Worse case at some later point we duplicate the static class/stubs and either translate it from one to the other in the EF provider or recognize both in the V3 provider.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
JamesKovacs: `MongoDBLinq.Functions.IsMissing` (and friends) isn't readily discoverable. Make sure you mark this as requiring documentation. I would have said the same thing about `MongoDBFunctions`. It's not the name change. It's the fact that it won't appear readily in Intellisense.

JamesKovacs (approved):
LGTM

JamesKovacs (approved):
Renaming LGTM.

BorisDog (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,66 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators.MethodTranslators
+{
+    internal static class FieldExistsOrIsMissingMethodToAggregationExpressionTranslator
```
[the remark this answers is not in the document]
rstam: We name these classes after the method(s) they translate.

We wouldn't want to group ALL of them into a single `MqlMethodsTranslator`.

We could separate translation of `Exists` and `IsMissing` but they have so much in common that there is good opportunity to share code here.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                var ast = method switch
+                {
+                    _ when method.Is(MqlMethod.Exists) => AstExpression.Ne(AstExpression.Type(fieldAst), "missing"),
+                    _ when method.Is(MqlMethod.IsMissing) => AstExpression.Eq(AstExpression.Type(fieldAst), "missing"),
+                    _ when method.Is(MqlMethod.IsNullOrMissing) => AstExpression.In(AstExpression.Type(fieldAst), new BsonArray { "null", "missing" }),
+                    _ => throw new InvalidOperationException("Unexpected method."),
+                }; ;
+
+                return new AggregationExpression(expression, ast, BooleanSerializer.Instance);
```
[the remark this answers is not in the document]
rstam: We haven't worried to much about this in other translators.

The initial test is to ensure it's one of the methods that this translator knows how to handle.

In other translators that would be it, but in this one there are slight variations depending on which method is being translated.

I will refactor a slightly different way.

src/MongoDB.Driver/Linq/Mql.cs
```diff
@@ -18,18 +18,17 @@
 namespace MongoDB.Driver.Linq
 {
     /// <summary>
-    /// Contains MongoDBLinqFunctions extension methods that can be used in LINQ queries against MongoDB.
+    /// Contains methods that can be used to access MongoDB specific functionality in LINQ queries.
     /// </summary>
-    public static class MongoDBLinqFunctionsExtensions
+    public static class Mql
```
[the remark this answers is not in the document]
rstam: Originally I would have said yes to putting this in the `Linq` namespace.

This was discussed and others felt it shouldn't be in the `Linq` namespace because you can use this with builders (and the LINQ machinery is incidental).

What do others think about which namespace this should be in?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4858Tests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using System.Linq;
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp4858Tests : Linq3IntegrationTest
```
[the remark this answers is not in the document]
rstam: Tests moved to `FieldExistsOrIsMissingMethodToAggregationExpressionTranslatorTests.cs`

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,66 @@
...
+                if (fieldAst.NodeType != AstNodeType.GetFieldExpression)
+                {
+                    throw new ExpressionNotSupportedException(expression, because: $"argument to {method.Name} must be a reference to a field");
+                }
+
+                var ast = method.Name switch
+                {
+                    nameof(Mql.Exists) => AstExpression.Ne(AstExpression.Type(fieldAst), "missing"),
+                    nameof(Mql.IsMissing) => AstExpression.Eq(AstExpression.Type(fieldAst), "missing"),
+                    nameof(Mql.IsNullOrMissing) => AstExpression.In(AstExpression.Type(fieldAst), new BsonArray { "null", "missing" }),
```
rstam, line 56: It's safe to use string comparisons here because line 42 verified that it REALLY is one of the supported methods and not just one that is named alike.

src/MongoDB.Driver/Linq/Mql.cs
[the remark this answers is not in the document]
rstam: Sorry I misunderstood your question.

It currently IS under the `MongoDB.Driver.Linq` namespace.

Should we move it up one level?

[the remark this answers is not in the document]
damieng: I'd go with bringing it up one level. Even when they are doing LINQ they may well have not brought the MongoDB.Driver.Linq namespace in.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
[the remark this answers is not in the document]
BorisDog: Got it. 
So any new method in MqlMethods, would require to implement a new translator, and could not take advantage of code reuse here? Or we don't expect new methods for the same form?

src/MongoDB.Driver/Linq/Mql.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Tests whether a field is null or missing.
+        /// </summary>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="field">The field.</param>
+        /// <returns>True if the field is null missing.</returns>
+        public static bool IsNullOrMissing<TField>(TField field)
+        {
+            throw new NotSupportedException("This method is not functional. It is only usable in MongoDB LINQ queries.");
```
damieng, line 55: Might be worth having the method name explicitly in the message, e.g.

```csharp
static InvalidOperationException BuildException([CallerMemberName] string memberName = null)
{
   return new InvalidOperationException($"Mql.{memberName} is a method for mapping queries to MongoDB and should not be called directly client-side.");
}
```

EF goes an `InvalidOperationException` with the message:

> The '{methodName}' method is not supported because the query has switched to client-evaluation. This usually happens when the arguments to the method cannot be translated to server. Rewrite the query to avoid client evaluation of arguments so that method can be translated to server.

src/MongoDB.Driver/Linq/Mql.cs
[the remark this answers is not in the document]
BorisDog: It seems that it should be moved up, to be more easily discoverable for Builder users. Unless it can be used only with Linq.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+* limitations under the License.
+*/
+
+using System.Linq;
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
```
BorisDog, line 22: Need to adjust the namespace as well.

BorisDog (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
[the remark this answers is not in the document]
rstam: Sharing a translator for multiple methods is only reasonable when all the methods have something in common. In this case the thing they have in common is that they all do some kind of existence tests on a field.

The usual case is that a new method requires a new translator.

src/MongoDB.Driver/Linq/Mql.cs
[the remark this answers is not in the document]
rstam: Done.

src/MongoDB.Driver/Linq/Mql.cs
rstam: This is a good suggestion. Maybe we need a separate JIRA ticket for it.

This exception was copied verbatim from existing custom methods.

Changing the exception type would normally be considered a breaking change. Changing the message is somewhat less breaking.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslatorTests.cs
rstam: Good catch.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslator.cs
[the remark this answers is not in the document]
BorisDog: Just trying to think what happens when a method that has similar translation will be added to `Mql`, and `FieldExistsOrIsMissingMethodToAggregationExpressionTranslator` would be the most natural place for it.

Which might be the case for many potential methods in Mql, or is it unlikely at this point?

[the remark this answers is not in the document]
rstam: In general when we add new MQL methods they won't have anything in common with these three.

There will be all sorts of combinations of return types, number of arguments, argument types, argument interpretations (i.e. in this case they must idenitfy a field) and all kinds of different translations.

I don't anticipate supporting any new methods in this translator.

src/MongoDB.Driver/Mql.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Contains methods that can be used to access MongoDB specific functionality in LINQ queries.
+    /// </summary>
+    public static class Mql
+    {
+        /// <summary>
+        /// Tests whether a field exists.
+        /// </summary>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="field">The field.</param>
+        /// <returns>True if the field is missing.</returns>
```
BorisDog, line 30: field exists

src/MongoDB.Driver/Mql.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            throw new NotSupportedException("This method is not functional. It is only usable in MongoDB LINQ queries.");
+        }
+
+        /// <summary>
+        /// Tests whether a field is null or missing.
+        /// </summary>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="field">The field.</param>
+        /// <returns>True if the field is null missing.</returns>
```
BorisDog, line 52: null or missing?

src/MongoDB.Driver/Mql.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            throw new NotSupportedException("This method is not functional. It is only usable in MongoDB LINQ queries.");
+        }
+
+        /// <summary>
+        /// Tests whether a field is missing.
+        /// </summary>
+        /// <typeparam name="TField">The type of the field.</typeparam>
+        /// <param name="field">The field.</param>
+        /// <returns>True if the field is missing.</returns>
```
BorisDog, line 41: `<c>true</c>` in all comments?

BorisDog (changes requested):
Few minor comments.

BorisDog (approved):
LGTM

damieng:
LGMT

damieng (approved):

sanych-sun (approved):
LGTM

JamesKovacs (approved):
LGTM

JamesKovacs (changes requested):
Just noticed that the patch is failing on Windows due to a path length issue.
```
[2024/01/19 17:16:15.096] LibGit2Sharp.LibGit2SharpException: path too long: 'C:/data/mci/98efc01bfb8b1056ce0fd6be52c3b4a2/mongo-csharp-driver/tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/FieldExistsOrIsMissingMethodToAggregationExpressionTranslatorTests.cs'
```

PR #1225, opened 2023-11-28
CSHARP-4380: Remove legacy shell from test scripts

sanych-sun:


JamesKovacs (approved):
Changes all look good. Our patch builds don't test AWS auth. Assuming you ran a full patch build that ran the AWS auth variants, then LGTM.

PR #1227, opened 2023-11-30
CSHARP-4861: Count comparison with constant on the left is not translated correctly.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionTranslators/CountComparisonExpressionToFilterTranslator.cs
```diff
@@ -68,30 +68,30 @@ public static bool CanTranslate(Expression leftExpression, Expression rightExpre
...
 
-        public static AstFilter Translate(TranslationContext context, BinaryExpression expression, Expression enumerableExpression, Expression sizeExpression)
+        public static AstFilter Translate(TranslationContext context, BinaryExpression expression, AstComparisonFilterOperator comparisonOperator, Expression enumerableExpression, Expression sizeExpression)
         {
             var field = ExpressionToFilterFieldTranslator.Translate(context, enumerableExpression);
 
             if (TryConvertSizeExpressionToBsonValue(sizeExpression, out var size))
             {
-                switch (expression.NodeType)
+                switch (comparisonOperator)
```
rstam, line 77: When the constant is on the left and the sides are swapped we need to use the comparison operator for the swapped operands, not the original comparison operator.

JamesKovacs: Oh! Good catch!

JamesKovacs (approved):
LGTM

PR #1230, opened 2023-12-05
CSHARP-4863: Refactor AggregateMethodToAggregationExpressionTranslator to produce simpler translations when possible.

rstam:


JamesKovacs (approved):
LGTM. That was quite the mindbender thinking through exactly what was being done and why. :)

PR #1231, opened 2023-12-06
CSHARP-3788: Use OP_MSG to authenticate if server supports OP_MSG

adelinowona:


JamesKovacs (changes requested):
Looking good. Minor feedback and simplifications.

JamesKovacs (approved):
LGTM

PR #1233, opened 2023-12-12
CSHARP-4862: Include TLS1.3 in the list of acceptable TLS protocols.

JamesKovacs:
We don't explicitly test different TLS versions as we rely on the underlying TLS libraries to negotiate the protocol. This commit opts into TLS1.3 as an option. I verified that a `netcoreapp2.1` could still connect to an Atlas cluster with TLS1.2 enabled after adding the enum value for TLS1.3 to the flags. The value of the `SslProtocols.Tls13` flag is from:

https://learn.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols?view=net-8.0

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Configuration/SslStreamSettings.cs
+++ b/src/MongoDB.Driver.Core/Core/Configuration/SslStreamSettings.cs
@@ -53,9 +53,9 @@ namespace MongoDB.Driver.Core.Configuration
             _checkCertificateRevocation = checkCertificateRevocation.WithDefault(false);
             _clientCertificates = Ensure.IsNotNull(clientCertificates.WithDefault(Enumerable.Empty<X509Certificate>()), "clientCertificates").ToList();
             _clientCertificateSelectionCallback = clientCertificateSelectionCallback.WithDefault(null);
-            // SslProtocols.Tls13 not available until netcoreapp3.1 (not in netstandard2.1) and net5.0
-            const SslProtocols SslProtocolsTls13 = (SslProtocols)12288;
-            _enabledSslProtocols = enabledProtocols.WithDefault(SslProtocolsTls13 | SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls);
+            // SslProtocols.Tls13 is not available until netcoreapp3.1 (but not part of netstandard2.1) and net5.0
+            const SslProtocols sslProtocolsTls13 = (SslProtocols)12288;
+            _enabledSslProtocols = enabledProtocols.WithDefault(sslProtocolsTls13 | SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls);
             _serverCertificateValidationCallback = serverCertificateValidationCallback.WithDefault(null);
         }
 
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver.Core/Core/Configuration/SslStreamSettings.cs
```diff
@@ -53,7 +53,9 @@ public class SslStreamSettings
             _checkCertificateRevocation = checkCertificateRevocation.WithDefault(false);
             _clientCertificates = Ensure.IsNotNull(clientCertificates.WithDefault(Enumerable.Empty<X509Certificate>()), "clientCertificates").ToList();
             _clientCertificateSelectionCallback = clientCertificateSelectionCallback.WithDefault(null);
-            _enabledSslProtocols = enabledProtocols.WithDefault(SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls);
+            // SslProtocols.Tls13 not available until netcoreapp3.1 (not in netstandard2.1) and net5.0
+            const SslProtocols SslProtocolsTls13 = (SslProtocols)12288;
+            _enabledSslProtocols = enabledProtocols.WithDefault(SslProtocolsTls13 | SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls);
```
BorisDog, line 58: @JamesKovacs Do you want to have this as opt-in option and not enabled by default in `SslSettings.EnabledSslProtocols` ?

BorisDog (approved):
LGTM

adelinowona (approved):
LGTM

src/MongoDB.Driver.Core/Core/Configuration/SslStreamSettings.cs
JamesKovacs: Great catch. I forgot that we need it in both `SslStreamSettings` (core) and `SslSettings` (high) due to the core/high split.

JamesKovacs:
I don't like scattering the numeric value for `SslProtocols.Tls13` across multiple classes, but because of the core/high split, I would need to make the constant public - which is an even worse choice given that it already exists in `SslProtocols` in newer framework versions. I choose the lesser evil of duplicating the constant value across core/high and tests. If anyone has a better idea, I'm willing to consider it.

BorisDog:
IMHO duplicating the constant is ok for now, as we can handle this in 3.0.
There is also an option of declaring `internal` constant (internals are visible to MongoDB.Driver)

JamesKovacs:
Good point. I thought that we had InternalsVisibleTo between Core/Core.Tests and Driver/Driver.Tests, but not Core/Driver.Tests. I was wrong and we have it between all 3. (Just not Driver/Core.Tests - which makes sense.) Refactored to use an internal constant `SslStreamSettings.SslProtocolsTls13` until we update to `net5.0` as our minimum TFM.

tests/MongoDB.Driver.Tests/SslSettingsTests.cs
```diff
@@ -22,6 +22,7 @@
 using System.Security.Cryptography.X509Certificates;
 using Xunit;
 using FluentAssertions;
+using MongoDB.Driver.Core.Configuration;
```
BorisDog, line 25: minor: sort usings.

BorisDog (approved):
LGTM (+minor comment)

tests/MongoDB.Driver.Tests/SslSettingsTests.cs
JamesKovacs: Done.

PR #1237, opened 2023-12-15
CSHARP-4615: Switch to polling monitoring on FaaS environment.

adelinowona:
- Adds serverMonitoringMode (stream|poll|auto) to URI options and MongoClientOptions
- Updates monitor behaviour to poll or stream based on options or environment.
- Ensures no dedicated connection is used for RTT pinging in polling mode.
- Updates Unified test runner to work with schema 1.17
- Syncs SDAM unified tests
- Syncs URI options tests

### Highlight

For users that want to control the behaviour of the server monitoring connection between each node in the topology, a new option, `serverMonitoringMode`, has been added. This defaults to auto but can be forced into a specific mode by providing a value of poll or stream. A polling monitor periodically issues a hello command to the node at an interval of `heartbeatFrequencyMS`. A streaming monitor sends an initial hello and then will automatically get a response from the Node when a change in server configuration occurs or at a maximum time of `heartbeatFrequencyMS`

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
@@ -27,6 +27,7 @@ namespace MongoDB.Driver.Core.Servers
     internal interface IRoundTripTimeMonitor : IDisposable
     {
         TimeSpan Average { get; }
+        bool Started { get; }
         void AddSample(TimeSpan roundTripTime);
         void Reset();
         void Start();
@@ -48,6 +49,8 @@ namespace MongoDB.Driver.Core.Servers
         private readonly ServerId _serverId;
         private readonly ILogger<RoundTripTimeMonitor> _logger;
 
+        private bool _started;
+
         public RoundTripTimeMonitor(
             IConnectionFactory connectionFactory,
             ServerId serverId,
@@ -63,6 +66,7 @@ namespace MongoDB.Driver.Core.Servers
             _serverApi = serverApi;
             _cancellationTokenSource = new CancellationTokenSource();
             _cancellationToken = _cancellationTokenSource.Token;
+            _started = false;
 
             _logger = logger;
         }
@@ -74,6 +78,17 @@ namespace MongoDB.Driver.Core.Servers
                 lock (_lock)
                 {
                     return _averageRoundTripTimeCalculator.Average;
+                }
+            }
+        }
+
+        public bool Started
+        {
+            get
+            {
+                lock (_lock)
+                {
+                    return _started;
                 }
             }
         }
@@ -99,6 +114,7 @@ namespace MongoDB.Driver.Core.Servers
         {
             _roundTripTimeMonitorThread = new Thread(ThreadStart) { IsBackground = true };
             _roundTripTimeMonitorThread.Start();
+            _started = true;
 
             void ThreadStart()
             {
--- a/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
@@ -27,7 +27,7 @@ namespace MongoDB.Driver.Core.Servers
     internal interface IRoundTripTimeMonitor : IDisposable
     {
         TimeSpan Average { get; }
-        bool Started { get; }
+        bool IsStarted { get; }
         void AddSample(TimeSpan roundTripTime);
         void Reset();
         void Start();
@@ -49,8 +49,6 @@ namespace MongoDB.Driver.Core.Servers
         private readonly ServerId _serverId;
         private readonly ILogger<RoundTripTimeMonitor> _logger;
 
-        private bool _started;
-
         public RoundTripTimeMonitor(
             IConnectionFactory connectionFactory,
             ServerId serverId,
@@ -66,7 +64,6 @@ namespace MongoDB.Driver.Core.Servers
             _serverApi = serverApi;
             _cancellationTokenSource = new CancellationTokenSource();
             _cancellationToken = _cancellationTokenSource.Token;
-            _started = false;
 
             _logger = logger;
         }
@@ -82,16 +79,7 @@ namespace MongoDB.Driver.Core.Servers
             }
         }
 
-        public bool Started
-        {
-            get
-            {
-                lock (_lock)
-                {
-                    return _started;
-                }
-            }
-        }
+        public bool IsStarted => _roundTripTimeMonitorThread != null;
 
         // public methods
         public void Dispose()
@@ -114,7 +102,6 @@ namespace MongoDB.Driver.Core.Servers
         {
             _roundTripTimeMonitorThread = new Thread(ThreadStart) { IsBackground = true };
             _roundTripTimeMonitorThread.Start();
-            _started = true;
 
             void ThreadStart()
             {
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -36,6 +36,7 @@ namespace MongoDB.Driver.Core.Servers
         private ServerDescription _currentDescription;
         private readonly EndPoint _endPoint;
         private HeartbeatDelay _heartbeatDelay;
+        private readonly bool _isStreamingEnabled;
         private readonly object _lock = new object();
         private readonly EventLogger<LogCategories.SDAM> _eventLoggerSdam;
         private readonly ILogger<IServerMonitor> _logger;
@@ -48,7 +49,6 @@ namespace MongoDB.Driver.Core.Servers
         private readonly ServerMonitorSettings _serverMonitorSettings;
 
         private Thread _serverMonitorThread;
-        private bool _streamingEnabled;
 
         public event EventHandler<ServerDescriptionChangedEventArgs> DescriptionChanged;
 
@@ -107,7 +107,7 @@ namespace MongoDB.Driver.Core.Servers
             _logger = loggerFactory?.CreateLogger<IServerMonitor>();
             _eventLoggerSdam = loggerFactory.CreateEventLogger<LogCategories.SDAM>(eventSubscriber);
 
-            _streamingEnabled = serverMonitorSettings.ServerMonitoringMode switch
+            _isStreamingEnabled = serverMonitorSettings.ServerMonitoringMode switch
             {
                 ServerMonitoringMode.Stream => true,
                 ServerMonitoringMode.Poll => false,
@@ -189,7 +189,7 @@ namespace MongoDB.Driver.Core.Servers
         {
             BsonDocument helloCommand;
             var commandResponseHandling = CommandResponseHandling.Return;
-            if (_streamingEnabled && connection.Description.HelloResult.TopologyVersion != null)
+            if (_isStreamingEnabled && connection.Description.HelloResult.TopologyVersion != null)
             {
                 connection.SetReadTimeout(_serverMonitorSettings.ConnectTimeout + _serverMonitorSettings.HeartbeatInterval);
                 commandResponseHandling = CommandResponseHandling.ExhaustAllowed;
@@ -425,12 +425,12 @@ namespace MongoDB.Driver.Core.Servers
                 var transitionedWithNetworkError =
                     (IsNetworkError(heartbeatException) && previousDescription.Type != ServerType.Unknown);
 
-                if (_streamingEnabled && serverSupportsStreaming && !_roundTripTimeMonitor.Started)
+                if (_isStreamingEnabled && serverSupportsStreaming && !_roundTripTimeMonitor.Started)
                 {
                     _roundTripTimeMonitor.Start(); // start RTT monitoring on separate thread
                 }
 
-                processAnother = (_streamingEnabled && (serverSupportsStreaming || connectionIsStreaming)) || transitionedWithNetworkError;
+                processAnother = (_isStreamingEnabled && (serverSupportsStreaming || connectionIsStreaming)) || transitionedWithNetworkError;
             }
 
             bool IsNetworkError(Exception ex)
@@ -446,7 +446,7 @@ namespace MongoDB.Driver.Core.Servers
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            _eventLoggerSdam.LogAndPublish(new ServerHeartbeatStartedEvent(connection.ConnectionId, _streamingEnabled && connection.Description.HelloResult.TopologyVersion != null));
+            _eventLoggerSdam.LogAndPublish(new ServerHeartbeatStartedEvent(connection.ConnectionId, _isStreamingEnabled && connection.Description.HelloResult.TopologyVersion != null));
 
             var stopwatch = Stopwatch.StartNew();
             try
@@ -455,19 +455,19 @@ namespace MongoDB.Driver.Core.Servers
                 stopwatch.Stop();
 
                 // RTT check if using polling monitoring
-                if ((!_streamingEnabled || (_streamingEnabled && connection.Description.HelloResult.TopologyVersion == null)) && helloResult != null)
+                if ((!_isStreamingEnabled || (_isStreamingEnabled && connection.Description.HelloResult.TopologyVersion == null)) && helloResult != null)
                 {
                     _roundTripTimeMonitor.AddSample(stopwatch.Elapsed);
                 }
 
-                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatSucceededEvent(connection.ConnectionId, stopwatch.Elapsed, _streamingEnabled && connection.Description.HelloResult.TopologyVersion != null, helloResult.Wrapped));
+                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatSucceededEvent(connection.ConnectionId, stopwatch.Elapsed, _isStreamingEnabled && connection.Description.HelloResult.TopologyVersion != null, helloResult.Wrapped));
 
                 return helloResult;
             }
             catch (Exception ex)
             {
                 stopwatch.Stop();
-                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatFailedEvent(connection.ConnectionId, stopwatch.Elapsed, ex, _streamingEnabled && connection.Description.HelloResult.TopologyVersion != null));
+                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatFailedEvent(connection.ConnectionId, stopwatch.Elapsed, ex, _isStreamingEnabled && connection.Description.HelloResult.TopologyVersion != null));
 
                 throw;
             }
--- a/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
+++ b/src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
@@ -48,6 +48,7 @@ namespace MongoDB.Driver.Core.Servers
         private readonly ServerMonitorSettings _serverMonitorSettings;
 
         private Thread _serverMonitorThread;
+        private bool _streamingEnabled;
 
         public event EventHandler<ServerDescriptionChangedEventArgs> DescriptionChanged;
 
@@ -105,6 +106,13 @@ namespace MongoDB.Driver.Core.Servers
 
             _logger = loggerFactory?.CreateLogger<IServerMonitor>();
             _eventLoggerSdam = loggerFactory.CreateEventLogger<LogCategories.SDAM>(eventSubscriber);
+
+            _streamingEnabled = serverMonitorSettings.ServerMonitoringMode switch
+            {
+                ServerMonitoringMode.Stream => true,
+                ServerMonitoringMode.Poll => false,
+                _ => !IsRunningInFaaS() // ServerMonitoringMode.Auto
+            };
         }
 
         public ServerDescription Description => Interlocked.CompareExchange(ref _currentDescription, null, null);
@@ -155,7 +163,6 @@ namespace MongoDB.Driver.Core.Servers
             {
                 _logger?.LogDebug(_serverId, "Initializing");
 
-                _roundTripTimeMonitor.Start();
                 _serverMonitorThread = new Thread(new ParameterizedThreadStart(ThreadStart)) { IsBackground = true };
                 _serverMonitorThread.Start(_monitorCancellationToken);
 
@@ -182,7 +189,7 @@ namespace MongoDB.Driver.Core.Servers
         {
             BsonDocument helloCommand;
             var commandResponseHandling = CommandResponseHandling.Return;
-            if (connection.Description.HelloResult.TopologyVersion != null)
+            if (_streamingEnabled && connection.Description.HelloResult.TopologyVersion != null)
             {
                 connection.SetReadTimeout(_serverMonitorSettings.ConnectTimeout + _serverMonitorSettings.HeartbeatInterval);
                 commandResponseHandling = CommandResponseHandling.ExhaustAllowed;
@@ -411,13 +418,19 @@ namespace MongoDB.Driver.Core.Servers
                     SetDescription(newDescription);
                 }
 
-                processAnother =
-                    // serverSupportsStreaming
-                    (newDescription.Type != ServerType.Unknown && heartbeatHelloResult != null && heartbeatHelloResult.TopologyVersion != null) ||
-                    // connectionIsStreaming
-                    (helloProtocol != null && helloProtocol.MoreToCome) ||
-                    // transitionedWithNetworkError
+                var serverSupportsStreaming = (newDescription.Type != ServerType.Unknown &&
+                                               heartbeatHelloResult != null &&
+                                               heartbeatHelloResult.TopologyVersion != null);
+                var connectionIsStreaming = (helloProtocol != null && helloProtocol.MoreToCome);
+                var transitionedWithNetworkError =
                     (IsNetworkError(heartbeatException) && previousDescription.Type != ServerType.Unknown);
+
+                if (_streamingEnabled && serverSupportsStreaming && !_roundTripTimeMonitor.Started)
+                {
+                    _roundTripTimeMonitor.Start(); // start RTT monitoring on separate thread
+                }
+
+                processAnother = (_streamingEnabled && (serverSupportsStreaming || connectionIsStreaming)) || transitionedWithNetworkError;
             }
 
             bool IsNetworkError(Exception ex)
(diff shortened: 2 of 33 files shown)


src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -203,17 +210,25 @@ private CommandWireProtocol<BsonDocument> InitializeHelloProtocol(IConnection co
         {
             var connection = _connectionFactory.CreateConnection(_serverId, _endPoint);
 
+            _eventLoggerSdam.LogAndPublish(new ServerHeartbeatStartedEvent(connection.ConnectionId, _streamingEnabled && _currentDescription.TopologyVersion != null));
```
adelinowona, line 213: Adding this here as the spec tests seemed to require that the first server heartbeat events have the awaited field as false. Before this change we were only started logging server heartbeat events after the initial handshake (by then we know if the server supports streaming thus the awaited field would not be false causing spec tests to fail). I don't know if that was a requirement but I couldn't find anything about not emitting heartbeat events in the initial handshake in the specs. 

TLDR; I added the logging of server heartbeats when we initialize the server monitor connection as it is technically a heartbeat event.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -163,8 +161,8 @@ public void Description_should_be_connected_after_successful_heartbeat()
             subject.Description.State.Should().Be(ServerState.Connected);
             subject.Description.Type.Should().Be(ServerType.Standalone);
 
-            // no ServerHeartbeat events should be triggered during initial handshake
```
adelinowona, line 166: Is this driver's requirement or just one set for the C# driver? I couldn't find reasoning for this in the specs.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedEventMatcher.cs
```diff
@@ -371,6 +410,30 @@ private string GetAssertionErrorMessage(List<object> actualEvents, BsonArray exp
...
+                        break;
+                    case ServerHeartbeatSucceededEvent serverHeartbeatSucceededEvent:
+                        var serverHeartbeatSucceededEventDocument = new BsonDocument
+                        {
+                            { "awaited", serverHeartbeatSucceededEvent.Awaited }
+                        };
+                        actualEventsDocuments.Add(new BsonDocument(actualEvent.GetType().Name,
+                            serverHeartbeatSucceededEventDocument));
+                        break;
+                    case ServerHeartbeatFailedEvent serverHeartbeatFailedEvent:
```
adelinowona, line 429: These cases are very similar to each other and thought of collapsing them to the same together but I don't know how to do that while keeping the captured variable like:
case Event1:
case Event2:
case Event3 variable:
   variable is either of type Event1, Event2, Event3.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -48,6 +48,7 @@ internal sealed class ServerMonitor : IServerMonitor
         private readonly ServerMonitorSettings _serverMonitorSettings;
 
         private Thread _serverMonitorThread;
+        private bool _streamingEnabled;
```
BorisDog, line 51: minor: `_isStreamingEnabled`

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
BorisDog: Yes this part is not consistent among drivers, and is being clarified in [DRIVERS-2677](https://jira.mongodb.org/browse/DRIVERS-2677).
This also should affect logging tests.
I would suggest to temporary adapt the tests accordingly, and make the code change if needed in [CSHARP-4831](https://jira.mongodb.org/browse/CSHARP-4831). Let's discuss more offline if needed.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -405,13 +420,19 @@ private void Heartbeat(CancellationToken cancellationToken)
...
                 }
 
-                processAnother =
-                    // serverSupportsStreaming
-                    (newDescription.Type != ServerType.Unknown && heartbeatHelloResult != null && heartbeatHelloResult.TopologyVersion != null) ||
-                    // connectionIsStreaming
-                    (helloProtocol != null && helloProtocol.MoreToCome) ||
-                    // transitionedWithNetworkError
+                var serverSupportsStreaming = (newDescription.Type != ServerType.Unknown &&
+                                               heartbeatHelloResult != null &&
```
BorisDog, line 424: formatting: new lines alignment (no need for parenthesis) 
```
var serverSupportsStreaming = newDescription.Type != ServerType.Unknown &&
   heartbeatHelloResult != null &&
   ...
```

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -471,6 +498,16 @@ private void SetDescription(ServerDescription oldDescription, ServerDescription
             OnDescriptionChanged(oldDescription, newDescription);
         }
 
+        private bool IsRunningInFaaS()
+        {
+            return (Environment.GetEnvironmentVariable("AWS_EXECUTION_ENV")?.StartsWith("AWS_Lambda_") ?? false) ||
```
BorisDog, line 503: I could not find this in spec PR, is this specified in spec somewhere?

BorisDog (changes requested):
Partial review, looks great so far.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -427,22 +448,28 @@ bool IsNetworkError(Exception ex)
...
 
             var stopwatch = Stopwatch.StartNew();
             try
             {
                 var helloResult = HelloHelper.GetResult(connection, helloProtocol, cancellationToken);
                 stopwatch.Stop();
 
-                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatSucceededEvent(connection.ConnectionId, stopwatch.Elapsed, connection.Description.HelloResult.TopologyVersion != null, helloResult.Wrapped));
+                // RTT check if using polling monitoring
+                if ((!_streamingEnabled || (_streamingEnabled && connection.Description.HelloResult.TopologyVersion == null)) && helloResult != null)
```
BorisDog, line 460: Equivalent to shorter  `(!_streamingEnabled || connection.Description.HelloResult.TopologyVersion == null)` ?

tests/MongoDB.Driver.Core.Tests/Core/Configuration/ServerSettingsTests.cs
```diff
@@ -85,6 +95,18 @@ public void constructor_with_heartbeatTimeout_should_initialize_instance()
...
+        [Fact]
+        public void constructor_with_serverMonitoringMode_should_initialize_instance()
+        {
+            var serverMonitoringMode = ServerMonitoringMode.Stream;
+
+            var subject = new ServerSettings(serverMonitoringMode: serverMonitoringMode);
+
+            subject.HeartbeatInterval.Should().Be(ServerSettings.DefaultHeartbeatInterval);
+            subject.HeartbeatTimeout.Should().Be(ServerSettings.DefaultHeartbeatTimeout);
+            subject.ServerMonitoringMode.Should().Be(ServerMonitoringMode.Stream);
```
BorisDog, line 107: We can verify just the `ServerMonitoringMode` here and in `With_serverMonitoringMode_should_return_expected_result`?

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
BorisDog: The spec is not clear about this, so other drivers might have different behavior. I think we should change this in different ticket.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -389,6 +415,34 @@ public void ServerMonitor_without_serverApi_but_with_loadBalancedConnection_shou
             sentMessages[0].Should().Be($"{{ opcode : \"opmsg\", requestId : {requestId}, responseTo : 0, exhaustAllowed : true, sections : [ {{ payloadType : 0, document : {{ hello : 1, helloOk: true, topologyVersion : {{ processId : ObjectId(\"000000000000000000000000\"), counter : NumberLong(0) }}, maxAwaitTimeMS : NumberLong(86400000), loadBalanced: true, $db : \"admin\" }} }} ] }}");
         }
 
+        [Fact]
+        public void Should_use_polling_protocol_if_running_in_FaaS_platform()
+        {
+            // set environment to simulate being on FaaS platform
+            Environment.SetEnvironmentVariable("VERCEL", "vercel");
```
BorisDog, line 422: Do we want to test all environment variables?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedEventMatcher.cs
BorisDog: You could use dynamic, but that's hacky.
Another option is to have `BsonDocument GetHeartbeatDocument(IEvent event)` helper.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -471,6 +498,16 @@ private void SetDescription(ServerDescription oldDescription, ServerDescription
             OnDescriptionChanged(oldDescription, newDescription);
         }
 
+        private bool IsRunningInFaaS()
+        {
+            return (Environment.GetEnvironmentVariable("AWS_EXECUTION_ENV")?.StartsWith("AWS_Lambda_") ?? false) ||
+                   Environment.GetEnvironmentVariable("AWS_LAMBDA_RUNTIME_API") != null ||
```
BorisDog, line 504: Maybe it's worth specifying the FAAS provider in a comment for each var.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedEntityMap.cs
```diff
@@ -703,6 +709,8 @@ private IGridFSBucket CreateBucket(BsonDocument entity, Dictionary<string, IMong
                         settings.WriteConcern = writeConcern;
                         settings.HeartbeatInterval = heartbeatFrequency.GetValueOrDefault(defaultValue: TimeSpan.FromMilliseconds(5)); // 5 ms default value for spec tests
                         settings.ServerApi = serverApi;
+                        settings.ServerMonitoringMode =
+                            serverMonitoringMode.GetValueOrDefault(settings.ServerMonitoringMode);
```
BorisDog, line 713: minor: could place assignment in the same line.

BorisDog (changes requested):

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -27,6 +27,7 @@ namespace MongoDB.Driver.Core.Servers
     internal interface IRoundTripTimeMonitor : IDisposable
     {
         TimeSpan Average { get; }
+        bool Started { get; }
```
sanych-sun: _Minor_: should it be named `IsStarted`?

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -78,6 +82,17 @@ public TimeSpan Average
             }
         }
 
+        public bool Started
+        {
+            get
+            {
+                lock (_lock)
```
sanych-sun, line 86: Are we expecting high concurrency here? If so - it's better to use ReaderWriterLockSlim so to have let multiple readers read the value in parallel.

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -99,6 +114,7 @@ public void Start()
         {
             _roundTripTimeMonitorThread = new Thread(ThreadStart) { IsBackground = true };
             _roundTripTimeMonitorThread.Start();
+            _started = true;
```
sanych-sun: Having lock on read and no-lock on write looks suspicious. We either need both or none.

sanych-sun (changes requested):

src/MongoDB.Driver.Legacy/MongoServerSettings.cs
```diff
@@ -610,6 +613,20 @@ public IEnumerable<MongoServerAddress> Servers
             }
         }
 
+        /// <summary>
+        /// Gets or sets the server monitoring mode.
+        /// </summary>
+        /// <exception cref="System.InvalidOperationException">MongoServerSettings is frozen.</exception>
+        public ServerMonitoringMode ServerMonitoringMode
```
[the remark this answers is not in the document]
JamesKovacs: Excellent point. Customers shouldn’t be using the legacy driver anymore, especially for new applications. FaaS apps are by definition new apps as cloud providers only introduced these capabilities relatively recently. We can remove the changes in legacy.

src/MongoDB.Driver.Core/Core/Servers/RoundTripTimeMonitor.cs
adelinowona: I don't think we are expecting high concurrency here, so I will lean towards leaving it as is for now.

BorisDog: Actually we don't need any kind of synchronization at all here.
Proposed refactoring would eliminate the need of this property. Otherwise `Started => _roundTripTimeMonitorThread != null`  would be sufficient.

BorisDog (changes requested):

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -178,11 +185,16 @@ public void RequestHeartbeat()
         }
 
         // private methods
+        private bool UsingStreamingProtocol(HelloResult helloResult)
```
BorisDog, line 188: 1. Alphabetical order.
2. Personal preference : consider short arrow form `=>`
3. Also consider `IsUsingStreamingProtocol` as just `Using` might be read as some action.

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -405,13 +417,19 @@ private void Heartbeat(CancellationToken cancellationToken)
...
                 }
 
-                processAnother =
-                    // serverSupportsStreaming
-                    (newDescription.Type != ServerType.Unknown && heartbeatHelloResult != null && heartbeatHelloResult.TopologyVersion != null) ||
-                    // connectionIsStreaming
-                    (helloProtocol != null && helloProtocol.MoreToCome) ||
-                    // transitionedWithNetworkError
-                    (IsNetworkError(heartbeatException) && previousDescription.Type != ServerType.Unknown);
+                var serverSupportsStreaming = newDescription.Type != ServerType.Unknown &&
```
BorisDog, line 420: Alignment of new lines should be: 
```
var serverSupportsStreaming = newDescription.Type != ServerType.Unknown &&
   heartbeatHelloResult != null &&
   heartbeatHelloResult.TopologyVersion != null;
```

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -471,6 +495,22 @@ private void SetDescription(ServerDescription oldDescription, ServerDescription
             OnDescriptionChanged(oldDescription, newDescription);
         }
 
+        private bool IsRunningInFaaS()
```
BorisDog, line 498: Place with `IsUsing`, in alphabetical order.

src/MongoDB.Driver/MongoDB.Driver.csproj
```diff
@@ -14,8 +14,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="MongoDB.Libmongocrypt" Version="1.8.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageReference Include="MongoDB.Libmongocrypt" Version="1.8.0" />
```
BorisDog, line 18: unrelated change.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -293,14 +293,16 @@ public void InitializeHelloProtocol_should_use_streaming_protocol_when_available
...
             var subject = CreateSubject(out var mockConnection, out _, out var mockRoundTripTimeMonitor);
 
-            SetupHeartbeatConnection(mockConnection);
+            SetupHeartbeatConnection(mockConnection, isStreamable: true, false);
+            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(true), null);
+            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(), null);
 
             subject.Initialize();
-            subject.Initialize();
+            SpinWait.SpinUntil(() => mockConnection.GetSentMessages().Count >= 1, TimeSpan.FromSeconds(5)).Should().BeTrue();
```
BorisDog, line 305: I don't understand how this test works, let's sync offline.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -327,6 +329,29 @@ public void RequestHeartbeat_should_force_another_heartbeat()
...
+            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage());
+
+            subject.Initialize();
+            SpinWait.SpinUntil(() => mockConnection.GetSentMessages().Count >= 2, TimeSpan.FromSeconds(5)).Should().BeTrue();
+
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatStartedEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatSucceededEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatStartedEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatSucceededEvent>().Subject.Awaited.Should().Be(false);
+        }
```
BorisDog, line 353: It this stable?
Waiting for >= 2 events, but checking for 4.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -389,6 +414,40 @@ public void ServerMonitor_without_serverApi_but_with_loadBalancedConnection_shou
...
+
+            subject.Initialize();
+            SpinWait.SpinUntil(() => mockConnection.GetSentMessages().Count >= 2, TimeSpan.FromSeconds(5)).Should().BeTrue();
+
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatStartedEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatSucceededEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatStartedEvent>().Subject.Awaited.Should().Be(false);
+            capturedEvents.Next().Should().BeOfType<ServerHeartbeatSucceededEvent>().Subject.Awaited.Should().Be(false);
+
+            Environment.SetEnvironmentVariable(environmentVariableName, null);
```
BorisDog, line 448: try/finally. Or even better move `DisposableEnvironmentVariable` to utils and reuse it.

BorisDog (changes requested):

src/MongoDB.Driver.Core/Core/Servers/ServerMonitor.cs
```diff
@@ -427,22 +445,28 @@ bool IsNetworkError(Exception ex)
...
 
             var stopwatch = Stopwatch.StartNew();
             try
             {
                 var helloResult = HelloHelper.GetResult(connection, helloProtocol, cancellationToken);
                 stopwatch.Stop();
 
-                _eventLoggerSdam.LogAndPublish(new ServerHeartbeatSucceededEvent(connection.ConnectionId, stopwatch.Elapsed, connection.Description.HelloResult.TopologyVersion != null, helloResult.Wrapped));
+                // RTT check if using polling monitoring
+                if (UsingStreamingProtocol(connection.Description.HelloResult) is not true)
```
sanych-sun, line 457: Minor: why do we need pattern matching here? Probably could be simplified with:
`if (!UsingStreamingProtocol(connection.Description.HelloResult))`

sanych-sun (approved):
LGTM + minor comment

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
adelinowona: Waiting for the 2 sent messages should generate the 4 heartbeat events. Indeed it was a bit unstable and I had already fixed it locally. I just have to push the changes. Basically now I am straight up waiting for the 4 events to be ready before proceeding.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
adelinowona: Discussed offline.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -470,7 +565,7 @@ private CommandResponseMessage CreateHeartbeatCommandResponseMessage(bool moreTo
                     new BsonDocument
                     {
                         { "processId", ObjectId.Parse("5ee3f0963109d4fe5e71dd28") },
-                        { "counter", 0 }
+                        { "counter", new BsonInt64(0) }
```
adelinowona, line 568: This was a bug as the `Topology` get method from `HelloResult` expects this to be a 64-bit int and thus silently fails omitting the topology version even though it was provided in the response. This caused my tests to not behave properly even when I knew the server supported streaming based on the responses.

BorisDog: nice find

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -300,18 +300,51 @@ public void InitializeHelloProtocol_should_use_streaming_protocol_when_available
...
 
             mockRoundTripTimeMonitor.Verify(m => m.Start(), Times.Once);
         }
 
+        [Fact]
+        public void RoundTripTimeMonitor_should_not_be_started_if_using_polling_protocol()
+        {
+            var serverMonitorSettings = new ServerMonitorSettings(
+                TimeSpan.FromSeconds(5),
+                TimeSpan.FromMilliseconds(100),
```
BorisDog, line 331: Can we try even lower heartbeat interval, like 5ms in all tests?

tests/MongoDB.Driver.Core.TestHelpers/DisposableEnvironmentVariable.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+namespace MongoDB.Driver.Core.TestHelpers;
+
+public class DisposableEnvironmentVariable : IDisposable
```
BorisDog, line 5: nice.
make is sealed for Bon tone.

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -1,17 +1,17 @@
...
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing,.Setup software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-*/
+ *
```
BorisDog, line 2: revert?

BorisDog (changes requested):

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -300,18 +300,51 @@ public void InitializeHelloProtocol_should_use_streaming_protocol_when_available
...
 
             subject.Initialize();
-            subject.Initialize();
+            // spin until the initial handshake has happened in which case the server description state should be Connected
+            SpinWait.SpinUntil(() => subject.Description.State == ServerState.Connected, TimeSpan.FromSeconds(5)).Should().BeTrue();
+
+            // RTT thread should be started after first heartbeat (initial handshake)
+            mockRoundTripTimeMonitor.Verify(m => m.Start(), Times.Once);
+
+            SpinWait.SpinUntil(() => capturedEvents.Count >= 4, TimeSpan.FromSeconds(5)).Should().BeTrue();
```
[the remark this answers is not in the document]
adelinowona: `IsStarted` is setup in the `CreateSubject` method, it switches to returning true after it is called once. I will add checking `IsStarted` call count as well.

BorisDog (approved):
LGTM! + minor comment

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedEntityMap.cs
```diff
@@ -18,13 +18,15 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using FluentAssertions;
```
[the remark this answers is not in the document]
adelinowona: Nope I'll remove it!

PR #1239, opened 2023-12-19
CSHARP-1893: New DistinctMany method which is like Distinct but for array fields.

rstam:


src/MongoDB.Driver/FilteredMongoCollectionBase.cs
```diff
@@ -212,6 +212,26 @@ public override Task<IAsyncCursor<TField>> DistinctAsync<TField>(IClientSessionH
             return _wrappedCollection.DistinctAsync(session, field, CombineFilters(filter), options, cancellationToken);
         }
 
+        public override IAsyncCursor<TItem> DistinctMany<TItem>(FieldDefinition<TDocument, IEnumerable<TItem>> field, FilterDefinition<TDocument> filter, DistinctOptions options = null, CancellationToken cancellationToken = default)
```
rstam, line 215: I also considered the name `DistinctItems` for this new method but ultimately stuck with Sergei's suggestion of `DistinctMany`. The name `DistinctMany` parallels the name of the existing `SelectMany` LINQ method which has a similar signature.

src/MongoDB.Driver/FilteredMongoCollectionBase.cs
```diff
@@ -212,6 +212,26 @@ public override Task<IAsyncCursor<TField>> DistinctAsync<TField>(IClientSessionH
...
+        {
+            return _wrappedCollection.DistinctMany(field, CombineFilters(filter), options, cancellationToken);
+        }
+
+        public override IAsyncCursor<TItem> DistinctMany< TItem>(IClientSessionHandle session, FieldDefinition<TDocument, IEnumerable<TItem>> field, FilterDefinition<TDocument> filter, DistinctOptions options = null, CancellationToken cancellationToken = default)
```
JamesKovacs, line 220: Extraneous space in `DistinctMany< TItem>`.

JamesKovacs (changes requested):
LGTM. Minor formatting issue.

damieng:
What's the difference between `SelectMany(a => a.Names).Distinct()` and `DistinctMany(a => a.Name)` ?

rstam:
> What's the difference between `SelectMany(a => a.Names).Distinct()` and `DistinctMany(a => a.Name)` ?

SelectMany/Distinct are LINQ methods.

DistinctMany is an IMongoCollection method (it's a variation of the existing Distinct method).

The also execute differently. The LINQ methods are translated to an aggregation pipeline. The Distinct and DistinctMany methods are executed using the distinct command (no aggregation pipeline involved).

damieng:
So it just performs a Distinct operation directly against an array? I'm not sure where the Many part comes from then. SelectMany is a way of effectively combining many arrays into a new array hence the "Many" semantics.

rstam:
> So it just performs a Distinct operation directly against an array? I'm not sure where the Many part comes from then. SelectMany is a way of effectively combining many arrays into a new array hence the "Many" semantics.

So the method name DistinctMany was proposed by the original submitter of an external PR. At first I wasn't sure how they had come up with the name, but thinking about it I realized that this method had a lot in common with SelectMany.

SelectMany takes each input object and produces multiple output objects (the result of the selector) and concatenates them all into one output sequence.

DistinctMany takes each input document (object) and produces multiple output values (the items of an array field) and concatenates them all into one output sequence and then returns only the distinct values in that sequence.

We don't have to call this new method DistinctMany, but we have to call it something different than Distinct to avoid potential ambiguities with the existing Distinct method. But the name DistinctMany sounds as good as any other to me.

damieng (approved):

JamesKovacs (changes requested):
`FilteredMongoCollectionBase.cs` line 220 has an extra space in `DistinctMany< TItem>`.`

Production code looks good, but testing code only verifies mocked interactions. We need some integration tests to validate the implementation.

rstam:
The tests for DistinctMany are just modifed versions of the tests for Distinct.

I can change them to be integrations tests. Do you want the tests for Distinct to be changed also?

JamesKovacs (changes requested):
Discussed in Slack. Given `DistinctMany` delegates to integration-tested distinct operations and the existing tests in this file are testing via mocks, we can forgo the integration tests for these new `DistinctMany` methods. Just need to fix the formatting issue in FilteredMongoCollectionBase.cs.

sanych-sun (approved):
LGTM

JamesKovacs (approved):
LGTM

PR #1240, opened 2023-12-20
CSHARP-4888: Fix known serializer issue with enum underlying types.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
```diff
@@ -122,10 +122,6 @@ private HashSet<IBsonSerializer> GetPossibleSerializersAtThisLevel(Type type)
             foreach (var serializer in _knownSerializers.Values.SelectMany(hashset => hashset))
             {
                 var valueType = serializer.ValueType;
-                if (valueType == type || valueType.IsEnum && Enum.GetUnderlyingType(valueType) == type)
```
rstam, line 125: This code was needed at one time (long ago) but is no longer needed because we now handle comparing enums using their underlying types differently.

While this may have been needed at one time, it actually causes this problem (and did all along, we just didn't have tests for it).

JamesKovacs (approved):
LGTM

JamesKovacs (approved):
LGTM

PR #1242, opened 2023-12-27
CSHARP-4540: Add full support for $dateFromString.

rstam:


src/MongoDB.Driver/Linq/MongoDBFunctionsExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+    public static class MongoDBFunctionsExtensions
+    {
+        /// <summary>
+        /// Converts a string to a DateTime using the $dateFromString aggregation operator.
+        /// </summary>
+        /// <param name="functions">The functions object.</param>
+        /// <param name="dateString">The string.</param>
+        /// <returns>A DateTime.</returns>
+        /// <exception cref="NotSupportedException"></exception>
+        public static DateTime DateFromString(
```
rstam, line 32: Expressions in LINQ queries can't have default parameters, so that's why there are multiple overloads.

src/MongoDB.Driver/Linq/MongoDBFunctionsExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+*/
+
+using System;
+
+namespace MongoDB.Driver.Linq
+{
+    /// <summary>
+    /// Contains MongoDBFunctions extension methods that can be used in LINQ queries against MongoDB.
+    /// </summary>
+    public static class MongoDBFunctionsExtensions
```
rstam, line 23: This approach for MongoDB specific extension methods to be used in LINQ queries is copied from the pending PR for CSHARP-4858. 

If the approach changes there we will change it here also.

rstam:
You should probably review the Exists/IsMissing PR first:

https://github.com/mongodb/mongo-csharp-driver/pull/1224

JamesKovacs (approved):
LGTM

JamesKovacs (changes requested):
See previous comment about `$dateFromString` failing in 3.6.

JamesKovacs (approved):
LGTM

rstam:
Rebased on master and 4548 approach using static Mql method and force pushed.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4540Tests.cs
```diff
@@ -0,0 +1,282 @@
...
+        [Theory]
+        [InlineData(1, "2023-12-26T12:34:56Z", LinqProvider.V2)]
+        [InlineData(1, "2023-12-26T12:34:56Z", LinqProvider.V3)]
+        [InlineData(2, "throws:FormatException", LinqProvider.V2)]
+        [InlineData(2, "throws:FormatException", LinqProvider.V3)]
+        [InlineData(3, "throws:FormatException", LinqProvider.V2)]
+        [InlineData(3, "throws:FormatException", LinqProvider.V3)]
+        [InlineData(4, "throws:MongoCommandException", LinqProvider.V2)]
+        [InlineData(4, "throws:MongoCommandException", LinqProvider.V3)]
+        public void Parse_should_work(int id, string expectedResult, LinqProvider linqProvider)
```
sanych-sun, line 39: Somehow I do not like the idea to mix positive and negative tests, but I do not have enough arguments to insist on separating them.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4540Tests.cs
```diff
@@ -0,0 +1,282 @@
...
+            if (expectedResult.StartsWith("throws:"))
+            {
+                var expectedExceptionType = expectedResult.Substring(7);
+                exception.GetType().Name.Should().Be(expectedExceptionType);
+            }
+            else
+            {
+                exception.Should().BeNull();
+                result.Should().Be(DateTime.Parse(expectedResult, null, DateTimeStyles.AdjustToUniversal));
+            }
```
sanych-sun, line 76: It might makes sense to move code on lines 67-76 into helper method (`AssertResult`?) and reuse it from all tests in the class.

sanych-sun (changes requested):
Looks good. Just some minor comments. Please take a look if you want to take the ideas on board.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4540Tests.cs
rstam: I hear you... but I also like that these are all variants of the exact same test scenario,  just with different data, and that in some cases the result is a string value and in some cases the result is an exception.

It makes it a lot easier to confirm that you are covering all desired scenarios (whether they throw or not) when all the `InlineData` lines are clustered together.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4540Tests.cs
rstam: I considered that but there are slight variations from test to test that make that hard.

One thing that makes it hard is that in some tests the exception is thrown during translation and in other tests the exception is thrown when executing the query.

I can pursue further if you really want me to. There are only 4 test methods so not a huge amount of sharing, and the sharing benefit is reduced by the added complication.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4540Tests.cs
```diff
@@ -0,0 +1,282 @@
...
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp4540Tests : Linq3IntegrationTest
```
sanych-sun, line 28: Consider to rename the class to something like `DateFromStringMethodTranslatorTests` or make tests name to include the `DateFromString` term (not just `Parse_should_work` but `Parse_from_date_string_should_work`) to provide some more context on what we are testing here.

JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/DateFromStringMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -0,0 +1,251 @@
...
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators.MethodTranslators
+{
+    public class DateFromStringMethodToAggregationExpressionTranslatorTests : Linq3IntegrationTest
```
rstam, line 29: @sanych-sun I moved the tests here as you suggested.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/DateFromStringMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                    $"{{ $match : {{ _id : {id} }} }}",
+                    "{ $project : { __fld0 : { $dateFromString : { dateString : '$S' } }, _id : 0 } }"
+                } :
+                new[]
+                {
+                    $"{{ $match : {{ _id : {id} }} }}",
+                    "{ $project : { _v : { $dateFromString : { dateString : '$S' } }, _id : 0 } }"
+                };
+
+            AssertOutcome(collection, queryable, expectedStages, expectedResult);
```
rstam, line 61: @sanych-sun I was able to share assertion code as you suggested in an earlier review but it took 5 methods instead of one. See what you think.

sanych-sun: If there is so much differences it might be easier to leave validation in each test method.  \I'm OK with both versions. Please choose what looks better for you.

sanych-sun (approved):
LGTM

PR #1245, opened 2024-01-12
CSHARP-4533: More informative message for invalid type

bradamelton:
This will cause the method call above to thow an "Unknown discriminator value [type name]" which is much more descriptive than the current behavior where System...GetType throws "System.ArgumentException: String cannot have zero length." Also by returning null rather than throwing, it is more compliant with the method description.

### DIFF
--- a/src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
+++ b/src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
@@ -56,6 +56,11 @@ namespace MongoDB.Bson.Serialization
         /// <returns>The type if type type name can be resolved; otherwise, null.</returns>
         public static Type GetActualType(string typeName)
         {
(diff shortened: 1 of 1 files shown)


BorisDog:
Thank you for your PR @bradamelton 
Could please file a Jira ticket describing this issue?

Please see [contribution guidelines](https://github.com/mongodb/mongo-csharp-driver/blob/master/CONTRIBUTING.md) for more details.

bradamelton:
Thanks! 

It looks like it could address this issue: https://jira.mongodb.org/browse/CSHARP-4533

I have added a new commit to link this in.

adelinowona:
Hi @bradamelton, thanks again for submitting this PR. We have revisited the issue in the ticket: https://jira.mongodb.org/browse/CSHARP-4533 and will make further changes on our end. So I will close this PR but we'll keep your commit and add you to the contributors list once the work is done.

PR #1247, opened 2024-01-15
CSHARP-4890: Search index management operations fail with a different error message

BorisDog:


adelinowona (approved):
LGTM! I see there is a search-index test that fails in the EG patch, is it related to these changes?

PR #1251, opened 2024-01-19
CSHARP-4894: Mark API that will be removed in 3.0 as obsolete: MongoDB.Driver.Core

BorisDog:


barbahgamesfounder (approved):

src/MongoDB.Driver.Core/Core/Misc/Ensure.cs
```diff
@@ -176,7 +177,7 @@ public static T IsEqualTo<T>(T value, T comparand, string paramName)
         /// <param name="value">The value of the parameter.</param>
         /// <param name="paramName">The name of the parameter.</param>
         /// <returns>The value of the parameter.</returns>
-        public static TimeSpan IsGreaterThanZero(TimeSpan value, string paramName) =>
+        public static TimeSpan IsGreaterThanZero(TimeSpan value, [CallerArgumentExpression("value")]string paramName = "value") =>
```
rstam, line 180: Once you use `CallerArgumentExpression` it's no longer a `paramName`, it's a string representing the expression the caller used to provide the argument when calling this method.

rstam: In the next line `IsGreaterThan` expects a parameter name as its last argument, not a string representing the caller's argument expression.

rstam: The net result is that the following line of code on line 97:

```throw new ArgumentOutOfRangeException(paramName, message)```

might be called with a value for  `paramName` that is not a parameter name, but rather a string representing an `Expression`. That seems wrong.

File: src/MongoDB.Driver/RenderContext.cs [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Linq;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Encapsulates classes needed for rendering Builder definitions.
+    /// </summary>
+    /// <typeparam name="TDocument">The type of the document.</typeparam>
+    public sealed record RenderContext<TDocument>
```
rstam, line 11: The old class name was `SearchDefinitionRenderContext`.

Are you implying that this `RenderContext` class will be used in more places?

rstam: Seems like you are proposing replacing multiple `Render` arguments with an instance of this class.

In which case the name `RenderArgs` might be more appropiate.

src/MongoDB.Driver/RenderContext.cs
```diff
@@ -0,0 +1,65 @@
...
+            SerializerRegistry = Ensure.IsNotNull(serializerRegistry, nameof(serializerRegistry));
+        }
+
+        /// <summary>
+        /// Gets the document serializer.
+        /// </summary>
+        public IBsonSerializer<TDocument> DocumentSerializer
+        {
+            get => _documentSerializer;
+            init => _documentSerializer = Ensure.IsNotNull(value, nameof(value));
```
rstam, line 43: A good example of why I prefer we standardize on using explicit backing fields (like `_documentSerializer`) instead of auto properties.

src/MongoDB.Driver/RenderContext.cs
rstam: And should probably be a a `struct` to avoid an extra memory allocation.

rstam (changes requested):
Overall I see no serious issues with using `record` for our settings classes (and other classes also).

There is still an issue when calling a constructor with how to tell the difference between the caller omitting an optional argument and passing `null`. In the first case we might want to supply a default value, in the second we would want to respect the explicitly passed `null`.

    new Settings() ;  // assigns a default value to `OptionalProperty`
    new Settings(optionalProperty: null); // also assigns a default value to `OptionalProperty` instead of null    


It might be that the only way a user can explicitly set a property to null is:

    new Settings() with { OptionalProperty = null };

BorisDog:
Closing this draft PR which started as POC for RenderArgs and using records, which was implemented in https://github.com/mongodb/mongo-csharp-driver/pull/1278.

PR #1254, opened 2024-01-24
CSHARP-4940: Update libmongocrypt version used in the driver

adelinowona:


tests/MongoDB.Driver.Tests/Packaging/PackagingTests.cs
```diff
@@ -38,7 +38,7 @@ public static void Libmongocrypt_library_should_provide_library_version()
         {
             var version = Library.Version;
 
-            version.Should().Be("1.8.0");
+            version.Should().Be("1.8.4");
```
adelinowona, line 41: `Library.Version` queries the version of the underlying C libmongrocrypt which is 1.8.4 hence why this is not 1.8.1

BorisDog (approved):
LGTM

PR #1255, opened 2024-01-30
CSHARP-4937: Support CompareTo via the IComparable interface.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/IComparableMethod.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using System.Reflection;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
+{
+    internal static class IComparableMethod
+    {
+        // public static methods
+        public static bool IsCompareToMethod(MethodInfo method)
```
rstam, line 24: Combines and enhances and replaces the two existing IsCompareToMethod methods.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/IComparableMethod.cs
```diff
@@ -0,0 +1,54 @@
+﻿/* Copyright 2010-present MongoDB Inc.
```
damieng, line 1: Nitpick: Should probably say 2024.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/IComparableMethod.cs
```diff
@@ -0,0 +1,54 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Reflection;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
+{
+    internal static class IComparableMethod
```
damieng, line 21: Kind of weird using the interface "ISomething" naming convention on a static class but if that's what we've done elsewhere it's fine.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4937Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+            var after = 1;
+
+            Implementation(collection, after);
+
+            void Implementation<TModel, TId>(IMongoCollection<TModel> collection, TId after)
+                where TModel : IIdentity<TId>
+                where TId : IComparable<TId>
+            {
+                var queryable = collection.AsQueryable()
+                    .Where(x => x.Id.CompareTo(after) > 0);
```
damieng, line 42: Do we support:

`x.Id.CompareTo(after) < 0`
`x.Id.CompareTo(after) <= 0`
`x.Id.CompareTo(after) == 0`
`x.Id.CompareTo(after) >= 0`
`x.IdCompareTo(after) == someValue`

If we do we should add test coverage.

damieng: I'm guessing we don't support:

`x,Id.CompareTo(x.SomeOtherId) > 0`

or

`x.Id.CompareTo(after) > x.SomeOtherId`

If we don't do we want to add "Throws NotSupportedException" tests for it?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4937Tests.cs
```diff
@@ -0,0 +1,104 @@
...
+            var after = 1;
+
+            Implementation(collection, after);
+
+            void Implementation<TModel, TId>(IMongoCollection<TModel> collection, TId after)
+                where TModel : IIdentity<TId>
+                where TId : IComparable<TId>
+            {
+                var queryable = collection.AsQueryable()
+                    .Select(x => x.Id.CompareTo(after) > 0);
```
damieng, line 67: Same comment about what combinations are supported here too.

damieng (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/IComparableMethod.cs
rstam: The standard we use is 2010-present. Even for new files.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/IComparableMethod.cs
rstam: This is a tough naming issue.

These classes are named XyzMethod for methods in class Xyz.

That's how we end up with IComparableMethod for methods in IComparable.

I know it conflicts with the naming convention for interfaces, but I think the alternative (mismatched names) is worse.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4937Tests.cs
rstam: All of those examples should be supported.

Some of them aren't possible in filter language and will end up using `$expr`.

Some of them don't make no sense but will still be translated.

rstam: I will add new tests for all those combinations to make sure they all also work with `IComparable`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4937Tests.cs
rstam: Same replies as above.

rstam (commented):
Good questions. I will add new tests and push a new commit later today.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4937Tests.cs
rstam: I decided to add more tests but only comparing the result of `CompareTo` to `0` (which is all we support at the moment).

We could create a JIRA ticket to support comparisons to values other than `0` but we could also just wait to see if anyone asks for it. Any comparison to `-1` or `1` can be rewritten as a comparison to `0`, so supporting that wouldn't add any new functionality. Worse, `CompareTo` is allowed to return any negative or positive number, so comparison to `-1` or `1` is technically not legal.

damieng (approved):

JamesKovacs (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1256, opened 2024-02-01
CSHARP-4797: Fix segfault when disposing of GSSAPI (Kerberos) security context on Linux.

JamesKovacs:
When calling `gss_delete_sec_context` we passed in a handle (aka `IntPtr` aka `void*`), but the function was expecting a `void**`. Strangely the function returned success even when passed a `void*` rather than a `void**`.


tests/MongoDB.Driver.Core.Tests/Core/Authentication/Libgssapi/GssapiSecurityCredentialTests.cs
```diff
@@ -55,7 +55,7 @@ public void Should_acquire_gssapi_security_credential_with_username_only()
         {
             RequireEnvironment.Check().EnvironmentVariable("GSSAPI_TESTS_ENABLED");
 
-            var credential = GssapiSecurityCredential.Acquire(_username);
+            using var credential = GssapiSecurityCredential.Acquire(_username);
```
BorisDog, line 58: good catch

BorisDog (approved):
LGTM!

PR #1258, opened 2024-02-02
CSHARP-4820: Always use the configured serializer.

rstam:


rstam:
This will be fixed by CSHARP-5572.

PR #1264, opened 2024-02-09
CSHARP-4962: Fix package build script on CI to produce 472 binaries

sanych-sun:
Run packages-pack on Windows host + set priority to 10 for all package-related tasks.

### DIFF
--- a/evergreen/evergreen.yml
+++ b/evergreen/evergreen.yml
@@ -944,7 +944,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: ${PROJECT_DIRECTORY}/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.nupkg
+        local_file: mongo-csharp-driver/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.nupkg
         remote_file: ${UPLOAD_BUCKET}/${revision}/${PACKAGE_ID}.${PACKAGE_VERSION}.nupkg
         bucket: mciuploads
         permissions: public-read
@@ -953,7 +953,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: ${PROJECT_DIRECTORY}/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.snupkg
+        local_file: mongo-csharp-driver/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.snupkg
         remote_file: ${UPLOAD_BUCKET}/${revision}/${PACKAGE_ID}.${PACKAGE_VERSION}.snupkg
         bucket: mciuploads
         permissions: public-read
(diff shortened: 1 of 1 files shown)


evergreen/evergreen.yml
```diff
@@ -944,7 +944,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: ${PROJECT_DIRECTORY}/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.nupkg
+        local_file: mongo-csharp-driver/artifacts/nuget/${PACKAGE_ID}.${PACKAGE_VERSION}.nupkg
```
sanych-sun, line 947: Windows hosts were not happy with the absolute path, it was threated it as relative.

evergreen/evergreen.yml
```diff
@@ -2331,22 +2331,24 @@ buildvariants:
 - matrix_name: build-packages
   matrix_spec:
     build-target: "release"
+    os: "windows-64" # should produce package on Windows to make sure full framework binaries created.
```
sanych-sun, line 2334: Setting host as a matrix_spec axis also set env variable, so platform detection works properly in sh-files.

evergreen/evergreen.yml
```diff
@@ -2331,22 +2331,24 @@ buildvariants:
...
   display_name: "Package Pack"
-  run_on: ubuntu2004-small
   tags: ["build-packages", "release-tag"]
   tasks:
     - name: build-packages
       git_tag_only: true
+      priority: 10
```
sanych-sun, line 2340: When someone starting a release, he/she have to wait for it's completion to move forward with after release manual steps. So I've decided to prioritize package-related steps.

evergreen/evergreen.yml
JamesKovacs: We use `${PROJECT_DIRECTORY}` throughout `evergreen.yml` and it runs successfully on Windows hosts. Any idea why it is problematic when used for the `local_file` argument?

evergreen/evergreen.yml
JamesKovacs: `matrix_name` allows us to run all combinations of the specified `matrix_spec`. Given that the package step is a single action, we can simply use `buildvariants`/`name` with `run_on` similar to `gssapi-auth-tests-windows` or `driver-performance-tests`. No need to invoke the `matrix_*` machinery.

evergreen/evergreen.yml
```diff
@@ -2331,22 +2331,24 @@ buildvariants:
...
 
 - matrix_name: generate-apidocs
   matrix_spec:
     build-target: "release"
+    os: "ubuntu-2004"
```
JamesKovacs, line 2345: Why are we moving from `ubuntu2004-small` to `ubuntu2004` for these variants?

JamesKovacs (changes requested):
Some questions and suggested simplifications.

evergreen/evergreen.yml
sanych-sun: But I was unable to properly set the OS expansion/env variable without axis, so the platform detection was not working properly.

evergreen/evergreen.yml
sanych-sun: Basically I've removed explicit run_on by using the matrix_spec (so it would set OS env variable and for consistency). Setting OS axis to "ubuntu-2004" is to be run on the ubuntu2004-small.
```
      - id: "ubuntu-2004"
        display_name: "Ubuntu 20.04"
        variables:
          OS: "ubuntu-2004"
          python3_binary: "/opt/python/3.8/bin/python3"
        run_on: ubuntu2004-small
```

evergreen/evergreen.yml
sanych-sun: As it was discussed offline: s3.put command somehow does not work very well with absolute path on Windows hosts. So will go with relative path as per proposed code.

JamesKovacs (approved):
LGTM

PR #1265, opened 2024-02-09
CSHARP-2442: Avoid extra memory allocations in StringBuilder in JsonWriter.

rstam:
This is the optimization that Boris suggested in our meeting today, namely that while we are examining the string to see if it needs to be escaped we can simultaneously calculate the length of the escaped string.

I haven't done any measurements to see whether this is faster or not. It's probably going to be faster with some string values than with others (depending on whether any characters need to be escaped or how many).

### DIFF
--- a/src/MongoDB.Bson/IO/JsonWriter.cs
+++ b/src/MongoDB.Bson/IO/JsonWriter.cs
@@ -814,12 +814,12 @@ namespace MongoDB.Bson.IO
         // private methods
         private string EscapedString(string value)
         {
-            if (value.All(c => !NeedsEscaping(c)))
+            if (!NeedsEscaping(value, out var escapedLength))
             {
                 return value;
             }
 
-            var sb = new StringBuilder(value.Length);
+            var sb = new StringBuilder(escapedLength);
 
             foreach (char c in value)
             {
@@ -935,7 +935,7 @@ namespace MongoDB.Bson.IO
             }
         }
 
-        private static bool NeedsEscaping(char c)
+        private static int EscapedLength(char c)
         {
             switch (c)
             {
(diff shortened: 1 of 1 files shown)


JamesKovacs (approved):
LGTM

src/MongoDB.Bson/IO/JsonWriter.cs
```diff
@@ -814,12 +814,12 @@ protected override void Dispose(bool disposing)
         // private methods
         private string EscapedString(string value)
         {
-            if (value.All(c => !NeedsEscaping(c)))
+            if (!NeedsEscaping(value, out var escapedLength))
```
BorisDog, line 817: Just a note that for strings that need escaping we'll run for the full length now. But I think this is negligible compared to overall work.

BorisDog (approved):
LGTM

src/MongoDB.Bson/IO/JsonWriter.cs
rstam: Yes... but for strings that do *not* need escaping we have to examine the whole string already. So for the common case of strings that do *not* need escaping there should be negligible overhead, and it should still be faster than the current code.

For strings that *do* need escaping we now have to examine the whole string, but in exchange we get to avoid memory copying in `StringBuilder`.

We could even not use `StringBuilder` at all and populate a `char[]` and use the `string(char[])` constructor.

BorisDog: Yes I agree, good point regarding `char[]`, this will eliminate any extra allocation. We could even keep almost the same syntax by using local function `Append`.

That might be next public PR in this area :)

rstam:
I was reluctant to push this change to master without measuring the impact.

Boris ran some initial measurements and found that the new code was slightly slower.

Leaving on the back burner for now.

rstam:
Closing PR as it did not actually improve performance.

PR #1266, opened 2024-02-09
CSHARP-4965: Bump maxWireVersion for MongoDB 7.2

adelinowona:
- Add wireVersions 22 and 23 to list of known wireVersions and updates tests as needed.

- Check that I updated everything that needs to be updated.


BorisDog (approved):
LGTM

PR #1267, opened 2024-02-14
CSHARP-4827: Verify that driver does not add RetryableWriteError when it should not.

rstam:
The gist of  DRIVERS-1641 is:

 Drivers MUST NOT add a RetryableWriteError label based on the following:

  - any `writeErrors[].code` fields in a mongod or mongos response
  - the `writeConcernError.code` field in a mongos response

when connected to a pre-4.4 server.

I added the new unified tests and they pass without any changes to the driver.

I searched the driver code base for uses of `.Code` and verified that we are not adding a `RetryableWriteError` anywhere based on the value of `Code`.


JamesKovacs (approved):
LGTM

PR #1273, opened 2024-02-22
CSHARP-4432: POC of standard implementations of Equals.

rstam:


rstam:
I'm about to implement a bunch of new `Equals` methods. 

Before doing so I'd like us to agree as a team on how we want to implement `Equals` going forward.

src/MongoDB.Bson/EqualsPoc/EquatableBaseClass.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public override bool Equals(object obj) =>
+            Equals(obj as EquatableBaseClass);
+
+        public bool Equals(EquatableBaseClass other) =>
+            object.ReferenceEquals(this, other) ||
+            !object.ReferenceEquals(other, null) &&
+            this.GetType() == other.GetType() &&
+            object.Equals(_ref1, other._ref1) &&
+            _value1.Equals(other._value1);
```
rstam, line 24: Notes:

Some very old articles claim the following is faster than using `object.ReferenceEquals`:

    (object)this == other

but most newer articles say that after JITing they are exactly the same thing. Using `object.ReferenceEquals` is more intention-revealing than the more obscure cast.

We must check that the types are equal. Objects of different types are never equal.

My recommendation is that we NEVER use `==` in the implementation fo `Equals`. Instead, we should always call `Equals` when comparing fields.

The reason is that in some cases `==` and `Equals` do NOT return the same value. When writing and reading `Equals` implementations it is better to standardize on calling `Equals` to compare fields instead of having to analyze field by field whether `==` semantics are correct.

Always use `object.Equals(_ref1, other._ref1)` instead of `_ref1.Equals(other._ref1)` when comparing reference values. That way nulls are handled correctly.

Always use `_value1.Equals(other._value1)` when comparing struct values. That way boxing is usually avoided (most struct classes will implement `IEquatable<T>`).

src/MongoDB.Bson/EqualsPoc/EquatableDerivedClass.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public override bool Equals(object obj) =>
+            Equals(obj as EquatableDerivedClass);
+
+        public bool Equals(EquatableDerivedClass other) =>
+            object.ReferenceEquals(this, other) ||
+            base.Equals(other) &&
+            object.Equals(_ref2, other._ref2) &&
+            _value2.Equals(other._value2);
```
rstam, line 24: Notes:

The `object.ReferenceEquals(this, other)` check must be coded at every level of a polymorphic class hierarchy for the optimization to apply at each level.

No need to check for null or compare types, that happens in `base.Equals`.

src/MongoDB.Bson/EqualsPoc/EquatableNonPolymorphicClass.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public override bool Equals(object obj) =>
+            Equals(obj as EquatableNonPolymorphicClass);
+
+        public bool Equals(EquatableNonPolymorphicClass other) =>
+            object.ReferenceEquals(this, other) ||
+            !object.ReferenceEquals(other, null) &&
+            object.Equals(_ref, other._ref) &&
+            _value.Equals(other._value);
```
rstam, line 23: No need to compare types because the class is non polymorphic.

This assumption is only valid is the class is `sealed`.

src/MongoDB.Bson/EqualsPoc/EquatableStruct.cs
```diff
@@ -0,0 +1,29 @@
...
+            _ref = @ref;
+            _value = value;
+        }
+
+        public override bool Equals(object obj) =>
+            obj is EquatableStruct other && Equals(other);
+
+        public bool Equals(EquatableStruct other) =>
+            object.Equals(_ref, other._ref) &&
+            _value.Equals(other._value);
```
rstam, line 21: No need to check for null because this is a struct.

No need to compare types because structs can't be polymorphic.

sanych-sun (approved):
LGTM

adelinowona (approved):
LGTM

src/MongoDB.Bson/EqualsPoc/EquatableBaseClass.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            _ref1 = ref1;
+            _value1 = value1;
+        }
+
+        public override bool Equals(object obj) =>
+            Equals(obj as EquatableBaseClass);
+
+        public bool Equals(EquatableBaseClass other) =>
+            object.ReferenceEquals(this, other) ||
```
BorisDog, line 20: I'd suggest to check for `null` first.
I think in general a comparison to a different non-derived type is more likely than to the same instance. 
This will also fail faster for `null`-case for derived classes (instead of two subsequent `object.ReferenceEquals(this, other)` calls)
Maybe this is one of the reasons that null check is first in MS examples.

BorisDog (approved):
Looks great overall, one small comment.

src/MongoDB.Bson/EqualsPoc/EquatableBaseClass.cs
rstam: Good suggestions.

I've refactored to compare to null first.

rstam:
I've re-requested a review from everyone on this PR.

BorisDog (approved):
LGTM.

damieng (approved):
Looks good. I did something similar a while back but had base classes implement CompareTo and get all the other operators https://github.com/damieng/DamienGKit/blob/master/CSharp/DamienG.Library/System/AutoOperators.cs

adelinowona (approved):
LGTM

src/MongoDB.Bson/EqualsPoc/EquatableDerivedClass.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public override bool Equals(object obj) =>
+            Equals(obj as EquatableDerivedClass);
+
+        public bool Equals(EquatableDerivedClass other)
+        {
+            if (object.ReferenceEquals(other, null)) { return false; }
+            if (object.ReferenceEquals(this, other)) { return true; }
+            return
```
sanych-sun, line 24: Should we check if `this` and `other` types are the same? What if someone try to compare `EquatableDerivedClass` with `EquatableDerivedFromDerivedClass`? As far as I understood the current implementation will returns `true` is all local variables would be equal.

sanych-sun (changes requested):

src/MongoDB.Bson/EqualsPoc/EquatableDerivedClass.cs
rstam: If the two objects are not of the same type then `base.Equals` will return `false`.

sanych-sun (approved):
LGTM

rstam:
Closing the PR as it is only a POC. See CSHARP-3315 for new examples of Equals methods implemented in this standard way.

PR #1274, opened 2024-02-27
CSHARP-4684: Provide a way for applications to see what MQL was executed for a LINQ query.

rstam:


JamesKovacs (approved):
LGTM

damieng:
Confirmed I'm seeing the MQL in the right place in the EF Provider that I may log it.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQuery.cs
```diff
@@ -62,13 +65,33 @@ public MongoQuery(MongoQueryProvider<TDocument> provider, Expression expression)
         public override IAsyncCursor<TOutput> Execute()
         {
             var executableQuery = ExpressionToExecutableQueryTranslator.Translate<TDocument, TOutput>(_provider, _expression);
-            return _provider.Execute(executableQuery);
+            return _provider.Execute(executableQuery, CancellationToken.None);
+        }
+
+        public (TResult Result, BsonDocument[] LoggedStages) ExecuteAndLog<TResult>(Func<IMongoQueryable<TOutput>, TResult> executor)
```
damieng, line 71: We don't need ExecuteAndLog or ExecuteAndLogAsync in the EF provider - we do a lot of what this does already for retargetting the query and can access the LoggedStaged directly.

rstam:
Removed ExecuteAndLog.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQuery.cs
```diff
@@ -54,6 +55,8 @@ public MongoQuery(MongoQueryProvider<TDocument> provider, Expression expression)
 
         public Expression Expression => _expression;
 
+        public BsonDocument[] LoggedStages => _provider.LoggedStages;
```
damieng, line 58: If the provider is storing the actual result won't this property return incorrect results where it isn't the most recently run query?

Would seem we have a couple of options depending on whether we want to support there being multiple around:

1. Remove the LoggedStages on IMongoQueryable and it's various implementations and let people access LoggedStages only via the provider for only the most recent executed
2. Move ExecutableQuery from the provider to the MongoQueryable and have people access the LoggedStages only on the MongoQueryable and not on the provider

damieng (approved):

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQuery.cs
JamesKovacs: I think we can live with this oddity. We should document the limitation to avoid confusion. Our recommended practice is to call `collection.AsQueryable()` for each query anyways. We should document this somewhere anyways.

JamesKovacs (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1275, opened 2024-02-27
Make ThreadingUtilities.ExecuteTasksOnNewThreads await on the operations completion.

sanych-sun:
Test console before the fix:
1 - sync method works as expected.
2 - async method "exit" before inner operations are completed.
image
image

Test console after the fix:
both sync and async methods work as expected:
image
image


tests/MongoDB.Bson.TestHelpers/Threading/ThreadingUtilities.cs
```diff
@@ -110,7 +110,7 @@ public static async Task<Exception[]> ExecuteTasksOnNewThreadsCollectExceptions(
                 }
             });
 
-            var taskAll = Task.WhenAll(allTasks);
+            var taskAll = Task.WhenAll(allTasks.Select(t => t.Result));
```
sanych-sun, line 113: Not sure if this is the best way to fix the problem, but this is the simplest way I could find.

BorisDog (approved):
LGTM!

PR #1277, opened 2024-02-29
CSHARP-4988: Fix flaky ServerMonitor test

sanych-sun:


tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -307,9 +307,10 @@ public void RoundTripTimeMonitor_should_be_started_only_once_if_using_streaming_
             var subject = CreateSubject(out var mockConnection, out _, out var mockRoundTripTimeMonitor, capturedEvents, serverMonitorSettings: serverMonitorSettings);
 
             SetupHeartbeatConnection(mockConnection, isStreamable: true, autoFillStreamingResponses: false);
-            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(), null);
-            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(), null);
-            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(), null);
+            // Have to setup small delay on the heartbeat responses to emulate streamableServer functionality.
+            mockConnection.EnqueueCommandResponseMessage(CreateHeartbeatCommandResponseMessage(), TimeSpan.FromMilliseconds(500));
```
adelinowona, line 311: Is it possible to reduce the delay a bit more to like maybe 100ms?

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -307,9 +307,10 @@ public void RoundTripTimeMonitor_should_be_started_only_once_if_using_streaming_
             var subject = CreateSubject(out var mockConnection, out _, out var mockRoundTripTimeMonitor, capturedEvents, serverMonitorSettings: serverMonitorSettings);
```
adelinowona, line 307: Also I think you might want remove the serverMonitorSettings variable before this. Looking at the code I don't think it is needed since we will be streaming heartbeats with the server.

adelinowona (changes requested):

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
BorisDog: I think it would be great if could not rely on timings at all, and then have much lower intervals like in the rest of tests (5ms).

sanych-sun: So after additional discussion with Boris we decided to remove delays at all, as well as verification of the Start method was called after the very first heartbeat (because the way we detecting the first heartbeat was done is not that reliable)

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
sanych-sun: Done

tests/MongoDB.Driver.Core.Tests/Core/Servers/ServerMonitorTests.cs
```diff
@@ -303,23 +303,17 @@ public void InitializeHelloProtocol_should_use_streaming_protocol_when_available
         public void RoundTripTimeMonitor_should_be_started_only_once_if_using_streaming_protocol()
         {
             var capturedEvents = new EventCapturer().Capture<ServerHeartbeatSucceededEvent>();
-            var serverMonitorSettings = new ServerMonitorSettings(TimeSpan.FromSeconds(5), TimeSpan.FromMilliseconds(10));
```
BorisDog, line 306: I think we still need this to override the higher default interval.

sanych-sun: OK, will revert this =)

BorisDog (approved):
LGTM

adelinowona (approved):
LGTM

PR #1279, opened 2024-03-01
CSHARP-4953: dot-net-driver Evergreen Configuration May Be Leaking Secrets

sanych-sun:


evergreen/evergreen.yml
```diff
@@ -1916,6 +1942,7 @@ task_groups:
         params:
           file: testazurekms-expansions.yml
     teardown_group:
+      - func: upload-test-results
```
sanych-sun, line 1945: This change unrelated to leaking secrets, but it fixes missed tests results for some build variants.

BorisDog (commented):
Double check also
GCPKMS_SERVICEACCOUNT and ATLAS_SEARCH

BorisDog (approved):
LGTM.
Do we need also the check the nested scripts as well?

JamesKovacs (approved):
LGTM

I considered suggesting YAML anchors/aliases to remove the duplication in env var names, but it adds complexity.

You could add a `variables section at the top:
```
variables:
- fle_env_vars: &fle_env_vars
  - "FLE_AWS_ACCESS_KEY_ID"
  - "FLE_AWS_SECRET_ACCESS_KEY"
  - "FLE_AZURE_TENANT_ID"
  - "FLE_AZURE_CLIENT_ID"
  - "FLE_AZURE_CLIENT_SECRET"
  - "FLE_GCP_EMAIL"
  - "FLE_GCP_PRIVATE_KEY"

- other_env_vars: &other_env_vars
  - "A"
  - "B"
  - "C"
```

And then use it later:

```
include_expansions_in_env: *fle_env_vars
```

I'm OK with keeping the repetition and simplicity of the current impl.

PR #1285, opened 2024-03-07
CSHARP-4993: Avoid TruncationException when deserializing Decimal from Decimal128.

rstam:


src/MongoDB.Bson/Serialization/Attributes/BsonRepresentationAttribute.cs
```diff
@@ -37,6 +37,8 @@ public class BsonRepresentationAttribute : BsonSerializationOptionsAttribute
         public BsonRepresentationAttribute(BsonType representation)
         {
             _representation = representation;
+            _allowOverflow = false;
+            _allowTruncation = representation == BsonType.Decimal128;
```
rstam, line 41: This change was needed in order for configuration like this to work:

```
[BonRepresentation(BsonType.Decimal128)] public decimal D { get; set; }
```

src/MongoDB.Bson/Serialization/Serializers/DecimalSerializer.cs
```diff
@@ -51,7 +51,7 @@ public DecimalSerializer()
         /// </summary>
         /// <param name="representation">The representation.</param>
         public DecimalSerializer(BsonType representation)
-            : this(representation, new RepresentationConverter(false, false))
+            : this(representation, new RepresentationConverter(allowOverflow: false, allowTruncation: true))
```
sanych-sun, line 54: Should we mention this behavioral change as a breaking change?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs
```diff
@@ -119,7 +119,7 @@ public static AggregationExpression Translate(TranslationContext context, Binary
                 Type t when t == typeof(ulong) => new UInt64Serializer(),
                 Type t when t == typeof(float) => new SingleSerializer(),
                 Type t when t == typeof(double) => new DoubleSerializer(),
-                Type t when t == typeof(decimal) => new DecimalSerializer(),
+                Type t when t == typeof(decimal) => DecimalSerializer.Decimal128Instance,
```
sanych-sun, line 122: As far as I remember default representation for decimals is string for now. Are we sure this change is safe?

src/MongoDB.Bson/Serialization/Serializers/DecimalSerializer.cs
rstam: I don't think this is a breaking change.

Something that used to fail now works. Nothing broke.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs
rstam: I believe this change is safe.

If the `decimal` values are stored as strings in the database the server will return an error when arithmetic is attempted on strings.

Note that we plan to change the default representation for `decimal` to `Decimal128` in 3.0 We should have changed this default long ago.

You might also be interested in CSHARP-4985 where we intend to verify client side that arguments to arithmetic operations are represented as numbers on the server.

src/MongoDB.Bson/Serialization/Serializers/DecimalSerializer.cs
rstam: I guess in the unlikely case that you were counting on this exception then you might consider this a breaking change, but I can't imagine anyone wanted an exception here!

sanych-sun (approved):
LGTM

PR #1286, opened 2024-03-08
CSHARP-3742: Sync BSON corpus tests; fix minor issue where JsonReader was allowing nulls in  pattern and option values.

JamesKovacs:


src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1812,6 +1812,10 @@ private BsonValue ParseRegularExpressionExtendedJsonCanonical()
                     throw new FormatException(message);
                 }
                 var value = nextToken.StringValue;
+                if (value.Contains("\0"))
```
sanych-sun, line 1815: _Minor_: It could be char, not a string.

sanych-sun (approved):
LGTM

src/MongoDB.Bson/IO/JsonReader.cs
JamesKovacs: Then I'd have to do `value.IndexOf('\0') != 0`, which is less readable.

PR #1287, opened 2024-03-08
CSHARP-5001: Add OS variable to PREPARE_SHELL

BorisDog:
- Fixes .NET8 installation issue with Plain Auth tests on M1
- Simplifies EG script


sanych-sun (approved):
LGTM

PR #1289, opened 2024-03-11
CSHARP-4997: Removed the long-deprecated 1.11 refdocs and apidocs.

JamesKovacs:
Taking a look at our `gh-pages` branch, the `1.11` directory is a hodgepodge of 1.11 and 2.0 docs. For example, both the 1.11 and 2.0 `What is New?` page refer to the 2.0 driver though the 1.11 version isn’t linked from anywhere. The same is true for many other pages in the 1.11 directory.
https://mongodb.github.io/mongo-csharp-driver/1.11/what_is_new/
https://mongodb.github.io/mongo-csharp-driver/2.0/what_is_new/

Given that the 1.11 driver is long-deprecated and no longer supported, it was easiest to simply delete the 1.11 directory and its contents. I could not find any external links into the old 1.11 docs. I could only find links between pages within 1.11/apidocs. The only way I could get Google to produce a link to the 1.11/apidocs is if I searched for a class name unique to the 1.X API.

My recommendation is to delete the entire `1.11` directory.


BorisDog (approved):
LGTM

adelinowona (approved):
LGTM

PR #1295, opened 2024-03-18
CSHARP-4718: Enable Container and kubernetes awareness

adelinowona:
Enables container and kubernetes awareness in the driver as specified in the [MongoDB Handshake Spec](https://github.com/abr-egn/specifications/blob/c4831b426c7eab8749723d7f3cd6bb5c237bcbbe/source/mongodb-handshake/handshake.rst#container) .

#### What is the motivation for this change?
 - Allows Product to track user usage of docker and kubernetes.

### DIFF
--- a/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
@@ -14,7 +14,6 @@
 */
 
 using System;
-using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
@@ -32,8 +31,8 @@ namespace MongoDB.Driver.Core.Connections
         private static Lazy<BsonDocument> __envDocument;
         private static Lazy<BsonDocument> __osDocument;
         private static Lazy<string> __platformString;
-        private static IEnvironmentVariableProvider __environmentVariableProvider;
-        private static IFileSystemProvider __filesystem;
+        private static Lazy<IEnvironmentVariableProvider> __environmentVariableProvider;
+        private static Lazy<IFileSystemProvider> __filesystem;
 
         private static void Initialize()
         {
@@ -41,20 +40,20 @@ namespace MongoDB.Driver.Core.Connections
             __envDocument = new Lazy<BsonDocument>(CreateEnvDocument);
             __osDocument = new Lazy<BsonDocument>(CreateOSDocument);
             __platformString = new Lazy<string>(GetPlatformString);
-            __environmentVariableProvider = new EnvironmentVariableProvider();
-            __filesystem = new FileSystemProvider();
+            __environmentVariableProvider = new Lazy<IEnvironmentVariableProvider>(() => new EnvironmentVariableProvider());
+            __filesystem = new Lazy<IFileSystemProvider>(() => new FileSystemProvider());
         }
 
         static ClientDocumentHelper() => Initialize();
 
         internal static void SetEnvironmentVariableProvider(IEnvironmentVariableProvider environmentVariableProvider)
         {
-            __environmentVariableProvider = environmentVariableProvider;
+            __environmentVariableProvider = new Lazy<IEnvironmentVariableProvider>(() => environmentVariableProvider);
         }
 
         internal static void SetFileSystemProvider(IFileSystemProvider fileSystemProvider)
         {
-            __filesystem = fileSystemProvider;
+            __filesystem = new Lazy<IFileSystemProvider>(() => fileSystemProvider);
         }
 
         // private static methods
@@ -196,8 +195,8 @@ namespace MongoDB.Driver.Core.Connections
 
             BsonDocument GetContainerDocument()
             {
-                var isExecutionContainerDocker = __filesystem.File.Exists("/.dockerenv");
-                var isOrchestratorKubernetes = __environmentVariableProvider.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST") != null;
+                var isExecutionContainerDocker = __filesystem.Value.File.Exists("/.dockerenv");
+                var isOrchestratorKubernetes = __environmentVariableProvider.Value.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST") != null;
 
                 if (isExecutionContainerDocker || isOrchestratorKubernetes)
                 {
--- a/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
+++ b/src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
@@ -32,7 +32,7 @@ namespace MongoDB.Driver.Core.Connections
         private static Lazy<BsonDocument> __osDocument;
         private static Lazy<string> __platformString;
         private static Lazy<IEnvironmentVariableProvider> __environmentVariableProvider;
-        private static Lazy<IFileSystemProvider> __filesystem;
+        private static Lazy<IFileSystemProvider> __fileSystemProvider;
 
         private static void Initialize()
         {
@@ -41,7 +41,7 @@ namespace MongoDB.Driver.Core.Connections
             __osDocument = new Lazy<BsonDocument>(CreateOSDocument);
             __platformString = new Lazy<string>(GetPlatformString);
             __environmentVariableProvider = new Lazy<IEnvironmentVariableProvider>(() => new EnvironmentVariableProvider());
-            __filesystem = new Lazy<IFileSystemProvider>(() => new FileSystemProvider());
+            __fileSystemProvider = new Lazy<IFileSystemProvider>(() => new FileSystemProvider());
         }
 
         static ClientDocumentHelper() => Initialize();
@@ -53,7 +53,7 @@ namespace MongoDB.Driver.Core.Connections
 
         internal static void SetFileSystemProvider(IFileSystemProvider fileSystemProvider)
         {
-            __filesystem = new Lazy<IFileSystemProvider>(() => fileSystemProvider);
+            __fileSystemProvider = new Lazy<IFileSystemProvider>(() => fileSystemProvider);
         }
 
         // private static methods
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -176,6 +179,23 @@ name switch
                     _ => null,
                 };
 
+            BsonDocument GetContainerDocument()
+            {
+                var isExecutionContainerDocker = File.Exists("/.dockerenv");
```
adelinowona, line 184: I hesitated a little about checking for existence of a file to detect if we are in a docker container. Doing this makes us a dependent on the file system which might have potential issues related to file permissions or availability. I did some digging and found there is a environment variable called `DOTNET_RUNNING_IN_CONTAINER ` which is set automatically by official Microsoft .Net images [More info](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_running_in_container-and-dotnet_running_in_containers) . Problem is using this environment variable relies on the .net runtime to set it which is guaranteed if users use the official .Net images for their docker containers. However, it doesn't cover cases where users don't use the official .net images or choose to manually install a .net runtime in their containers. So I think checking for the file existence is our best bet unless we want to use both strategies and let them compensate for each other?

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
```diff
@@ -132,6 +132,17 @@ public void CreateDriverDocument_should_return_expected_result()
             result.Should().Be($"{{ name : 'mongo-csharp-driver', version : '{driverVersion}' }}");
         }
 
+        [Fact]
+        public void CreateEnvDocument_should_return_expected_result()
```
adelinowona, line 136: Using file existence as check for docker container presence means we can't really unit test it because we can't easily mock `FIle.Exist` as it is a static class. There is a way to bypass that but it will mean complicating the code for checking docker container presence just for the sake of testing. However, I manually tested by running a mongodb .net application in a docker container with the changes to test and the docker detection logic works.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -176,6 +179,23 @@ name switch
...
+                var isOrchestratorKubernetes = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST") != null;
```
sanych-sun, line 185: It might makes sense to use `EnvironmentVariableProvider` here. It might be substitutable and it will help us when we will approach task to run tests in parallel.

sanych-sun: We probably ought to implement something similar for file system. With the same motivation: having such substitutable components could help us to isolate each test without necessarily to have file existence on the host machine or setting environment variables, as each such host adjustment could affect other tests.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
sanych-sun: See mine comment above.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
sanych-sun, line 184: Is Windows file system OK with such check? Should we wrap this with try-catch (some permissions restriction could lead to exception I suppose)?

BTW, this check looks like a Linux container detection. What about Windows containers? Should we try to detect them as well?

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
```diff
@@ -132,6 +132,17 @@ public void CreateDriverDocument_should_return_expected_result()
...
+        {
+            using (new DisposableEnvironmentVariable("VERCEL"))
+            using (new DisposableEnvironmentVariable("KUBERNETES_SERVICE_HOST"))
+            {
+                var result = ClientDocumentHelper.CreateEnvDocument();
+                result.Should().Be($"{{ name : 'vercel', container : {{ orchestrator : 'kubernetes' }} }}");
+            }
+        }
+
```
JamesKovacs, line 145: We should also verify:
- `client.env.container.runtime` is set to `docker` when `/.dockerenv` is present.
- `client.env.container` is not present if neither Docker nor Kubernetes is deteced.

JamesKovacs (changes requested):
Looking good. I'm OK with the current impl or switching to our own mocks for `IFileSystem` and `IEnvironment` (or whatever we want to call it. We do need the two additional test cases noted.

As for Windows container detection, if we find a way to reliably detect Docker on Windows, then we need to create a new DRIVERS ticket and update the spec so all drivers perform the detection in the same way.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
adelinowona: Done. Also I think I found a reliable way to detect Docker for Windows. I will reach out the owners of the DRIVERS ticket for this work to discuss further if spec needs to be updated or if we are ok with drivers not detecting Windows containers given their low usage.

JamesKovacs (approved):
LGTM

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -30,17 +31,31 @@ internal class ClientDocumentHelper
...
+        private static Lazy<IEnvironmentVariableProvider> __environmentVariableProvider;
+        private static Lazy<IFileSystemProvider> __fileSystemProvider;
 
         private static void Initialize()
         {
             __driverDocument = new Lazy<BsonDocument>(CreateDriverDocument);
             __envDocument = new Lazy<BsonDocument>(CreateEnvDocument);
             __osDocument = new Lazy<BsonDocument>(CreateOSDocument);
             __platformString = new Lazy<string>(GetPlatformString);
+            __environmentVariableProvider = new Lazy<IEnvironmentVariableProvider>(() => new EnvironmentVariableProvider());
```
sanych-sun, line 43: Why it's Lazy? It could point to `EnvironmentVariableProvider.Instance` by default.

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -30,17 +31,31 @@ internal class ClientDocumentHelper
...
+            __fileSystemProvider = new Lazy<IFileSystemProvider>(() => new FileSystemProvider());
```
sanych-sun, line 44: Same here: I suppose we do not need Lazy, just use Instance by default.

sanych-sun (changes requested):
Looks good, minor changes requested.

sanych-sun (approved):
LGTM

src/MongoDB.Driver.Core/Core/Misc/FileWrapper.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// Abstraction for static methods in <see cref="System.IO.File" />.
+    /// </summary>
+    internal interface IFile
+    {
+        bool Exists(string name);
+    }
+
+    internal class FileWrapper : IFile
```
BorisDog, line 28: minor: sealed?

src/MongoDB.Driver.Core/Core/Misc/FileSystemProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    internal interface IFileSystemProvider
+    {
+        IFile File { get; }
+    }
+
+    internal class FileSystemProvider : IFileSystemProvider
+    {
+        private static readonly IFileSystemProvider __instance = new FileSystemProvider();
+        public static IFileSystemProvider Instance => __instance;
```
BorisDog, line 29: minor: maybe read-only property is sufficient ?
`public static IFileSystemProvider Instance { get; } = new FileSystemProvider();`

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -30,17 +31,31 @@ internal class ClientDocumentHelper
         private static Lazy<BsonDocument> __envDocument;
         private static Lazy<BsonDocument> __osDocument;
         private static Lazy<string> __platformString;
+        private static IEnvironmentVariableProvider __environmentVariableProvider;
```
BorisDog, line 34: Do we want to use `__environmentVariableProvider` for all other variables in this file?
Follow up ticket for that? This is also pre-requisite for future tests parallelization.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
```diff
@@ -132,6 +134,44 @@ public void CreateDriverDocument_should_return_expected_result()
...
+            }
+            else
+            {
+                fileSystemProviderMock.Setup(p => p.File.Exists("/.dockerenv")).Returns(false);
+
+            }
+
+            ClientDocumentHelper.SetEnvironmentVariableProvider(environmentVariableProviderMock.Object);
+            ClientDocumentHelper.SetFileSystemProvider(fileSystemProviderMock.Object);
+            using (new DisposableEnvironmentVariable("VERCEL"))
```
BorisDog, line 168: Ideally we'd just use `environmentVariableProviderMock` for "VERCEL".

src/MongoDB.Driver.Core/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -176,6 +193,23 @@ name switch
...
                 };
 
+            BsonDocument GetContainerDocument()
+            {
+                var isExecutionContainerDocker = __fileSystemProvider.File.Exists("/.dockerenv");
+                var isOrchestratorKubernetes = __environmentVariableProvider.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST") != null;
+
+                if (isExecutionContainerDocker || isOrchestratorKubernetes)
+                {
+                    return new BsonDocument
```
BorisDog, line 203: minor: styling shortcut for your consideration
```
return new() 
{
 ....
}
```

BorisDog (approved):
LGTM, some minor comments for your consideration.
File and environment vars mocking approach is a great addition.

tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
adelinowona: Yeah this is connected to replacing uses of Environment with EnvironmentVariableProvider in ClientDocumentHelper.cs

PR #1296, opened 2024-03-18
CSHARP-5004: Invoke all Drivers Evergreen Tools Scripts with Bash

sanych-sun:
Change sh to bash as per Driver ticket request.


adelinowona (approved):
LGTM

PR #1299, opened 2024-03-25
CSHARP-3995: Fix flaky pool-checkout-maxConnecting-is-enforced.json:maxConnecting_is_enforced

BorisDog:
These timings seem to solve the issue.

### DIFF
--- a/specifications/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-maxConnecting-is-enforced.json
+++ b/specifications/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-maxConnecting-is-enforced.json
@@ -19,7 +19,7 @@
       ],
       "closeConnection": false,
       "blockConnection": true,
-      "blockTimeMS": 750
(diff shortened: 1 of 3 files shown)


JamesKovacs (changes requested):
Given that these are unified test files, we shouldn't diverge from the tests in the specs repo. We should create a DRIVERS ticket and PR against the specs repo.

BorisDog:
Good question @JamesKovacs 
We are tracking this change in [DRIVERS-2874](https://jira.mongodb.org/browse/DRIVERS-2874).
To minimize the chance of further changes for all drivers, in addition to standalone experiments I suggested to run our tests for a week before changing the test for all drivers.

JamesKovacs (approved):
Given the plan is to run these changes in EG for a week before merging upstream in DRIVERS-2874, I'm good with these changes. Also discussed the switch to polling in the CMAP tests, which is alright since we are mostly testing connection pool logic, not SDAM logic.

PR #1301, opened 2024-04-01
Reduce dependencies by making them framework specific

thompson-tomo:
This will reduce the dependencies for net standard projects


BorisDog:
Thanks for submitting this PR @thompson-tomo.
In the upcoming major release we'll be updating our target frameworks, and also optimizing our dependencies accordingly.

We'll be closing this PR as this or similar optimization will be done as part of an upcoming change.

PR #1302, opened 2024-04-01
CSHARP-5033: Switch to License Expression

thompson-tomo:
This result's in the packages using a license expression so that license analysis can occur etc.


BorisDog:
Thanks @thompson-tomo for this PR.
It's has been merged as part of https://github.com/mongodb/mongo-csharp-driver/pull/1347 with minor addition.

PR #1303, opened 2024-04-02
Add solution DotSettings file

sanych-sun:
Added solution-level layer contains following customization:
- Live code templates for standard Equals implementation (based on #1273)
- File header template


JamesKovacs (approved):
LGTM

PR #1309, opened 2024-04-13
Update performance tests to run on net8.0

sanych-sun:
Change performance test to be run on net8.0 to follow the framework installed by install-dotnet function.

I've configured PR patch to include the perf tests variant to show it working now.


adelinowona (approved):
LGTM

PR #1305, opened 2024-04-11
CSHARP-5021: Replace use of DisposableEnvironmentVariable with EnvironmentVariableProvider

adelinowona:

PR #1311, opened 2024-04-18
CSHARP-5021: Use EnvironmentVariableProvider where relevant

adelinowona:
PR https://github.com/mongodb/mongo-csharp-driver/pull/1305 was accidentally merged when there was a failing test in evergreen. The merged was reverted and this PR is essentially same but with the fix for the failing test.


tests/MongoDB.Driver.Core.Tests/Core/Connections/ClientDocumentHelperTests.cs
```diff
@@ -31,6 +31,7 @@ public class ClientDocumentHelperTests
         [Fact]
         public void CreateClientDocument_should_return_expected_result()
         {
+            ClientDocumentHelper.Initialize();
```
adelinowona, line 34: Ensures that ClientDocumentHelper is starting from a clean slate and not using values set by other tests since ClientDocumentHelper is used as a static class. This brings the question of should we just make ClientDocumentHelper a static class?  because everything in it is defined as static yet it is not defined as a static class itself.

BorisDog (approved):
LGTM

PR #1312, opened 2024-04-18
CSHARP-4828: Handling errors in WithTransaction

adelinowona:
### Add error handling information to withTransaction docs
The withTransaction function docs now has more information on error handling to avoid an application entering an infinite loop in the event that a failed command aborts the transaction and causes the transaction to be retried. I've added "Docs Changes Needed" to the ticket description, asking the docs team to call this out when the transaction page is created.

### DIFF
--- a/src/MongoDB.Driver/IClientSession.cs
+++ b/src/MongoDB.Driver/IClientSession.cs
@@ -139,6 +139,15 @@ namespace MongoDB.Driver
         /// <summary>
         /// Executes a callback within a transaction, with retries if needed.
         /// </summary>
+        /// <remarks>
+        /// If a command inside the callback fails, it may cause the transaction on the server to be
+        /// aborted. This situation is normally handled transparently by the driver. However, if the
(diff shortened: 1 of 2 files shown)


JamesKovacs (changes requested):
`<remarks/>` look good. I would suggest expanding our `WithTransactionExample1.cs` to include proper error handling. Something like this:

```
                result = session.WithTransaction(
                    (s, ct) =>
                    {
                        try
                        {
                            collection1.InsertOne(s, new BsonDocument("abc", 1), cancellationToken: ct);
                            collection2.InsertOne(s, new BsonDocument("xyz", 999), cancellationToken: ct);
                        }
                        catch (MongoWriteException ex)
                        {
                            // Do something in response to the exception
                            throw; // NOTE: You must rethrow the exception otherwise an infinite loop can occur.
                        }
                        return "Inserted into collections in different databases";
                    },
                    transactionOptions,
                    cancellationToken);
```

JamesKovacs (approved):
LGTM

PR #1313, opened 2024-04-26
CSHARP-5067: Filters with boolean fields or properties should use correct representation of true.

rstam:


rstam:
Most of the work in this PR was writing new tests.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5067Tests.cs
```diff
@@ -0,0 +1,490 @@
...
+            var collection = GetCollection(linqProvider);
+
+            var queryable = justField ?
+                collection.AsQueryable().Where(x => x.BoolFieldRepresentedAsString) :
+                collection.AsQueryable().Where(x => x.BoolFieldRepresentedAsString == true);
+
+            var stages = Translate(collection, queryable);
+            var results = queryable.ToList();
+
+            if (linqProvider == LinqProvider.V2 && justField)
```
damieng, line 90: If the LINQ v2 provider is wrong and we don't intend on fixing it is it worth having tests that ensure it's wrong?

rstam: That's a fair question and I've asked myself that before.

Here's my thinking:

If this were a feature that was only supported in LINQ3 of course we wouldn't have tests showing that it didn't work in LINQ2.

But in cases like this one it mostly works in LINQ2, and it seems useful to document the few exceptions in the tests. If we didn't have these tests where LINQ2 does the wrong thing you might ask the reverse question: why aren't we testing this scenario in LINQ2.

Also, when LINQ2 is removed this anomaly will go away.

damieng (approved):

PR #1317, opened 2024-05-03
CSHARP-5071: Support string concatenation of mixed types.

rstam:


rstam:
Added tests for all combinations of string concatenation using `+` or `string.Concat` and implemented any missing support.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5071Tests.cs
```diff
@@ -0,0 +1,446 @@
...
+        [InlineData("stringproperty+stringproperty+intconstant", "{ $project : { _v : { $concat : ['$A', '$B', '3'] }, _id : 0 } }", "AB3", LinqProvider.V3)]
+        [InlineData("stringproperty+stringproperty+intproperty", "throws:InvalidOperationException", null, LinqProvider.V2)]
+        [InlineData("stringproperty+stringproperty+intproperty", "{ $project : { _v : { $concat : ['$A', '$B', { $toString : '$K' }] }, _id : 0 } }", "AB3", LinqProvider.V3)]
+        [InlineData("stringproperty+stringproperty+stringconstant", "{ $project : { __fld0 : { $concat : ['$A', '$B', 'Z'] }, _id : 0 } }", "ABZ", LinqProvider.V2)]
+        [InlineData("stringproperty+stringproperty+stringconstant", "{ $project : { _v : { $concat : ['$A', '$B', 'Z'] }, _id : 0 } }", "ABZ", LinqProvider.V3)]
+        [InlineData("stringproperty+stringproperty+stringproperty", "{ $project : { __fld0 : { $concat : ['$A', '$B', '$C'] }, _id : 0 } }", "ABC", LinqProvider.V2)]
+        [InlineData("stringproperty+stringproperty+stringproperty", "{ $project : { _v : { $concat : ['$A', '$B', '$C'] }, _id : 0 } }", "ABC", LinqProvider.V3)]
```
damieng, line 98: Given how much you need in an inline dat a+ the need for the big switch to convert I wonder if these would be better of as individual tests?

rstam: I started that way: writing a different test for each scenario, but quickly realized that the tests were all the same except for a few lines, so I prefer refactoring it this way.

The other reason I went this way is that I could scan the scenario strings in the InlineData and quickly confirm that I was in fact testing all the combinations I intended to.

rstam: I'm aware of MemberData and we've certainly used it before, but it's not usually my go to choice because it feels awkward to me.

And worse, Theory(s) that have non-primitive parameters aren't well supported by TestExplorer. There are ways around this but they require more code where you create an xUnit TestCase and calculate the test name yourself (as a string).

I can change this if you insist...

damieng (approved):

damieng (changes requested):
Just need to address those test failures.

damieng (approved):

PR #1318, opened 2024-05-03
CSHARP-5070: Add SBOM file

adelinowona:
Adds the generated SBOM file as part of the Silk prep work.

purls.txt - this file houses the [package URLs](https://github.com/package-url/purl-spec) for the third-party dependencies bundled in the driver. Presently, it includes the C# wrapper libmongocrypt nuget package since we bundle the libmongocrypt library dlls in that package. This is used when generating/updating sbom.json with the third-party dependencies.

sbom.json - This file serves as the SBOM-lite, documenting the third-party dependencies bundled within the shipped driver. Its purpose is vital for DevProd, as it's required to furnish to Silk.

For now, generating/updating sbom.json is to be done manually every time we update our third-party dependencies.

### DIFF
--- a/purls.txt
+++ b/purls.txt
@@ -1,0 +1,1 @@
+pkg:nuget/MongoDB.Libmongocrypt@1.8.2
--- a/sbom.json
+++ b/sbom.json
@@ -1,0 +1,82 @@
+{
+  "components": [
+    {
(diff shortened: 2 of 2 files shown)


JamesKovacs (approved):
LGTM. Please add a note to the C# Driver Release Process doc noting that these particular files must be updated when we update Libmongocrypt (if you haven't already).

PR #1321, opened 2024-05-07
CSHARP-5065: Dispose RawBsonDocument in ExplicitEncryptionLibMongoCryptController.cs

adelinowona:
- wrap uses of `RawBsonDocument` appropriately in using statements.

- Switch to using `BsonSerializer.Deserialize` in the `UnwrapValue` function instead of using `RawBsonDocument`. This helps us since we couldn't dispose of `RawBsonDocument` in this case as the returned value would be disposed as well. Also avoids us having to do cloning as a remedy if we were to still use `RawBsonDocument`.

Just want input on whether I'm missing something here by not using RawBsonDocument

### DIFF
--- a/src/MongoDB.Driver/Encryption/ExplicitEncryptionLibMongoCryptController.cs
+++ b/src/MongoDB.Driver/Encryption/ExplicitEncryptionLibMongoCryptController.cs
@@ -103,7 +103,7 @@ namespace MongoDB.Driver.Encryption
                 {
                     var wrappedKeyBytes = ProcessStates(context, _keyVaultNamespace.DatabaseNamespace.DatabaseName, cancellationToken);
 
-                    using var wrappedKeyDocument = new RawBsonDocument(wrappedKeyBytes);
+                    var wrappedKeyDocument = BsonSerializer.Deserialize<BsonDocument>(wrappedKeyBytes);
                     var keyId = UnwrapKeyId(wrappedKeyDocument);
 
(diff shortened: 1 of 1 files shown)


sanych-sun (approved):
LGTM

src/MongoDB.Driver/Encryption/ExplicitEncryptionLibMongoCryptController.cs
```diff
@@ -568,8 +568,8 @@ private Guid UnwrapKeyId(RawBsonDocument wrappedKeyDocument)
 
         private BsonValue UnwrapValue(byte[] encryptedWrappedBytes)
         {
-            var rawDocument = new RawBsonDocument(encryptedWrappedBytes);
-            return rawDocument["v"];
+            var bsonDocument = BsonSerializer.Deserialize<BsonDocument>(encryptedWrappedBytes);
```
BorisDog, line 571: Looks safe to me. I don't recall what was the motivation to use `RawBsonDocument`  as in most cases the `v` value will being read completely but consumer.
Maybe @rstam has some ideas.

rstam: I don't know why a `RawBsonDocument` is being used here. If the goal is to deserialize the key bytes we should just do that directly.

Not sure that is the goal though.

rstam (changes requested):

BorisDog (approved):
LGTM

rstam (approved):
LGTM

PR #1322, opened 2024-05-08
CSHARP-5066: Onboard .NET Driver to Papertrail Service

BorisDog:


evergreen/evergreen.yml
```diff
@@ -1785,7 +1807,10 @@ tasks:
           vars:
             PACKAGES_SOURCE: "https://www.myget.org/F/mongodb/api/v3/index.json"
             PACKAGES_SOURCE_KEY: ${myget_api_key}
-
+        - func: trace-artifacts
```
sanych-sun, line 1810: Please double-check if we need this step for myget packages, as such packages are intermediate and do not intended for production usage.

sanych-sun (approved):
LGTM + please check if we have to report MyGet packages.

evergreen/evergreen.yml
BorisDog: Thanks for pointing this out, no need for dev packages.

PR #1327, opened 2024-05-22
CSHARP-5087: Bump MacOS version used for tests

sanych-sun:
Remove workaround to set open files limit for Macos-14-ARM hosts.


JamesKovacs (commented):
LGTM, but we have a number of new variant failures on previously passing spec tests on MacOS. They don't seem related to this change, but should be investigated to verify.

PR #1328, opened 2024-05-23
CSHARP-5081: Support SelectMany inside Project/Select.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5081Tests.cs
```diff
@@ -0,0 +1,210 @@
...
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp5081Tests : Linq3IntegrationTest
+    {
+        [Theory]
+        [ParameterAttributeData]
+        public void SelectMany_chained_should_work(
```
rstam, line 30: This test merely confirms that SelectMany at the top level already worked.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5081Tests.cs
```diff
@@ -0,0 +1,210 @@
...
+                "Book 1 Chapter 2",
+                "Book 2 Chapter 1",
+                "Book 2 Chapter 2",
+                "Book 3 Chapter 1",
+                "Book 3 Chapter 2");
+        }
+
+        [Theory]
+        [ParameterAttributeData]
+        public void SelectMany_nested_should_work(
```
rstam, line 74: This is the test for a nested SelectMany in a LINQ query.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5081Tests.cs
```diff
@@ -0,0 +1,210 @@
...
+                var exception = Record.Exception(() => Translate(collection, queryable));
+                exception.Should().BeOfType<NotSupportedException>();
+            }
+            else
+            {
+                var stages = Translate(collection, queryable);
+                AssertStages(
+                    stages,
+                    "{ $project : { _v : { $reduce : { input : { $map : { input : '$Books', as : 'book', in : '$$book.Chapters.Title' } }, initialValue : [], in : { $concatArrays : ['$$value', '$$this'] } } }, _id : 0 } }",
+                    "{ $unwind : '$_v' }");
```
rstam, line 95: In the JIRA ticket you said that the two LINQ queries are "equivalent".

They are equivalent in the sense that they return the same results. But they are not equivalent in the sense that they express the query in completely different ways that end up in completely different MQL.

Clearly the first form is going to be much better. The second form is going to result in much larger intermediate documents in the pipeline, possibly to the extent that the query fails.

I don't think there is any point in trying to recognize that this second query could potentially be rewritten as the first one. I think that sort of analysis would be extremely difficult.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5081Tests.cs
```diff
@@ -0,0 +1,210 @@
...
+                var results = aggregate.ToList();
+                results.Should().HaveCount(2);
+                results[0].Should().Equal(
+                    "Book 1 Chapter 1",
+                    "Book 1 Chapter 2",
+                    "Book 2 Chapter 1",
+                    "Book 2 Chapter 2");
+                results[1].Should().Equal(
+                    "Book 3 Chapter 1",
+                    "Book 3 Chapter 2");
```
rstam, line 141: This test is similar to the second one, except that it uses `Project` instead of `SelectMany` at the top level. That means that the results have a different shape because the multiple result arrays are not being concatenated into a single array.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodCallExpressionToAggregationExpressionTranslator.cs
```diff
@@ -68,6 +68,7 @@ public static AggregationExpression Translate(TranslationContext context, Method
                 case "Reverse": return ReverseMethodToAggregationExpressionTranslator.Translate(context, expression);
                 case "Round": return RoundMethodToAggregationExpressionTranslator.Translate(context, expression);
                 case "Select": return SelectMethodToAggregationExpressionTranslator.Translate(context, expression);
+                case "SelectMany": return SelectManyMethodToAggregationExpressionTranslator.Translate(context, expression);
```
rstam, line 71: Support for nested `SelectMany` is a new feature.

LINQ2 does not support nested `SelectMany`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,65 @@
...
+                var asVar = selectorParameterSymbol.Var;
+                var valueVar = AstExpression.Var("value");
+                var thisVar = AstExpression.Var("this");
+                var ast = AstExpression.Reduce(
+                    input: AstExpression.Map(
+                        input: sourceTranslation.Ast,
+                        @as: asVar,
+                        @in: selectorTranslation.Ast),
+                    initialValue: new BsonArray(),
+                    @in: AstExpression.ConcatArrays(valueVar, thisVar));
```
rstam, line 58: This is a pretty inefficient implementation because it keeps creating longer and longer arrays as it reduces the multiple intermediate arrays into a single one.

I can't think of any other way to do this though.

JamesKovacs (approved):
LGTM

PR #1332, opened 2024-05-27
CSHARP-5338: Improve integration test performance by test fixtures

sanych-sun:
Most of the integration tests we have currently re-inserting the same data for each run. For the tests that is reading data (like a most LINQ tests) it's inefficient. Proposed changes let us utilize xUnit's Fixture to share the test data for the test collection. It means that Fixture will hold the test data, which will be inserted only one before tests will be run for a test class. Also fixture will be disposed after all tests were run, so we can cleanup the test collection.

I've choose 20 last tests in Jira folder because all of them are integration and refactored to use the approach. On mine local machine I saw like a 4 times improvement for this tests.
Unit-tests-before
Unit-tests-after

### DIFF
--- a/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
+++ b/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
@@ -1,0 +1,41 @@
+/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using MongoDB.Driver.Core.TestHelpers.Logging;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MongoDB.Driver.Tests
+{
+    [IntegrationTest]
+    public abstract class IntegrationTest<TFixture> : LoggableTestClass, IClassFixture<TFixture>
+        where TFixture : MongoDatabaseFixture
+    {
+        private readonly TFixture _fixture;
+
+        protected IntegrationTest(ITestOutputHelper testOutput, TFixture fixture, Action<RequireServer> requireServerCheck = null)
+            : base(testOutput, fixture.LogsAccumulator)
+        {
+            _fixture = fixture;
+            requireServerCheck?.Invoke(RequireServer.Check());
+            _fixture.Initialize();
+        }
+
+        public TFixture Fixture => _fixture;
+    }
+}
--- a/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
+++ b/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
@@ -30,7 +30,7 @@ namespace MongoDB.Driver.Tests
         {
             _fixture = fixture;
             requireServerCheck?.Invoke(RequireServer.Check());
-            _fixture.Initialize();
+            _fixture.BeforeTestCase();
         }
 
         public TFixture Fixture => _fixture;
--- a/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
+++ b/tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
@@ -14,22 +14,19 @@
 */
 
 using System;
-using MongoDB.Driver.Core.TestHelpers.Logging;
 using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
 using MongoDB.TestHelpers.XunitExtensions;
 using Xunit;
-using Xunit.Abstractions;
 
 namespace MongoDB.Driver.Tests
 {
     [IntegrationTest]
-    public abstract class IntegrationTest<TFixture> : LoggableTestClass, IClassFixture<TFixture>
+    public abstract class IntegrationTest<TFixture> : IClassFixture<TFixture>
         where TFixture : MongoDatabaseFixture
     {
         private readonly TFixture _fixture;
 
-        protected IntegrationTest(ITestOutputHelper testOutput, TFixture fixture, Action<RequireServer> requireServerCheck = null)
-            : base(testOutput, fixture.LogsAccumulator)
+        protected IntegrationTest(TFixture fixture, Action<RequireServer> requireServerCheck = null)
         {
             _fixture = fixture;
             requireServerCheck?.Invoke(RequireServer.Check());
--- a/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
@@ -21,8 +21,9 @@ namespace MongoDB.Driver.TestHelpers
 {
     public abstract class MongoCollectionFixture<TDocument> : MongoDatabaseFixture
     {
+        private readonly Lazy<IMongoCollection<TDocument>> _collection;
         private readonly string _collectionName;
-        private bool _collectionInitialized;
+        private bool _dataInitialized;
 
         protected MongoCollectionFixture(string collectionName = null)
         {
@@ -31,42 +32,35 @@ namespace MongoDB.Driver.TestHelpers
             {
                 throw new ArgumentNullException(nameof(collectionName) , "Cannot resolve the collection name. Try to specify the parameter explicitly");
             }
+
+            _collection = new Lazy<IMongoCollection<TDocument>>(CreateCollection);
         }
 
+        public IMongoCollection<TDocument> Collection => _collection.Value;
         public string CollectionName => _collectionName;
-
         protected abstract IEnumerable<TDocument> InitialData { get; }
 
-        public virtual bool ResetOnEachTestCase => false;
+        public virtual bool InitializeDataBeforeEachTestCase => false;
 
-        public override IMongoCollection<T> GetCollection<T>(string collectionName = null)
+        protected virtual IMongoCollection<TDocument> CreateCollection()
         {
-            if (typeof(T) != typeof(TDocument))
-            {
-                throw new ArgumentException("T must be of TDocument type");
-            }
-
-            if (!string.IsNullOrEmpty(collectionName))
-            {
-                throw new NotSupportedException("CollectionFixture does not support explicit collection name.");
-            }
-
-            var db = GetDatabase();
-            return db.GetCollection<T>(CollectionName);
+            return CreateCollection<TDocument>(_collectionName);
         }
 
         protected override void InitializeTestCase()
         {
-            if (ResetOnEachTestCase || !_collectionInitialized)
+            if (InitializeDataBeforeEachTestCase || !_dataInitialized)
             {
+                Collection.Database.DropCollection(_collectionName);
+                Collection.Database.CreateCollection(_collectionName);
+
                 var initialData = InitialData;
                 if (initialData != null)
                 {
-                    var collection = base.GetCollection<TDocument>(CollectionName);
-                    collection.InsertMany(initialData);
+                    Collection.InsertMany(initialData);
                 }
 
-                _collectionInitialized = true;
+                _dataInitialized = true;
             }
         }
 
--- a/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
@@ -39,12 +39,6 @@ namespace MongoDB.Driver.TestHelpers
 
         public virtual bool ResetOnEachTestCase => false;
 
-        public override void Dispose()
-        {
-            GetDatabase().DropCollection(_collectionName);
-            base.Dispose();
-        }
-
         public override IMongoCollection<T> GetCollection<T>(string collectionName = null)
         {
             if (typeof(T) != typeof(TDocument))
--- a/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
@@ -1,0 +1,85 @@
+/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using MongoDB.Driver.Tests;
+
+namespace MongoDB.Driver.TestHelpers
+{
+    public abstract class MongoCollectionFixture<TDocument> : MongoDatabaseFixture
+    {
+        private readonly string _collectionName;
+        private bool _collectionInitialized;
+
+        protected MongoCollectionFixture(string collectionName = null)
+        {
+            _collectionName = collectionName ?? GetCollectionName();
+            if (string.IsNullOrEmpty(_collectionName))
+            {
+                throw new ArgumentNullException(nameof(collectionName) , "Cannot resolve the collection name. Try to specify the parameter explicitly");
+            }
+        }
+
+        public string CollectionName => _collectionName;
+
+        protected abstract IEnumerable<TDocument> InitialData { get; }
+
+        public virtual bool ResetOnEachTestCase => false;
+
+        public override void Dispose()
+        {
+            GetDatabase().DropCollection(_collectionName);
+            base.Dispose();
+        }
+
+        public override IMongoCollection<T> GetCollection<T>(string collectionName = null)
+        {
+            if (typeof(T) != typeof(TDocument))
+            {
+                throw new ArgumentException("T must be of TDocument type");
+            }
+
+            if (!string.IsNullOrEmpty(collectionName))
+            {
+                throw new NotSupportedException("CollectionFixture does not support explicit collection name.");
+            }
+
+            var db = GetDatabase();
+            return db.GetCollection<T>(CollectionName);
+        }
+
+        protected override void InitializeTestCase()
+        {
+            if (ResetOnEachTestCase || !_collectionInitialized)
+            {
+                var initialData = InitialData;
+                if (initialData != null)
+                {
+                    var collection = base.GetCollection<TDocument>(CollectionName);
+                    collection.InsertMany(initialData);
+                }
+
+                _collectionInitialized = true;
+            }
+        }
+
+        private string GetCollectionName()
+        {
+            var currentType = GetType();
+            return currentType.DeclaringType?.Name;
+        }
+    }
+}
--- a/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
@@ -22,52 +22,54 @@ namespace MongoDB.Driver.TestHelpers
     public abstract class MongoCollectionFixture<TDocument> : MongoDatabaseFixture
     {
         private readonly Lazy<IMongoCollection<TDocument>> _collection;
-        private readonly string _collectionName;
         private bool _dataInitialized;
 
-        protected MongoCollectionFixture(string collectionName = null)
+        protected MongoCollectionFixture()
         {
-            _collectionName = collectionName ?? GetCollectionName();
-            if (string.IsNullOrEmpty(_collectionName))
-            {
-                throw new ArgumentNullException(nameof(collectionName) , "Cannot resolve the collection name. Try to specify the parameter explicitly");
-            }
-
             _collection = new Lazy<IMongoCollection<TDocument>>(CreateCollection);
         }
 
         public IMongoCollection<TDocument> Collection => _collection.Value;
-        public string CollectionName => _collectionName;
+
         protected abstract IEnumerable<TDocument> InitialData { get; }
 
         public virtual bool InitializeDataBeforeEachTestCase => false;
-
-        protected virtual IMongoCollection<TDocument> CreateCollection()
-        {
-            return CreateCollection<TDocument>(_collectionName);
-        }
 
         protected override void InitializeTestCase()
         {
             if (InitializeDataBeforeEachTestCase || !_dataInitialized)
             {
-                Collection.Database.DropCollection(_collectionName);
-                Collection.Database.CreateCollection(_collectionName);
+                Collection.Database.DropCollection(Collection.CollectionNamespace.CollectionName);
 
-                var initialData = InitialData;
-                if (initialData != null)
+                if (InitialData == null)
                 {
-                    Collection.InsertMany(initialData);
+                    Collection.Database.CreateCollection(Collection.CollectionNamespace.CollectionName);
+                }
+                else
+                {
+                    Collection.InsertMany(InitialData);
                 }
 
                 _dataInitialized = true;
             }
         }
 
-        private string GetCollectionName()
+        protected virtual string GetCollectionName()
         {
             var currentType = GetType();
-            return currentType.DeclaringType?.Name;
+            var result = currentType.DeclaringType?.Name;
+
+            if (string.IsNullOrEmpty(result))
+            {
+                throw new InvalidOperationException("Cannot resolve the collection name. Try to override GetCollectionName for custom collection name resolution.");
+            }
+
+            return result;
+        }
+
+        private IMongoCollection<TDocument> CreateCollection()
+        {
+            return CreateCollection<TDocument>(GetCollectionName());
         }
     }
 }
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -22,14 +22,19 @@ namespace MongoDB.Driver.Tests
 {
     public class MongoDatabaseFixture : IDisposable
     {
-        private static readonly string __timeStamp = DateTime.Now.ToString("yyyyMMdd-HHmmss");
+        private static readonly string __timeStamp = DateTime.Now.ToString("MMddHHmm");
 
-        private readonly string _databaseName = $"CSharpDriver-{__timeStamp}";
+        private readonly Lazy<IMongoClient> _client;
+        private readonly Lazy<IMongoDatabase> _database;
+        private readonly string _databaseName = $"Tests{__timeStamp}";
+        private bool _fixtureInialized;
         private readonly HashSet<string> _usedCollections = new();
-        private IMongoClient _client;
 
         public MongoDatabaseFixture()
         {
+            _client = new Lazy<IMongoClient>(CreateClient);
+            _database = new Lazy<IMongoDatabase>(CreateDatabase);
+
             var logCategoriesToExclude = new[]
             {
                 "MongoDB.Command",
@@ -39,61 +44,66 @@ namespace MongoDB.Driver.Tests
             LogsAccumulator = new XUnitOutputAccumulator(logCategoriesToExclude);
         }
 
+        public IMongoClient Client => _client.Value;
+        public IMongoDatabase Database => _database.Value;
+
         internal XUnitOutputAccumulator LogsAccumulator { get; }
 
         public virtual void Dispose()
         {
-            if (_client == null)
+            var client = _client.IsValueCreated ? _client.Value : null;
+            var database = _database.IsValueCreated ? _database.Value : null;
+
+            if (database != null)
             {
-                return;
+                foreach (var collection in _usedCollections)
+                {
+                    database.DropCollection(collection);
+                }
             }
 
-            var database = GetDatabase();
-            foreach (var collection in _usedCollections)
-            {
-                database.DropCollection(collection);
-            }
-
-            var client = _client;
-            _client = null;
-            client.Dispose();
+            client?.Dispose();
         }
 
-        public IMongoClient GetClient()
-            => _client;
-
-        public virtual IMongoDatabase GetDatabase()
-            => GetClient().GetDatabase(_databaseName);
-
-        public virtual IMongoCollection<T> GetCollection<T>(string collectionName = null)
+        public virtual IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
         {
             if (string.IsNullOrEmpty(collectionName))
             {
                 var stack = new System.Diagnostics.StackTrace();
-                var frame = stack.GetFrame(2); // skip 2 frames to get the calling method info (this method and IntegrationTest method)
+                var frame = stack.GetFrame(1); // skip 1 frame to get the calling method info
                 var method = frame.GetMethod();
-                collectionName = $"{method.DeclaringType.Name}.{method.Name}_{typeof(T).Name}";
+                collectionName = $"{method.DeclaringType.Name}_{method.Name}";
             }
 
-            var db = GetDatabase();
-            db.DropCollection(collectionName);
+            Database.DropCollection(collectionName);
             _usedCollections.Add(collectionName);
-            return db.GetCollection<T>(collectionName);
+
+            return Database.GetCollection<TDocument>(collectionName);
         }
 
-        protected virtual void ConfigureMongoClient(MongoClientSettings settings)
+        protected virtual void ConfigureMongoClientSettings(MongoClientSettings settings)
         {
             settings.LoggingSettings = new LoggingSettings(new XUnitLoggerFactory(LogsAccumulator), 10000); // Spec test require larger truncation default
         }
 
+        protected virtual IMongoClient CreateClient()
+        {
+            var clientSettings = DriverTestConfiguration.GetClientSettings();
+            ConfigureMongoClientSettings(clientSettings);
+            return new MongoClient(clientSettings);
+        }
+
+        protected virtual IMongoDatabase CreateDatabase()
+        {
+            return Client.GetDatabase(_databaseName);
+        }
+
         internal void Initialize()
         {
-            if (_client == null)
+            if (!_fixtureInialized)
             {
-                var clientSettings = DriverTestConfiguration.GetClientSettings();
-                ConfigureMongoClient(clientSettings);
-                _client = new MongoClient(clientSettings);
                 InitializeFixture();
+                _fixtureInialized = true;
             }
 
             InitializeTestCase();
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -1,0 +1,108 @@
+/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.TestHelpers.Logging;
+
+namespace MongoDB.Driver.Tests
+{
+    public class MongoDatabaseFixture : IDisposable
+    {
+        private static readonly string __timeStamp = DateTime.Now.ToString("yyyyMMdd-HHmmss");
+
+        private readonly string _databaseName = $"CSharpDriver-{__timeStamp}";
+        private readonly HashSet<string> _usedCollections = new();
+        private IMongoClient _client;
+
+        public MongoDatabaseFixture()
+        {
+            var logCategoriesToExclude = new[]
+            {
+                "MongoDB.Command",
+                "MongoDB.Connection"
+            };
+
+            LogsAccumulator = new XUnitOutputAccumulator(logCategoriesToExclude);
+        }
+
+        internal XUnitOutputAccumulator LogsAccumulator { get; }
+
+        public virtual void Dispose()
+        {
+            if (_client == null)
+            {
+                return;
+            }
+
+            var database = GetDatabase();
+            foreach (var collection in _usedCollections)
+            {
+                database.DropCollection(collection);
+            }
+
+            var client = _client;
+            _client = null;
+            client.Dispose();
+        }
+
+        public IMongoClient GetClient()
+            => _client;
+
+        public virtual IMongoDatabase GetDatabase()
+            => GetClient().GetDatabase(_databaseName);
+
+        public virtual IMongoCollection<T> GetCollection<T>(string collectionName = null)
+        {
+            if (string.IsNullOrEmpty(collectionName))
+            {
+                var stack = new System.Diagnostics.StackTrace();
+                var frame = stack.GetFrame(2); // skip 2 frames to get the calling method info (this method and IntegrationTest method)
+                var method = frame.GetMethod();
+                collectionName = $"{method.DeclaringType.Name}.{method.Name}_{typeof(T).Name}";
+            }
+
+            var db = GetDatabase();
+            db.DropCollection(collectionName);
+            _usedCollections.Add(collectionName);
+            return db.GetCollection<T>(collectionName);
+        }
+
+        protected virtual void ConfigureMongoClient(MongoClientSettings settings)
+        {
+            settings.LoggingSettings = new LoggingSettings(new XUnitLoggerFactory(LogsAccumulator), 10000); // Spec test require larger truncation default
+        }
+
+        internal void Initialize()
+        {
+            if (_client == null)
+            {
+                var clientSettings = DriverTestConfiguration.GetClientSettings();
+                ConfigureMongoClient(clientSettings);
+                _client = new MongoClient(clientSettings);
+                InitializeFixture();
+            }
+
+            InitializeTestCase();
+        }
+
+        protected virtual void InitializeFixture()
+        {}
+
+        protected virtual void InitializeTestCase()
+        {}
+    }
+}
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -53,7 +53,7 @@ namespace MongoDB.Driver.Tests
         {
             if (string.IsNullOrEmpty(collectionName))
             {
-                throw new ArgumentNullException(nameof(collectionName));
+                throw new ArgumentException($"{nameof(collectionName)} should be non-empty string.", nameof(collectionName));
             }
 
             Database.DropCollection(collectionName);
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -73,7 +73,7 @@ namespace MongoDB.Driver.Tests
             return Client.GetDatabase(_databaseName);
         }
 
-        internal void Initialize()
+        internal void BeforeTestCase()
         {
             if (!_fixtureInialized)
             {
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -26,7 +26,7 @@ namespace MongoDB.Driver.Tests
 
         private readonly Lazy<IMongoClient> _client;
         private readonly Lazy<IMongoDatabase> _database;
-        private readonly string _databaseName = $"Tests{__timeStamp}";
+        private readonly string _databaseName = $"CSTests{__timeStamp}";
         private bool _fixtureInialized;
         private readonly HashSet<string> _usedCollections = new();
 
@@ -65,7 +65,7 @@ namespace MongoDB.Driver.Tests
             client?.Dispose();
         }
 
-        public virtual IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
+        protected IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
         {
             if (string.IsNullOrEmpty(collectionName))
             {
--- a/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
+++ b/tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
@@ -15,8 +15,6 @@
 
 using System;
 using System.Collections.Generic;
-using MongoDB.Driver.Core.Configuration;
-using MongoDB.Driver.Core.TestHelpers.Logging;
 
 namespace MongoDB.Driver.Tests
 {
@@ -34,26 +32,14 @@ namespace MongoDB.Driver.Tests
         {
             _client = new Lazy<IMongoClient>(CreateClient);
             _database = new Lazy<IMongoDatabase>(CreateDatabase);
-
-            var logCategoriesToExclude = new[]
-            {
-                "MongoDB.Command",
-                "MongoDB.Connection"
-            };
-
-            LogsAccumulator = new XUnitOutputAccumulator(logCategoriesToExclude);
         }
 
         public IMongoClient Client => _client.Value;
         public IMongoDatabase Database => _database.Value;
 
-        internal XUnitOutputAccumulator LogsAccumulator { get; }
-
         public virtual void Dispose()
         {
-            var client = _client.IsValueCreated ? _client.Value : null;
             var database = _database.IsValueCreated ? _database.Value : null;
-
             if (database != null)
             {
                 foreach (var collection in _usedCollections)
@@ -61,8 +47,6 @@ namespace MongoDB.Driver.Tests
                     database.DropCollection(collection);
                 }
             }
-
-            client?.Dispose();
         }
 
         protected IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
@@ -81,17 +65,8 @@ namespace MongoDB.Driver.Tests
             return Database.GetCollection<TDocument>(collectionName);
         }
 
-        protected virtual void ConfigureMongoClientSettings(MongoClientSettings settings)
-        {
-            settings.LoggingSettings = new LoggingSettings(new XUnitLoggerFactory(LogsAccumulator), 10000); // Spec test require larger truncation default
-        }
-
         protected virtual IMongoClient CreateClient()
-        {
-            var clientSettings = DriverTestConfiguration.GetClientSettings();
-            ConfigureMongoClientSettings(clientSettings);
-            return new MongoClient(clientSettings);
-        }
+            => DriverTestConfiguration.Client;
 
         protected virtual IMongoDatabase CreateDatabase()
         {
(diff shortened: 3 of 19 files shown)


tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,191 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Linq;
+using MongoDB.Driver.Linq.Linq3Implementation;
+using MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToExecutableQueryTranslators;
+using Xunit;
+
+namespace MongoDB.Driver.Tests
+{
+    [Trait("Category", "Integration")]
```
sanych-sun, line 28: I would like to have this category assigned to all tests that require database connection. In future we might use it to optimize test variants. So we can filter down functional tests from integration tests.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,191 @@
...
+        where TFixture : TemporaryDatabaseFixture
+    {
+        protected IntegrationTest(TFixture fixture)
+        {
+            Fixture = fixture;
+        }
+
+        protected TFixture Fixture { get; }
+
+        protected void AssertStages(IEnumerable<BsonDocument> stages, params string[] expectedStages)
```
sanych-sun, line 39: All `Assert-` and `Translate-` methods were simply copied from existing `Linq3IntegrationTest` base class (which probably has to be removed)

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Concurrent;
+using MongoDB.Driver.Linq;
+
+namespace MongoDB.Driver.Tests
+{
+    public class TemporaryDatabaseFixture: IDisposable
```
sanych-sun, line 22: There is no direct usage for this fixture for now, but the idea was to have it for test classes that require more comprehensive test data (more then 1 collection).

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
```diff
@@ -18,19 +18,25 @@
...
 using MongoDB.Bson;
 using MongoDB.Driver.Linq;
+using MongoDB.Driver.TestHelpers;
 using MongoDB.TestHelpers.XunitExtensions;
 using Xunit;
 
 namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
 {
-    public class CSharp702Tests : Linq3IntegrationTest
+    public class CSharp702Tests : IntegrationTest<CSharp702Tests.TestDataFixture>
```
sanych-sun, line 27: The main difference with comparing to the current test class structure:
1) Test class should be inherited from `IntegrationTest` base class
2) For typical test class that uses only 1 set of data we should create TestDataFixture as nested class and provide it as a type parameter for base IntegrationTest
3) Test class should have ctor to pass the Fixture created by xUnit infrastructure
4) Test collection is available through the Fixture property
5) Test data should be moved from GetCollection method to TestDataFixture

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
```
sanych-sun, line 40: BTW, we already have 2 set of such methods: in `Linq3IntegrationTest` class and also in `Linq3TestHelpers` static class. We probably ought to standardize the approach and have either static helper class or use base `IntegrationTest` class.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected TemporaryCollectionFixture()
+        {
+            _initCollection = new Lazy<bool>(InitCollection, LazyThreadSafetyMode.ExecutionAndPublication);
+            _collectionName = new Lazy<string>(GetDeclaringTypeName);
+        }
+
+        protected abstract IEnumerable<TModel> GetInitialData();
+
+        protected virtual string CollectionName => _collectionName.Value;
```
sanych-sun, line 37: Default collection name would be taken from the parent class name, as I suggest to put all test fixtures as a nested class of the test class.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4772Tests.cs
```diff
@@ -192,49 +198,49 @@ public class CSharp4772Tests : Linq3IntegrationTest
...
         {
             public int Id { get; set; }
-            public Profile[] ProfilesArray { get; set; }
-            public List<Profile> ProfilesList { get; set; }
+            public ProfileModel[] ProfilesArray { get; set; }
+            public List<ProfileModel> ProfilesList { get; set; }
         }
 
-        private class Profile
+        public class ProfileModel
```
rstam, line 208: Do you remember why you renamed the class?

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,191 @@
...
+            LinqProvider linqProvider)
+        {
+            var documentSerializer = collection.DocumentSerializer;
+            var serializerRegistry = BsonSerializer.SerializerRegistry;
+            var renderedProjection = projection.RenderForFind(documentSerializer, serializerRegistry, linqProvider);
+            return renderedProjection.Document;
+        }
+    }
+
+    public abstract class IntegrationTest : IntegrationTest<TemporaryDatabaseFixture>
```
rstam, line 184: This class is not used.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using MongoDB.Driver.Linq;
+using MongoDB.Driver.Tests;
+
+namespace MongoDB.Driver.TestHelpers
+{
+    public abstract class TemporaryCollectionFixture<TModel> : TemporaryDatabaseFixture
```
rstam, line 24: We should use `TDocument` instead of `TModel`.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        private readonly Lazy<string> _collectionName;
+        private readonly Lazy<bool> _initCollection;
```
rstam, line 27: I would name this field `_collectionInitializer`. Fields names are usually nouns.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        protected virtual string CollectionName => _collectionName.Value;
+
+        public IMongoCollection<TModel> GetCollection(LinqProvider provider = LinqProvider.V3)
+        {
+            // should always call this to ensure the collection is being initialized with the data
+            _ = _initCollection.Value;
+            return GetCollection<TModel>(CollectionName, provider);
+        }
+
+        public IMongoCollection<T> GetCollection<T>(LinqProvider provider = LinqProvider.V3)
```
rstam, line 46: I would get rid of this overload.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        private readonly Lazy<string> _collectionName;
+        private readonly Lazy<bool> _initCollection;
+
+        protected TemporaryCollectionFixture()
+        {
+            _initCollection = new Lazy<bool>(InitCollection, LazyThreadSafetyMode.ExecutionAndPublication);
+            _collectionName = new Lazy<string>(GetDeclaringTypeName);
+        }
+
+        protected abstract IEnumerable<TModel> GetInitialData();
```
rstam, line 35: Add another optional overload:

    protected virtual IEnumerable<BsonDocument> GetRawInitialData();

to be used in the rare cases where we need to initialize test data with raw BsonDocuments.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
BorisDog, line 184: This one is not needed?

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected virtual string CollectionName => _collectionName.Value;
+
+        public IMongoCollection<TModel> GetCollection(LinqProvider provider = LinqProvider.V3)
+        {
+            // should always call this to ensure the collection is being initialized with the data
+            _ = _initCollection.Value;
```
rstam, line 42: I would replace the previous two lines with:

    EnsureCollectionIsInitialized();

and add a helper method:

    private void EnsureCollectionIsInitialized()
    {
        _ = _collectionInitializer.Value;
    }

to encapsulate the trickiness of using a Lazy field to ensure the collection is initialized.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public IMongoCollection<T> GetCollection<T>(LinqProvider provider = LinqProvider.V3)
+        {
+            // should always call this to ensure the collection is being initialized with the data
+            _ = _initCollection.Value;
+            return GetCollection<T>(CollectionName, provider);
+        }
+
+        private string GetDeclaringTypeName()
```
rstam, line 53: I would rename this method `GetCollectionName`.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            var currentType = GetType();
+            return currentType.DeclaringType?.Name;
+        }
+
+        private bool InitCollection()
```
rstam, line 59: I would rename this method `InitializeCollection`.

Eschew abbreviations.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            if (string.IsNullOrEmpty(CollectionName))
+            {
+                throw new InvalidOperationException("Cannot resolve the collection name. Try to override the CollectionName property and return the name explicitly");
+            }
+
+            GetDatabase().DropCollection(CollectionName);
+            var collection = GetCollection<TModel>(CollectionName);
```
rstam, line 67: `var collection` can be moved inside the `if` statement.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+            var initialData = GetInitialData();
+            if (initialData != null)
+            {
+                collection.InsertMany(initialData);
+            }
```
rstam, line 72: Add:

    var rawInitialData = GetRawInitialData();
    if (rawInitialData != null)
    {
        var rawCollection = GetCollection<BsonDocument>(CollectionName);
        rawCollection.InsertMany(rawInitialData);
    }

to handle the rare cases where we want to initialize the collection with raw BsonDocuments.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private static readonly string __timeStamp = DateTime.Now.ToString("yyyyMMdd-HHmmss");
+
+        protected readonly string _databaseName;
+        private readonly ConcurrentBag<string> _createdCollections = new();
```
rstam, line 27: Rename field `_createdCollectionNames` because we are only storing the names, not the collections.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+        public virtual void Dispose()
+        {
+            var database = MongoDbClient.GetDatabase(_databaseName);
+            foreach (var collection in _createdCollections)
+            {
+                database.DropCollection(collection);
+            }
+        }
+
+        public IMongoClient MongoDbClient => DriverTestConfiguration.Client;
```
rstam, line 43: I would remove this property.

We want to encourage using `GetClient(linqProvider)` and not using this property by mistake.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
rstam: Maybe fix the class name while we're at it? Doesn't match the filename.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
BorisDog, line 24: We might need to readjust the naming, as more tests are migrated to use this approach. Something like ReadOnlyCollectionFixture...

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
```diff
@@ -111,27 +117,26 @@ public class CSharp702Tests : Linq3IntegrationTest
...
-        }
-
         private class Model
         {
             public int Id { get; set; }
             public List<string> List { get; set; }
             public HashSet<string> HashSet { get; set; }
         }
+
+        public class TestDataFixture : TemporaryCollectionFixture<BsonDocument>
```
rstam, line 127: public class CollectionFixture : TemporaryCollectionFixture<MyModel>

I would name this class `CollectionFixture` rather than `TestDataFixture`. To me the most important thing is that it is a "collection fixture". Initializing the test data is a secondary function.

Or `Model` if you don't get the same naming conflict with `Amazon,SecurityServices` that I was getting.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
```diff
@@ -111,27 +117,26 @@ public class CSharp702Tests : Linq3IntegrationTest
...
+        {
+            protected override IEnumerable<BsonDocument> GetInitialData()
```
rstam, line 129: `GetRawInitialData`

BorisDog (commented):
Looks like a great improvement to our tests. Waiting to see how faster our tests get when the rest of the tests are migrated.

rstam (commented):
Overall I think this approach is great! Nice!

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I like the idea to derive `Linq3IntegrationTest` from `IntegrationTest`. Will check if we can do it now, or should do it as the very last step of the refactoring if there are any conflicts.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: It will be used for tests that uses `TemporaryDatabaseFixture` instead of `TemporaryCollectionFixture`. I'll show an example in next commit.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
```diff
@@ -48,7 +54,7 @@ public class CSharp702Tests : Linq3IntegrationTest
         public void Query2_using_list_should_work(
             [Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
         {
-            var collection = GetCollection(linqProvider);
+            var collection = Fixture.GetCollection<Model>(linqProvider);
```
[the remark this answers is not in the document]
sanych-sun: Yes, in this particular test we are inserting the data using one model, but want to read the data as another.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4772Tests.cs
sanych-sun: The name conflict with some namespace. I did not investigate this deep, just renamed the model to get rid of conflict.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I'll show the usage for the case when test class do not want to initialize the collection, but want to have db for inserts or another data modification operations.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: I'll try to remove it by your GetRawInitialData suggestion. Have to investigate existing tests if it will cover all existing cases.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
rstam: Another option would be to make `InitializeCollection` virtual and in the rare case where a test class needs to initialize the collection with non-standard documents it can simply override `InitializeCollection` instead of `GetInitialDocuments`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
sanych-sun: Class was renamed.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
sanych-sun: Done

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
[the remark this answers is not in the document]
sanych-sun: I've decided to make InitializeCollection virtual, so derived fixtures are free to use any initialization logic.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4702Tests.cs
sanych-sun: InitializeCollection is virtual now as per your another suggestion.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: Removed.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: InitializeCollection is virtual now as per your another suggestion.

tests/MongoDB.Driver.TestHelpers/TemporaryCollectionFixture.cs
sanych-sun: I like idea of making `InitializeCollection` virtual more, because in such case derived fixture free to have any custom initialization logic, not just `BsonDocument` scenario.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: Please find example of usage in `CSharp4066Tests` test class.

sanych-sun: Please find example of usage in `CSharp4066Tests` test class.

tests/MongoDB.TestHelpers/XunitExtensions/IntegrationTestAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+namespace MongoDB.TestHelpers.XunitExtensions
+{
+    public class IntegrationTestAttribute : CategoryAttribute
+    {
+        public IntegrationTestAttribute()
+            : base("Integration")
+        {
+
```
BorisDog, line 23: minor: remove empty line

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
BorisDog: Maybe a better naming to reflect that. And then why not having matching class for `TemporaryCollectionFixture` ?

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var database = GetDatabase();
+            foreach (var collection in _createdCollectionNames)
+            {
+                database.DropCollection(collection);
+            }
+        }
+
+        public IMongoClient GetClient(LinqProvider provider)
+            => DriverTestConfiguration.GetLinqClient(provider);
```
BorisDog, line 39: The question regarding the client still remains. Do we want to use this opportunity to start working with disposable clients?

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+        private readonly ConcurrentBag<string> _createdCollectionNames = new();
+
+        public virtual void Dispose()
+        {
+            var database = GetDatabase();
+            foreach (var collection in _createdCollectionNames)
+            {
+                database.DropCollection(collection);
+            }
+        }
```
BorisDog, line 36: Why not to drop the database instead?

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public IMongoDatabase GetDatabase(LinqProvider provider = LinqProvider.V3)
+            => GetClient(provider).GetDatabase(_databaseName);
+
+        public IMongoCollection<T> GetCollection<T>(string collectionName = null, LinqProvider provider = LinqProvider.V3)
+        {
+            if (string.IsNullOrEmpty(collectionName))
+            {
+                var stack = new System.Diagnostics.StackTrace();
+                var frame = stack.GetFrame(1); // skip 1 frame to get the calling method info
```
BorisDog, line 49: Nice,
I wanted to suggest `[CallerMemberName]` , but we having type name is better.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,192 @@
...
+using MongoDB.Driver.Linq;
+using MongoDB.Driver.Linq.Linq3Implementation;
+using MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToExecutableQueryTranslators;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests
+{
+    [IntegrationTest]
+    public abstract class IntegrationTest<TFixture> : IClassFixture<TFixture>
```
BorisDog, line 30: Probably worth inheriting for `LoggableTestClass`, to have logging and timeouts for all.
Ideally all test would derive from   `LoggableTestClass`.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+                var method = frame.GetMethod();
+                collectionName = $"{method.DeclaringType.Name}.{method.Name}";
+            }
```
BorisDog, line 52: It's worth having some tests that consume same collection, overwrite same data. To see how we avoid collisions. 
So we are ready for parallelization.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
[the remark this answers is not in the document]
sanych-sun: Done. Client(s) will be created per fixture and will be disposed together with fixture. See TemporaryDatabaseFixture class for implementation.

tests/MongoDB.TestHelpers/XunitExtensions/IntegrationTestAttribute.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
sanych-sun: Done. See `TemporaryDatabaseFixture` for implementation.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
sanych-sun: Currently it would be almost the same, but if we approach running in parallel task - then it will be safer to remove the collection as database might be in use by other tests/fixtures.

tests/MongoDB.Driver.TestHelpers/TemporaryDatabaseFixture.cs
sanych-sun: For test Theories - yep. It will be more complicated, as we will need to add some extra suffixes to the names probably (based on test parameters? or test sequential number). Will try to find a way to work with this.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: `IntegrationTest` is not inhireted from `LoggableTestClass`. Let's discuss offline if we need some extra functionality around.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I've removed the class for now to avoid confusions. We might consider to re-introduce it later.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I've decided to introduce new `LinqIntegrationTest` base class.For 2 reasons:

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I'll try to consolidate all usages into the just introduced LinqIntegrationTest base class and remove "old" methods.

CSharpDriver.sln.DotSettings
```diff
@@ -1,4 +1,14 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
sanych-sun, line 1: This file is really hard to review, but the needed to let Rider/ReSharper add Integration Tests from template. I can demo how it works on some stand up.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: I've removed the class to avoid confusion.

tests/MongoDB.Driver.TestHelpers/CollectionFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                throw new ArgumentNullException(nameof(collectionName) , "Cannot resolve the collection name. Try to specify the parameter explicitly");
+            }
+        }
+
+        public string CollectionName => _collectionName;
+
+        public virtual bool ResetOnEachGet => false;
+
+        public override IMongoCollection<T> GetCollection<T>(Action<MongoClientSettings> configure = null, string collectionName = null)
```
BorisDog, line 40: Do we have cases where T != TDocument?

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            return _fixture.GetMongoClient(settings =>
+            {
+                settings.LoggingSettings = LoggingSettings;
+                configure?.Invoke(settings);
+            });
+        }
+
+        public IMongoDatabase GetDatabase(Action<MongoClientSettings> configure = null)
+        {
+            return _fixture.GetDatabase(settings =>
```
BorisDog, line 47: minor: expression body ?

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public virtual IMongoDatabase GetDatabase(Action<MongoClientSettings> configure = null)
+            => GetMongoClient(configure).GetDatabase(_databaseName);
+
+        public virtual IMongoCollection<T> GetCollection<T>(Action<MongoClientSettings> configure = null, string collectionName = null)
+        {
+            if (string.IsNullOrEmpty(collectionName))
+            {
+                var stack = new System.Diagnostics.StackTrace();
+                var frame = stack.GetFrame(2); // skip 2 frame to get the calling method info (this method and IntegrationTest method)
```
BorisDog, line 52: nit: `frames`

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly HashSet<string> _usedCollections = new();
+
+        public virtual void Dispose()
+        {
+            var database = GetDatabase();
+            foreach (var collection in _usedCollections)
+            {
+                database.DropCollection(collection);
+            }
+        }
```
BorisDog, line 35: Don't we want just to drop the DB instead?

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace MongoDB.Driver.Tests
+{
+    public class DatabaseFixture: IDisposable
+    {
+        private static readonly string __timeStamp = DateTime.Now.ToString("yyyyMMdd-HHmmss");
+
+        private readonly string _databaseName = $"CsharpDriver-{__timeStamp}";
```
BorisDog, line 25: nit: `CSharp`

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly HashSet<string> _usedCollections = new();
+
+        public virtual void Dispose()
+        {
```
BorisDog, line 29: Dispose the client.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,63 @@
...
+using MongoDB.Driver.Core.TestHelpers.Logging;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MongoDB.Driver.Tests
+{
+    [IntegrationTest]
+    public abstract class IntegrationTest<TFixture> : LoggableTestClass, IClassFixture<TFixture>
+        where TFixture : DatabaseFixture
```
BorisDog, line 26: Do we want to consider reusing the client between tests as well (`ICollectionFixture`)?
So we don't dispose it after each tests class. For example single client for all `LinqIntegrationTest`?

BorisDog (changes requested):

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: Done. Now we have a single client per Fixture.

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
sanych-sun: Done.

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
sanych-sun: Done

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
sanych-sun: Discussed, it's better to remove collections just if we will ever decide to implement tests in parallel.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
sanych-sun: This method is gone, by exposing Fixture itself.

tests/MongoDB.Driver.TestHelpers/CollectionFixture.cs
sanych-sun: This method signature is inherited from DatabaseFixture. Currently we throw if  T != TDocument, but it might be changed as we proceed to convert more tests.

tests/MongoDB.Driver.TestHelpers/IntegrationTest.cs
```diff
@@ -0,0 +1,63 @@
...
+        public IMongoDatabase GetDatabase(Action<MongoClientSettings> configure = null)
+        {
+            return _fixture.GetDatabase(settings =>
+            {
+                settings.LoggingSettings = LoggingSettings;
+                configure?.Invoke(settings);
+            });
+        }
+
+        public IMongoCollection<T> GetCollection<T>(Action<MongoClientSettings> configure = null, string collectionName = null)
```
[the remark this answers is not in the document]
sanych-sun: I've decided to expose Fixture itself from the IntegrationTest class. The decision is motivated by the idea of introducing using fixtures for tests that need more then just single collection. Example  could be find in `CSharp5231ests`.

tests/MongoDB.Driver.TestHelpers/DatabaseFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+
+namespace MongoDB.Driver.Tests
+{
+    public class DatabaseFixture: IDisposable
```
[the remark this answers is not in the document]
sanych-sun: Fixed

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+                var clientSettings = DriverTestConfiguration.GetClientSettings();
+                ConfigureMongoClient(clientSettings);
+                _client = new MongoClient(clientSettings);
+                InitializeFixture();
+            }
+
+            InitializeTestCase();
+        }
+
+        protected virtual void InitializeFixture()
```
sanych-sun, line 102: Probably have to add some inline documentation here, but the idea was: `InitializeFixture` method is being executed only once per fixture lifetime (per test class in most cases) when `InitializeTestCase` is being executed for each test. So combining this two methods it's easy to implement custom initialization logic depending on nature of test collection.

tests/MongoDB.Driver.Tests/app.config
```diff
@@ -2,5 +2,6 @@
 <configuration>
   <appSettings>
     <add key="xunit.methodDisplay" value="method"/>
+    <add key="diagnosticMessages" value="true" />
```
sanych-sun, line 5: Revert this change.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5188Tests.cs
```diff
@@ -270,5 +270,7 @@ private class H : G
         {
             public int Z { get; set; }
         }
+
```
papafe, line 273: I suppose these two lines can be removed

papafe (commented):
Overall it makes sense for me

papafe (commented):
Overall it makes sense for me

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+*/
+
+using System;
+using System.Collections.Generic;
+using MongoDB.Driver.Tests;
+
+namespace MongoDB.Driver.TestHelpers
+{
+    public abstract class MongoCollectionFixture<TDocument> : MongoDatabaseFixture
+    {
```
rstam, line 23: Add:

    private readonly Lazy<IMongoCollection<TDocument>> _collection;

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+        private readonly string _collectionName;
+        private bool _collectionInitialized;
```
rstam, line 25: Rename to `_dataInitialized`.

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        protected MongoCollectionFixture(string collectionName = null)
+        {
+            _collectionName = collectionName ?? GetCollectionName();
+            if (string.IsNullOrEmpty(_collectionName))
+            {
+                throw new ArgumentNullException(nameof(collectionName) , "Cannot resolve the collection name. Try to specify the parameter explicitly");
+            }
```
rstam, line 33: Add:

    _collection = new Lazy<IMongoCollection<TDocument>>(CreateCollection);

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
```
rstam, line 35: Add:

    public IMongoCollection<TDocument> Collection => _collection.Value;

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+        public string CollectionName => _collectionName;
+
+        protected abstract IEnumerable<TDocument> InitialData { get; }
+
+        public virtual bool ResetOnEachTestCase => false;
```
rstam, line 40: Rename to `InitializeDataBeforeEachTestCase`.

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public override IMongoCollection<T> GetCollection<T>(string collectionName = null)
```
rstam, line 42: Replace `GetCollection` method witth `Collection` property. The property can return the same `IMongoCollection<TDocumnt>` instance every time.

Even if we do need to initialize the data before every test case, we don't have to create a new `MongoCollection` instance.

tests/MongoDB.Driver.TestHelpers/MongoCollectionFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (!string.IsNullOrEmpty(collectionName))
+            {
+                throw new NotSupportedException("CollectionFixture does not support explicit collection name.");
+            }
+
+            var db = GetDatabase();
+            return db.GetCollection<T>(CollectionName);
+        }
+
+        protected override void InitializeTestCase()
```
rstam, line 58: Refactor to:

        protected override void InitializeTestCase()
        {
            if (InitializeDataBeforeEachTestCase || !_dataInitialized)
            {
                Collection.Database.DropCollection(_collectionName);
                Collection.Database.CreateCollection(_collectionName);

                var initialData = InitialData;
                if (initialData != null)
                {
                    Collection.InsertMany(initialData);
                }

                _dataInitialized = true;
            }
        }

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+using System;
+using System.Collections.Generic;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.TestHelpers.Logging;
+
+namespace MongoDB.Driver.Tests
+{
+    public class MongoDatabaseFixture : IDisposable
+    {
+        private static readonly string __timeStamp = DateTime.Now.ToString("yyyyMMdd-HHmmss");
```
rstam, line 25: Use `DateTime.Now.ToString("MMddHHmm");` like we currently do.

The idea is to have the shortest database name possible but still be unique.

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var database = GetDatabase();
+            foreach (var collection in _usedCollections)
+            {
+                database.DropCollection(collection);
+            }
+
+            var client = _client;
+            _client = null;
+            client.Dispose();
```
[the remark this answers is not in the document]
rstam: Thought for the future: 

Currently calling `client.Dispose` doesn't actually do much. It certainly doesn't shut down the connection pool.

However, if in the future `MongoClient` becomes truly disposable then calling `Dispose` will shut down the connection pool, which in turn will result in each test class starting up a NEW cluster, which can take a long time since it has to create a new connection pool, open a bunch of sockets and let SDAM discover a writable server. 

We want all the tests to use the *same* `MongoClient` instance for efficiency (unless of course a test requires a specially configured client, which is a special case).

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+            _usedCollections.Add(collectionName);
+            return db.GetCollection<T>(collectionName);
+        }
+
+        protected virtual void ConfigureMongoClient(MongoClientSettings settings)
+        {
+            settings.LoggingSettings = new LoggingSettings(new XUnitLoggerFactory(LogsAccumulator), 10000); // Spec test require larger truncation default
+        }
+
+        internal void Initialize()
```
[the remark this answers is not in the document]
rstam: Use `!_fixtureInitialized` instead of `_client == null` to control only initializing the fixture once.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5163Tests.cs
```diff
@@ -13,19 +13,26 @@
...
+    public CSharp5163Tests(ITestOutputHelper testOutputHelper, TestDataFixture fixture)
+        : base(testOutputHelper, fixture)
+    {
+    }
+
     [Fact]
     public void Select_muliply_int_long_should_work()
     {
-        var collection = GetCollection();
+        var collection = Fixture.GetCollection<MyModel>();
```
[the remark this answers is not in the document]
rstam: Same in many other places.

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,108 @@
...
+            client.Dispose();
+        }
+
+        public IMongoClient GetClient()
+            => _client;
+
+        public virtual IMongoDatabase GetDatabase()
+            => GetClient().GetDatabase(_databaseName);
+
+        public virtual IMongoCollection<T> GetCollection<T>(string collectionName = null)
```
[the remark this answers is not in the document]
rstam: I'm recommending `CreateCollection` instead of `GetCollection` because this actually modifies the collection (`database.GetCollection` doesn't actually do anything on the server, it just returns a `IMongoCollection` instances).

This method could also be called `CreateEmptyCollection`.

rstam (changes requested):
I like this approach a LOT!!! Well done!

Sorry the number of requested changes looks like a lot. It's really just:

1. A bunch of renaming suggestions
2. Changing some methods to properties with Lazy values

The whole basic approach you are proposing seems 100% sound and I am not requesting any changes to the essence of your approach.

rstam:
My requested changes are prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp5338-0124

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4066Tests.cs
```diff
@@ -13,37 +13,37 @@
...
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Driver.TestHelpers;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
 {
-    public class CSharp4066Tests : Linq3IntegrationTest
+    public class CSharp4066Tests : IntegrationTest<CSharp4066Tests.TestDataFixture>
```
rstam, line 26: Consider renaming` TestDataFixture` to `CollectionFixture`.

That way we know immediately that we are using a collection fixture or a database fixture.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5231Tests.cs
```diff
@@ -16,21 +16,23 @@
...
 using MongoDB.Driver.Core.Misc;
-using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
-using MongoDB.Driver.Linq;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
 {
-    public class CSharp5231ests : Linq3IntegrationTest
+    public class CSharp5231Tests : LinqIntegrationTest<CSharp5231Tests.TestDataFixture>
```
rstam, line 24: Consider renaming `TestDataFixture` to `DatabaseFixture`.

That way we know immediately that we are using a collection fixture or a database fixture.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4066Tests.cs
rstam: Same in other test files.

rstam (changes requested):
Minor request pending. Looks great!

BorisDog (changes requested):
Partial review.
Requested to evaluating the approach of having a collection fixture instead.

papafe (approved):
LGTM! I also like the new shared`MongoClientFixture`

tests/MongoDB.Driver.TestHelpers/MongoClientFixture.cs
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+using System;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.TestHelpers.Logging;
+using MongoDB.Driver.Tests;
+
+namespace MongoDB.Driver.TestHelpers;
+
+public class MongoClientFixture : IDisposable
```
BorisDog, line 23: nit: sealed?

tests/MongoDB.Driver.TestHelpers/MongoClientFixture.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    public MongoClientFixture()
+    {
+        _client = new Lazy<IMongoClient>(CreateClient);
+        LogsAccumulator = new XUnitOutputAccumulator(__logCategoriesToExclude);
+    }
+
+    public IMongoClient Client => _client.Value;
+
+    internal XUnitOutputAccumulator LogsAccumulator { get; }
```
BorisDog, line 37: Not sure what was my motivation for making `XUnitOutputAccumulator` internal.

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public MongoDatabaseFixture(MongoClientFixture clientFixture = null)
+        {
+            _disposeClientFixture = clientFixture == null;
+            _mongoClientFixture = clientFixture ?? new MongoClientFixture();
+        }
+
+        public IMongoDatabase Database { get; private set; }
+
+        internal IMongoClient MongoClient => _mongoClientFixture.Client;
```
BorisDog, line 39: Most internals can be public?

BorisDog (changes requested):
Looks great! Few minor comments.

BorisDog (approved):
LGTM!

adelinowona (approved):
LGTM

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,93 @@
...
+                    database.DropCollection(collection);
+                }
+            }
+        }
+
+        protected IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
+        {
+            if (string.IsNullOrEmpty(collectionName))
+            {
+                var stack = new System.Diagnostics.StackTrace();
```
rstam: This code is not reached.

Throw an exception instead.

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (database != null)
+            {
+                foreach (var collection in _usedCollections)
+                {
+                    database.DropCollection(collection);
+                }
+            }
+        }
+
+        protected IMongoCollection<TDocument> CreateCollection<TDocument>(string collectionName = null)
```
rstam: `collectionName` parameter should be mandatory.

rstam (changes requested):

tests/MongoDB.Driver.TestHelpers/MongoDatabaseFixture.cs
sanych-sun: Done

rstam (approved):
LGTM

This is nice!

PR #1333, opened 2024-05-30
CSHARP-2187: Flatten CombinedProjectionDefinitions to reduce recursive calls to Render.

JamesKovacs:
When fluently building a projection, each call to `Include` or `Exclude` creates a new `CombinedProjectionDefinition<T>` with the projection to that point and the new include/exclude projection. This results in a chain of nested projections. When this chain is rendered, `CombinedProjectionDefinition<T>.Render` is called recursively. With a large number of included/excluded fields, this can result in significant callstack depth. Although it is unlikely in common cases to overflow the stack, it is possible as demonstrated by the test case. This issue was resolved by flattening nested `CombinedProjectionDefinition<T>` instances.


sanych-sun (approved):
LGTM

PR #1348, opened 2024-06-14
CSHARP-2659: Adding support for $sample aggregation pipeline stage.

JamesKovacs:
Added `$sample` stage to `Aggregate()`. Reorganized external PR, added some tests, and changed the `size` from `int` to `long`. The server accepts `NumberLong` for the `size` and throws on negative. Although the docs say that `size` must be a positive integer, empirical testing shows that zero is permissible and returns no documents as expected.


sanych-sun (approved):
LGTM

PR #1353, opened 2024-06-17
CSHARP-5139: Coverity analysis defect 127428: Explicit null dereferenced.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/WindowMethodToAggregationExpressionTranslator.cs
```diff
@@ -292,6 +292,7 @@ public static AggregationExpression Translate(TranslationContext context, Method
 
                 if (method.IsOneOf(__unaryMethods))
                 {
+                    ThrowIfSelectorTranslationIsNull(selectorTranslation);
```
rstam, line 295: We know that `selectorTranslation` is not null here, but adding this line of code lets Coverity know also.

adelinowona (approved):
LGTM

PR #1355, opened 2024-06-17
CSHARP-5140: Coverity analysis defect 127372: Explicit null dereferenced.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/WindowMethodToAggregationExpressionTranslator.cs
```diff
@@ -504,6 +504,10 @@ private static AstWindow ToAstWindow(SetWindowFieldsWindow window, object sortBy
         private static IBsonSerializer GetSortBySerializer(object sortBy, IBsonSerializer inputSerializer, BsonSerializerRegistry serializerRegistry)
         {
             // use reflection to call GetSortBySerializerGeneric because we don't know TDocument until runtime
+            if (sortBy == null)
+            {
+                throw new Exception("sortBy was unexpectedly null.");
```
rstam, line 509: We know that this method is never called with `sortBy` equal to `null`.

I think turning the `NullReferenceException` into any other kind of exception is enough to satisfy Coverity.

sanych-sun: Should we use ArgumentNullException instead?

rstam: I considered that but decided not to because `ArgumentNullException` is normally used in public APIs to inform the user (not us) that they passed `null` when they shouldn't have.

In this case we *know* (or think we do) that none of our code ever calls this method with a `null` value for `sortBy`, so this is actually something that should *never* happen. The code is fine as is, we will never actually throw a `NullReferenceException`. But Coverity doesn't know that.

Having said that, if you prefer an `ArgumentNullException` I can go along with that.

sanych-sun: I would say that `Exception` is "too wide" exception to use here. For me we should either use some exception of `ArgumentException` family (as we validate argument of the method) or something like `InvalidOperationException` (if we see this as a situation when something leads us to wrong system state).

rstam: I don't disagree for the most part. We should never throw an `Exception` (and we DON'T!!!).

In this case we could use `ArgumentNullException` because it just so happens that `sortBy` is an argument. My objection was that this was not a public API.

We could consider creating a new `ShouldNeverHappenException` or `InternalException` for exceptions that we don't ever expect to throw.

`InvalidOperationException` is not correct here (because it's not tied to the current internal state of the object).

Let me know where you end up on this... or if you can think of a better way to shup Coverity up?

sanych-sun: I would prefer to have `ArgumentNullException` then.

rstam: Done.

adelinowona (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1356, opened 2024-06-17
CSHARP-5143: Coverity analysis defect 127277: Dereference before null check.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstMapExpression.cs
```diff
@@ -52,7 +52,7 @@ public override BsonValue Render()
                 { "$map", new BsonDocument
                     {
                         { "input", _input.Render() },
-                        { "as", _as.Name, _as != null },
+                        { "as", _as?.Name, _as != null },
```
rstam, line 55: An alternative would have been:

    { "as", () => _as.Name, _as != null },

but this is lighter weight.

adelinowona (approved):
LGTM

PR #1357, opened 2024-06-17
CSHARP-5150: Coverity analysis defect 102677: Copy-paste error.

rstam:


src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1961,13 +1961,13 @@ private BsonValue ParseTimestampConstructor()
...
             var incrementToken = PopToken();
-            if (secondsSinceEpochToken.IsNumber)
+            if (incrementToken.IsNumber)
             {
                 increment = incrementToken.Int32Value;
             }
             else
             {
-                var message = string.Format("JSON reader expected a number but found '{0}'.", secondsSinceEpochToken.Lexeme);
+                var message = string.Format("JSON reader expected a number but found '{0}'.", incrementToken.Lexeme);
```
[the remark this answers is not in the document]
adelinowona: I would say add tests. Looking at `TestTimestampConstructor` in `JsonReaderTests.cs` which implicitly tests the `ParseTimestampConstructor` method, we don't test for potential invalid json input which is probably how this error wasn't caught before. 

Something similar to `TestTimestampExtendedJsonNewRepresentationWhenInvalid` should be enough.

adelinowona (changes requested):

tests/MongoDB.Bson.Tests/IO/JsonReaderTests.cs
```diff
@@ -1396,6 +1396,26 @@ public void TestTimestampConstructor()
...
+        // truncated input
+        [InlineData("Timestamp(")]
+        [InlineData("Timestamp()")]
+        [InlineData("Timestamp(1")]
+        [InlineData("Timestamp(1,")]
+        [InlineData("Timestamp(1, 2")]
+        // valid JSON but not a valid extended JSON BsonTimestamp
+        [InlineData("Timestamp('abc', 2)")]
+        [InlineData("Timestamp(2, 'abc')")]
+        public void TestTimestampConstructorWhenInvalid(string json)
```
adelinowona, line 1409: `TestTimestampExtendedJsonNewRepresentationWhenInvalid` seems to expect to fail if the increment value is not an integer. Should we add that check as well for the TimeStampConstructor?

adelinowona (changes requested):

tests/MongoDB.Bson.Tests/IO/JsonReaderTests.cs
rstam: Did you mean this test?

    [InlineData("{ \"$timestamp\" : { \"t\" : 1, \"i\" : 2.0 } }")]

rstam: Looks like `ParseTimestampConstructor` and `ParseTimestampExtendedJsonNewRepresentation` behave differently in the presence of a non-integer value for `timestamp` or `increment`.

`ParseTimestampConstructor`  only requires a number, and converts it to an int32, but `ParseTimestampExtendedJsonNewRepresentation` throws an exception if it's not an integer.

So I can't add this new test unless we also change the implementation of `ParseTimestampConstructor`.

src/MongoDB.Bson/IO/JsonReader.cs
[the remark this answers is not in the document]
rstam: Done

tests/MongoDB.Bson.Tests/IO/JsonReaderTests.cs
rstam: I added these two new test cases expecting them to pass but they don't, because they aren't invalid according to the implementation of `ParseTimestampConstructor`

    [InlineData("Timestamp(1.0, 2)")]
    [InlineData("Timestamp(1, 2.0)")]

adelinowona (approved):
LGTM

PR #1359, opened 2024-06-17
CSHARP-5149: Add lock in MongoUrl.ClearCache

BorisDog:


damieng (approved):

damieng:
Other option would be to switch to ConcurrentDictionary but probably more effort than it's worth.

rstam (approved):
LGTM

PR #1363, opened 2024-06-18
CSHARP-5151: Coverity analysis defect 102658: Dereference after null check.

sanych-sun:
I've figured out that problematic code is being used only from the not-used internal class - so decide to remove all this dead code.


damieng (approved):
Nice find!

BorisDog (approved):
LGTM

PR #1365, opened 2024-06-19
CSHARP-5145: Coverity analysis defect 127214: Dereference null return value

BorisDog:


src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
```diff
@@ -436,6 +436,12 @@ private TClass CreateInstanceUsingCreator(Dictionary<string, object> values)
                 var value = keyValuePair.Value;
 
                 var memberMap = _classMap.GetMemberMapForElement(elementName);
+
+                if (memberMap == null)
+                {
+                    throw new FormatException($"Element '{elementName}' does not match any field or property of class {_classMap.ClassType.FullName}");
+                }
```
BorisDog, line 443: This should never happen, as `CreateInstanceUsingCreator` is called from `DeserializeClass` with values checked against `_classMap.ElementTrie`. This check will prevent from accicendantal bugs `DeserializeClass`.

damieng (approved):

rstam (approved):
LGTM

PR #1366, opened 2024-06-20
CSHARP-5154: Coverity analysis defect 138095: Dereference null return value.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslator.cs
```diff
@@ -39,7 +39,13 @@ public static AstPipeline Translate(TranslationContext context, MethodCallExpres
                 if (secondValue is IMongoQueryable secondQueryable)
                 {
                     var secondProvider = (IMongoQueryProviderInternal)secondQueryable.Provider;
-                    var secondCollectionName = secondProvider.CollectionNamespace.CollectionName;
+                    var secondCollectionNamespace = secondProvider.CollectionNamespace;
+                    if (secondCollectionNamespace == null)
+                    {
+                        throw new ExpressionNotSupportedException(expression, because: "second argument must be an IMongoQueryable against a collection");
+                    }
```
rstam, line 46: This was a good catch by Coverity. The secondCollectionNamespace could indeed be null if the argument passed in was an IMongoQueryable against a database instead of a collection.

BorisDog (approved):
LGTM

PR #1373, opened 2024-06-27
POC of Mql.Field (could be used by EF Core Provider to support shadow properties).

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/MqlFieldTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            result.Should().Be(2);
+        }
+
+        [Fact]
+        public void Select_Mql_Field_should_work_with_POCO()
+        {
+            var collection = GetCollection<C>();
+
+            var queryable = collection.AsQueryable()
+                .Select(root => Mql.Field(root, "X", Int32Serializer.Instance) + 1); // like root.X except C does not have a property called X
```
rstam, line 48: We need to know the container, even if it is only the `root`.

`X` is like a shadow property. It exists in the document in the database but not in the class `C`.

rstam:
The serializer is important because it defines how the field is represented in the database.

If it is not provided, where would we get it from? The only reasonable place would be the `BsonSerializerRegistry`, but there is no guarantee that the serializer registered there is the correct one for this particular hidden field in this particular POCO.

damieng:
Tesetd and working well with the EF Shadow Properties implementation and also for select projection rewriting.

damieng:
Two open questions here:

1. Should the string "fieldName" second parameter of Mql.Field be elementName or even elementPath?
2. When using Mql.Field I'm seeing:

```json
	{
		"$match": { 
			"$and": [
					{ "$expr" : { "$eq" : ["$planetId",	 { "$oid" : "621ff30d2a3e781873fcb663" }] } },
				 	{ "$expr" : { "$eq" : ["$label", "VI"] } }
				 ] 
		} 
	},
	{ "$limit" : 1 }"
```

But when executing against the same properties as CLR properties I get a quite different query:

```json
  { "$match" :
  	{ "_id.planetId" : { "$oid" : "621ff30d2a3e781873fcb663" },
  	{ "_id.label" : "VI" } }, 
  { "$limit" : 1 }
```

Why are the ones from MqlFied using $expr and why do they get wrapped in an additional $and ?

rstam:
> Why are the ones from MqlFied using $expr and why do they get wrapped in an additional $and ?

Because initially I only implemented `Mql.Field` for aggregation expressions and not for filters. 

The `$and` is required because implicit and can't be used when there are two `$expr` in the same document (that would be duplicate field names).

rstam:
> Should the string "fieldName" second parameter of Mql.Field be elementName or even elementPath?

It's a bit arbitrary when we use "field" and when we use "element".

I don't think the server team ever uses the term "element". To them it's always a "field".

rstam:
> or even elementPath?

Not sure about paths... I was envisioning that `Mql.Field` would descend EXACTLY ONE level. If there are more levels you can compose multiple `Mql.Field` uses to descend to the nested field.

rstam:
> or even elementPath?

Looking at this in more detail I believe that `Mql.Field` can only descend one level.

`Mql.Field` translates to the MQL `$getField` operator which only descends one level at a time.

The server expects you to nest calls to `$getField` to descend multiple levels.

rstam:
I've pushed a new commit adding support for Mql.Field in filters.

damieng:
POC working great here now as the backing for EF-42 (shadow properties).

rstam:
The POC is concluded and we have created JIRA ticket CSHARP-5241 for this new feature.

The PR is now here:

https://github.com/mongodb/mongo-csharp-driver/pull/1554

PR #1374, opened 2024-06-28
CSHARP-5168: Fixed using some LINQ function after a Union throws an Exception

616b2f:
For details see Jira Ticket pls.

### DIFF
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
@@ -48,6 +48,14 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Serializers.KnownSerializers
             foreach (var type in child.KnownSerializers.Keys)
             foreach (var serializer in child.KnownSerializers[type])
             {
+                if (_knownSerializers.TryGetValue(type, out var set))
+                {
+                    if (set.Any(x => x.GetHashCode() == serializer.GetHashCode()))
+                    {
+                        continue;
+                    }
+                }
+
                 AddKnownSerializer(type, serializer);
             }
         }
--- a/src/MongoDB.Driver/Linq/MongoQueryable.cs
+++ b/src/MongoDB.Driver/Linq/MongoQueryable.cs
@@ -52,23 +52,6 @@ namespace MongoDB.Driver.Linq
         }
 
         /// <summary>
-        /// Combine two collections
-        /// </summary>
-        /// <typeparam name="TSource">The type of the elements of <paramref name="source1" /> and <paramref name="source1" />.</typeparam>
-        /// <param name="source1">A sequence to being merged.</param>
-        /// <param name="source2">A sequence to being merged.</param>
-        /// <returns>
-        /// cobined collection.
-        /// </returns>
-        public static IMongoQueryable<TSource> Union<TSource>(this IMongoQueryable<TSource> source1, IEnumerable<TSource> source2)
-        {
-            Ensure.IsNotNull(source1, nameof(source1));
-            Ensure.IsNotNull(source2, nameof(source2));
-
-            return (IMongoQueryable<TSource>)Queryable.Union(source1, source2);
-        }
-
-        /// <summary>
         /// Determines whether any element of a sequence satisfies a condition.
         /// </summary>
         /// <typeparam name="TSource">The type of the elements of <paramref name="source" />.</typeparam>
--- a/src/MongoDB.Driver/Linq/MongoQueryable.cs
+++ b/src/MongoDB.Driver/Linq/MongoQueryable.cs
@@ -52,6 +52,23 @@ namespace MongoDB.Driver.Linq
         }
 
         /// <summary>
+        /// Combine two collections
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source1" /> and <paramref name="source1" />.</typeparam>
+        /// <param name="source1">A sequence to being merged.</param>
+        /// <param name="source2">A sequence to being merged.</param>
+        /// <returns>
+        /// cobined collection.
+        /// </returns>
+        public static IMongoQueryable<TSource> Union<TSource>(this IMongoQueryable<TSource> source1, IEnumerable<TSource> source2)
+        {
+            Ensure.IsNotNull(source1, nameof(source1));
+            Ensure.IsNotNull(source2, nameof(source2));
+
+            return (IMongoQueryable<TSource>)Queryable.Union(source1, source2);
+        }
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
```diff
@@ -48,6 +48,14 @@ public void AddKnownSerializersFromChild(KnownSerializersNode child)
...
             foreach (var serializer in child.KnownSerializers[type])
             {
+                if (_knownSerializers.TryGetValue(type, out var set))
+                {
+                    if (set.Any(x => x.GetHashCode() == serializer.GetHashCode()))
+                    {
+                        continue;
+                    }
+                }
+
```
rstam, line 58: I couldn't figure out why this change was made.

I commented it out and all tests still pas..

src/MongoDB.Driver/Linq/MongoQueryable.cs
```diff
@@ -51,6 +51,23 @@ public static Task<bool> AnyAsync<TSource>(this IMongoQueryable<TSource> source,
...
+        /// <returns>
+        /// cobined collection.
+        /// </returns>
+        public static IMongoQueryable<TSource> Union<TSource>(this IMongoQueryable<TSource> source1, IEnumerable<TSource> source2)
+        {
+            Ensure.IsNotNull(source1, nameof(source1));
+            Ensure.IsNotNull(source2, nameof(source2));
+
+            return (IMongoQueryable<TSource>)Queryable.Union(source1, source2);
+        }
```
rstam, line 69: Good call. This change is necessary.

However: all tests (including the new one you wrote below) pass without this.

This change is simply about getting the compile time type of the LINQ query to be `IMongoQueryable<T>` instead of `IQueryable<T>`. The runtime type was already `IMongoQueryable<T>`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslatorTests.cs
```diff
@@ -131,6 +131,29 @@ public void Union_should_support_projection_to_anonymous()
...
 
+        [Fact]
+        public void Union_should_be_usable_in_another_pipeline_step()
+        {
+            RequireServer.Check().Supports(Feature.AggregateUnionWith);
+
+            var queryable = _firstCollection
+                .AsQueryable()
+                .Where(c => c.Name.StartsWith("second"))
+                .Union(_secondCollection.AsQueryable()
```
rstam, line 142: I had to add:

    using MongoDB.Driver.Linq;

above to get the compile time type of `Union` to be `IMongoQueryable<T>` instead of `IQueryable<T>`.

However: this test and all other tests in this file pass with our without the new `using` statement. The `using` statement doesn't change the meaning of any queries, it just changes the compile time type of `Union`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslatorTests.cs
```diff
@@ -131,6 +131,29 @@ public void Union_should_support_projection_to_anonymous()
...
+                    .Where(c => c.Name.StartsWith("another")))
+                .Where(x => x.Id == 2);
+
+            var stages = Translate(_firstCollection, queryable);
+            AssertStages(stages,
+                "{ $match : { Name : /^second/s } }",
+                "{ $unionWith : { coll : 'partners', pipeline : [{ $match : { Name : /^another/s } }] } }",
+                "{ $match : { Id : 2 } }"
```
rstam, line 150: I had to change this to:

    "{ $match : { _id : 2 } }"

to get the test to pass. I think `_id` is the correct field name.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
rstam: I'm testing against `master` which has some other changes that might have made this change unnecessary.

By "using master" I mean that I rebased the PR on `master` on my machine.

rstam (changes requested):
Thank you for the PR. Please see the comments.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
616b2f: I got the test now running locally, I can see that the test I provided is not showing how the issue can be reproduced. I changed it, now you should be able to reproduce it when you comment out this part here.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslatorTests.cs
616b2f: Removed it, I could not up with the test that represents why I needed this change. When I find that happening again I create a separate issue for that.

616b2f:
@rstam Please have another look, I fixed some shortcomings of my initial request.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/KnownSerializers/KnownSerializersNode.cs
```diff
@@ -48,6 +48,14 @@ public void AddKnownSerializersFromChild(KnownSerializersNode child)
             foreach (var type in child.KnownSerializers.Keys)
             foreach (var serializer in child.KnownSerializers[type])
             {
+                if (_knownSerializers.TryGetValue(type, out var set))
+                {
+                    if (set.Any(x => x.GetHashCode() == serializer.GetHashCode()))
```
rstam, line 53: You are sort of on the right track here. But the real fix is to get some broken `Equals` methods to return true for two serializers that are "equal" so they only get added to the HastSet once.

Just because two serializers have the same hash code doesn't mean they are equal.

I have created a new PR that addresses the root issue here and has a very different minimal reproduction.

See CSHARP-5168.

rstam: Here's the new PR:

https://github.com/mongodb/mongo-csharp-driver/pull/1413

rstam:
Thank you for submitting this PR! Even though we didn't use it as submitted it was very helpful in finding the underlying issue (which wasn't actually related to Union after all).

PR #1379, opened 2024-07-04
CSHARP-4017: Fix flaky Aggregate_with__out_includes_read_preference_for_5.0__server error

sanych-sun:
The issue is reproducible on replica set environments because of race condition between replication and executing $out aggregation on secondary node.


BorisDog (approved):
LGTM. Waiting to see those tests finally fixed!

PR #1381, opened 2024-07-05
CSHARP-4017: Fix flaky Aggregate_with__out_includes_read_preference_for_5.0__server error

sanych-sun:
Found test that is duplicated with spec test "Aggregate with $out includes read preference for 5.0+ server" (defined in specifications\crud\tests\unified\aggregate-write-readPreference.yml)


rstam (approved):
LGTM

PR #1390, opened 2024-07-16
CSHARP-4957: Fix issue with new array expression.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -24,14 +27,29 @@ internal static class NewArrayInitExpressionToAggregationExpressionTranslator
...
+
+                // make sure all items are serialized using the same serializer
+                if (!itemTranslation.Serializer.Equals(itemSerializer))
+                {
+                    throw new ExpressionNotSupportedException(expression, because: "all items in the array must be serialized using the same serializer");
+                }
             }
+
             var ast = AstExpression.ComputedArray(items);
-            var serializer = context.KnownSerializersRegistry.GetSerializer(expression);
```
rstam, line 33: KnownSerializer was returning the wrong serializer here.

I think we need to reevaluate whether the known serializer strategy is working.

In this particular case we should just determine the required result serializer here without help from the known serializer registry.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -24,14 +27,29 @@ internal static class NewArrayInitExpressionToAggregationExpressionTranslator
...
         {
             var items = new List<AstExpression>();
+            IBsonSerializer itemSerializer = null;
             foreach (var itemExpression in expression.Expressions)
             {
                 var itemTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, itemExpression);
                 items.Add(itemTranslation.Ast);
+                itemSerializer ??= itemTranslation.Serializer;
+
+                // make sure all items are serialized using the same serializer
```
rstam, line 37: This goes slightly beyond fixing the reported bug, but it's in the same logic path.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -24,14 +27,29 @@ internal static class NewArrayInitExpressionToAggregationExpressionTranslator
...
+                if (!itemTranslation.Serializer.Equals(itemSerializer))
+                {
+                    throw new ExpressionNotSupportedException(expression, because: "all items in the array must be serialized using the same serializer");
+                }
```
JamesKovacs, line 41: Items must use compatible serializers but not necessarily the same serializer. For example I would expect the following to work:

```
var queryable = collection.AsQueryable()
    .Select(x => new Animal[] { x.Cat, x.Dog });
```

JamesKovacs (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
JamesKovacs: After investigating further and discussing in Slack, we are going to leave this as a known limitation that we don't support arrays of polymorphic types in projections.

JamesKovacs (approved):
LGTM

PR #1392, opened 2024-07-17
CSHARP-5162: Continue to use FindExpressionProjectionDefinition to avoid regression against LINQ2

rstam:
About a year ago we added `RenderForFind` and stopped using `FindExpressionProjectionDefinition`. That resulted in a subtle regression affecting LINQ2 that was only discovered relatively recently.

This PR fixes the regression by reverting back to using `FindExpressionProjectionDefinition` wherever it was used before, BUT then making `FindExpressionProjectionDefinition` behave the same as it used when using LINQ2 while behaving the SAME as `ExpressionProjectionDefinition` when using LINQ3.

The net effect is that when using LINQ2 everything works the same as it used to, and when using LINQ3 we have effectively (but not actually) replaced `FindExpressionProjectionDefinition` with `ExpressionProjectionDefinition`.

### DIFF
--- a/tests/MongoDB.Driver.Tests/FindExpressionProjectionDefinitionTests.cs
+++ b/tests/MongoDB.Driver.Tests/FindExpressionProjectionDefinitionTests.cs
@@ -17,35 +17,69 @@ using System;
 using System.Linq.Expressions;
 using FluentAssertions;
 using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Linq;
+using MongoDB.TestHelpers.XunitExtensions;
 using Xunit;
 
 namespace MongoDB.Driver.Tests
 {
     public class FindExpressionProjectionDefinitionTests
     {
-        [Fact]
-        public void Projection_to_class_should_work()
-            => AssertProjection(
+        [Theory]
+        [ParameterAttributeData]
+        public void Projection_to_class_should_work(
+            [Values(false, true)] bool renderForFind,
+            [Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
+        {
+            var expectedRenderedProjection = (linqProvider, renderForFind) switch
+            {
+                (LinqProvider.V2, _) => "{ A : 1, B : 1, _id : 0 }", // note: result serializer does client-side projection
+                (LinqProvider.V3, true) => "{ A : 1, X : '$B', _id : 0 }",
+                (LinqProvider.V3, false) => "{ A : '$A', X : '$B', _id : 0 }",
+                _ => throw new Exception()
+            };
+            AssertProjection(
                 x => new Projection { A = x.A, X = x.B },
-                "{ A : 1, X : '$B', _id : 0 }");
+                renderForFind,
+                linqProvider,
+                expectedRenderedProjection);
+        }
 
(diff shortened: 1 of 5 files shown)


JamesKovacs (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4681Tests.cs
```diff
@@ -36,7 +36,14 @@ public class CSharp4681Tests : Linq3IntegrationTest
...
             var renderedProjection = fluentFind.Options.Projection.Render(documentSerializer, serializerRegistry, linqProvider);
 
-            renderedProjection.Document.Should().Be("{ _id : 1 }");
+            if (linqProvider == LinqProvider.V2)
+            {
+                renderedProjection.Document.Should().Be("{ _id : 1 }");
+            }
+            else
+            {
+                renderedProjection.Document.Should().Be("{ _v : '$_id', _id : 0 }");
```
sanych-sun, line 45: Does it mean Linq3 provider will generate find projection that is not compatible with MongoDB before 4.4?

rstam: I don't think so. I think I need to change the test.

The test is calling `Render` instead of `RenderForFind`. But when executing an actual `Find` we would call `RenderForFind`.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/FindExpressionProjectionDefinitionTests.cs
```diff
@@ -17,35 +17,69 @@
...
+        [ParameterAttributeData]
+        public void Projection_to_class_should_work(
+            [Values(false, true)] bool renderForFind,
+            [Values(LinqProvider.V2, LinqProvider.V3)] LinqProvider linqProvider)
+        {
+            var expectedRenderedProjection = (linqProvider, renderForFind) switch
+            {
+                (LinqProvider.V2, _) => "{ A : 1, B : 1, _id : 0 }", // note: result serializer does client-side projection
+                (LinqProvider.V3, true) => "{ A : 1, X : '$B', _id : 0 }",
+                (LinqProvider.V3, false) => "{ A : '$A', X : '$B', _id : 0 }",
```
rstam, line 38: `FindExpressionProjectionDefinition` has some weird behavior when used in LINQ2. For one it ignores the distinction between `Render` and `RenderForFind`. And also, it uses a serializer that does client-side projection so the MQL is a little different from what you'd expect.

The LINQ3 behavior is more like what you'd expect, and also the same as `ExpressionProjectionDefinition`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4681Tests.cs
rstam: Fixed the test.

sanych-sun (approved):
LGTM

JamesKovacs (approved):
LGTM

PR #1393, opened 2024-07-19
CSHARP-1276: Provide Strong-Named Assemblies

sanych-sun:


benchmarks/MongoDB.Driver.Benchmarks/MongoDB.Driver.Benchmarks.csproj
```diff
@@ -20,11 +20,13 @@
...
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <IsPackable>false</IsPackable>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\MongoDB.Driver.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.13.8" />
-    <PackageReference Include="NDesk.Options" Version="0.2.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
```
sanych-sun, line 29: Have to replace `NDesk.Options` with `System.CommandLine`, because there is no strong-named version of `NDesk.Options`.

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkRunner.cs
```diff
@@ -13,48 +13,53 @@
...
 using BenchmarkDotNet.Running;
 using MongoDB.Benchmarks.Exporters;
-using NDesk.Options;
 
 namespace MongoDB.Benchmarks
 {
     public class BenchmarkRunner
     {
-        public static void Main(string[] args)
+        public static int Main(string[] args)
```
sanych-sun, line 26: @adelinowona Could you please double-check if new code properly process all command line parameters?

nuget.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sanych-sun, line 1: SHOULD NOT MERGE THIS FILE! It's here just to test not-yet-published libmongocrypt version.

src/MongoDB.Bson/Serialization/TypeNameDiscriminator.cs
```diff
@@ -145,7 +145,9 @@ public static string GetDiscriminator(Type type)
                 if (match.Success)
                 {
                     var publicKeyToken = match.Groups["token"].Value;
-                    if (publicKeyToken == "null")
+                    if (publicKeyToken == "null" ||
+                        // MongoDB's assemblies should use "type name, assembly name" discriminator format for backward compatibility
+                        (assembly.FullName.StartsWith("MongoDB") && publicKeyToken == "94992a530f44e321"))
```
sanych-sun, line 150: Have to add this condition for backward compatibility, because now we have strongly named assemblies and without this changes we would use fully qualified type name here, what would be a breaking change.

src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj
```diff
@@ -28,7 +28,7 @@
     <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.100.14" />
     <PackageReference Include="DnsClient" Version="1.6.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
-    <PackageReference Include="MongoDB.Libmongocrypt" Version="1.10.0" />
+    <PackageReference Include="MongoDB.Libmongocrypt" Version="1.11.0" />
```
sanych-sun, line 31: Do not forget to publish MongoDB.Libmongocrypt 1.11.0 before merging this PR.

src/MongoDB.Bson/Properties/AssemblyInfo.cs
```diff
@@ -16,6 +16,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography.X509Certificates;
```
BorisDog, line 19: Is this using needed?

BorisDog (changes requested):

BorisDog (approved):
LGTM

benchmarks/MongoDB.Driver.Benchmarks/BenchmarkRunner.cs
adelinowona: all good

adelinowona (approved):
LGTM

adaskos-signal:
I think public signing an assembly is a breaking change and should have been part of a new major version.

A library built with any of the previous unsigned versions will not be able to be consumed by a project which upgrades to the new, signed version. The two are not assembly compatible, and this applies to .NET Core as well. 
It won't even built, if any Mongo related class is not encapsulated and needs to cross the library boundaries.

Mongo2Go library for example seems ok, but see this issue for example: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2265
Same goes for any internal projects too.

I'm not sure the implications of this change were all assessed before proceeding.

mhbuck:
The Mongo diagnostic source is also impacted by this change
https://github.com/jbogard/MongoDB.Driver.Core.Extensions.DiagnosticSources/issues/34

I would expect a number of projects that use the MongoDB drivers are going to be impacted.

Droxx:
To add to the above comments. This has also broken some of our dependencies. I agree that this should have been part of a major version. We now have to pin ourselves to 2.27.0 until all of our dependencies are also updated:

https://github.com/ChilliCream/graphql-platform/issues/7306

aj-scram:
just echoing that this broke things for my org to cause we use: https://github.com/ChangemakerStudios/serilog-sinks-mongodb

PR #1399, opened 2024-07-26
CSHARP-5172: Updating an array fails if array contains values rather than documents.

rstam:


src/MongoDB.Driver/RenderArgs.cs
```diff
@@ -30,13 +30,14 @@ public record struct PathRenderArgs(string PathPrefix = null, bool AllowScalarVa
     /// Encapsulates settings needed for rendering Builder definitions.
     /// </summary>
     /// <typeparam name="TDocument">The type of the document.</typeparam>
-    public record struct RenderArgs<TDocument>
+    public record RenderArgs<TDocument>
```
rstam, line 33: @BorisDog add you as a reviewer to weigh in on whether `RenderArgs` should be a `class` or a `struct`.

src/MongoDB.Driver/RenderArgs.cs
```diff
@@ -33,10 +33,11 @@ public record struct PathRenderArgs(string PathPrefix = null, bool AllowScalarVa
     public record struct RenderArgs<TDocument>
     {
         private readonly IBsonSerializer<TDocument> _documentSerializer = default;
-        private readonly bool _renderForFind = false;
         private readonly LinqProvider _linqProvider = default;
         private readonly PathRenderArgs _pathRenderArgs = default;
         private readonly bool _renderDollarForm = default;
+        private readonly bool _renderForElemMatch = false;
+        private readonly bool _renderForFind = false;
```
sanych-sun, line 40: Should we introduce Enum instead? Like a 
```
enum RenderingContainer {
  Default = 0,
  Find = 1,
  ElemMatch = 2
}

```
 Or validate the parameters to avoid case when both `_renderForFind` and `_renderForElemMatch` set to `true`

sanych-sun (changes requested):

src/MongoDB.Driver/RenderArgs.cs
rstam: I don't think so.

In this case I think `RenderForFind` and `RenderForElemMatch` might BOTH be true.

rstam:
@BorisDog if you want you can just review the `RenderArgs` changes.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5172Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp5172Tests : Linq3IntegrationTest
+    {
+        [Theory]
+        [ParameterAttributeData]
+        public void Filter_ElemMatch_should_work(
```
BorisDog, line 32: Ideally we'd organize such tests in more thematic manner. 
For example this test would belong to hypothetical `BuildersFilterTests`  or `BuildersFilterElemMatchTests`

BorisDog (approved):
RenderArgs LGTM. 
Left optional suggestion for tests organization.

sanych-sun (approved):
LGTM

PR #1400, opened 2024-07-29
CSHARP-2348: Switch to $expr when $elemMatch filter is not supported by server.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstFilter.cs
```diff
@@ -92,6 +93,11 @@ public static AstFieldOperationFilter Compare(AstFilterField field, AstCompariso
 
         public static AstFieldOperationFilter ElemMatch(AstFilterField field, AstFilter filter)
         {
+            if (!ServerSupportsElemMatchFilter(filter))
+            {
+                throw new ExpressionNotSupportedException($"$elemMatch does not support filter: {filter}");
```
rstam, line 98: This exception will be caught at a higher level and the translation will switch to using `$expr`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstFilter.cs
```diff
@@ -245,6 +251,22 @@ public static AstFieldOperationFilter Type(AstFilterField field, BsonType type)
         {
             return new AstFieldOperationFilter(field, new AstTypeFilterOperation(type));
         }
+
+        private static bool ServerSupportsElemMatchFilter(AstFilter filter)
+        {
+            if (filter is AstOrFilter orFilter &&
+                orFilter.Filters.Any(IsImpliedElementFilter))
```
rstam, line 258: This only checks the specific thing not supported that was reported in CSHARP-2348.

There are probably many other things that are not supported by `$elemMatch` but we don't know about them yet. So this method might check for more unsupported things in the future.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp2348Tests.cs
```diff
@@ -0,0 +1,99 @@
...
+            var results = find.ToList().OrderBy(x => x.Id).ToList();
+
+            if (linqProvider == LinqProvider.V2)
+            {
+                renderedFilter.Should().Be("{ Roles : { $elemMatch : { $or : [{ '' : 1 }, { '' : 2 }] } } }"); // LINQ2 translation is wrong
+                results.Should().BeEmpty(); // LINQ2 result is wrong
+            }
+            else
+            {
+                // the ideal translation would be { Roles : { $elemMatch : { $or : [{ $eq : 1 }, { $eq : 2 }] } } }
```
rstam, line 69: @rishitb-mongodb let us know what you find out after talking to the server team.

This simpler translation really SHOULD be supported by the server.

rishitb-mongodb (approved):
Csharp2348 looks good as per my understanding

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstFilter.cs
```diff
@@ -92,7 +93,30 @@ public static AstFieldOperationFilter Compare(AstFilterField field, AstCompariso
...
         public static AstFieldOperationFilter ElemMatch(AstFilterField field, AstFilter filter)
         {
+            if (!ServerSupportsElemMatchFilter(filter))
+            {
+                throw new ExpressionNotSupportedException($"$elemMatch does not support filter: {filter}");
+            }
+
             return new AstFieldOperationFilter(field, new AstElemMatchFilterOperation(filter));
+
+            static bool ServerSupportsElemMatchFilter(AstFilter filter)
```
sanych-sun, line 103: Somehow I do not like term 'Server' in the method name. Probably because it makes me think we will check the server version. Should we rename it to something like `IsElemMatchFilterValid` or `IsValidForElemMatch`?

sanych-sun (approved):
LGTM + minor suggestion

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Filters/AstFilter.cs
rstam: You are right that we don't know the server version during LINQ translation (at least not now, that might change in the future?).

But `IsElemMatchFilterValid` is not quite right. The filter in question here IS valid, it's just a limitation of the server that it doesn't currently support it.

PR #1406, opened 2024-08-08
CSHARP-5215: Hide Operations/IBinding APIs

BorisDog:


tests/MongoDB.Driver.Tests/FindFluentTests.cs
```diff
@@ -322,13 +318,6 @@ public class FindFluentTests
...
             subject.Options.Max = new BsonDocument("max", 5);
             subject.Options.MaxTime = TimeSpan.FromSeconds(2);
-#pragma warning disable 618
-            subject.Options.Modifiers = new BsonDocument
-            {
-                { "$explain", true },
-                { "$hint", "ix_1" }
-            };
-#pragma warning restore 618
             subject.Options.Min = new BsonDocument("min", 2);
```
sanych-sun, line 321: It looks like there is no other way to pass { "$explain", true }. Should we add appropriate option then?

sanych-sun: If so, should we double-check if other options is available as well? https://www.mongodb.com/docs/manual/reference/method/db.collection.find/#std-label-find-options

src/MongoDB.Driver/Core/Operations/ExplainOperation.cs
```diff
@@ -65,54 +45,37 @@ public ExplainOperation(DatabaseNamespace databaseNamespace, IExplainableOperati
...
-        /// <value>
-        /// The verbosity.
-        /// </value>
         public ExplainVerbosity Verbosity
         {
             get { return _verbosity; }
             set { _verbosity = value; }
+        }
+
+        public BsonDocument CreateExplainCommand(ConnectionDescription connectionDescription, ICoreSession session)
```
sanych-sun, line 69: Should the method name follow the convention in other similar classes? Rename it to `CreateCommand`.

src/MongoDB.Driver/Core/Operations/MapReduceOperationBase.cs
```diff
@@ -19,15 +19,14 @@
...
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Core.WireProtocol.Messages.Encoders;
-using MongoDB.Shared;
 
 namespace MongoDB.Driver.Core.Operations
 {
     /// <summary>
     /// Represents a base class for map-reduce operations.
     /// </summary>
     [Obsolete("Use Aggregation pipeline instead.")]
```
sanych-sun, line 28: Is this to be removed later?

tests/MongoDB.Driver.Tests/Core/Operations/OperationTestBase.cs
```diff
@@ -136,7 +136,7 @@ protected void EnsureDatabaseExists(string databaseName)
             ExecuteOperation(operation);
         }
 
-        protected TResult ExecuteOperation<TResult>(IReadOperation<TResult> operation)
+        private protected TResult ExecuteOperation<TResult>(IReadOperation<TResult> operation)
```
sanych-sun, line 139: Just curious: why do we made it private protected?

sanych-sun (approved):
LGTM + minor comments

src/MongoDB.Driver/Core/Operations/MapReduceOperationBase.cs
BorisDog: Yes, we'll remove it when dropping 4.4 support.

tests/MongoDB.Driver.Tests/Core/Operations/OperationTestBase.cs
BorisDog: It has to be limited to this assembly (IReadOperation is internal), this is achieved by `private protected`.

tests/MongoDB.Driver.Tests/MongoCollectionImplTests.cs
```diff
@@ -1515,9 +1505,6 @@ private class ClassForDistinctWithArrayField
                 Limit = 30,
                 MaxAwaitTime = TimeSpan.FromSeconds(4),
                 MaxTime = TimeSpan.FromSeconds(3),
-#pragma warning disable 618
-                Modifiers = BsonDocument.Parse("{ $snapshot : true }"),
```
rstam, line 1519: Is this a real modifier that we need to support?

rstam (approved):
LGTM on removing `Modifiers` from `FindOptions`.

I glanced at some of the other files and the changes I did look at are good. It's basically the same change over and over.

PR #1411, opened 2024-08-14
CSHARP-5228: Fix OperatingSystemHelper for MACCATALYS (MAUI)

elias-ik:
### Background

OperatingSystemHelper is supposed to throw when an unsupported platform is detected, eg. WASM. 
While MacOS is a supported platform, when running in MAUI MacCatalyst the  `DetermineOperatingSystemPlatform` still throws, since `OSPlatform.OSX` is not the current platform. 
According to `System.OperatingSystem.cs` `IsOSPlatform(string platform)` the current Platform is "MACCATALYST" which is not a static member to `OSPlatform.cs` thus we create a matching OSPlatform with `OSPlatform.Create(..)`

This might need to be raised as an issue in the dotnet repository.


BorisDog:
Thanks for raising this issue @elias-ik, this commit was cherrypicked and merged in https://github.com/mongodb/mongo-csharp-driver/pull/1511.

PR #1412, opened 2024-08-14
CSHARP-5230: Removed uuidhelpers.js as they're now available as a mongosh snippet.

JamesKovacs:
Removing uuidhelpers.js as it is no longer required. It is now available as a mongosh snippet. See https://jira.mongodb.org/browse/MONGOSH-1847 for details.


rstam (approved):
LGTM

PR #1419, opened 2024-08-16
CSHARP-4935: Support casting from an interface to a type that implements that interface in LINQ queries

rstam:


src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -96,7 +109,7 @@ public DiscriminatedInterfaceSerializer(IDiscriminatorConvention discriminatorCo
 
             _interfaceType = typeof(TInterface);
             _discriminatorConvention = discriminatorConvention ?? BsonSerializer.LookupDiscriminatorConvention(typeof(TInterface));
-            _objectSerializer = BsonSerializer.LookupSerializer<object>();
+            _objectSerializer = objectSerializer ?? new ObjectSerializer(allowedTypes: type => typeof(TInterface).IsAssignableFrom(type));
```
rstam, line 112: Do you see any backward compatibility issues with this change?

rstam (commented):
The LINQ changes in this PR look fine to me.

The concern we had that has kept this on the back burner is whether the change to `DiscriminatedInterfaceSerializer` would be backward breaking in any way.

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
sanych-sun: It might be a problem if someone registered their own object serializer for some reasons. Can we implement method like a `WithAllowedTypes` on objectSerializer? So we lookup the same way as it was before - and then try to use this new method, similar to this:
```
var objectSerializer = BsonSerializer.LookupSerializer<object>();
if (objectSerializer is ObjectSerializer builtInSerializer)
{
    objectSerializer  = builtInSerializer.WithAllowedTypes(allowedTypes: type => typeof(TInterface).IsAssignableFrom(type));
}
```

sanych-sun (changes requested):

src/MongoDB.Bson/Serialization/Serializers/DiscriminatedInterfaceSerializer.cs
```diff
@@ -97,20 +110,25 @@ public DiscriminatedInterfaceSerializer(IDiscriminatorConvention discriminatorCo
...
             _discriminatorConvention = discriminatorConvention ?? interfaceSerializer.GetDiscriminatorConvention();
-            _objectSerializer = BsonSerializer.LookupSerializer<object>();
-            if (_objectSerializer is ObjectSerializer standardObjectSerializer)
-            {
-                _objectSerializer = standardObjectSerializer.WithDiscriminatorConvention(_discriminatorConvention);
-            }
-            else
+            _interfaceSerializer = interfaceSerializer;
+
+            if (objectSerializer == null)
```
rstam, line 115: If a user passes in an `objectSerializer` we should use it as-is with no further configuration on our part. It is up to the user to pass in a properly configured seiralizer.

sanych-sun (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1422, opened 2024-08-19
CSHARP-5235: Refactor/Merge helpers projects

BorisDog:
- Introduce `MongoDB.Shared `project for shared items, for better maintainability:
   For example `Ensure` moved to` MongoDB.Shared` and made internal

- `MongoDB.Bson.TestHelpers` merged with` MongoDB.TestHelpers`. Now there are only two helper projects

- Logging dependency was moved to BSON, and Xunit logging helpers to `MongoDB.TestHelpers`. Not we can start adding logs in BSON tests as well.


rstam:
Top level observation: I don't think we should remove the `MongoDB.Bson.TestHelpers` project.

The overall idea is that:

* `MongoDB.TestHelpers` only depends on external packages
* `MongoDB.Bson.TestHelpers` depends on `MongoDB.TestHelpers` and `MongoDB.Bson`
* `MongoDB.Driver.TestHelpers` depends on `MongoDB.Bson.TestHelpers` and `MongoDB.Driver`

What we can and should do is move some things down to the correct level:

* anything in `MongoDB.Driver.TestHelpers` that does not depend on `MongoDB.Driver` can be moved down to `MongoDB.Bson.TestHelpers` or `MongoDB,TestHelpers`
* anything in `MongoDB.Bson.TestHelpers` that does not depend on `MongoDB.Bson` can be moved down to `MongoDB.TestHelpers`

rstam (changes requested):
Please request a re-review when you are ready.

BorisDog:
This work has been postponed, will submit a new PR instead.

PR #1420, opened 2024-08-16
CSHARP-5233: Remove IMongoQueryable interface.

rstam:

PR #1425, opened 2024-08-22
CSHARP-3310: Add ExpressionTranslationOptions to MongoClientSettings.

rstam:


src/MongoDB.Driver/ExpressionTranslationOptions.cs
```diff
@@ -14,18 +14,39 @@
...
-        /// Gets or sets the string translation mode.
+        /// Gets or sets the server wire version to target.
         /// </summary>
-        public AggregateStringTranslationMode? StringTranslationMode { get; set; }
+        public int? TargetWireVersion { get; set; }
+    }
+
+    internal static class ExpressionTranslationOptionsExtensions
+    {
+        public static ExpressionTranslationOptions SetTargetWireVersionIfMissing(this ExpressionTranslationOptions options, int? targetWireVersion)
```
rstam, line 36: This method takes an existing `options` instance and sets the `TargetWireVersion` if it is missing.

It also creates a new instance if the `options` argument is `null`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/TranslationContext.cs
```diff
@@ -44,24 +49,33 @@ public static TranslationContext Create(Expression expression, IBsonSerializer s
         private readonly KnownSerializersRegistry _knownSerializersRegistry;
         private readonly NameGenerator _nameGenerator;
         private readonly SymbolTable _symbolTable;
+        private readonly ExpressionTranslationOptions _translationOptions;
```
rstam, line 52: New field added to `TranslationContext`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/TranslationContext.cs
```diff
@@ -44,24 +49,33 @@ public static TranslationContext Create(Expression expression, IBsonSerializer s
...
         {
             _symbolTable = Ensure.IsNotNull(symbolTable, nameof(symbolTable));
             _nameGenerator = Ensure.IsNotNull(nameGenerator, nameof(nameGenerator));
             _knownSerializersRegistry = Ensure.IsNotNull(knownSerializersRegistry, nameof(knownSerializersRegistry));
+            _translationOptions = Ensure.IsNotNull(translationOptions, nameof(translationOptions));
             _data = data; // can be null
+
+            if (!translationOptions.TargetWireVersion.HasValue)
+            {
+                throw new ArgumentException("TargetWireVersion is not set.", nameof(translationOptions));
```
rstam, line 69: While the `translationOptions` might be `null` or have a missing `TargetWireVersion` before this point, by the time code reaches here we insist on having a defined `TargetWireVersion` so that during translation we always have a defined target.

Callers of this code have to ensure that the `translationOptions` are not `null` and have a defined `TargetWireVersion`.

Typically that would be determined as follows:

1. The user specified a `TargetWireVersion` (e.g. in `AggregateOptions`)
2. The target wire version is deduced from the `ClusterDescription`
3. The target wire version is defaulted to `Server40` if nothing else set it

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -42,7 +42,6 @@ internal sealed class MongoCollectionImpl<TDocument> : MongoCollectionBase<TDocu
         private readonly IOperationExecutor _operationExecutor;
         private readonly IBsonSerializer<TDocument> _documentSerializer;
         private readonly MongoCollectionSettings _settings;
-        private readonly RenderArgs<TDocument> _args;
```
rstam, line 45: We can't use a cached copy because the new `TargetWireVersion` needs to be determined as late as possible and could change from time to time.

src/MongoDB.Driver/PipelineDefinition.cs
```diff
@@ -101,9 +101,21 @@ public override string ToString()
         /// <returns>
         /// A <see cref="System.String" /> that represents this instance.
         /// </returns>
-        public string ToString(IBsonSerializer<TInput> inputSerializer, IBsonSerializerRegistry serializerRegistry)
+        public string ToString(IBsonSerializer<TInput> inputSerializer, IBsonSerializerRegistry serializerRegistry) =>
```
rstam, line 104: An alternative to a new overload would be to add make the new `translationOptions` parameter optional using `translationOptions = null` in the method signature.

src/MongoDB.Driver/RenderArgs.cs
```diff
@@ -37,6 +37,7 @@ public record struct RenderArgs<TDocument>
         private readonly bool _renderForElemMatch = false;
         private readonly bool _renderForFind = false;
         private readonly IBsonSerializerRegistry _serializerRegistry = default;
+        private readonly ExpressionTranslationOptions _translationOptions = default;
```
rstam, line 40: New field for `RenderArgs`. This will eventually get copied to the `TranslationContext` if rendering involves invoking a LINQ translator.

tests/MongoDB.Driver.Tests/FindFluentTests.cs
```diff
@@ -1,407 +1,407 @@
...
-        public void Count_should_call_collection_Count(
-            [Values(false, true)] bool usingSession,
-            [Values(false, true)] bool async)
-        {
-            var session = CreateSession(usingSession);
-            var filter = new BsonDocumentFilterDefinition<Person>(new BsonDocument("filter", 1));
-            var hint = new BsonDocument("hint", 1);
-            var findOptions = new FindOptions<Person>
-            {
+/* Copyright 2010-present MongoDB Inc.
```
rstam, line 1: I think this file had inconsistent line ending prior to me editing it and asking Visual Studio to make the line endings consistent.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp3933Tests.cs
```diff
@@ -80,7 +80,7 @@ public void PipelineDefinitionBuilder_Group_with_projection_to_implied_BsonDocum
 
             var pipeline = emptyPipeline.Group("{ _id : 1, Count : { $sum : 1 } }");
 
-            var stages = Linq3TestHelpers.Render(pipeline, BsonDocumentSerializer.Instance);
+            var stages = Linq3TestHelpers.Render(pipeline, BsonDocumentSerializer.Instance, ExpressionTranslationOptions.Default);
```
rstam, line 83: There is no way to compute the `TargetWireVersion` here so we just use the default `ExpressionTranslationOptions`.

The default has `TargetWireVersion` set to `Server40`.

JamesKovacs (changes requested):
I’ve been going through the PR and have a few overall concerns:

- Potential implications of blocking to determine `maxWireVersion`. - Query rendering behaviour changing depending on connectivity to certain cluster members.
- Impossible to reason about the resulting MQL in a deterministic fashion.
- Problems for the Analyzer since at design time in the IDE you aren’t connected to any servers. The Analyzer would have to add a configuration option to specify maxWireVersion to be used when rendering out queries for tooltips.

Let’s discuss over Zoom.

src/MongoDB.Driver/Core/Clusters/ClusterDescription.cs
```diff
@@ -293,6 +293,33 @@ public ClusterType Type
...
+                    }
+                }
+                else
+                {
+                    targetWireVersion = serverWireVersion;
+                }
+            }
+
+            return targetWireVersion;
+        }
```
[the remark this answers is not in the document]
rstam: Sorry I missed this comment. I will add the tests and push a new commit.

src/MongoDB.Driver/IMongoDatabaseExtensions.cs
```diff
@@ -166,6 +166,13 @@ public static Task<IChangeStreamCursor<ChangeStreamDocument<BsonDocument>>> Watc
             return database.WatchAsync(session, emptyPipeline, options, cancellationToken);
         }
 
+        // internal static methods
+        internal static int GetTargetWireVersion(this IMongoDatabase database) =>
+            database.Client.GetTargetWireVersion();
```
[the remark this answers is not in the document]
rstam: Note that I would have used `ServerVersion` as the name of the `enum` but that type name was already taken to `TargetServerVersion` was the best alternative I could come up with.

rstam:
This PR is still blocked on https://github.com/mongodb/mongo-csharp-driver/pull/1420 (remove `IMongoQueryable`)

src/MongoDB.Driver/TargetServerVersion.cs
```diff
@@ -0,0 +1,185 @@
...
+using System;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// Represents target server versions when translation Expressions to MQL.
+    /// </summary>
+    /// <note>Only major/minor versions are represented.</note>
+    public enum TargetServerVersion
```
sanych-sun, line 25: I like enum better then set of int constants. But do we really need both of them? Can we eliminate constants in `WireVersion` by replacing with this new enum?

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
```diff
@@ -210,6 +211,24 @@ public static MongoClientSettings GetClientSettings()
             return clientSettings;
         }
 
+        public static int GetTargetWireVersion()
```
sanych-sun, line 214: Should we call the method from `GetClientSettings` for proper defaults in tests?

sanych-sun (changes requested):

src/MongoDB.Driver/ExpressionTranslationOptions.cs
```diff
@@ -19,13 +19,39 @@ namespace MongoDB.Driver
...
+    public sealed class ExpressionTranslationOptions
     {
         internal static ExpressionTranslationOptions Default = new ExpressionTranslationOptions();
 
         /// <summary>
-        /// Gets or sets the string translation mode.
+        /// Gets or sets the server version to target when translating Expressions.
         /// </summary>
-        public AggregateStringTranslationMode? StringTranslationMode { get; set; }
+        public TargetServerVersion? TargetServerVersion { get; set; }
```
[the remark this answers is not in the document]
rstam: It is named `ExpressionTranslationOptions.TargetServerVersion` because this option means:

"When translating an Expression target this server version"

So it sounded quite natural to me.

src/MongoDB.Driver/TargetServerVersion.cs
rstam: Possibly... I didn't want to touch all the code that currently uses `WireVersion`.

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
rstam: No. 

`MongoClientSettings` is where the user tells us what server version to target. That's not necessarily the same as the server version currently connected to. If they don't tell us anything we default to target server 4.0. Almost all our tests should run with default settings (i.e. target server 4.0), and the few tests that need to target a newer server version will set their own `TargetServerVersion`.

This method determines the `TargetWireVersions` of the servers we are CURRENTLY connected to. It will be used in a subsequent commit related to `Join/GroupJoin` to determine what the current server is capable of, and to set the `TargetWireVersion` accordingly in some new tests.

src/MongoDB.Driver/ExpressionTranslationOptions.cs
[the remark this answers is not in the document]
sanych-sun: I'm not insisting, but term "Target" sounds strict for me (like "I expect the exact version out there"), and not obvious what system will do if there is different server version discovered. When term "Compatibility" - sounds more understandable for me like "expect at least this level of API available on that side of wire".

src/MongoDB.Driver/TargetServerVersion.cs
sanych-sun: Ok, Can we at least redefine WireVersion constants referencing new enum values? like 
`public const int Server26 = (int)ServerVersionEnum.Server26;`

This way it will be obvious that both places are connected and should have the same set of versions.

src/MongoDB.Driver/ExpressionTranslationOptions.cs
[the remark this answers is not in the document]
rstam: Discussed in Slack and we settled on `CompatibilityLevel`.

src/MongoDB.Driver/TargetServerVersion.cs
rstam: I've actually gone a step further and removed the integer values from the `ServerVersion` enum.

My thinking is that `ServerVersion` and `WireVersion` are actually two completely separate concepts and there is no reason they should have the same numeric values.

There is a mapping between them but it is not 1-to-1. There are many more server versions than there are wire versions.

When mapping from server version to wire version the result is the the wire version supported by that server version.

When mapping from wire version to server version the result is the FIRST server version that supported that wire version.

rstam:
Looks like a lot of failed variants due to some failing `SemanticVersionTests`.

I'll fix those... but they are basically orthogonal to the real work in this PR. They are related to the removal of the `ServerVersion` class, which we removed to free up the name for the new `ServerVersion` enum.

rstam:
I've fixed the SemanticVersion CompareTo issue.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1322,6 +1399,18 @@ private IBsonSerializer<TItem> GetItemSerializerForDistinctMany<TItem>(RenderedF
...
+        {
+            var translationOptions = _database.Client.Settings.TranslationOptions;
+            return new RenderArgs<TDocument>(_documentSerializer, _settings.SerializerRegistry, translationOptions: translationOptions);
+        }
+
+        private RenderArgs<TDocument> GetRenderArgs(ExpressionTranslationOptions translationOptions)
+        {
+            translationOptions = translationOptions.AddMissingOptionsFrom(_database.Client.Settings.TranslationOptions);
+            return new RenderArgs<TDocument>(_documentSerializer, _settings.SerializerRegistry, translationOptions: translationOptions);
+        }
```
JamesKovacs, line 1412: These methods may no longer be required or can simply return `_args`.

src/MongoDB.Driver/PipelineDefinition.cs
```diff
@@ -317,7 +329,7 @@ public override RenderedPipelineDefinition<TOutput> Render(RenderArgs<TInput> ar
             IBsonSerializer currentSerializer = args.DocumentSerializer;
             foreach (var stage in _stages)
             {
-                var renderedStage = stage.Render(currentSerializer, args.SerializerRegistry);
+                var renderedStage = stage.Render(currentSerializer, args.SerializerRegistry, args.TranslationOptions);
```
JamesKovacs, line 332: Given that `currentSerializer` is `args.DocumentSerializer`, we could simply pass `args`to `Render` instead of deconstructing it to pull out specific parameters.

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
```diff
@@ -210,6 +211,24 @@ public static MongoClientSettings GetClientSettings()
...
+            }
+
+            int? supportedWireVersion = null;
+            SpinWait.SpinUntil(() => (supportedWireVersion = cluster.Description.GetSupportedWireVersion()).HasValue, TimeSpan.FromSeconds(10));
+            if (!supportedWireVersion.HasValue)
+            {
+                throw new InvalidOperationException($"Unable to determine supported wire version: {cluster.Description}.");
+            }
+            return supportedWireVersion.Value;
+        }
```
JamesKovacs, line 230: Unused. Remove.

tests/MongoDB.Driver.Tests/Core/Clusters/ClusterDescriptionTests.cs
```diff
@@ -140,6 +140,25 @@ public void Equals_should_return_true_if_all_fields_are_equal()
...
+        {
+            var clusterDescription = CreateSubject();
+            var server0 = clusterDescription.Servers[0].With(wireVersionRange: maxWireVersion0.HasValue ? new Range<int>(WireVersion.Server36, maxWireVersion0.Value) : null);
+            var server1 = clusterDescription.Servers[1].With(wireVersionRange: maxWireVersion1.HasValue ? new Range<int>(WireVersion.Server36, maxWireVersion1.Value) : null);
+            var subject = clusterDescription.WithServerDescription(server0).WithServerDescription(server1);
+
+            var result = subject.GetSupportedWireVersion();
+
+            result.Should().Be(expectedWireVersion);
+        }
```
JamesKovacs, line 160: No longer used. Remove.

JamesKovacs (changes requested):
Overall looking good. There are still some remanants of trying to figure out `ServerVersion` from the `ClusterDescription` and those should be removed. Also there are a number of places where we pull apart `RenderArgs` to pass into a method whereas we could simply pass in `RenderArgs` directly.

src/MongoDB.Driver/Core/Clusters/ClusterDescription.cs
```diff
@@ -317,6 +317,13 @@ public override bool Equals(object obj)
             return Equals(obj as ClusterDescription);
         }
 
+         /// <summary>
+        /// Gets the highest wire version that can be safely targeted because it is the highest wire version supported by all connected servers.
+        /// </summary>
+        /// <returns>The highest wire version that is supported by all connected servers, or null if not connected to any servers.</returns>
+        public int? GetTargetWireVersion() =>
+            _servers.Min(s => s.WireVersionRange?.Max);
```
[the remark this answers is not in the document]
rstam: We will need this in the PRs that are blocked on this one.

This is how we determine at runtime what capabilities the servers we are connected to have.

src/MongoDB.Driver/Core/Misc/SemanticVersion.cs
```diff
@@ -39,9 +40,33 @@ private static void LookForPreReleaseNumericSuffix(string preRelease, out string
                 preReleaseNumericSuffix = null;
             }
         }
+
+        private static bool TryParseInternalPrelease(string preReleaseIn, out string preReleaseOut, out int? commitsAfterRelease, out string commitHash)
```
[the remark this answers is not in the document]
rstam: Oh interesting. I copy/pasted this from the deleted `ServerVersion.cs` file and didn't even  notice the typo!

Done.

src/MongoDB.Driver/ExpressionTranslationOptions.cs
```diff
@@ -19,13 +19,39 @@ namespace MongoDB.Driver
     /// <summary>
     /// Options for controlling translation from .NET expression trees into MongoDB expressions.
     /// </summary>
-    public class ExpressionTranslationOptions
+    public sealed class ExpressionTranslationOptions
     {
         internal static ExpressionTranslationOptions Default = new ExpressionTranslationOptions();
```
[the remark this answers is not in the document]
rstam: OK. It was being used for awhile but I switched over to using `null` to mean use defaults.

src/MongoDB.Driver/FieldDefinition.cs
```diff
@@ -242,7 +242,7 @@ public LambdaExpression Expression
         /// <inheritdoc />
         public override RenderedFieldDefinition Render(RenderArgs<TDocument> args)
         {
-            return LinqProviderAdapter.TranslateExpressionToField(_expression, args.DocumentSerializer, args.SerializerRegistry);
+            return LinqProviderAdapter.TranslateExpressionToField(_expression, args.DocumentSerializer, args.SerializerRegistry, args.TranslationOptions);
```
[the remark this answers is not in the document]
rstam: I briefly considered that, but these aren't `Render` methods.

I don't want to count on any coincidental overlap between translation arguments and render arguments.

src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
```diff
@@ -56,5 +56,11 @@ internal interface IMongoQueryProviderInternal : IMongoQueryProvider
         /// Gets the pipeline input serializer (the DocumentSerializer for collection queries and NoPipelineInputSerializer for database queries).
         /// </summary>
         IBsonSerializer PipelineInputSerializer { get; }
+
+        /// <summary>
+        /// Gets the translation options.
+        /// </summary>
+        /// <returns>The translation options.</returns>
+        ExpressionTranslationOptions GetTranslationOptions();
```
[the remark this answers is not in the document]
rstam: Why do you think it is unused? I find multiple references to it.

src/MongoDB.Driver/MongoCollectionImpl.cs
rstam: It's still true but for a slightly different reason. It can be overridden by the various Aggregate methods.

It still feels safer to determine the `RenderArgs` dynamically when needed. 

It would be extremely confusing if we sometimes used cached `RenderArgs` and other times did not.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -60,7 +59,6 @@ private MongoCollectionImpl(IMongoDatabase database, CollectionNamespace collect
             _documentSerializer = Ensure.IsNotNull(documentSerializer, nameof(documentSerializer));
 
             _messageEncoderSettings = GetMessageEncoderSettings();
-            _args = new(_documentSerializer, _settings.SerializerRegistry);
```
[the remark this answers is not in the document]
rstam: See comment above.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -103,9 +101,12 @@ public override MongoCollectionSettings Settings
         public override IAsyncCursor<TResult> Aggregate<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
         {
             Ensure.IsNotNull(session, nameof(session));
-            var renderedPipeline = Ensure.IsNotNull(pipeline, nameof(pipeline)).Render(_args);
+            Ensure.IsNotNull(pipeline, nameof(pipeline));
             options = options ?? new AggregateOptions();
 
+            var renderArgs = GetRenderArgs(options.TranslationOptions);
+            var renderedPipeline = pipeline.Render(renderArgs);
```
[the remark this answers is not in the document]
rstam: See comment above.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -138,9 +139,12 @@ public override MongoCollectionSettings Settings
         public override async Task<IAsyncCursor<TResult>> AggregateAsync<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
         {
             Ensure.IsNotNull(session, nameof(session));
-            var renderedPipeline = Ensure.IsNotNull(pipeline, nameof(pipeline)).Render(_args);
+            Ensure.IsNotNull(pipeline, nameof(pipeline));
             options = options ?? new AggregateOptions();
 
+            var renderArgs = GetRenderArgs(options.TranslationOptions);
+            var renderedPipeline = pipeline.Render(renderArgs);
```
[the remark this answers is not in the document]
rstam: Same here.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -173,9 +177,12 @@ public override MongoCollectionSettings Settings
         public override void AggregateToCollection<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
         {
             Ensure.IsNotNull(session, nameof(session));
-            var renderedPipeline = Ensure.IsNotNull(pipeline, nameof(pipeline)).Render(_args);
+            Ensure.IsNotNull(pipeline, nameof(pipeline));
             options = options ?? new AggregateOptions();
 
+            var renderArgs = GetRenderArgs(options.TranslationOptions);
+            var renderedPipeline = pipeline.Render(renderArgs);
```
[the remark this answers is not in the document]
rstam: And here.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -195,9 +202,12 @@ public override MongoCollectionSettings Settings
         public override async Task AggregateToCollectionAsync<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
         {
             Ensure.IsNotNull(session, nameof(session));
-            var renderedPipeline = Ensure.IsNotNull(pipeline, nameof(pipeline)).Render(_args);
+            Ensure.IsNotNull(pipeline, nameof(pipeline));
             options = options ?? new AggregateOptions();
 
+            var renderArgs = GetRenderArgs(options.TranslationOptions);
+            var renderedPipeline = pipeline.Render(renderArgs);
```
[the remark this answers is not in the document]
rstam: See previous reply.

src/MongoDB.Driver/MongoCollectionImpl.cs
rstam: Pay close attention to the fact there there are TWO methods and in one case the `translationOptions` are computed by merging two different sets of translation options.

src/MongoDB.Driver/PipelineDefinition.cs
rstam: `currentSerializer` changes for every iteration of the loop.

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
rstam: This is required in the next two PRs for tests that need to determine at runtime what the servers they are connected to support.

tests/MongoDB.Driver.Tests/Core/Clusters/ClusterDescriptionTests.cs
rstam: See previous comments about why this method exists.

src/MongoDB.Driver/FieldDefinition.cs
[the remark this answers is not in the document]
rstam: Also, `RenderArgs` have a bunch of other things that have nothing to do with LINQ translation.

src/MongoDB.Driver/Core/Clusters/ClusterDescription.cs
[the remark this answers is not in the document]
rstam: I thought about this some more and for testing purposes we can use `CoreTestConfiguration.MaxWireVersion` to determine the capabilities of the server we are connected to.

While `CoreTestConfiguration.MaxWireVersion` does not acount for clusters with mixed server versions, that's OK because our test clusters never have mixed server versions.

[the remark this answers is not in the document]
rstam: Removed

tests/MongoDB.Driver.TestHelpers/DriverTestConfiguration.cs
rstam: As mentioned in a previous comment I thought about this and decided that we can use the existing `CoreTestConfiguration.MaxWireVersion` property instead.

rstam: Removed.

tests/MongoDB.Driver.Tests/Core/Clusters/ClusterDescriptionTests.cs
rstam: Removed. See updates to previous comments.

JamesKovacs (approved):
LGTM

PR #1430, opened 2024-08-28
CSHARP-5251: Fix tests of Decimal128 and DateTime with locales

papafe:


tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -17,6 +17,7 @@
 using System.Dynamic;
 using System.IO;
 using System.Linq;
+using System.Threading;
```
rstam, line 20: This using statement did not need to be added.

rstam (changes requested):
Looks good but one small change requested.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
papafe: Done

PR #1432, opened 2024-08-28
CSHARP-5239: Clarify Reauthentication and Speculative Authentication combination behavior

sanych-sun:
The issue itself was fixed in scope of AWS separation/Auth refactoring PR. This PR is to add the new test to verify the fix.


papafe (approved):
LGTM

PR #1433, opened 2024-08-29
CSHARP-5252: Fix possible null reference

adelinowona:
When calling ComputeRescanDelay, there is a possibility that srvRecords is null so we should conditionally access it here.


BorisDog (approved):
LGTM

PR #1435, opened 2024-09-03
CSHARP-1309: Added unit tests to demonstrate that strongly-typed index definitions on POCO arrays now work with LINQ3.

JamesKovacs:
This PR only adds test cases to ensure we don't have a future regression.


papafe (approved):
LGTM

PR #1436, opened 2024-09-04
CSHARP-5225: Mongo Serialization Deadlock

sanych-sun:


tests/MongoDB.Bson.Tests/DeadLockInvestigation.cs
```diff
@@ -0,0 +1,73 @@
...
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Options;
+using Xunit;
+
+namespace MongoDB.Bson.Tests
+{
+    public class DeadLockInvestigation
+    {
+        [Fact]
+        public void TestDeadLock()
```
sanych-sun, line 33: This test will be removed before merging, because we have to add thread.sleep in factory method to make the problem visible.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -347,19 +347,26 @@ public static BsonClassMap LookupClassMap(Type classType)
                 BsonSerializer.ConfigLock.ExitReadLock();
             }
 
+            if (classMap == null)
```
rstam, line 350: I don't think the check for `null` makes sense.

The only time it won't be `null` is if it *was* registered but *not* frozen. I don't think that happens very often.

In this rare case we can just look it up again.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -347,19 +347,26 @@ public static BsonClassMap LookupClassMap(Type classType)
...
+                else
                 {
-                    // automatically create a classMap for classType and register it
-                    var classMapDefinition = typeof(BsonClassMap<>);
-                    var classMapType = classMapDefinition.MakeGenericType(classType);
-                    classMap = (BsonClassMap)Activator.CreateInstance(classMapType);
-                    classMap.AutoMap();
                     RegisterClassMap(classMap);
                 }
+
```
rstam, line 369: I would propose the following variation for lines 333-368:

    BsonSerializer.ConfigLock.EnterReadLock();                                                             
    try                                                                                                    
    {                                                                                                      
        if (__classMaps.TryGetValue(classType, out var classMap))                                          
        {                                                                                                  
            if (classMap.IsFrozen)                                                                         
            {                                                                                              
                return classMap;                                                                           
            }                                                                                              
        }                                                                                                  
    }                                                                                                      
    finally                                                                                                
    {                                                                                                      
        BsonSerializer.ConfigLock.ExitReadLock();                                                          
    }                                                                                                      
                                                                                                           
    // automatically create a new classMap for classType and register it (unless another thread does first)
    // do the work of speculatively creating a new class map outside of holding any lock                   
    var classMapDefinition = typeof(BsonClassMap<>);                                                       
    var classMapType = classMapDefinition.MakeGenericType(classType);                                      
    var newClassMap = (BsonClassMap)Activator.CreateInstance(classMapType);                                
    newClassMap.AutoMap();                                                                                 
                                                                                                           
    BsonSerializer.ConfigLock.EnterWriteLock();                                                            
    try                                                                                                    
    {                                                                                                      
        if (!__classMaps.TryGetValue(classType, out var classMap))                                         
        {                                                                                                  
            RegisterClassMap(newClassMap);                                                                 
            classMap = newClassMap;                                                                        
        }                                                                                                  
                                                                                                           
        return classMap.Freeze();                                                                          
    }                                                                                                      
    finally                                                                                                
    {                                                                                                      
        BsonSerializer.ConfigLock.ExitWriteLock();                                                         
    }

rstam (changes requested):
Let me know what you think of my comments. I might be off base... :)

src/MongoDB.Bson/Serialization/BsonClassMap.cs
sanych-sun: Sure can remove that check, I was not sure how often that might be happening. If it's absolutely rare case - then yep, definitely can simplify.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
sanych-sun: Done

rstam (approved):
Don't forget to remove the code that isn't supposed to be merged.

LGTM

PR #1438, opened 2024-09-04
CSHARP-5266: Switch to Ubuntu for package creation

sanych-sun:


src/Directory.Build.props
```diff
@@ -4,8 +4,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(IsWindows)'=='true'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net472</TargetFrameworks>
```
BorisDog, line 7: same change for tests?

BorisDog (changes requested):

src/Directory.Build.props
sanych-sun: Done

sanych-sun: Changes to tests were reverted. We need that condition there, so tests would not failure locally on non-Windows platforms.

BorisDog (approved):
LGTM! Please double check that the failures are not related to the changes.

PR #1440, opened 2024-09-05
CSHARP-5261: Add DateOnly support

papafe:
This PR is still in draft because we need support for the .NET 6 target framework before merging. I have made the changes for .NET 6 locally, but I did not push them to avoid polluting the PR. 

The behaviour of `DateOnlySerializer` has taken mostly inspiration from `DateTimeSerializer` with `DateOnly = true`, including throwing an exception when deserializing a date time with a non-zero time component.

### DIFF
--- a/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
@@ -15,6 +15,7 @@
 
 using System;
 using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization.Options;
 
 namespace MongoDB.Bson.Serialization.Serializers
 {
@@ -42,6 +43,7 @@ namespace MongoDB.Bson.Serialization.Serializers
         // private fields
         private readonly BsonType _representation;
         private readonly SerializerHelper _helper;
+        private readonly RepresentationConverter _converter;
 
         // constructors
         /// <summary>
@@ -71,6 +73,7 @@ namespace MongoDB.Bson.Serialization.Serializers
             }
 
             _representation = representation;
+            _converter = new RepresentationConverter(false, false);
 
             _helper = new SerializerHelper
             (
@@ -113,10 +116,19 @@ namespace MongoDB.Bson.Serialization.Serializers
                     break;
 
                 case BsonType.Decimal128:
+                    value = VerifyAndMakeDateOnly(new DateTime(_converter.ToInt64(bsonReader.ReadDecimal128()), DateTimeKind.Utc));
+                    break;
+
                 case BsonType.Double:
+                    value = VerifyAndMakeDateOnly(new DateTime(_converter.ToInt64(bsonReader.ReadDouble()), DateTimeKind.Utc));
+                    break;
+
                 case BsonType.Int32:
+                    value = VerifyAndMakeDateOnly(new DateTime(bsonReader.ReadInt32(), DateTimeKind.Utc));
+                    break;
+
                 case BsonType.Int64:
-                    value = VerifyAndMakeDateOnly(new DateTime(Int64Serializer.Instance.Deserialize(context), DateTimeKind.Utc));
+                    value = VerifyAndMakeDateOnly(new DateTime(bsonReader.ReadInt64(), DateTimeKind.Utc));
                     break;
 
                 case BsonType.String:
--- a/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
@@ -23,7 +23,7 @@ namespace MongoDB.Bson.Serialization.Serializers
     /// <summary>
     /// Represents a serializer for DateOnlys.
     /// </summary>
-    public class DateOnlySerializer : StructSerializerBase<DateOnly>, IRepresentationConfigurable<DateOnlySerializer>
+    public sealed class DateOnlySerializer : StructSerializerBase<DateOnly>, IRepresentationConfigurable<DateOnlySerializer>
     {
         // static
         private static readonly DateOnlySerializer __instance = new DateOnlySerializer();
@@ -41,9 +41,9 @@ namespace MongoDB.Bson.Serialization.Serializers
         }
 
         // private fields
-        private readonly BsonType _representation;
-        private readonly SerializerHelper _helper;
         private readonly RepresentationConverter _converter;
+        private readonly SerializerHelper _helper;
+        private readonly BsonType _representation;
 
         // constructors
         /// <summary>
--- a/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
+++ b/src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
@@ -1,0 +1,170 @@
+/* Copyright 2010-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using MongoDB.Bson.IO;
+
+namespace MongoDB.Bson.Serialization.Serializers
+{
+    #if NET6_0_OR_GREATER
+    /// <summary>
+    /// Represents a serializer for DateOnlys.
+    /// </summary>
+    public class DateOnlySerializer : StructSerializerBase<DateOnly>, IRepresentationConfigurable<DateOnlySerializer>
+    {
+        #region static
+        private static readonly DateOnlySerializer __instance = new DateOnlySerializer();
+
+        /// <summary>
+        /// Gets the default DateOnlySerializer.
+        /// </summary>
+        public static DateOnlySerializer Instance => __instance;
+        #endregion
+
+        private static class Flags
+        {
+            public const long DateTime = 1;
+            public const long Ticks = 2;
+        }
+
+        private readonly SerializerHelper _helper;
+
+        /// <inheritdoc />
+        public BsonType Representation { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DateOnlySerializer"/> class.
+        /// </summary>
+        public DateOnlySerializer()
+            : this(BsonType.DateTime)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DateOnlySerializer"/> class.
+        /// </summary>
+        /// <param name="representation">The representation.</param>
+        public DateOnlySerializer(BsonType representation)
+        {
+            switch (representation)
+            {
+                case BsonType.DateTime:
+                case BsonType.Document:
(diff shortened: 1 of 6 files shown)


src/MongoDB.Bson/Serialization/Serializers/DateTimeSerializer.cs
```diff
@@ -208,7 +208,7 @@ public override DateTime Deserialize(BsonDeserializationContext context, BsonDes
                     break;
 
                 case BsonType.Document:
-                    value = default(DateTime);
+                    value = default;
                     _helper.DeserializeMembers(context, (elementName, flag) =>
                     {
                         switch (flag)
```
papafe, line 214: I am not sure why here for `Ticks` we use `Int64Serializer`, while some lines down we use `bsonReader.ReadInt64`. Is it because here we want to be a little more permissive with the actual type used?

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -0,0 +1,198 @@
...
+                    bsonWriter.WriteInt64("Ticks", utcDateTime.Ticks);
+                    bsonWriter.WriteEndDocument();
+                    break;
+
+                case BsonType.Int64:
+                    bsonWriter.WriteInt64(utcDateTime.Ticks);
+                    break;
+
+                case BsonType.String:
+                    bsonWriter.WriteString(value.ToString("yyyy-MM-dd"));
```
rstam, line 176: Is this culture invariant?

rstam (changes requested):

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -0,0 +1,198 @@
...
+            _helper = new SerializerHelper
+            (
+                new SerializerHelper.Member("DateTime", Flags.DateTime),
+                new SerializerHelper.Member("Ticks", Flags.Ticks)
+            );
+        }
+
+        // public properties
+        /// <inheritdoc />
+        public BsonType Representation { get; }
```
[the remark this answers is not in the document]
papafe: Fixed

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -0,0 +1,198 @@
...
+                return DateOnly.FromDateTime(dt);
+            }
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj)
+        {
+            return base.Equals(obj) &&
+                obj is DateOnlySerializer other &&
+                Representation.Equals(other.Representation);
```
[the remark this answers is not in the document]
papafe: Fixed the `Equals` method. 
And I should already have the indent set to spaces.

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
papafe: Reading the documentation at https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings, this seems to be the case

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -0,0 +1,198 @@
...
+                        }
+                    });
+                    break;
+
+                case BsonType.Int64:
+                    value = VerifyAndMakeDateOnly(new DateTime(bsonReader.ReadInt64(), DateTimeKind.Utc));
+                    break;
+
+                case BsonType.String:
+                    value = DateOnly.ParseExact(bsonReader.ReadString(), "yyyy-MM-dd");
```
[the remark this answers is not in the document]
papafe: Reading the documentation at https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings, this seems to be the case

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -0,0 +1,198 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using MongoDB.Bson.IO;
+
+namespace MongoDB.Bson.Serialization.Serializers
+{
+    #if NET6_0_OR_GREATER
```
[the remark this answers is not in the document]
papafe: Fixed

papafe:
@rstam I've fixed the things you noted in your review and that we discussed yesterday. 
The only modification I have done is to use the `Int64Serializer` instead of spelling out the different cases for the various `BsonType` in the deserializer. I have decided to do so mostly to avoid code repetition and keeping the code for the deserialization of numeric types in the `Int64Serializer`, so in case we decide to do that kind of deserialization in other ways we can modify only that serializer and not also all the others. I can definitely be convinced otherwise though 😄

rstam:
> The only modification I have done is to use the `Int64Serializer` instead of spelling out the different cases for the various `BsonType` in the deserializer.

It's not wrong to do this, but we expect that 99% of the time the `BsonType` will actually be `Int64` and this approach involves several more levels of indirection before finally calling `ReadInt64` after all.

It's a tradeoff between conciseness and efficiency.

I made the comment in our Zoom conversation that serialization is the slowest part of our driver so maybe we need to worry more about efficiency.

rstam:
> Another question. I have added the `BsonType.Document` representation to align with the `DateTimeSerializer`. I remember you saying that it was a legacy representation for `DateTime`. Should we maybe avoid this representation at least for the new `DateOnlySerializer`?

The `Document` representation for `DateTime` is NOT legacy. It has a purpose.

The purpose is this: when representing a `DateTime` as a `BsonDateTime` there is a loss of precision. When saving a `DateTime` as an `Int64` we maintain full precision but make it hard to do MQL queries based on dates.

The `Document` representation stores the `DateTime` twice: once for easy of querying in MQL and again for keeping full precision.

That being said, IF we have a `Document` representation for `DateOnly` it would probably look different than the `Document` representation for `DateTime`.

papafe:
> > The only modification I have done is to use the `Int64Serializer` instead of spelling out the different cases for the various `BsonType` in the deserializer.
> 
> It's not wrong to do this, but we expect that 99% of the time the `BsonType` will actually be `Int64` and this approach involves several more levels of indirection before finally calling `ReadInt64` after all.
> 
> It's a tradeoff between conciseness and efficiency.
> 
> I made the comment in our Zoom conversation that serialization is the slowest part of our driver so maybe we need to worry more about efficiency.



> > The only modification I have done is to use the `Int64Serializer` instead of spelling out the different cases for the various `BsonType` in the deserializer.
> 
> It's not wrong to do this, but we expect that 99% of the time the `BsonType` will actually be `Int64` and this approach involves several more levels of indirection before finally calling `ReadInt64` after all.
> 
> It's a tradeoff between conciseness and efficiency.
> 
> I made the comment in our Zoom conversation that serialization is the slowest part of our driver so maybe we need to worry more about efficiency.

I have corrected this.

papafe:
@rstam The work for .NET 6 has been merged to main and I've rebased on top of that, so this should be mergeable after your review.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
```diff
@@ -0,0 +1,259 @@
...
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using FluentAssertions;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
```
rstam, line 27: Remove unused usings.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            [Values(BsonType.DateTime, BsonType.String, BsonType.Int64, BsonType.Document)]
+            BsonType representation)
+        {
+            var subject = new DateOnlySerializer(representation);
+
+            subject.Representation.Should().Be(representation);
+        }
+
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDouble" : "638649792000000000" } }""","10/20/2024" )]
```
rstam, line 58: I would use `"2024-10-20"` as the string for a date to avoid confusion about whether month or date goes first.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
```diff
@@ -0,0 +1,259 @@
...
+        public void Deserialize_should_throw_when_date_has_time(string json)
+        {
+            var subject = new DateOnlySerializer();
+
+            using var reader = new JsonReader(json);
+            reader.ReadStartDocument();
+            reader.ReadName("x");
+            var context = BsonDeserializationContext.CreateRoot(reader);
+
+            Action action = () => subject.Deserialize(context);
```
rstam, line 118: We use the following:

```
var exception = Record.Exception(() => subject.Deserialize(context));

exception.Should().BeOfType<FormatException>();
exception.Message.Should().Be("...");
```

```

**Review by rstam — CHANGES_REQUESTED** [review]

**Comment by papafe** [comment]
@rstam I've done the corrections following your points

File: tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs [hunk 10]
```diff
@@ -0,0 +1,241 @@
...
+            [Values(BsonType.DateTime, BsonType.String, BsonType.Int64, BsonType.Document)]
+            BsonType representation)
+        {
+            var subject = new DateOnlySerializer(representation);
+
+            subject.Representation.Should().Be(representation);
+        }
+
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDouble" : "638649792000000000" } }""","2024/10/20" )]
```
rstam, line 51: The ISO standard for culture invariant dates is to use dashes: `2024-10-20`

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            var subject = new DateOnlySerializer();
+
+            using var reader = new JsonReader(json);
+            reader.ReadStartDocument();
+            reader.ReadName("x");
+            var context = BsonDeserializationContext.CreateRoot(reader);
+
+            var exception = Record.Exception(() => subject.Deserialize(context));
+            exception!.Should().BeOfType<FormatException>();
```
rstam, line 112: I understand what the `!` is doing here but I didn't think we needed it with our current project settings.

rstam (changes requested):

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
papafe: Sorry, when you wrote it the first time I did not notice the dashes, I've just corrected id

papafe:
Corrected according to notes

tests/MongoDB.Bson.Tests/Serialization/Serializers/DateOnlySerializerTests.cs
papafe: You're right, corrected it.

rstam (approved):
LGTM

PR #1441, opened 2024-09-05
CSHARP-5250: Change Decimal128 default serialization to Bson.Decimal128

papafe:


papafe:
@rstam `Decimal128Serializer` has a static property called `Decimal128Instance` that is a "a cached instance of a Decimal128Serializer with Decimal128 representation". Given that now the `Decimal128` is the default representation, should we change the name of the static property to just `Instance` like with other serializers?

rstam:
> Given that now the `Decimal128` is the default representation, should we change the name of the static property to just `Instance` like with other serializers?

Sure. That makes sense.

src/MongoDB.Bson/Serialization/Serializers/Decimal128Serializer.cs
```diff
@@ -42,7 +42,7 @@ public class Decimal128Serializer : StructSerializerBase<Decimal128>, IRepresent
         /// Initializes a new instance of the <see cref="Decimal128Serializer"/> class.
         /// </summary>
         public Decimal128Serializer()
-            : this(BsonType.String)
+            : this(BsonType.Decimal128)
```
rstam, line 45: We need to make the same changes to `DecimalSerializer`.

There actually is no `Decimal128` datatype in .NET or a `decimal` type in BSON so we map between them when serializing and deserializing.

rstam (changes requested):
Looking good!

src/MongoDB.Bson/Serialization/Serializers/Decimal128Serializer.cs
papafe: Makes sense. Done

rstam (approved):
LGTM

PR #1448, opened 2024-09-13
CSHARP-4494: Consider marking serializers sealed

papafe:


papafe:
@rstam I have made all the possible serializers sealed, apart from `EnumSerializer` as there is a derived type in `CSharp4172Tests`. Should we use a different serializer for the test or should we keep the serializer open? In this case the derived type, `CamelCaseEnumSerializer` could be implemented with a parameter on the base serializer to decide the casing as well. 
I think this also points out that we do not know exactly if someone is subclassing the serializers now and I suppose we will discover it only with the new version.

rstam:
> here is a derived type in `CSharp4172Tests`. Should we use a different serializer for the test

Yes. We can seal `EnumSerializer` and change `CamelCaseEnumSerializer` in `CSharp4172.cs` to derive directly from `StructSerializerBase`. Not sure why it ever derived from `EnumSerializer`, it doesn't seem to need to.

rstam (changes requested):
I think we should seal `EnumSerializer` also.

papafe:
@rstam I've sealed also the `EnumSerializer`

rstam (approved):
LGTM

Skyppid:
May I ask why this has been done? As a consuming developer this has caused us quite some issues upgrading to 3.0.0. We actually inherited the `EnumSerializer` to extend it's functionality (to be specific, we had cases were we renamed enum values and have been using string-based representation, so we wrote an extended serializer that could handle these scenarios by tagging enum fields with alternative names that the serializer considers if old names are used in the database while deserializing). Marking it sealed forced us to rewrite the whole serializer (for ease of use I just copied it and removed the sealed argument). Since our inherited serializer basically just builds on top of the original's deserialization logic, it's useless having to reimplement serialization as well for that case.

The serializer is already implemented well, so why not allow developers to extend upon that?

BorisDog:
Hi @Skyppid 
We understand that this is a major breaking change. But for better support and maintainability in the long term, in 3.0 we've eliminated most of the API that was not designed to be public. This also includers the serializers that were never intended to be extended directly. The extension point for custom serializers is implementing IBsonSerializer. 
This is part of our general recommendation of using "composition over inheritance" approach as well as extension methods for extending the driver functionality.

PR #1449, opened 2024-09-17
CSHARP-4909: Add serialization support for new Numeric types

papafe:


tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,438 @@
...
+            var result = subject.WithRepresentation(newRepresentation);
+
+            result.Representation.Should().Be(newRepresentation);
+            if (newRepresentation == oldRepresentation)
+            {
+                result.Should().BeSameAs(subject);
+            }
+        }
+
+        private static void TestDeserialize(HalfSerializer subject, Half expectedValue, string json)
```
rstam, line 384: I would order the parameters like this:

* `subject`
* `json`
* `expectedResult`

(note also renaming of `expectedValue` to `expectedResult` which is our common naming convention)

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,438 @@
...
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDecimal" : "18" } }""")]
+        [InlineData("""{ "x" : { "$numberDouble" : "18" } }""")]
+        [InlineData("""{ "x" : { "$numberLong" : "18" } }""")]
+        [InlineData("""{ "x" : { "$numberInt" : "18" } }""")]
+        [InlineData("""{ "x" : "18" }""")]
+        public void Deserialize_should_have_expected_result(string json)
+        {
+            var subject = new HalfSerializer(BsonType.Decimal128, new RepresentationConverter(true, true));
+            var expectedValue = (Half)18;
```
rstam, line 59: We normally use `expectedResult` as the name of this variable.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,438 @@
...
+            result.Should().Be(expectedValue);
+        }
+
+        private static void TestDeserializeWithException<T>(HalfSerializer subject, string json) where T : Exception
+        {
+            using var reader = new JsonReader(json);
+            reader.ReadStartDocument();
+            reader.ReadName("x");
+            var context = BsonDeserializationContext.CreateRoot(reader);
+            Action action = () => subject.Deserialize(context);
```
rstam, line 402: Use FluentAssertions: `Record.Exception` and `exception.Should().BeOfType<>`.

rstam (changes requested):
Overall looks good. Minor changes requested.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,440 @@
...
+            string expectedResult)
+        {
+            var subject = new HalfSerializer(representation, new RepresentationConverter(true, true));
+            var halfValue = Half.MinValue;
+
+            TestSerialize(subject, halfValue, expectedResult);
+        }
+
+        // Int32 and Int64 are not tested because the conversion between NaN/Infinite values and integral values
+        // gives an indefinite result, and as such we cannot guarantee correct deserialization
```
rstam, line 325: We should test that trying to serialize a `NaN` throws an exception and that the exception is of the expected type and with a correct message.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,440 @@
...
+            string expectedResult)
+        {
+            var subject = new HalfSerializer(representation, new RepresentationConverter(true, true));
+            var halfValue = Half.NaN;
+
+            TestSerialize(subject, halfValue, expectedResult);
+        }
+
+        // Int32 and Int64 are not tested because the conversion between NaN/Infinite values and integral values
+        // gives an indefinite result, and as such we cannot guarantee correct deserialization
```
rstam, line 340: Same here.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,440 @@
...
+            string expectedResult)
+        {
+            var subject = new HalfSerializer(representation, new RepresentationConverter(true, true));
+            var halfValue = Half.NegativeInfinity;
+
+            TestSerialize(subject, halfValue, expectedResult);
+        }
+
+        // Int32 and Int64 are not tested because the conversion between NaN/Infinite values and integral values
+        // gives an indefinite result, and as such we cannot guarantee correct deserialization
```
rstam, line 355: And here.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
rstam: This variable and others with the same name still need to be renamed.

rstam (changes requested):

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
papafe: You're right, my bad!

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
papafe: The reason why I kept it like this is to be consistent with the behaviour of the floating point serializer (`SingleSerializer`, `DoubleSerializer`). With those, there's no exception when serializing `NaN` to an integer value (independently from `allowOverflow` value). Should we have those throw an exception as well?

Regarding the serialization of infinity, I have realised I have misinterpreted the behaviour of the floating point serializers, so now an `ArithmeticException` will be thrown when trying to serialize an infinity value and `allowOverflow == false`. I have also added tests for those

rstam (changes requested):

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
papafe: Sorry, you're correct, I'm throwing an `OverflowException`, not an `ArithmeticException`. 

And regarding what to do when serializing `Infinity` and `allowOverflow == true`, that is a good question. Personally I would throw also in this case (possibly an `OverflowException` as well) as we cannot guarantee a good serialization.

src/MongoDB.Bson/IO/JsonConvert.cs
```diff
@@ -200,6 +212,18 @@ public static string ToString(float value)
...
 
+#if NET5_0_OR_GREATER
+        /// <summary>
+        /// Converts a Half to a string.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>A string.</returns>
+        public static string ToString(Half value)
+        {
+            return value.ToString("R", NumberFormatInfo.InvariantInfo);
```
rstam, line 223: Recommend we use "G5" instead of "R".

That way `Half.MaxValue` converts to the expected `65504` and not the weird `65500`.

This page recommends against using "R":

https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+
+            TestDeserialize(subject, json, expectedResult);
+        }
+
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDecimal" : "9.999999999999999999999999999999999E+6144" } }""")] //Decimal128.MaxValue
+        [InlineData("""{ "x" : { "$numberDouble" : "1.7976931348623157E+308" } }""")] //double.MaxValue
+        [InlineData("""{ "x" : { "$numberLong" : "65504" } }""")]
+        [InlineData("""{ "x" : { "$numberInt" : "65504" } }""")]
+        [InlineData("""{ "x" : "65500" }""")]
```
rstam, line 81: If we fix `JsonConverter.ToString` we can use 65504 instead of 65500.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+            var expectedResult = (Half)18.5;
+
+            TestDeserialize(subject, json, expectedResult);
+        }
+
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDecimal" : "9.999999999999999999999999999999999E+6144" } }""")] //Decimal128.MaxValue
+        [InlineData("""{ "x" : { "$numberDouble" : "1.7976931348623157E+308" } }""")] //double.MaxValue
+        [InlineData("""{ "x" : { "$numberLong" : "65504" } }""")]
+        [InlineData("""{ "x" : { "$numberInt" : "65504" } }""")]
```
rstam, line 80: Previous two lines should use max values for long and int.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+            var expectedResult = Half.MaxValue;
+
+            TestDeserialize(subject, json, expectedResult);
+        }
+
+        [Theory]
+        [InlineData("""{ "x" : { "$numberDecimal" : "-9.999999999999999999999999999999999E+6144" } }""")] //Decimal128.MinValue
+        [InlineData("""{ "x" : { "$numberDouble" : "-1.7976931348623157E+308" } }""")] //double.MinValue
+        [InlineData("""{ "x" : { "$numberLong" : "-65504" } }""")]
+        [InlineData("""{ "x" : { "$numberInt" : "-65504" } }""")]
```
rstam, line 94: Previous two lines should use min values for long and int.

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+        [InlineData("""{ "x" : "-65500" }""")]
```
rstam, line 95: Should be `"-65504"`

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+
+            TestSerializeWithException<TruncationException>(subject, halfValue);
+        }
+
+        [Theory]
+        [InlineData(BsonType.Decimal128, """{ "x" : { "$numberDecimal" : "9.999999999999999999999999999999999E+6144" } }""")] //Decimal128.MaxValue
+        [InlineData(BsonType.Double, """{ "x" : { "$numberDouble" : "1.7976931348623157E+308" } }""")] //double.MaxValue
+        [InlineData(BsonType.Int64, """{ "x" : { "$numberLong" : "65504" } }""")]
+        [InlineData(BsonType.Int32, """{ "x" : { "$numberInt" : "65504" } }""")]
+        [InlineData(BsonType.String, """{ "x" : "65500" }""")]
```
rstam, line 293: Should be `"65504"`

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
```diff
@@ -0,0 +1,464 @@
...
+
+            TestSerialize(subject, halfValue, expectedResult);
+        }
+
+        [Theory]
+        [InlineData(BsonType.Decimal128, """{ "x" : { "$numberDecimal" : "-9.999999999999999999999999999999999E+6144" } }""")] //Decimal128.MinValue
+        [InlineData(BsonType.Double, """{ "x" : { "$numberDouble" : "-1.7976931348623157E+308" } }""")] //double.MinValue
+        [InlineData(BsonType.Int64, """{ "x" : { "$numberLong" : "-65504" } }""")]
+        [InlineData(BsonType.Int32, """{ "x" : { "$numberInt" : "-65504" } }""")]
+        [InlineData(BsonType.String, """{ "x" : "-65500" }""")]
```
rstam, line 308: Should be `"-65504"`

rstam (changes requested):
Small changes requested.

papafe:
Corrected the string formatting

papafe:
@rstam I think now the PR should be ready for a final review

tests/MongoDB.Bson.Tests/Serialization/Serializers/HalfSerializerTests.cs
papafe: We decided to keep the conversion between `Min/MaxValues` only between floating point, as it is now

rstam (approved):
Nice!

PR #1451, opened 2024-09-17
CSHARP-4586: Skip test on servers before 4.4.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4586Tests.cs
```diff
@@ -14,6 +14,8 @@
 */
 
 using FluentAssertions;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
 using Xunit;
 
 #if NET6_0_OR_GREATER
```
rstam, line 21: This #if is here because the test using an init property.

But... up until just now we've never actually run tests against net6.

BorisDog (approved):
LGTM

PR #1453, opened 2024-09-18
CSHARP-5201: Allow update to supply sort option

papafe:


src/MongoDB.Driver/MongoCollectionBase.cs
```diff
@@ -856,8 +856,10 @@ private UpdateResult UpdateOne(FilterDefinition<TDocument> filter, UpdateDefinit
                 ArrayFilters = options.ArrayFilters,
                 Collation = options.Collation,
                 Hint = options.Hint,
-                IsUpsert = options.IsUpsert
+                IsUpsert = options.IsUpsert,
+                Sort = options.Sort
             };
+            //TODO Why don't UpdateOne and UpdateOneAsync have a common path for creating the UpdateOneModel and BulkWriteOptions (if not more)?
```
BorisDog, line 862: In that case that can be done elegantly, we should do that in a separate ticket for other methods as well. Usually we can address such improvements on green-build Fridays.

src/MongoDB.Driver/Core/Operations/UpdateRequest.cs
```diff
@@ -130,6 +131,15 @@ public UpdateType UpdateType
             get { return _updateType; }
         }
 
+        /// <summary>
+        /// Gets or sets the sort document.
+        /// </summary>
+        public BsonDocument Sort
```
BorisDog, line 137: Alphabetical order.

BorisDog (commented):
Looks good overall! Now we need to wait for spec finalization.

src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
```diff
@@ -95,6 +95,11 @@ public BulkWriteReplaceOneModel(
         /// </summary>
         public TDocument Replacement { get; }
 
+        /// <summary>
+        /// The sort document to use.
+        /// </summary>
+        public BsonDocument Sort { get; init; }
```
sanych-sun, line 101: I suppose it should be `SortDefinition<TDocument>` instead of BsonDocument.

src/MongoDB.Driver/BulkWriteUpdateOneModel.cs
```diff
@@ -26,6 +26,7 @@ namespace MongoDB.Driver
     /// <typeparam name="TDocument">The type of the document.</typeparam>
     public sealed class BulkWriteUpdateOneModel<TDocument> : BulkWriteModel
     {
+        //TODO Maybe we need to pass the sort argument also to the constructors...?
```
sanych-sun, line 29: I suppose we should have both: `init` and constructor. Init - it's a convenient modern way, ctor - it's needed for older dev environments where `init` is not supported.

src/MongoDB.Driver/BulkWriteUpdateOneModel.cs
```diff
@@ -101,6 +102,11 @@ public BulkWriteUpdateOneModel(
         /// </summary>
         public bool IsUpsert { get; init; }
 
+        /// <summary>
+        /// The sort document to use.
+        /// </summary>
+        public BsonDocument Sort { get; init; }
```
sanych-sun, line 108: Should we use `SortDefinition<TDocument>`?

src/MongoDB.Driver/ReplaceOneModel.cs
```diff
@@ -31,6 +31,7 @@ public sealed class ReplaceOneModel<TDocument> : WriteModel<TDocument>
         private BsonValue _hint;
         private bool _isUpsert;
         private readonly TDocument _replacement;
+        private BsonDocument _sort;
```
sanych-sun, line 34: Should we use `SortDefinition<TDocument>`?

src/MongoDB.Driver/ReplaceOptions.cs
```diff
@@ -117,5 +119,14 @@ public BsonDocument Let
             get { return _let; }
             set { _let = value; }
         }
+
+        /// <summary>
+        /// Gets or sets the sort document.
+        /// </summary>
+        public BsonDocument Sort
```
sanych-sun, line 126: Use `SortDefinition<TDocument>`.

src/MongoDB.Driver/UpdateOneModel.cs
```diff
@@ -95,6 +96,15 @@ public bool IsUpsert
             set { _isUpsert = value; }
         }
 
+        /// <summary>
+        /// Gets or sets the sort document.
+        /// </summary>
+        public BsonDocument Sort
```
sanych-sun, line 102: Use `SortDefinition<TDocument>`.

src/MongoDB.Driver/UpdateOptions.cs
```diff
@@ -98,5 +99,14 @@ public BsonDocument Let
             get { return _let; }
             set { _let = value; }
         }
+
+        /// <summary>
+        /// Gets or sets the sort document.
+        /// </summary>
+        public BsonDocument Sort
```
sanych-sun, line 106: Use `SortDefinition<TDocument>`.

sanych-sun (changes requested):

src/MongoDB.Driver/UpdateOptions.cs
papafe: Unfortunately `UpdateOptions` and `ReplaceOptions` are not generic classes, so we can't parametrize the properties. We can add the parameter, but I think it would be a breaking change. 
We could also add the generic `UpdateOptions<TDocument>` alongside the non generic version and make new methods that use this, while deprecating the methods that are using the non generico version. This would be the long road

papafe: I have done this only for `ReplaceOptions` for now. If all I've done makes sense, I'll do the same for `UpdateOptions`. 
This required to create a new class `ReplaceOptions<T>` and an implicit conversion operator from `ReplaceOptions` to `ReplaceOptions<T>`.

src/MongoDB.Driver/ReplaceOptions.cs
```diff
@@ -118,4 +86,82 @@ public BsonDocument Let
             set { _let = value; }
         }
     }
+
+    /// <summary>
+    /// Options for replacing a single document.
+    /// </summary>
+    public sealed class ReplaceOptions : ReplaceOptionsBase
```
BorisDog, line 93: Need to ensure that this change is not breaking.

src/MongoDB.Driver/IMongoCollection.cs
```diff
@@ -972,8 +972,7 @@ public interface IMongoCollection<TDocument>
         /// <returns>
         /// The result of the replacement.
         /// </returns>
-        ReplaceOneResult ReplaceOne(FilterDefinition<TDocument> filter, TDocument replacement, ReplaceOptions options = null, CancellationToken cancellationToken = default(CancellationToken));
-
+        ReplaceOneResult ReplaceOne(FilterDefinition<TDocument> filter, TDocument replacement, ReplaceOptions<TDocument> options = null, CancellationToken cancellationToken = default(CancellationToken));
```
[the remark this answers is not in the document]
BorisDog: Yup we avoid breaking changes.

BorisDog (changes requested):

src/MongoDB.Driver/IMongoCollection.cs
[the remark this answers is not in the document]
rstam: This looks like a binary breaking change but not a source code breaking change.

There is also an implicit conversion operator from `ReplaceOptions` to `ReplaceOptions<TDocument>` that provides backward compatibility.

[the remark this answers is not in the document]
papafe: Yes, I discussed this with Robert and this was an attempt to avoid creating more methods/overloads and keep source compatibility.
I was reading a little bit around online about semver in regards to binary compatibility. It seems that there is not a complete agreement on it, but overall it's recommended to increase major version with a binary-incompatible change. I'm not sure what's our stance on this debate, but I think I'll make an overload to be on the safe side 😁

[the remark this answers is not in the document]
rstam: I would prefer a solution that doesn't involve doubling (or worse) the number of overloads. Too many overloads is a bad thing.

The current approach uses an implicit conversion from `ReplaceOptions` to `ReplaceOptions<TDocument>` to preserve backward compatibility.

Another approach could be to use subclassing: `ReplaceOptions<TDocument> : ReplaceOptions`. Then we could leave the current overloads unchanged at the expense of some type safety. The code that uses the `replaceOptions` would have to test whether the passed in options are the subclass or not.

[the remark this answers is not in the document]
papafe: Given what we discussed at Triage today I'll park this ticket until next when we can have a more general discussion about breaking changes in general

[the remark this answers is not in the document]
sanych-sun: I have to apologize, I did not notice the implicit conversion. So this is not code breaking change.
On other hand I suppose it's better to follow Robert's suggestion with using of subclasses as clearer way to do the same thing. At least because this way we could avoid changes to public API.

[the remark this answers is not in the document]
papafe: I went for Robert suggestion and made the generic options derive from the base class we have now

src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
```diff
@@ -34,14 +34,16 @@ public sealed class BulkWriteReplaceOneModel<TDocument> : BulkWriteModel
...
             string collectionNamespace,
             FilterDefinition<TDocument> filter,
             TDocument replacement,
             Collation collation = null,
             BsonValue hint = null,
-            bool isUpsert = false)
-            : this(CollectionNamespace.FromFullName(collectionNamespace), filter, replacement, collation, hint, isUpsert)
+            bool isUpsert = false,
+            SortDefinition<TDocument> sort = null)
+            : this(CollectionNamespace.FromFullName(collectionNamespace), filter, replacement, collation, hint, isUpsert, sort)
```
papafe, line 46: This is not binary compatible.
The alternative would be to add other two constructors (given that we already have two) but:
- Then we have 4 very similar constructors
- Developers using those constructors will have errors due to constructor ambiguity if they do not specify all parameters. This is theoretically still source compatible, but definitely annoying
Any suggestions about this? Personally I'd just keep the sort definition out of the constructor if we cannot find another alternative

sanych-sun: What if we add new constructor with `sort` as the very last non-optional parameter? In such way it would not be a breaking change and there would not be an ambiguity.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/IMongoCollectionExtensionsTests.cs
```diff
@@ -1132,7 +1132,7 @@ public void ReplaceOne_should_call_collection_with_expected_arguments(
 
             assertReplaceOne();
 
-            var replaceOptions = new ReplaceOptions();
+            var replaceOptions = new ReplaceOptions<Person>();
```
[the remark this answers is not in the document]
papafe: Definitely

src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
papafe: That's a good idea, I didn't think about it :)
I've added two new constructor to mirror the ones we have already, and added a new private one that is being called by all the others

sanych-sun: Nice! And I suppose the plan should be to eliminate this copied ctors in next major release, by making the `sort` parameter as optional. So the question is: how we going to track this and not forget to do?

sanych-sun (commented):
I'm ready to LGTM the changes, but could you please rebase before that? Because in the current main we already have that apicompat script, just want to be sure we did everything right in terms of backward compatibility.

src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
papafe: Yup, that would be the plan! I've added the constructors and the non generic options to the list of breaking changes for next version.

papafe:
@sanych-sun just rebased on main, curious to see what will the script say 😁

papafe:
@sanych-sun it seems the validate-apicompat task gave its ok 😁 
Do you think this is mergeable now?

sanych-sun (approved):
LGTM

PR #1454, opened 2024-09-18
CSHARP-5286: Add IHierarchicalDiscriminatorConvention and IScalarDiscriminatorConvention interfaces.

rstam:


rstam:
Here is a high level overview of the changes:

Added new interfaces:
* IHierarchicalDiscriminatorConvention
* IScalarDiscriminatorConvention
* IHasDiscriminatorConvention (and new GetDiscriminatorConvention helper extension method)

Rules about how discriminators work:
* discriminator can be omitted when nominal type == actual type (i.e. root class)
* discriminator for root class can be scalar even for hierarchical convention
* discriminator value of null means don't serialize _t

Additional notes:
* added SetDiscriminatorConvention to BsonClassMap
* the existing HierarchichalDiscriminatorConvention implements the new IHierarchicalDiscrminatorConvention interface
* the existing ScalarDiscriminatorConvention does not implement the new IScalarDiscriminatorConvention (yet)
* the only initial user of the in IScalarDiscriminatorConvention will be EF

src/MongoDB.Bson/Serialization/IHasDiscriminatorConvention.cs
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+using MongoDB.Bson.Serialization.Conventions;
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// Represents a serializer that has a DiscriminatorConvention property.
+    /// </summary>
+    public interface IHasDiscriminatorConvention
```
rstam, line 23: Named after the existing `IHasRepresentationSerializer`.

Just noticed I didn't add the `Serializer` suffix. Should I add it?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/GetTypeComparisonExpressionToAggregationExpressionTranslator.cs
```diff
@@ -40,15 +40,14 @@ public static AggregationExpression Translate(TranslationContext context, Binary
                 var nominalType = objectExpression.Type;
                 var actualType = comparandType;
 
-                var discriminatorConvention = objectTranslation.Serializer is ObjectSerializer objectSerializer ?
-                    objectSerializer.DiscriminatorConvention :
-                    BsonSerializer.LookupDiscriminatorConvention(nominalType);
+                var discriminatorConvention = objectTranslation.Serializer.GetDiscriminatorConvention();
                 var discriminatorField = AstExpression.GetField(objectTranslation.Ast, discriminatorConvention.ElementName);
-                var discriminatorValue = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
```
rstam, line 45: Standardizing on `discriminator` instead of `discriminatorValue` everywhere.

The return value of `GetDiscriminator` is a `discriminator`...

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/TypeIsExpressionToAggregationExpressionTranslator.cs
```diff
@@ -29,24 +30,41 @@ public static AggregationExpression Translate(TranslationContext context, TypeBi
...
-                objectSerializer.DiscriminatorConvention :
-                BsonSerializer.LookupDiscriminatorConvention(nominalType);
-            var discriminatorField = AstExpression.GetField(objectTranslation.Ast, discriminatorConvention.ElementName);
-            var discriminatorValue = discriminatorConvention.GetDiscriminator(nominalType, actualType);
-            if (discriminatorValue is BsonArray array)
+            AstExpression ast;
+            if (nominalType == actualType)
             {
-                discriminatorValue = array.Last();
+                ast = AstExpression.Constant(true);
```
rstam, line 40: Optimization for trivial case.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/TypeIsExpressionToAggregationExpressionTranslator.cs
```diff
@@ -29,24 +30,41 @@ public static AggregationExpression Translate(TranslationContext context, TypeBi
...
+                if (discriminatorConvention is IHierarchicalDiscriminatorConvention)
+                {
+                    var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                    var actualTypeDiscriminator = discriminator is BsonArray array ? array.Last() : discriminator;
+                    ast = AstExpression.Cond(
+                        AstExpression.Eq(AstExpression.Type(discriminatorField), "array"),
+                        AstExpression.In(actualTypeDiscriminator, discriminatorField),
+                        AstExpression.Eq(discriminatorField, actualTypeDiscriminator));
+                }
+                else if (discriminatorConvention is IScalarDiscriminatorConvention scalarDiscriminatorConvention)
```
rstam, line 56: This is new MQL to support queries against scalar discriminators.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/TypeIsExpressionToAggregationExpressionTranslator.cs
```diff
@@ -29,24 +30,41 @@ public static AggregationExpression Translate(TranslationContext context, TypeBi
...
-                    AstExpression.IsArray(discriminatorField),
-                    AstExpression.In(discriminatorValue, discriminatorField)));
+                if (discriminatorConvention is IHierarchicalDiscriminatorConvention)
+                {
+                    var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                    var actualTypeDiscriminator = discriminator is BsonArray array ? array.Last() : discriminator;
+                    ast = AstExpression.Cond(
+                        AstExpression.Eq(AstExpression.Type(discriminatorField), "array"),
+                        AstExpression.In(actualTypeDiscriminator, discriminatorField),
+                        AstExpression.Eq(discriminatorField, actualTypeDiscriminator));
```
rstam, line 54: I believe this MQL is slightly better than what we were generating.

Thoughts?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/AllOrAnyMethodToFilterTranslator.cs
```diff
@@ -115,10 +115,10 @@ public static (AstFilterField, AstFilter) Translate(TranslationContext context,
 
                     var nominalType = ArraySerializerHelper.GetItemSerializer(sourceField.Serializer).ValueType;
                     var actualType = method.GetGenericArguments()[0];
-                    var discriminatorConvention = BsonSerializer.LookupDiscriminatorConvention(actualType);
+                    var discriminatorConvention = sourceField.Serializer.GetDiscriminatorConvention();
                     var discriminatorField = AstFilter.Field(discriminatorConvention.ElementName, BsonValueSerializer.Instance);
-                    var discriminatorValue = discriminatorConvention.GetDiscriminator(nominalType, actualType);
-                    var ofTypeFilter = AstFilter.Eq(discriminatorField, discriminatorValue);
+                    var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+                    var ofTypeFilter = AstFilter.Eq(discriminatorField, discriminator); // TODO: hierarchical vs scalar?
```
rstam, line 121: Feels like there is some more work to do here.

I'll have to figure this one out.

tests/MongoDB.Driver.Tests/Jira/CSharp2622Tests.cs
```diff
@@ -47,6 +49,8 @@ public void InsertOne_to_oftype_collection_should_generate_id()
         }
 
         // nested types
+        [BsonDiscriminator(RootClass = true)]
+        [BsonKnownTypes(typeof(D))]
```
rstam, line 53: This class was misconfigured.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4691Tests.cs
```diff
@@ -37,6 +37,12 @@ static CSharp4691Tests()
                 var objectSerializer = new ObjectSerializer(discriminatorConvention, allowedTypes);
                 cm.MapMember(x => x.Object).SetSerializer(objectSerializer);
             });
+
+            BsonClassMap.RegisterClassMap<ActivityObject<int>>(cm =>
+            {
+                cm.AutoMap();
+                cm.SetIsRootClass(true);
+            });
```
rstam, line 45: Not sure this configuration could have been done with attributes.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5188Tests.cs
```diff
@@ -189,7 +189,7 @@ public void Where_GetType_equals_G_should_work()
                 .Where(x => x.GetType() == typeof(G));
 
             var stages = Translate(collection, queryable);
-            AssertStages(stages, "{ $match : { '_t.0' : { $exists : false }, '_t' : 'G' } }");
+            AssertStages(stages, "{ $match : { '_t' : 'G' } }");
```
rstam, line 192: `F`, `G` and `H` use a scalar discriminator so the query can be simplified.

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationWithLinq2Tests/Translators/PredicateTranslatorTests.cs
```diff
@@ -171,7 +171,7 @@ public void Any_with_a_predicate_on_documents_itself_and_ClassEquals()
             Assert(
                 x => x.G.Any(g => g == c1),
                 1,
-                "{ \"G\" : { \"Ids\" : null, \"D\" : \"Dolphin\", \"E\" : { \"F\" : 55, \"H\" : 66, \"S\": null, \"I\" : [\"insecure\"], \"C\" : null }, \"S\" : null, \"X\" : null, \"Y\" : null, \"Z\" : null } }");
+                "{ \"G\" : { \"Ids\" : null, \"D\" : \"Dolphin\", \"E\" : { _t : \"E\", \"F\" : 55, \"H\" : 66, \"S\": null, \"I\" : [\"insecure\"], \"C\" : null }, \"S\" : null, \"X\" : null, \"Y\" : null, \"Z\" : null } }");
```
rstam, line 174: `E` is now configured to use a hierarchical discriminator with the discriminator required even for the root class.

Note: `OfType` and `is` queries against hierarchical discriminators only work if the discriminator for an instance of the root class is not omitted, that's why it is required.

tests/MongoDB.Driver.Tests/FilterDefinitionBuilderTests.cs
```diff
@@ -843,10 +843,10 @@ public void OfType_Typed()
             Assert(personFilter, "{ $or : [{ _t : \"Twin\", wasBornFirst : true }, { _t : \"Triplet\", birthOrder : 1 }] }");
 
             // test OfType overloads that apply to a field of the document
-            Assert(subject.OfType<Animal, Cat>("favoritePet"), "{ \"favoritePet._t\" : \"Cat\" }");
-            Assert(subject.OfType<Animal, Cat>("favoritePet", Builders<Cat>.Filter.Eq(c => c.LivesLeft, 9)), "{ \"favoritePet._t\" : \"Cat\", \"favoritePet.livesLeft\" : 9 }");
-            Assert(subject.OfType<Animal, Cat>("favoritePet", "{ livesLeft : 9 }"), "{ \"favoritePet._t\" : \"Cat\", \"favoritePet.livesLeft\" : 9 }");
-            Assert(subject.OfType<Animal, Cat>("favoritePet", BsonDocument.Parse("{ livesLeft : 9 }")), "{ \"favoritePet._t\" : \"Cat\", \"favoritePet.livesLeft\" : 9 }");
+            Assert(subject.OfType<Animal, Cat>("FavoritePet"), "{ \"favoritePet._t\" : \"Cat\" }");
```
rstam, line 846: Not sure how this worked with the lower case name.

tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -871,10 +871,12 @@ private OfTypeMongoCollection<A, B> CreateSubject()
         {
             var mockRootCollection = new Mock<IMongoCollection<TRoot>>();
             mockRootCollection.SetupGet(c => c.CollectionNamespace).Returns(CollectionNamespace.FromFullName("foo.bar"));
+            mockRootCollection.SetupGet(c => c.DocumentSerializer).Returns(BsonSerializer.LookupSerializer<TRoot>());
```
rstam, line 874: I don't like testing with mocks because tests fail in weird ways when the mock is not properly set up.

src/MongoDB.Bson/Serialization/IHasDiscriminatorConvention.cs
damieng: I could go either way but have a preference for not having Serializer in the name. Originally I'd proposed IDiscriminationConventionProvider to match some of the other interfaces but I wasn't aware of IHasRepresentationSerializer.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -398,13 +398,13 @@ public static IDiscriminatorConvention LookupDiscriminatorConvention(Type type)
...
                         // inherit the discriminator convention from the closest parent (that isn't object) that has one
-                        // otherwise default to the standard hierarchical convention
+                        // otherwise default to the standard scalar convention
                         Type parentType = typeInfo.BaseType;
                         while (true)
                         {
                             if (parentType == typeof(object))
                             {
-                                convention = StandardDiscriminatorConvention.Hierarchical;
+                                convention = StandardDiscriminatorConvention.Scalar;
```
damieng, line 407: Is this a breaking change? Or is it safe given that heirarchical would only have emitted a single value and that it did that without wrapping it in a bsonarray?

src/MongoDB.Bson/Serialization/Conventions/ScalarDiscriminatorConvention.cs
```diff
@@ -43,7 +43,14 @@ public override BsonValue GetDiscriminator(Type nominalType, Type actualType)
         {
             // TODO: this isn't quite right, not all classes are serialized using a class map serializer
             var classMap = BsonClassMap.LookupClassMap(actualType);
-            return classMap.Discriminator;
+            if (actualType != nominalType || classMap.DiscriminatorIsRequired)
```
damieng, line 46: Why don't we want to store the discriminator in this case? Wouldn't it be beneficial to store it anyway?

damieng (commented):
Haven't had chance to go through the query/visitor stuff yet, a few comments for now. Will try this version with my EF fork to see how it goes in a couple of hours.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
rstam: It might be breaking but I'm not sure how to analyze it.

The issue is that a hierarchical discriminator MUST have a root class that is the beginning of the hierarchy, so defaulting to hierarchical here was just plain wrong because there is no known root.

src/MongoDB.Bson/Serialization/Conventions/ScalarDiscriminatorConvention.cs
rstam: In order to preserve existing behavior.

I don't' recall exactly why we didn't store the discriminator for the root class when not using the hierarchical discriminator convention. Could have been because it's redundant and not storing it saves a few bytes?

The other reason might be that we don't actually know when serializing an instance of the root class whether it's part of a polymorphic hierarchy or not. So for example we want to avoid storing `{ _t : "Employee" }` in a collection of employees where every document is an instance of `Employee` and there are no subclasses.

src/MongoDB.Bson/Serialization/IHasDiscriminatorConvention.cs
rstam: Our interfaces that end with `Provider` are more like factory interfaces that create new objects of some kind.

damieng:
Can confirm all EF Provider functional tests passing with these new changes including adding the IHasDiscriminator onto the EntitySerializer and no longer registering them on `BsonSerializer.__discriminatorConventions` via reflection. 💯

JamesKovacs (changes requested):
We should avoid switching on types and simply implement a method to encapsulate the discriminator convention logic.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -2183,17 +2204,42 @@ public FieldOfTypeFilterDefinition(FieldDefinition<TDocument, TField> field, Fil
...
+                }
 
-            var renderedField = _field.Render(args);
-            var discriminatorElementName = renderedField.FieldName + "." + discriminatorConvention.ElementName;
-            var discriminatorValue = discriminatorConvention.GetDiscriminator(typeof(TField), typeof(TDerived));
-            var renderedDiscriminatorFilter = new BsonDocument(discriminatorElementName, discriminatorValue);
+                var discriminatorElementName = renderedField.FieldName + "." + discriminatorConvention.ElementName;
+                var discriminator = discriminatorConvention.GetDiscriminator(typeof(TField), typeof(TDerived));
+
+                if (discriminatorConvention is IHierarchicalDiscriminatorConvention)
```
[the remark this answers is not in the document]
rstam: I don't think we want LINQ translation code in the discriminator convention.

The discriminator convention is ONLY about providing the values for `_t` during serialization and determining the *actual* type during deserialization.

How those `_t` values might or might not be used in the MQL generated by a LINQ query is not a concern of the discriminator convention.

Remember also that  users might want to implement their own discriminator conventions and in case they do we want to only burden them with the most minimal implementation demands possible.

Having said that, I do acknowledge the repetitive nature of some of this code. I will look for creating ways to refactor.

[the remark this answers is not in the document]
rstam: Note: `IDiscriminatorConvention` *already* has a method called `GetDiscriminator` whose purpose is to supply the value to store in the `_t` discriminator field.

rstam:
> We should avoid switching on types and simply implement a method to encapsulate the discriminator convention logic.

I don't think we can avoid switching on types, but I will look for a way to refactor to reduce the repetitive parts.

I had in fact already started looking at that but what looks repetitive at first glance is not 100% the same in each case so refactoring the repetitive part is not trivial.

rstam:
@JamesKovacs I've worked on addressing your concerns about code duplication.

The approach I came up with was to move the generation logic to the Ast classes. This is makes sense in general, but is slightly at odds with most Ast methods which tend to match closely to server MQL features. The new `TypeEquals` and `TypeIs` methods are at a slightly higher level because they can generate slightly more complex MQL. You can think of them as "macros".

I also modifed the Builders that were generating MQL by generating BsonDocuments directly to instead call the new Ast methods and then Rendering the Ast. This further reduced code duplication.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -856,6 +857,32 @@ public static AstExpression Type(AstExpression arg)
             return new AstUnaryExpression(AstUnaryOperator.Type, arg);
         }
 
+        public static AstExpression TypeEquals(AstGetFieldExpression discriminatorField, IDiscriminatorConvention discriminatorConvention, Type nominalType, Type actualType)
```
rstam, line 860: These new methods might be better as internal extension methods on the discriminator conventions (with suitably changed names).

It sort of feels wrong to me that the AST classes would know anything about discriminator conventions, because the AST mirrors server MQL and discriminator conventions are just something the driver knows about and AST shouldn't know about them.

Thoughts?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/DiscriminatorAstExpression.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Linq;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Translators;
+
+internal static class DiscriminatorAstExpression
+{
+    public static AstExpression TypeEquals(AstGetFieldExpression discriminatorField, IDiscriminatorConvention discriminatorConvention, Type nominalType, Type actualType)
```
rstam, line 26: I moved these methods here from `AstExpression.cs` because the Ast classes should not have any knowledge of how discriminators work.

Since there was no existing good place to put these methods I created a new class.

I named the class `DiscriminatorAstExpression` because it's like the `AstExpression` factory class but with knowledge of how discriminators are represented.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/DiscriminatorAstFilter.cs
```diff
@@ -0,0 +1,69 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Filters;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Translators;
+
+internal static class DiscriminatorAstFilter
+{
+    public static AstFilter TypeEquals(AstFilterField discriminatorField, IHierarchicalDiscriminatorConvention discriminatorConvention, Type nominalType, Type actualType)
```
rstam, line 27: I moved these methods here from `AstFilter.cs` because the Ast classes should not have any knowledge of how discriminators work.

Since there was no existing good place to put these methods I created a new class.

I named the class `DiscriminatorAstFilter` because it's like the `AstFilter` factory class but with knowledge of how discriminators are represented.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -2137,24 +2142,40 @@ public DocumentOfTypeFilterDefinition(FilterDefinition<TDerived> derivedDocument
...
+                }
+
+                var discriminator = discriminatorConvention.GetDiscriminator(typeof(TDocument), typeof(TDerived));
+                if (discriminator == null)
+                {
+                    throw new NotSupportedException($"OfType requires that documents of type {BsonUtils.GetFriendlyTypeName(typeof(TDerived))} have a discriminator value.");
+                }
+
+                var discriminatorField = new AstFilterField(discriminatorConvention.ElementName, BsonValueSerializer.Instance);
+                ofTypeFilter= discriminatorConvention switch
```
damieng, line 2169: Nit: Missing space.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/DiscriminatorAstExpression.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        var discriminator = discriminatorConvention.GetDiscriminator(nominalType, actualType);
+        return discriminator == null ?
+            AstExpression.IsMissing(discriminatorField) :
```
damieng, line 30: Under what scenario does this get hit? We'd need to say we have a discriminator convention... but then return a null? And that would then exclude all docs with a discriminator value I presume. Wondering what query gets us here and the use case.

src/MongoDB.Driver/OfTypeMongoCollection.cs
```diff
@@ -82,7 +82,7 @@ protected override UpdateDefinition<TDerivedDocument> AdjustUpdateDefinition(Upd
                                             new BsonArray
                                             {
                                                 new BsonDocument("$eq", new BsonArray { new BsonDocument("$type", "$_id"), "missing" }), // if "_id" is missed
```
damieng, line 84: Is `$type` remapped elsewhere or should this be using `discriminatorConvention.ElementName` ?

damieng (approved):
One nit and a few questions but nothing blocking.

Tested this version with latest EF type discriminator branch and still passing tests.

JamesKovacs (approved):
LGTM

PR #1456, opened 2024-09-19
Optimize GetFriendlyTypeName method for performance & allocation

wioniqle-q:
Hi, 

This pull request optimizes the `GetFriendlyTypeName` method in the `BsonUtils` class to improve performance and reduce memory allocations, especially for complex generic types.  You can merge if appropriate.

Testing:
- Ran unit tests(20 pieces) covering various scenarios, including:
  - Simple types
  - Generic types
  - Nested generic types
  - Array types
  - Edge cases (null, deeply nested generics)
  
Performance improvements:
 - for SimpleType : %15.4
 - for GenericType: %68.7
 - for NestedGenericType: %71.4 
  
**Perf Test 1**
performance_test
**Perf Test 2** 
performance_test_2

### DIFF
--- a/src/MongoDB.Bson/BsonUtils.cs
+++ b/src/MongoDB.Bson/BsonUtils.cs
@@ -32,23 +32,9 @@ namespace MongoDB.Bson
         /// </summary>
         /// <param name="type">The type.</param>
         /// <returns>A friendly class name.</returns>
-        public static string GetFriendlyTypeName(Type type)
+        public static string GetFriendlyTypeName(this Type type)
         {
-            var typeInfo = type.GetTypeInfo();
-            if (!typeInfo.IsGenericType)
-            {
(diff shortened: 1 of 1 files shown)


damieng:
Unless I'm mistake this function is only ever used when we're throwing exceptions?

If so, I don't think there are going to be any real-world performance gains given exceptions shouldn't be happening in regular/hot paths and produce a whole bunch of allocations and performance issues of their own.

wioniqle-q:
> Unless I'm mistake this function is only ever used when we're throwing exceptions?
> 
> If so, I don't think there are going to be any real-world performance gains given exceptions shouldn't be happening in regular/hot paths and produce a whole bunch of allocations and performance issues of their own.

Yes sir, my aim was to optimise the existing function, if it is unnecessary we can close it.

PR #1457, opened 2024-09-20
CSHARP-3601: Removed [Serializable] attribute

papafe:


papafe:
In the process of removing the attribute I had to remove quite a lot of tests, including the test that was verifying that [CSHARP-74](https://jira.mongodb.org/browse/CSHARP-74) had been fixed.

adelinowona (approved):
LGTM

PR #1459, opened 2024-09-21
CSHARP-5263: Remove TLS1.0 and 1.1

BorisDog:


JamesKovacs (approved):
LGTM.

I considered whether to use `SslProtocols.Tls13` with our `net6.0` rather than `SslStreamSettings` line 38, but I couldn't think of a clean way to `#if NET6_0` it. This just seemed ugly:

```
#if NET6_0_OR_GREATER
        internal const SslProtocols SslProtocolsTls13 = SslProtocols.Tls13;
#else
        internal const SslProtocols SslProtocolsTls13 = (SslProtocols)12288;
#endif
```

And `#if NET6_0` around its usages was even worse.

I think it's fine leaving the hard-coded constant in the code until .NET 6 is our lowest TFM.

PR #1467, opened 2024-09-29
CSHARP-5262: Add TimeOnly support

papafe:


rstam (changes requested):
Please re-request my review after the branch is linear.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,139 @@
...
+        private static readonly TimeOnlySerializer __instance = new ();
+
+        /// <summary>
+        /// Gets the default TimeOnlySerializer
+        /// </summary>
+        public static TimeOnlySerializer Instance => __instance;
+
+        // private fields
+        private readonly RepresentationConverter _converter;
+        private readonly BsonType _representation;
```
rstam, line 37: Do we want to allow the user to configure `units` like we do in the `TimeSpanSerializer`?

A `TimeOnly` value is kind of like a `TimeSpan`.

I could imagine someone wanting the `TimeOnly` value stored in the database as hours or minutes instead of some oddball unit like ticks.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeOnlySerializer"/> class.
+        /// </summary>
+        /// <param name="representation">The representation.</param>
+        public TimeOnlySerializer(BsonType representation)
+        {
+            switch (representation)
+            {
+                case BsonType.Int64:
+                case BsonType.String:
```
rstam, line 57: What about `Int32` or `Double`?

`Double` would be important if we also allow them to specify the units.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,139 @@
...
+        // public methods
+        /// <inheritdoc/>
+        public override TimeOnly Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
+        {
+            var bsonReader = context.Reader;
+            var bsonType = bsonReader.GetCurrentBsonType();
+
+            return bsonType switch
+            {
+                BsonType.String => TimeOnly.ParseExact(bsonReader.ReadString(), "o"),
```
rstam, line 81: Is this culture invariant?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,139 @@
...
+            var bsonWriter = context.Writer;
+
+            switch (_representation)
+            {
+                case BsonType.Int64:
+                    bsonWriter.WriteInt64(value.Ticks);
+                    break;
+
+                case BsonType.String:
+                    bsonWriter.WriteString(value.ToString("o"));
```
rstam, line 118: Is this culture invariant?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,139 @@
...
+        public override bool Equals(object obj)
+        {
+            if (object.ReferenceEquals(obj, null)) { return false; }
+            if (object.ReferenceEquals(this, obj)) { return true; }
+
+            return
+                base.Equals(obj) &&
+                obj is TimeOnlySerializer other &&
+                _representation.Equals(other._representation) &&
+                _converter.Equals(other._converter);
```
rstam, line 100: The `_converter` is not user configurable and therefore doesn't actually need to be compared here.

`_converter.Equals(other._converter)` will always return `true`.

rstam (changes requested):
Some questions.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Yes, it should be: https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-round-trip-o-o-format-specifier

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: As before

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: I did not think about it... Do we want to give the same kind of units for this serializer too?

papafe: I am adding configurable `units` too. Something that I've noticed is that for the string representation, the `TimeSpanSerializer` does not distinguish between the units. I think this makes sense, but I just want to be sure we want to do the same for the `TimeOnlySerializer` too

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: I have added `Int32` and `Double` as a consequence of adding various `units`

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Ah right 🤦 . Corrected

papafe:
I have added the possibility of choosing different units, as well as tests for those

src/MongoDB.Bson/Serialization/Options/TimeOnlyUnits.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        Seconds,
+        /// <summary>
+        /// Use milliseconds as the units.
+        /// </summary>
+        Milliseconds,
+        /// <summary>
+        /// Use microseconds as the units.
+        /// </summary>
+        Microseconds
```
rstam, line 46: Curious why you omitted `Nanoseconds` (which are present in `TimeSpanUnits`)?

Would it be a good idea to have the *same* units for `TimeOnly` as we already have for `TimeSpan`?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,231 @@
...
+        // static
+        private static readonly TimeOnlySerializer __instance = new ();
+
+        /// <summary>
+        /// Gets the default TimeOnlySerializer
+        /// </summary>
+        public static TimeOnlySerializer Instance => __instance;
+
+        // private fields
+        private readonly RepresentationConverter _converter;
```
rstam, line 36: `_converter` is not used?

rstam (changes requested):

papafe:
@rstam Added nanoseconds as a unit, as we discussed and removed the unused converter

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -0,0 +1,241 @@
...
+                return this;
+            }
+
+            return new TimeOnlySerializer(representation, units);
+        }
+
+        // private methods
+        private TimeOnly FromDouble(double value, TimeOnlyUnits units)
+        {
+            return units is TimeOnlyUnits.Nanoseconds
```
rstam, line 180: Kind of odd to use `is` instead of `==` ...

rstam (approved):
LGTM with minor comment

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Yeah, in this case it doesn't make much of a difference, but recently I have started using the pattern matching `is` compared to `==` because with this it is possible do things like:
```
if (units is Units.Microseconds or Units.Nanoseconds or Units.Days)
```

PR #1471, opened 2024-10-01
CSHARP-5295: Improve error message when subtracting two decimal values stored as strings in the database.

rstam:


BorisDog (approved):
LGTM

rstam:
I'm looking into the failing tests.

Looks like we are removing the widening converts too early which results in some errors. The approach I'm working on is to remove them at the last minute when generating the MQL.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/ConvertHelper.cs
```diff
@@ -149,20 +151,20 @@ public static Expression RemoveConvertToObject(Expression expression)
             return expression;
         }
 
-        public static Expression RemoveWideningConvert(Expression expression)
+        public static AstExpression RemoveWideningConvert(AggregationExpression translation)
```
rstam, line 154: By returning `AstExpression` instead of `Expression` we ensure that the removal of the widening convert only happens at the MQL level.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs
```diff
@@ -53,12 +53,6 @@ public static AggregationExpression Translate(TranslationContext context, Binary
                 throw new ExpressionNotSupportedException(expression, because: "operand types are not compatible with each other");
             }
 
-            if (IsArithmeticExpression(expression))
-            {
-                leftExpression = ConvertHelper.RemoveWideningConvert(leftExpression);
-                rightExpression = ConvertHelper.RemoveWideningConvert(rightExpression);
```
rstam, line 59: The problem with doing it this early (and at this level) is that removing the `Convert` expression also resulted in the translation having a different serializer which resulted in an exception being thrown at the new line 84 below.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs
```diff
@@ -81,38 +75,43 @@ public static AggregationExpression Translate(TranslationContext context, Binary
...
             }
 
+            var leftAst = leftTranslation.Ast;
+            var rightAst = rightTranslation.Ast;
             if (IsArithmeticExpression(expression))
             {
-                SerializationHelper.EnsureRepresentationIsNumeric(leftExpression, leftTranslation);
-                SerializationHelper.EnsureRepresentationIsNumeric(rightExpression, rightTranslation);
+                SerializationHelper.EnsureRepresentationIsNumeric(expression, leftExpression, leftTranslation);
+                SerializationHelper.EnsureRepresentationIsNumeric(expression, rightExpression, rightTranslation);
```
rstam, line 83: I think you were right that requiring the left and the right expressions to uses the *same* serializer is too strict.

Instead let's just check that they both have a numeric representation.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/SerializationHelper.cs
```diff
@@ -36,17 +36,17 @@ public static void EnsureRepresentationIsArray(Expression expression, IBsonSeria
...
-        public static void EnsureRepresentationIsNumeric(Expression expression, IBsonSerializer serializer)
+        public static void EnsureRepresentationIsNumeric(Expression expression, Expression argumentExpression, IBsonSerializer argumentSerializer)
         {
-            var representation = GetRepresentation(serializer);
-            if (!IsNumericRepresentation(representation))
+            var argumentRepresentation = GetRepresentation(argumentSerializer);
+            if (!IsNumericRepresentation(argumentRepresentation))
             {
-                throw new ExpressionNotSupportedException(expression, because: $"serializer for type {serializer.ValueType} uses a non-numeric representation: {representation}");
+                throw new ExpressionNotSupportedException(expression, because: $"{argumentExpression} uses a non-numeric representation: {argumentRepresentation}");
```
rstam, line 49: The key change here is that we include the whole expression for context and call out the specific argument expression that is not represented numerically in the database.

rstam:
@BorisDog I pushed a new commit that relaxes the check in `BinaryExpressionToAggregationExpressionTranslator` that you correctly identified as being too strict.

I think this new commit might be more likely to be included in 3.0.

The other work besides improving the error messages is to refactor the removal of the widening convert from the `Expression` level to the `Ast` level. I don't think this will change any behavior, but it is the more correct thing to do because the goal is to simplify the AST so that should probably be done at the AST level.

BorisDog (approved):
LGTM

PR #1473, opened 2024-10-01
CSHARP-3181: Change DateTimeOffset default serialization

papafe:
This was needed because the default DateTimeOffset serialization makes it impossible to search across timezones


papafe:
@rstam I have merged the two `DateTimeOffsetSerializerTests` classes, and I have removed the `TestSerializeDateTimeOffset` testing method. I have done so as the standalone testing class already has enough tests, and the test method seemed to follow the old testing conventions.

rstam (approved):
LGTM

PR #1474, opened 2024-10-01
CSHARP-2480: Fix ambiguous API calls

papafe:
Regarding FindOneAndReplace and similar methods


papafe:
I have tried to reproduce the same issue with other methods, but those seem the only one affected.

sanych-sun (approved):
LGTM

PR #1482, opened 2024-10-07
CSHARP-5311: Remove deprecation from BSON area

papafe:


src/MongoDB.Bson/ObjectModel/BsonBinarySubType.cs
```diff
@@ -33,7 +33,7 @@ public enum BsonBinarySubType
         /// <summary>
         /// Obsolete binary data subtype (use Binary instead).
         /// </summary>
-        [Obsolete("Use Binary instead")]
+        [Obsolete("Use Binary instead")]  //TODO Should we remove this?
```
papafe, line 36: Not sure if we should remove this here

src/MongoDB.Bson/ObjectModel/BsonDocument.cs
```diff
@@ -167,7 +132,7 @@ public BsonDocument(IEnumerable<BsonElement> elements)
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2214:DoNotCallOverridableMethodsInConstructors")]
         public BsonDocument(params BsonElement[] elements)
```
papafe, line 133: Not sure if we want to remove this constructor, as in some cases it's easier than to create an enumerable and pass it. On the other hand maybe we do not want to have so many similar constructor

src/MongoDB.Bson/IO/JsonOutputMode.cs
```diff
@@ -25,7 +25,7 @@ public enum JsonOutputMode
         /// <summary>
         /// Output strict JSON (an obsolete output mode that is similar to canonical extended JSON).
         /// </summary>
-        [Obsolete("Use CanonicalExtendedJson instead.")]
+        [Obsolete("Use CanonicalExtendedJson instead.")]  //TODO Should we remove this?
```
papafe, line 28: Not sure if we should remove this

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
```diff
@@ -605,7 +605,7 @@ protected virtual ICollection<KeyValuePair<TKey, TValue>> CreateAccumulator()
         /// Creates the instance.
         /// </summary>
         /// <returns>The instance.</returns>
-        [Obsolete("CreateInstance is deprecated. Please use CreateAccumulator instead.")]
+        [Obsolete("CreateInstance is deprecated. Please use CreateAccumulator instead.")]  //TODO Not sure what we want to do with this, it's used by the CreateAccumulator method
```
papafe, line 608: I'm not sure about this, the `CreateAccumulator` method is actually using this one  in its body.

papafe:
@rstam I have removed everything with the `Obsolete` attribute from almost everywhere in the BSON package, apart from a couple of places where I'm unsure. I've left a TODO and a comment in those places.

src/MongoDB.Bson/IO/JsonOutputMode.cs
rstam: Good question! Not sure either.

src/MongoDB.Bson/ObjectModel/BsonBinarySubType.cs
rstam: We could, but... if we do anyone who has really old data might not be able to read it anymore.

I'd ask Boris and Rishit.

src/MongoDB.Bson/ObjectModel/BsonDocument.cs
```diff
@@ -159,6 +123,7 @@ public BsonDocument(IEnumerable<BsonElement> elements)
             AddRange(elements);
         }
 
+        //TODO Are we sure we want to remove this? Having a constructor with params is useful
```
rstam, line 126: Agreed.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
rstam: I think we do want to remove this method and change `CreateAccumulator` to throw `NotImplementedException`.

src/MongoDB.Bson/IO/JsonOutputMode.cs
rstam: FYI `Strict` is a format the C# driver used BEFORE there was any extended JSON standard, but it is very similar to `CanonicalExtendedJson`.

rstam (changes requested):
Overall looks great!

src/MongoDB.Bson/ObjectModel/BsonBinarySubType.cs
papafe: @BorisDog @rishitb-mongodb What do you think, should we remove this?

src/MongoDB.Bson/IO/JsonOutputMode.cs
papafe: @BorisDog @rishitb-mongodb Also here, do you think we should remove `JsonOutputMode.Strict`?

src/MongoDB.Bson/ObjectModel/BsonDocument.cs
BorisDog: Either way it fine IMHO.

src/MongoDB.Bson/ObjectModel/BsonDocument.cs
papafe: Removed the comment and left the constructor

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
papafe: Done

src/MongoDB.Bson/IO/JsonOutputMode.cs
papafe: We decided to remove it

src/MongoDB.Bson/ObjectModel/BsonBinarySubType.cs
papafe: We decided to keep it, for compatibility reasons

PR #1484, opened 2024-10-08
CSHARP-5304: Mark MD5 support in GridFS as Obsolete in 2.x

papafe:


src/MongoDB.Driver.GridFS/GridFSDownloadOptions.cs
```diff
@@ -31,6 +33,7 @@ public class GridFSDownloadOptions
         /// <value>
         ///   <c>true</c> if the MD5 value should be checked; otherwise, <c>false</c>.
         /// </value>
+        [Obsolete("MD5 support will be removed in a later release")]
```
BorisDog, line 36: No building errors in tests?

papafe: I did not think about it to be honest 🤦  
I've just added the necessary pragmas

BorisDog (approved):
LGTM

PR #1486, opened 2024-10-09
CSHARP-5315:  Mark API as obsolete: Auth

papafe:
Mark API that will be remove in 3.0 as obsolete: Auth


papafe:
I have removed removed the CS0618 (use of obsolete members) error from both the libraries and the tests. I wanted to do it only for the tests, but we got so many references also in the libraries. This is the last PR where I have added the pragmas in the libraries, as I have removed the error after I've finished already the work

PR #1489, opened 2024-10-09
CSHARP-5318: Mark API as obsolete: Server + Sessions

papafe:


papafe:
@BorisDog I noticed that actually `IServerSelector` and all the implementing classes are not internal on 3.0. Are they supposed to be internal or public?

MongoDB.ruleset
```diff
@@ -105,5 +105,6 @@
     <Rule Id="CA2242" Action="Warning" />
     <Rule Id="CA5350" Action="None" />
     <Rule Id="CA5351" Action="None" />
+    <Rule Id="CS0618" Action="None"/>
```
rstam, line 108: I understand why you did this but I'm not sure it's a good idea. It could allow us to make mistakes on this branch. The ONLY reason this might be OK is because we are about to abandon work on this branch.

Let's never do this on the `main` branch!!!

rstam (approved):
LGTM with the caveat that turning off warnings is a very bad idea in general and only possibly justified because we are about to abandon work on this branch.

MongoDB.ruleset
papafe: I totally agree with you. I proposed it exactly because this branch is going to be abandoned and the alternative is to add thousands of pragmas all around the libraries

papafe:
I have just made another pass at the review comparing it to main, and I've removed the attribute where the member is public on main

src/MongoDB.Driver/IClientSession.cs
```diff
@@ -89,6 +89,7 @@ public interface IClientSession : IDisposable
         /// <value>
         /// The wrapped core session.
         /// </value>
+        [Obsolete("This property will be removed in a later release.")]
```
BorisDog, line 92: Unfortunately we did not remove this in 3.0.

papafe: Removed

rstam (approved):
LGTM

PR #1491, opened 2024-10-10
CHARP-4752: Remove deprecated GridFS fields

papafe:


src/MongoDB.Driver/GridFS/GridFSFileInfoCompat.cs
```diff
@@ -40,19 +40,6 @@ public GridFSFileInfo(BsonDocument backingDocument)
...
 
-        // public properties
-        /// <summary>
-        /// Gets the aliases.
-        /// </summary>
-        /// <value>
-        /// The aliases.
-        /// </value>
-        [Obsolete("Place aliases inside metadata instead.")]
-        public IEnumerable<string> Aliases
```
BorisDog, line 51: I think there are more  places, like `GridFSUploadOptions`, `GridFSFileInfo`...

PR #1494, opened 2024-10-10
CSHARP-5319: Remove AsDateTime and similar from BsonValue

papafe:


src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -227,15 +218,6 @@ public int AsInt32
             get { return ((BsonInt32)this).Value; }
         }
 
-        /// <summary>
-        /// Casts the BsonValue to a DateTime in the local timezone (throws an InvalidCastException if the cast is not valid).
-        /// </summary>
-        [Obsolete("Use ToLocalTime instead.")]
-        public DateTime AsLocalTime
```
rstam, line 234: I think this was obsoleted by mistake and we should keep the property and remove the `[Obsolete]` attribute.

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -341,15 +314,6 @@ public string AsString
             get { return ((BsonString)this).Value; }
         }
 
-        /// <summary>
-        /// Casts the BsonValue to a DateTime in UTC (throws an InvalidCastException if the cast is not valid).
-        /// </summary>
-        [Obsolete("Use ToUniversalTime instead.")]
```
rstam, line 347: I think this was obsoleted by mistake and we should keep the property and remove the `[Obsolete]` attribute.

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -252,15 +234,6 @@ public bool? AsNullableBoolean
             get { return (BsonType == BsonType.Null) ? null : (bool?)AsBoolean; }
         }
 
-        /// <summary>
-        /// Casts the BsonValue to a Nullable{DateTime} (throws an InvalidCastException if the cast is not valid).
-        /// </summary>
-        [Obsolete("Use ToNullableUniversalTime instead.")]
-        public DateTime? AsNullableDateTime
```
rstam, line 259: This property should be replaced with two properties: `AsNullableLocalTime` and `AsNullableUniversalTime`.

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -178,15 +178,6 @@ public byte[] AsByteArray
             get { return ((BsonBinaryData)this).Bytes; }
         }
 
-        /// <summary>
-        /// Casts the BsonValue to a DateTime in UTC (throws an InvalidCastException if the cast is not valid).
-        /// </summary>
-        [Obsolete("Use ToUniversalTime instead.")]
```
rstam, line 184: This obsoletion was proper and the member can be deleted.

The error message was wrong though. It should have been "Use AsUniversalTime" instead.

rstam (changes requested):
There is a discussion going on about what exactly we should do.

I am proposing that we should keep `AsLocalTime` and `AsUniversalTime` because that it is the existing pattern established by all our other `AsXyz` properties.

rstam:
Given our time zone difference and that I'm out of the office tomorrow I have prototyped in code what we were discussing earlier today:

https://github.com/rstam/mongo-csharp-driver/commits/csharp5319/

Summary of proposed changes:

Remove `AsDateTime` (use `AsUniversalTime` instead)
Remove `AsNullableDateTime` (use `AsNullableUniversalTime` instead)

Keep `AsLocalTime` (remove `[Obsolete]` attribute)
Keep `AsUniversalTime` (remove `[Obsolete]` attribute)

Add `AsNullableLocalTime`
Add `AsNullableUniversalTime`

Leave any existing "to date" methods alone.

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -245,20 +228,20 @@ public long AsInt64
         }
 
         /// <summary>
-        /// Casts the BsonValue to a Nullable{Boolean} (throws an InvalidCastException if the cast is not valid).
+        /// Casts the BsonValue to a DateTime in the local timezone (throws an InvalidCastException if the cast is not valid).
         /// </summary>
-        public bool? AsNullableBoolean
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
rstam, line 233: I think we want `DebuggerBrowsable` here.

Also, we want it for ALL `AsXyz` properties, not just the DateTime ones.

rstam (changes requested):

src/MongoDB.Bson/ObjectModel/BsonValue.cs
papafe: Yes, you're right, `DebuggerBrowsable` is the correct one, I've misunderstood the situation

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -77,6 +82,7 @@ public BsonBinaryData AsBsonBinaryData
         /// <summary>
         /// Casts the BsonValue to a BsonDateTime (throws an InvalidCastException if the cast is not valid).
         /// </summary>
+        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
rstam, line 85: Missing attribute on line 94.

src/MongoDB.Bson/ObjectModel/BsonValue.cs
```diff
@@ -165,6 +180,7 @@ public BsonUndefined AsBsonUndefined
         /// <summary>
         /// Casts the BsonValue to a BsonValue (a way of upcasting subclasses of BsonValue to BsonValue at compile time).
         /// </summary>
+        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
```
rstam, line 183: Missing attribute on line 192.

rstam (changes requested):

src/MongoDB.Bson/ObjectModel/BsonValue.cs
papafe: 🤦 just added

rstam (approved):
LGTM

PR #1495, opened 2024-10-11
CSHARP-5326: Mark API as obsolete: Guids

papafe:


papafe:
@BorisDog it seems that most of the things mentioned in the deprecation mentioned have already been done, so this is a very small PR

src/MongoDB.Bson/ObjectModel/GuidRepresentationMode.cs
```diff
@@ -13,12 +13,15 @@
...
 
+using System;
+
 namespace MongoDB.Bson
 {
     /// <summary>
     /// Whether to handle GuidRepresentation using the v2.x mode of the v3.x mode.
     /// See the reference documentation for details.
     /// </summary>
+    [Obsolete("This enum will be made removed in a later release, with V3 being the only representation mode.")]
```
BorisDog, line 24: Please make sure that compilation succeeds before merging

BorisDog (approved):
LGTM

src/MongoDB.Bson/ObjectModel/GuidRepresentationMode.cs
papafe: Sure

PR #1506, opened 2024-10-18
CSHARP-5342: Make sure ClientBulkWrite supports retryability properly

sanych-sun:


src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
```diff
@@ -95,6 +95,8 @@ public BulkWriteReplaceOneModel(
         /// </summary>
         public TDocument Replacement { get; }
 
+        internal override bool IsRetryable() => true;
```
BorisDog, line 98: Consider
`internal abstract bool IsMulti { get; }`
instead.

So `retryablity` decision is not in the model

BorisDog (changes requested):

src/MongoDB.Driver/BulkWriteReplaceOneModel.cs
sanych-sun: Done.

BorisDog (approved):
LGTM.
Please check the failures before merging.

PR #1509, opened 2024-10-21
CSHARP-5348: Avoid allocations for Bson*Context

obligaron:
See [CSHARP-5348](https://jira.mongodb.org/browse/CSHARP-5348) for details.

This PR is split into three commits
- First, adds PopContext if it is missing from Bson*Context (no logic change)
- Second, adds a PushContext method (no logic change)
- The third adds the caching logic to Bson*Context.

If the first and second commits should be separated, I can reduce the PR to just the caching logic. For our environment, the `BsonBinaryReader`/`BsonBinaryWriter` are also the most important ones.

### DIFF
--- a/benchmarks/MongoDB.Driver.Benchmarks/README.md
+++ b/benchmarks/MongoDB.Driver.Benchmarks/README.md
@@ -13,7 +13,7 @@ This suite implements the benchmarks described in this [spec](https://github.com
    (e.g `dotnet run -c Release -- --driverBenchmarks --envVars MONGODB_URI:"ConnectionString"`)
 
 You can also select the benchmarks to run directly on the command for running the benchmarks as such
-`dotnet run -c Release -- --driverBenchmarks --fitler "*BenchmarkClassName*"`. The benchmarks are also grouped into categories namely: BSONBench, WriteBench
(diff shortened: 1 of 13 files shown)


papafe:
Hi @obligaron thanks a lot for your PR. I'll take a look, but in the meantime could you rebase on the latest main? We did some changes to the CI and most tests will fail if you don't.

obligaron:
I rebased the PR. Thanks @papafe for looking at the code. 🙂

papafe:
Hey @obligaron thanks a lot for this PR too. 
I understand what you are trying to go but we need to have an internal discussion about how do we want to tackle the issue, so we'll handle this according to our priorities

obligaron:
Thanks a lot for reviewing the PRs. ❤️ 

Regarding this one:
The Bson*Context was the top 1 or 2 allocation (depending on the scenario) when profiling our workloads.

We mainly use BsonBinary, so just improving those parts alone would help us a lot.
I could create a new PR with just the BsonBinary(Reader|Writer)Context changes. This would reduce the code changes and hopefully be easier to review.
If this would be a good direction, or if I can help in any other way, please feel free to contact me. 🙂

papafe:
@obligaron I think this PR makes sense overall, and I understand what you're trying to but we need to have an internal discussion about how we want to implement this, and I'll keep you updated if there are any news.

PR #1512, opened 2024-10-23
CSHARP-5349: Fix discriminator convention inheritance.

rstam:


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1324,10 +1324,31 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
+                while (classMap != null)
+                {
+                    if (classMap._discriminatorConvention != null)
+                    {
+                        return classMap._discriminatorConvention;
+                    }
+
+                    if (classMap._isRootClass)
+                    {
+                        return StandardDiscriminatorConvention.Hierarchical;
```
rstam, line 1344: I think this is still wrong... there just isn't a test to show that it is.

Consider this scenario: there might be a registered discriminator convention at this level and we're not looking it up.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1323,11 +1323,40 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
+                    if (BsonSerializer.IsDiscriminatorConventionRegisteredAtThisLevel(classMap._classType))
+                    {
+                        // in this case LookupDiscriminatorConvention below will find it
+                        break;
+                    }
+
+                    if (classMap._isRootClass)
+                    {
+                        // in this case auto-register a hierarchical convention for the root class and look it up as usual below
+                        BsonSerializer.GetOrRegisterDiscriminatorConvention(classMap._classType, StandardDiscriminatorConvention.Hierarchical);
```
rstam, line 1351: Calling `GetOrRegisterDiscriminatorConvention` instead of `RegsiterDiscriminatorConvention` is in order to avoid collisions between threads.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -269,6 +269,43 @@ public static object Deserialize(TextReader textReader, Type nominalType, Action
             }
         }
 
+        internal static IDiscriminatorConvention GetOrRegisterDiscriminatorConvention(Type type, IDiscriminatorConvention discriminatorConvention)
```
rstam, line 272: These don't need to be `public`.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
adelinowona: How does it avoid collisions?

rstam: The new `GetOrRegisterDiscriminatorConvention` method uses the same lock as `RegsiterDiscriminatorConvention` to handle thread safety.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -269,6 +269,43 @@ public static object Deserialize(TextReader textReader, Type nominalType, Action
...
+            {
+                if (__discriminatorConventions.TryGetValue(type, out var registeredConvention))
+                {
+                    return registeredConvention;
+                }
+            }
+            finally
+            {
+                __configLock.ExitReadLock();
+            }
```
adelinowona, line 285: Is this whole block really needed? Seems a bit redundant to check for a registered convention here and then check again later on below when trying to register.

rstam: It may seem redundant but it's not. All the other methods use the same pattern.

The idea is that after the initial configuration phase is over everything can be done with read lock only, eliminating lock contention between threads.

The second check below is because of multithreading. Between releasing the read lock and acquiring the write lock some other thread might have snuck in and registered the convention.

adelinowona (approved):
LGTM

papafe (approved):

src/MongoDB.Bson/Serialization/BsonSerializer.cs
sanych-sun, line 272: I would suggest to have slightly different method signature to mimic `ConcurrentDictionary.GetOrAdd` method signature:

```
internal static IDiscriminatorConvention GetOrAddDiscriminatorConvention(Type type, Func<Type, IDiscriminatorConvention> discriminatorConventionFactory)
```

sanych-sun (changes requested):

sanych-sun (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1515, opened 2024-10-28
CSHARP-5368: Eliminiate allocations in Bson.Decimal128 to decimal conversion

obligaron:
See [CSHARP-5368](https://jira.mongodb.org/browse/CSHARP-5368) for details.

This PR removes allocations that happens for

- [`Enum.CompareTo(object? enum)`](https://learn.microsoft.com/de-de/dotnet/api/system.enum.compareto?view=net-8.0), because this results in boxing of both enums.
The alternative is `Comparer<T>.Default.Compar` for .NET 6.0 and manual int-casts for .NET 4.x.
- [`decimal.GetBits(decimal d)`](https://learn.microsoft.com/en-us/dotnet/api/system.decimal.getbits?view=net-8.0#system-decimal-getbits(system-decimal)) allocates an `int[4]` array on every call.
The alternative is [`decimal.GetBits(decimal d, Span<int> destination)`](https://learn.microsoft.com/en-us/dotnet/api/system.decimal.getbits?view=net-8.0#system-decimal-getbits(system-decimal-system-span((system-int32)))), which allows you to pass a [stack allocated](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc) int-array. This method is not available in .NET 4.x.

### DIFF
--- a/src/MongoDB.Bson/ObjectModel/Decimal128.cs
+++ b/src/MongoDB.Bson/ObjectModel/Decimal128.cs
@@ -1909,7 +1909,7 @@ namespace MongoDB.Bson
             {
                 var xType = GetDecimal128Type(x);
                 var yType = GetDecimal128Type(y);
-// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
+				// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
 #if NET6_0_OR_GREATER
                 var result = Comparer<Decimal128Type>.Default.Compare(xType, yType);
 #else
--- a/src/MongoDB.Bson/ObjectModel/Decimal128.cs
+++ b/src/MongoDB.Bson/ObjectModel/Decimal128.cs
@@ -1465,7 +1465,12 @@ namespace MongoDB.Bson
         /// <param name="value">The value.</param>
         public Decimal128(decimal value)
         {
+#if NET6_0_OR_GREATER
+            Span<int> bits = stackalloc int[4];
+            _ = decimal.GetBits(value, bits);
+#else
             var bits = decimal.GetBits(value);
+#endif
             var isNegative = (bits[3] & 0x80000000) != 0;
             var scale = (short)((bits[3] & 0x00FF0000) >> 16);
             var exponent = (short)-scale;
@@ -1904,7 +1909,11 @@ namespace MongoDB.Bson
             {
                 var xType = GetDecimal128Type(x);
                 var yType = GetDecimal128Type(y);
-                var result = xType.CompareTo(yType);
+#if NET6_0_OR_GREATER
+                var result = Comparer<Decimal128Type>.Default.Compare(xType, yType);
+#else
+                var result = ((int)xType).CompareTo((int)yType);
+#endif
                 if (result == 0 && xType == Decimal128Type.Number)
                 {
                     return CompareNumbers(x, y);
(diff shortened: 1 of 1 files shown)


papafe:
Hi @obligaron thanks a lot for your PR. I'll take a look, but in the meantime could you rebase on the latest main? We did some changes to the CI and most tests will fail if you don't.

src/MongoDB.Bson/ObjectModel/Decimal128.cs
```diff
@@ -1904,7 +1909,11 @@ public int Compare(Decimal128 x, Decimal128 y)
             {
                 var xType = GetDecimal128Type(x);
                 var yType = GetDecimal128Type(y);
-                var result = xType.CompareTo(yType);
+#if NET6_0_OR_GREATER
+                var result = Comparer<Decimal128Type>.Default.Compare(xType, yType);
```
papafe, line 1913: It seems that `Comparer.Default` is available since .NET Framework 2.0. Is there a specific reason why we're making a cutoff point at .NET 6.0 here?

obligaron: .NET Framework lacks some optimisations for enums (e.g. https://github.com/dotnet/coreclr/pull/5503 or https://github.com/dotnet/coreclr/pull/14125).
This results in boxing and lookup overhead for the default comparer on .NET Framework.
That's why I specially cased it to avoid this overhead.

Alternatives could be
- Change the .NET Framework implementation to `Comparer.Default` and accept the boxing/overhead (this is slightly slower than the current implementation in main).
- Change the .NET 6 implementation to the int-cast logic of the .NET Framework (this is almost identical, but less readable).

I'm open to all three (keep it or change to one of the alternatives). 🙂

src/MongoDB.Bson/ObjectModel/Decimal128.cs
```diff
@@ -1928,7 +1937,11 @@ private int CompareNumbers(Decimal128 x, Decimal128 y)
             {
                 var xClass = GetNumberClass(x);
                 var yClass = GetNumberClass(y);
-                var result = xClass.CompareTo(yClass);
+#if NET6_0_OR_GREATER
```
papafe, line 1940:
```suggestion
// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
#if NET6_0_OR_GREATER
```

src/MongoDB.Bson/ObjectModel/Decimal128.cs
```diff
@@ -1904,7 +1909,12 @@ public int Compare(Decimal128 x, Decimal128 y)
             {
                 var xType = GetDecimal128Type(x);
                 var yType = GetDecimal128Type(y);
-                var result = xType.CompareTo(yType);
+// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
```
papafe, line 1912:
```suggestion
				// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
```

src/MongoDB.Bson/ObjectModel/Decimal128.cs
```diff
@@ -1928,7 +1938,12 @@ private int CompareNumbers(Decimal128 x, Decimal128 y)
             {
                 var xClass = GetNumberClass(x);
                 var yClass = GetNumberClass(y);
-                var result = xClass.CompareTo(yClass);
+// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
```
papafe, line 1941:
```suggestion
				// .NET Framework lacks some optimizations for enums that would result in boxing and lookup overhead for the default comparer
```

papafe (approved):
LGTM! 

Thanks a lot for this PR @obligaron

PR #1517, opened 2024-10-28
CSHARP-5355: Add Unsubscribe method for ClusterBuilder

adelinowona:


adelinowona:
After further review this ticket will need more work, potentially revamping our whole events API so there's no value from this PR as of now.

PR #1519, opened 2024-10-29
CSHARP-5360: Fix resource leaks in LibMongoCryptControllerBase::SendKmsRequest

adelinowona:


src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -283,8 +283,9 @@ private void SendKmsRequest(KmsRequest request, CancellationToken cancellation)
             var tlsStreamSettings = GetTlsStreamSettings(request.KmsProvider);
             var sslStreamFactory = new SslStreamFactory(tlsStreamSettings, _networkStreamFactory);
             using (var sslStream = sslStreamFactory.CreateStream(endpoint, cancellation))
+            using (var binary = request.Message)
             {
-                var requestBytes = request.Message.ToArray();
+                var requestBytes = binary.ToArray();
```
BorisDog, line 288: I didn't realize that `request.Message` actually creates a new `Binary` each time, 
Should we change it to something like `GetMessage()` so the ownership is more clear?

adelinowona: Yep done!

BorisDog (approved):
LGTM + small comment

PR #1525, opened 2024-10-31
CSHARP-5098: Server 8.0+ requires tests for aggregation with $out to …

sanych-sun:
…not pre-create output collection


papafe (approved):
LGTM

PR #1526, opened 2024-10-31
CSHARP-5046: Test that durations are included on relevant pool events

sanych-sun:


src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -180,17 +180,15 @@ public AcquireConnectionHelper(ExclusiveConnectionPool pool)
 
             public IConnectionHandle AcquireConnection(CancellationToken cancellationToken)
             {
-                Stopwatch stopwatch = new();
+                var stopwatch = Stopwatch.StartNew();
```
sanych-sun, line 183: Have to start the stopwatch before all checks in StartCheckingOut happened. Otherwise ConnectionPoolCheckingOutConnectionFailedEvent could have zero duration for closed pool tests.

tests/MongoDB.Driver.TestHelpers/Core/MockConnection.cs
```diff
@@ -147,7 +147,7 @@ public void Dispose()
         {
             _closingEventHandler?.Invoke(new ConnectionClosingEvent(_connectionId, EventContext.OperationId));
             IsExpired = true;
-            _closedEventHandler?.Invoke(new ConnectionClosedEvent(_connectionId, TimeSpan.Zero, EventContext.OperationId));
+            _closedEventHandler?.Invoke(new ConnectionClosedEvent(_connectionId, TimeSpan.FromMilliseconds(42), EventContext.OperationId));
```
sanych-sun, line 150: There are new tests that ensure Duration has value, so cannot use TimeSpan.Zero anymore. So decide to use 42, because why not =)

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: Good catch, looks like the previous implementation was actually a bug.

I think it would be more accurate to start the timer after emitting `ConnectionPoolCheckingOutConnectionEvent`.
so `StartCheckingOut` can return or receive the `stopwatch`.

tests/MongoDB.Driver.TestHelpers/Core/MockConnection.cs
BorisDog: So the duration has a value here, it's just 0. Do we need to check that for mockConnection for Closed event?

tests/MongoDB.Driver.Tests/Specifications/connection-monitoring-and-pooling/ConnectionMonitoringAndPoolingTestRunner.cs
```diff
@@ -259,6 +259,35 @@ private void AssertEvent(object actualEvent, BsonDocument expectedEvent)
...
+                    ConnectionClosedEvent connectionClosedEvent => connectionClosedEvent.Duration,
+                    _ => throw new ArgumentOutOfRangeException()
+                };
+
+                if (expectedDuration == 42)
+                {
+                    if (actualDuration == TimeSpan.Zero)
+                    {
+                        int a = 10;
+                        a++;
```
BorisDog, line 280: `a += 42` ? :)

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Specifications/connection-monitoring-and-pooling/ConnectionMonitoringAndPoolingTestRunner.cs
sanych-sun: Ups, some debug code that does not intent to be pushed =)

tests/MongoDB.Driver.TestHelpers/Core/MockConnection.cs
sanych-sun: The test assert if the value is not `default`, when we use TimeSpan here which has TimeSpan.Zero (or simply 0) as default value.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: I've decided to pass `stopwatch` into the `StartCheckingOut` method, to reduce the affected areas.

tests/MongoDB.Driver.TestHelpers/Core/MockConnection.cs
BorisDog: So maybe let's use a different value here? like 1ms (also make it consistent with code below that uses `FromTicks`)?
It might be confusing to a reader to have two different meanings for `42` value.

BorisDog (changes requested):
Small change in tests requested

tests/MongoDB.Driver.TestHelpers/Core/MockConnection.cs
sanych-sun: Sure. Will do =)

sanych-sun: Done

BorisDog (approved):
LGTM

PR #1527, opened 2024-11-01
CSHARP-5200: Allow valid SRV hostnames with less than 3 parts

papafe:


tests/MongoDB.Driver.Tests/Specifications/initial-dns-seedlist-discovery/prose-tests/SrvPartsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        [Theory]
+        [InlineData("mongodb+srv://localhost", "localhost")]
+        [InlineData("mongodb+srv://mongo.local", "mongo.local")]
+        public void Throw_when_return_address_is_identical_to_srv_hostname(string uri, string resolvedHost)
+        {
+            var connectionString = new ConnectionString(uri, true, new MockDnsResolver(resolvedHost));
+            var exception = Record.Exception(() => connectionString.Resolve());
+
+            exception.Should().BeOfType<MongoConfigurationException>().
+                Which.Message.Should().Be("Hosts in the SRV record must have the same parent domain as the seed host.");
```
papafe, line 62: How much specific should we be with the error message? If we wanted to be very precise it could become:
"Hosts in the SRV record must have the same parent domain as the seed host. If the seed host has less than three separated dot separated parts, then the returned hostname must have at least one more domain level than the seed".
Would it be too much?

tests/MongoDB.Driver.Tests/Specifications/initial-dns-seedlist-discovery/prose-tests/SrvPartsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using FluentAssertions;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.Misc;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Specifications.initial_dns_seedlist_discovery.prose_tests
+{
+    public class SrvPartsTests
+    {
+        [Theory]
```
sanych-sun, line 30: Could you please add permalink to the prose test description in comment before the test so we can easily find what this test about?

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -1358,33 +1358,32 @@ private void ValidateResolvedHosts(string original, List<string> resolved)
 
         internal static bool HasValidParentDomain(string original, DnsEndPoint resolvedEndPoint)
         {
-            // Helper functions...
-            Func<string, string[]> getParentParts = x => x.Split('.').Skip(1).ToArray();
+            var host = resolvedEndPoint.Host;
+
+            var hostParts = host.Split('.').ToArray();
```
sanych-sun, line 1363: Can we avoid array allocation here? By counting number of '.' chars for length and checking if resolved host ends with original where we are comparing parts now.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/Specifications/initial-dns-seedlist-discovery/prose-tests/SrvPartsTests.cs
papafe: Just added them

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: Makes sense! I've rewritten the method, hope it's readable enough

tests/MongoDB.Driver.Tests/Specifications/initial-dns-seedlist-discovery/prose-tests/SrvPartsTests.cs
papafe: @sanych-sun What do you think?

sanych-sun: I'm OK with the current message.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -1358,33 +1358,35 @@ private void ValidateResolvedHosts(string original, List<string> resolved)
...
+            if (hostDotCount < originalDotCount)
+            {
+                return false;
+            }
 
-                // loop from back to front making sure that all of b is at the back of a, in order.
-                for (int ai = a.Length - 1, bi = b.Length - 1; bi >= 0; ai--, bi--)
+            // We check that all the parts are equal, going from the end and arriving at the
+            // first dot of original (where the subdomain is, if we continued)
+            for (int hostIndex = host.Length - 1, originalIndex = original.Length - 1;
```
sanych-sun, line 1380: Can this loop also be replaced with EndsWith?

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: You're right. I've rewritten the method to use `EndsWith` and use spans as much as possible to avoid string allocations

PR #1531, opened 2024-11-04
CSHARP-5396: Add support for arbitrary join inner expression arguments in LINQ

kerem-acer:
This PR add support for arbitrary join inner expression arguments in LINQ.

# Problem
The MongoDB LINQ translator currently throws an exception when attempting to join a foreign collection queryable that includes additional query operations (like `.Where()`).

## Example of the Issue
When trying to join two collections with filtering conditions, the following code throws an exception:

```csharp
// Collection setup and mapping code...

var pagesWithContainers = await (from page in pages.AsQueryable().Where(x => x.TenantId == 1)
    join container in pageContainers.AsQueryable().Where(x => x.TenantId == 1) 
    on page.Id equals container.PageId into containers
    select new
    {
        page.Id,
        page.Title,
        Containers = containers
    }).ToListAsync();
```

The code only works if you remove the `.Where()` clause from `pageContainers`. This PR adds support for such filtering operations on joined collections.

## Current Error
Without this PR, the above code throws the following exception:
```
MongoDB.Driver.Linq.ExpressionNotSupportedException: Expression not supported: [...] 
because inner expression must be a MongoDB IQueryable against a collection.
```

# Todos
- [ ] Write comprehensive tests

### DIFF
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/GroupJoinMethodToPipelineTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/GroupJoinMethodToPipelineTranslator.cs
@@ -13,6 +13,7 @@
 * limitations under the License.
 */
 
+using System;
 using System.Linq.Expressions;
 using MongoDB.Bson.Serialization;
 using MongoDB.Driver.Linq.Linq3Implementation.Ast;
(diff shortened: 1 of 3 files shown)


kerem-acer:
Even though now it doesn't throw exceptions, Where()'s of foreign collection queryable are ignored. Probably more work is needed to get this job done. 

As I understand, in JoinMethodToPipelineTranslator.Translate method line 66, innerExpression(foreign queryable expression) is used only for collectionName. So JoinMethodToPipelineTranslator.Translate does not respect the content of innerExpression.

kerem-acer:
Okay. Now, this this is a new feature rather than a bug. I will investigate and try to implement this feature this week. I will update you here.

Also, should I update the title and description of this PR and Jira issue? The scope of this PR has changed.

kerem-acer:
I implemented this feature and changed content of this PR. 

fyi @rstam

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/GroupJoinMethodToPipelineTranslator.cs
```diff
@@ -59,19 +60,36 @@ public static AstPipeline Translate(TranslationContext context, MethodCallExpres
...
+                {
+                    lookupStage = AstStage.Lookup(
+                        from: queryProvider.CollectionNamespace.CollectionName,
+                        localField,
+                        foreignField,
+                        @as: "_inner");
+                }
+                else
+                {
+                    var lookupPipeline = ExpressionToPipelineTranslator.Translate(context, innerExpression);
```
kerem-acer, line 83: Should we create a new TranslationContext for innerExpression?

616b2f:
Is there any way we could push this changes throw? We have the same issue not being able to join two queries with ".Where()" expressions in each.

PR #1534, opened 2024-11-05
CSHARP-5345: Unable to deserialize C# DateOnly objects

papafe:
[JIRA ticket](https://jira.mongodb.org/browse/CSHARP-5345)

### DIFF
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
@@ -1,0 +1,81 @@
+﻿/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using MongoDB.Bson.Serialization.Options;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Bson.Serialization.Attributes
+{
+#if NET6_0_OR_GREATER
+    /// <summary>
+    /// Specifies the external representation and related options for this field or property.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
+    public class BsonDateOnlyOptionsAttribute : BsonSerializationOptionsAttribute
+    {
+        // private fields
+        private BsonType _representation;
+        private DateOnlyDocumentFormat _documentFormat;
+
+        // constructors
+        /// <summary>
+        /// Initializes a new instance of the BsonDateOnlyOptionsAttribute class.
+        /// </summary>
+        /// <param name="representation">The external representation.</param>
+        public BsonDateOnlyOptionsAttribute(BsonType representation)
+        {
+            _representation = representation;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the BsonDateOnlyOptionsAttribute class.
+        /// </summary>
+        /// <param name="representation">The external representation.</param>
+        /// <param name="documentDocumentFormat">The format to use with document representation.</param>
+        public BsonDateOnlyOptionsAttribute(BsonType representation, DateOnlyDocumentFormat documentDocumentFormat)
+        {
+            _representation = representation;
+            _documentFormat = documentDocumentFormat;
+        }
+
+        // public properties
+        /// <summary>
+        /// Gets the external representation.
+        /// </summary>
+        public BsonType Representation => _representation;
+
+        /// <summary>
+        /// Gets or sets the TimeOnlyUnits.
+        /// </summary>
+        public DateOnlyDocumentFormat DocumentDocumentFormat => _documentFormat;
+
+        /// <summary>
+        /// Reconfigures the specified serializer by applying this attribute to it.
+        /// </summary>
+        /// <param name="serializer">The serializer.</param>
+        /// <returns>A reconfigured serializer.</returns>
+        protected override IBsonSerializer Apply(IBsonSerializer serializer)
+        {
+            if (serializer is DateOnlySerializer dateOnlySerializer)
+            {
+                return dateOnlySerializer.WithRepresentation(_representation, _documentFormat);
+            }
+
+            return base.Apply(serializer);
+        }
+    }
+#endif
+}
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
@@ -61,7 +61,7 @@ namespace MongoDB.Bson.Serialization.Attributes
         /// <returns>A reconfigured serializer.</returns>
         protected override IBsonSerializer Apply(IBsonSerializer serializer)
         {
-            var reconfiguredSerializer = serializer.GetReconfigured<DateOnlySerializer>(s => s.WithRepresentation(_representation, _documentFormat));
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer<DateOnlySerializer>(serializer, s => s.WithRepresentation(_representation, _documentFormat));
             return reconfiguredSerializer ?? base.Apply(serializer);
         }
     }
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
@@ -61,7 +61,7 @@ namespace MongoDB.Bson.Serialization.Attributes
         /// <returns>A reconfigured serializer.</returns>
         protected override IBsonSerializer Apply(IBsonSerializer serializer)
         {
-            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer<DateOnlySerializer>(serializer, s => s.WithRepresentation(_representation, _documentFormat));
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer(serializer, (DateOnlySerializer s) => s.WithRepresentation(_representation, _documentFormat));
             return reconfiguredSerializer ?? base.Apply(serializer);
         }
     }
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
@@ -31,22 +31,13 @@ namespace MongoDB.Bson.Serialization.Attributes
         private DateOnlyDocumentFormat _documentFormat;
 
         // constructors
-        /// <summary>
-        /// Initializes a new instance of the BsonDateOnlyOptionsAttribute class.
-        /// </summary>
-        /// <param name="representation">The external representation.</param>
-        public BsonDateOnlyOptionsAttribute(BsonType representation)
-            : this(representation, DateOnlyDocumentFormat.Classic)
-        {
-
-        }
 
         /// <summary>
         /// Initializes a new instance of the BsonDateOnlyOptionsAttribute class.
         /// </summary>
         /// <param name="representation">The external representation.</param>
         /// <param name="documentDocumentFormat">The format to use with document representation.</param>
-        public BsonDateOnlyOptionsAttribute(BsonType representation, DateOnlyDocumentFormat documentDocumentFormat)
+        public BsonDateOnlyOptionsAttribute(BsonType representation, DateOnlyDocumentFormat documentDocumentFormat = DateOnlyDocumentFormat.DateTimeTicks)
         {
             _representation = representation;
             _documentFormat = documentDocumentFormat;
@@ -70,12 +61,8 @@ namespace MongoDB.Bson.Serialization.Attributes
         /// <returns>A reconfigured serializer.</returns>
         protected override IBsonSerializer Apply(IBsonSerializer serializer)
         {
-            if (serializer is DateOnlySerializer dateOnlySerializer)
-            {
-                return dateOnlySerializer.WithRepresentation(_representation, _documentFormat);
-            }
-
-            return base.Apply(serializer);
+            var reconfiguredSerializer = serializer.GetReconfigured<DateOnlySerializer>(s => s.WithRepresentation(_representation, _documentFormat));
+            return reconfiguredSerializer ?? base.Apply(serializer);
         }
     }
 #endif
--- a/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
+++ b/src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
@@ -21,7 +21,7 @@ namespace MongoDB.Bson.Serialization.Attributes
 {
 #if NET6_0_OR_GREATER
     /// <summary>
-    /// Specifies the external representation and related options for this field or property.
+    /// Specifies the external representation and related options for a DateOnly field or property.
     /// </summary>
     [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
     public class BsonDateOnlyOptionsAttribute : BsonSerializationOptionsAttribute
@@ -36,8 +36,9 @@ namespace MongoDB.Bson.Serialization.Attributes
         /// </summary>
         /// <param name="representation">The external representation.</param>
         public BsonDateOnlyOptionsAttribute(BsonType representation)
+            : this(representation, DateOnlyDocumentFormat.Classic)
         {
-            _representation = representation;
+
         }
 
         /// <summary>
@@ -58,9 +59,9 @@ namespace MongoDB.Bson.Serialization.Attributes
         public BsonType Representation => _representation;
 
         /// <summary>
-        /// Gets or sets the TimeOnlyUnits.
+        /// Gets the document format.
         /// </summary>
-        public DateOnlyDocumentFormat DocumentDocumentFormat => _documentFormat;
+        public DateOnlyDocumentFormat DocumentFormat => _documentFormat;
 
         /// <summary>
         /// Reconfigures the specified serializer by applying this attribute to it.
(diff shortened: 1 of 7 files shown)


src/MongoDB.Bson/Serialization/Serializers/SerializerHelper.cs
```diff
@@ -25,7 +25,6 @@ namespace MongoDB.Bson.Serialization.Serializers
     public class SerializerHelper
     {
         // private fields
-        private readonly long _allMemberFlags;
```
papafe, line 28: This was assigned but not used

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -102,17 +140,37 @@ public override DateOnly Deserialize(BsonDeserializationContext context, BsonDes
                     break;
 
                 case BsonType.Document:
-                    value = default;
-                    _helper.DeserializeMembers(context, (_, flag) =>
+                    if (_documentFormat is DateOnlyDocumentFormat.Classic)
```
rstam, line 143: Do we want to be more forgiving and deserialize whatever we find instead of requiring that the stored documents match the configured format?

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -73,19 +92,38 @@ public DateOnlySerializer(BsonType representation)
...
             _representation = representation;
+            _documentFormat = documentFormat;
             _converter = new RepresentationConverter(false, false);
 
-            _helper = new SerializerHelper
-            (
-                new SerializerHelper.Member("DateTime", Flags.DateTime),
-                new SerializerHelper.Member("Ticks", Flags.Ticks)
-            );
+            if (_documentFormat is DateOnlyDocumentFormat.Classic)
```
rstam, line 98: It really throws me off to see `is` instead of `==`

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -200,6 +267,24 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
...
+        /// The reconfigured serializer.
+        /// </returns>
+        public DateOnlySerializer WithRepresentation(BsonType representation, DateOnlyDocumentFormat documentFormat)
+        {
+            if (representation == _representation && documentFormat == _documentFormat)
+            {
+                return this;
+            }
+
+            return new DateOnlySerializer(representation, documentFormat);
```
rstam, line 285: Probably should include `_documentFormat` in line 291 below.

rstam (changes requested):

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
papafe: I started doing that mostly because I think it is convenient to use this form when one must check if the enum is one of multiple values:
`v is Enum.A or Enum.B` vs `v == Enum.A || v == Enum.B`
And from there I started using also when there is only value, for consistency, even though there is no particular advantage here. 
Of course I can put `==` if we prefer.

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
papafe: Definitely

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
rstam: Developers should ensure that their documents are all serialized in the same way in the database.

However, we almost always are willing to deserialize whatever we find. I don't see why deserializing whatever document format we find is any different than deserializing `Double` vs `DateTime` vs `Document` etc...

Maybe we need other people to chime in also.

rstam (changes requested):

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
papafe: I corrected this part. The correction is kinda convoluted, but I'm not sure if we can make it easier on the eyes

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -26,31 +27,39 @@ namespace MongoDB.Bson.Serialization.Serializers
...
             public const long DateTime = 1;
             public const long Ticks = 2;
         }
 
+        private static class HumanReadableFormatFlags
+        {
+            public const long Year = 4;
+            public const long Month = 8;
+            public const long Day = 16;
+        }
```
rstam, line 49: No need for two sets of flags. We can do this:

```
private static class Flags
{
    public const long DateTime = 1;
    public const long Ticks = 2;
    public const long Year = 4;
    public const long Month = 8;
    public const long Day = 16;
                               
    public const long DateTimeTicks = DateTime | Ticks;
    public const long YearMonthDay = Year | Month | Day;
}                                                         
```

See below for how the last two are used.

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -102,17 +129,52 @@ public override DateOnly Deserialize(BsonDeserializationContext context, BsonDes
...
+
+                    if ((humanReadableFormatFound && (tickFound || dateTimeFound))
+                        || (classicFormatFound && (yearFound || monthFound || dayFound)))
+                    {
+                        throw new FormatException("Invalid document format.");
+                    }
+
+                    value = classicFormatFound ? VerifyAndMakeDateOnly(new DateTime(tickValue, DateTimeKind.Utc))
+                        : new DateOnly(year, month, day);
+
```
rstam, line 177: This whole `case BsonType.Document` can be improved.

I suggest the following refactoring for this whole method:

```
public override DateOnly Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
{
                                    
    var bsonType = bsonReader.GetCurrentBsonType();
    switch (bsonType)
    {
        case BsonType.DateTime:
            return VerifyAndMakeDateOnly(BsonUtils.ToDateTimeFromMillisecondsSinceEpoch(bsonReader.ReadDateTime()));
                                                                                                                    
        case BsonType.Document:                                                                                     
            var ticks = 0L;                                                                                         
            var year = 0;                                                                                           
            var month = 0;                                                                                          
            var day = 0;                                                                                            
                        
            var foundMemberFlags = _helper.DeserializeMembers(context, (_, flag) =>
            {
                switch (flag)
                {
                    case Flags.DateTime: bsonReader.SkipValue();  break; // ignore value (use Ticks instead)
                    case Flags.Ticks: ticks = Int64Serializer.Instance.Deserialize(context); break;
                    case Flags.Year: year = Int32Serializer.Instance.Deserialize(context); break;
                    case Flags.Month: month = Int32Serializer.Instance.Deserialize(context); break;
                    case Flags.Day: day = Int32Serializer.Instance.Deserialize(context); break;
                }
            });
               
            if (foundMemberFlags == Flags.DateTimeTicks)
            {
                return VerifyAndMakeDateOnly(new DateTime(ticks, DateTimeKind.Utc));
            }
            else if (foundMemberFlags == Flags.YearMonthDay)
            {
                return new DateOnly(year, month, day);
            }
            else
            {
                throw new FormatException("Invalid document format.");
            }
             
        case BsonType.Decimal128:
            return VerifyAndMakeDateOnly(new DateTime(_converter.ToInt64(bsonReader.ReadDecimal128()), DateTimeKind.Utc));
                                                                                                                          
        case BsonType.Double:                                                                                             
            return VerifyAndMakeDateOnly(new DateTime(_converter.ToInt64(bsonReader.ReadDouble()), DateTimeKind.Utc));    
                                                                                                                          
        case BsonType.Int32:                                                                                              
            return VerifyAndMakeDateOnly(new DateTime(bsonReader.ReadInt32(), DateTimeKind.Utc));                         
                                                                                                                          
        case BsonType.Int64:                                                                                              
            return VerifyAndMakeDateOnly(new DateTime(bsonReader.ReadInt64(), DateTimeKind.Utc));                         
                                                                                                                          
        case BsonType.String:                                                                                             
            return DateOnly.ParseExact(bsonReader.ReadString(), "yyyy-MM-dd");                                            
                                                                                                                          
        default:                                                                                                          
            throw CreateCannotDeserializeFromBsonTypeException(bsonType);                                                 
    }                                                                                                                     
                                                                                                                          
    DateOnly VerifyAndMakeDateOnly(DateTime dt)                                                                           
    {                                                                                                                     
        if (dt.TimeOfDay != TimeSpan.Zero)                                                                                
        {                                                                                                                 
            throw new FormatException("Deserialized value has a non-zero time component.");                               
        }                                                                                                                 
                                                                                                                          
        return DateOnly.FromDateTime(dt);                                                                                 
    }                                                                                                                     
}
```

rstam: Don't forget to modify `Equals` below to also compare the `_documentFormam`

```_documentFormat.Equals(other._documentFormat)```

**Review by rstam — CHANGES_REQUESTED** [review]

File: src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs (hunk 6 above)
Comment (papafe) [reply src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs:177]: Sure! This looks very nice

File: src/MongoDB.Bson/Serialization/Options/DateOnlyDocumentFormat.cs [hunk 7]
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The document will contain "DateTime" (BsonType.DateTime) and "Ticks" (BsonType.Int64).
+        /// </summary>
+        Classic,
+
+        /// <summary>
+        /// The document will contain "Year", "Month" and "Day" (all BsonType.Int32).
+        /// </summary>
+        HumanReadable
```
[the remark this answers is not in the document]
papafe: I think it makes sense. A small thing I don't like is that `DateTimeTicks` at first glance looks like 3 elements, instead of two, but it's a small thing 😁

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -60,6 +60,37 @@ serializer is IHasDiscriminatorConvention hasDiscriminatorConvention
                 ? hasDiscriminatorConvention.DiscriminatorConvention
                 : BsonSerializer.LookupDiscriminatorConvention(serializer.ValueType);
 
+        /// <summary>
```
papafe, line 63: @rstam do you think this would be an appropriate place for this method?

rstam (changes requested):

rstam:
Please rebase on main and force push to fix the failing auth tests in the patch build.

src/MongoDB.Bson/Serialization/Serializers/DateOnlySerializer.cs
```diff
@@ -129,53 +128,29 @@ public override DateOnly Deserialize(BsonDeserializationContext context, BsonDes
...
 
                 case BsonType.Document:
-                    var tickFound = false;
-                    var dateTimeFound = false;
-                    var yearFound = false;
-                    var monthFound = false;
-                    var dayFound = false;
-
-                    var tickValue = 0L;
+                    var ticks = 0L;
```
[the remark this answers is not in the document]
papafe: Makes sense! My bad, I didn't notice the returns in your suggestion before 🤦

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -60,6 +60,37 @@ serializer is IHasDiscriminatorConvention hasDiscriminatorConvention
...
+        /// <param name="serializer">The input serializer to be reconfigured.</param>
+        /// <param name="reconfigure">A function that defines how the serializer of type <typeparamref name="T"/> should be reconfigured.</param>
+        /// <param name="shouldReconfigure">
+        /// An optional predicate to determine if the reconfiguration should be applied to the current serializer.
+        /// </param>
+        /// <typeparam name="T">The specific type of serializer to be reconfigured.</typeparam>
+        /// <returns>
+        /// The reconfigured serializer, or <c>null</c> if no leaf serializer could be reconfigured.
+        /// </returns>
+        internal static IBsonSerializer GetReconfigured<T>(this IBsonSerializer serializer, Func<T, IBsonSerializer> reconfigure, Func<IBsonSerializer, bool> shouldReconfigure = null )
```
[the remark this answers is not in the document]
papafe: Regarding both points, I was actually having in mind the PR for configuring the `EnumRepresentationConvention` to work with collections of enums that we parked until the new major version, in particular [this method](https://github.com/papafe/mongo-csharp-driver/blob/cba6a12ed7949156f370f452bd5468d28f6bba51/src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs#L67).

For that case, because `EnumSerializer` is a generic type we can't simply use `serializer is TSerializer` and we need to do something else. Using the `GetReconfigured` method in that case would look something like this:

```csharp
var reconfigured = serializer.GetReconfigured<IRepresentationConfigurable>(
    configurable => configurable.WithRepresentation(_representation),
    bsonSerializer => serializer.ValueType.IsEnum);
```

This is one case where the `shouldReconfigure` method is used, because we cannot simply check the serializer type as we do it in this PR. 
Also, this shows why I didn't use `TSerializer` in the first place, because I thought it could cause some confusion given that it is not necessarily a serializer. Maybe it would be worth to call it `TSerializer` in any case to make it clearer?
I've also changed a little bit the xml docs because they were deceiving.

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
rstam: Suggested new method:

```
namespace MongoDB.Bson.Serialization
{
    internal static class SerializerConfigurator
    {
        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure)
        {                                                                                                                                             
            switch (serializer)                                                                                                                       
            {                                                                                                                                         
                case IChildSerializerConfigurable childSerializerConfigurable:                                                                        
                    var childSerializer = childSerializerConfigurable.ChildSerializer;                                                                
                    var reconfiguredChildSerializer = ReconfigureSerializer(childSerializer, reconfigure);                                            
                    return reconfiguredChildSerializer != null? childSerializerConfigurable.WithChildSerializer(reconfiguredChildSerializer) : null;  
                                                                                                                                                      
                case TSerializer typedSerializer:                                                                                                     
                    return reconfigure(typedSerializer);                                                                                              
                                                                                                                                                      
                default:                                                                                                                              
                    return null;                                                                                                                      
            }                                                                                                                                         
        }                                                                                                                                             
    }                                                                                                                                                 
}                                                                                                                                                     
```

Note the name of the method and the removal of the unused `shouldReconfigure` parameter.

While you may have a future use in mind for `shouldReconfigure` as a general rule don't add anything until you actually need it. You might change your mind about it between now and then.

rstam (changes requested):

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
papafe: Makes sense! I've added this and corrected the call.

src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+        public DateOnlyDocumentFormat DocumentFormat => _documentFormat;
+
+        /// <summary>
+        /// Reconfigures the specified serializer by applying this attribute to it.
+        /// </summary>
+        /// <param name="serializer">The serializer.</param>
+        /// <returns>A reconfigured serializer.</returns>
+        protected override IBsonSerializer Apply(IBsonSerializer serializer)
+        {
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer<DateOnlySerializer>(serializer, s => s.WithRepresentation(_representation, _documentFormat));
```
rstam, line 64: I kind of liked my suggestion in an earlier review to put the `DateOnlySerializer` on the lambda parameter and let the compiler infer the generic method type from that:

```
var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer(serializer, (DateOnlySerializer s) => s.WithRepresentation(_representation, _documentFormat));
```

Don't know if you just didn't notice that or didn't like it.

If you didn't like it we can go with what you have now.

rstam (changes requested):

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+using System;
+
+namespace MongoDB.Bson.Serialization
+{
+    internal static class SerializerConfigurator
+    {
+        /// <summary>
```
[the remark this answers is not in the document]
papafe: They don't need it, but I think it's not bad to have it. Maybe it helps at understanding what this method does without trying to read too much the implementation.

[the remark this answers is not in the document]
rstam: Sure you can leave these comments, but we're not going to start writing doc comments for internal classes in general. In general code should be written with proper naming that makes code self-describing.

Occasionally some small comments here and there may be helpful, but full-blown doc comments are not necessary.

rstam (approved):
LGTM

sguryev:
Hey guys.
I have come here looking for the fix for the `An error occurred while deserializing the DateOfBirth property of class CustomerApp.Infrastructure.Entities.Individual: Invalid element: 'year'. Invalid element: 'year'.` error.

How it was before:
image

How it is now:
image

My question is: will we be able to preserve backward compatibility and set the old format globally instead of adding attribute over every single `DateOnly` property or even do nothing and preserve old format?

Is it the correct final way to preserve the old behavior:
https://jira.mongodb.org/browse/CSHARP-5345?focusedId=6802536&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6802536
```
var dateOnlyClassMap = new BsonClassMap<DateOnly>(cm => cm.AutoMap()).Freeze();
var dateOnlyClassMapSerializer = new BsonClassMapSerializer<DateOnly>(dateOnlyClassMap);
BsonSerializer.RegisterSerializer<DateOnly>(dateOnlyClassMapSerializer); 
```

The solution above helped me to fix the issue I think
image

papafe:
@sguryev Yes, what you posted is a good "temporary" fix for the problem. When the content of this PR will be included with the next version you can either:
- Use the new attribute: 
```
[BsonDateOnlyOptions(BsonType.Document, DateOnlyDocumentFormat.YearMonthDay)]
 public DateOnly DateOfBirth { get; set; }
```
- Register a static serializer and set the correct format:
```
BsonSerializer.RegisterSerializer(new DateOnlySerializer(BsonType.Document, DateOnlyDocumentFormat.YearMonthDay));
```

PR #1535, opened 2024-11-05
CSHARP-2096: Make EnumRepresentationConvention also affect collection…

papafe:
…s of Enums

### DIFF
--- a/src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
+++ b/src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
@@ -15,7 +15,7 @@
 
 using System;
 using System.Reflection;
-using MongoDB.Bson.Serialization.Options;
+using MongoDB.Bson.Serialization.Serializers;
 
 namespace MongoDB.Bson.Serialization.Conventions
 {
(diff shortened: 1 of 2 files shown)


papafe:
This is still in draft because I need to check what happens with dictionaries

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -50,60 +49,47 @@ public EnumRepresentationConvention(BsonType representation)
...
-                if (representationConfigurableSerializer != null)
-                {
-                    var reconfiguredSerializer = representationConfigurableSerializer.WithRepresentation(_representation);
-                    memberMap.SetSerializer(reconfiguredSerializer);
-                }
+                memberMap.SetSerializer(representationConfigurableSerializer.WithRepresentation(_representation));
                 return;
             }
 
-            if (IsNullableEnum(memberType))
```
papafe, line 68: I have removed the "nullable enum" case, since it is covered by new code

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -50,60 +49,47 @@ public EnumRepresentationConvention(BsonType representation)
...
-                    }
-                }
-                return;
+                memberMap.SetSerializer(reconfiguredSerializer);
             }
         }
 
-        // private methods
-        private bool IsNullableEnum(Type type)
+        private IBsonSerializer Reconfigure(IBsonSerializer serializer)
```
papafe, line 67: This method here is recursively  trying to fin an `EnumSerializer`. If it finds it, it changes its representation and sets it as the new child serializer.
If no `EnumSerializer` "leaf" is found, it will return null.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -50,60 +49,47 @@ public EnumRepresentationConvention(BsonType representation)
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
-            var memberType = memberMap.MemberType;
-            var memberTypeInfo = memberType.GetTypeInfo();
+            var serializer = memberMap.GetSerializer();
 
-            if (memberTypeInfo.IsEnum)
+            if (memberMap.MemberType.IsEnum && serializer is IRepresentationConfigurable representationConfigurableSerializer)
```
papafe, line 54: This part here of the code could be removed, as it is a special case of `Reconfigure`. I decided to keep it as maybe it's slightly faster than `Reconfigure` and it's possibly a much more common case. 

I've removed the "nullable enum" case though, but we can decide it needs to have the same treatment as this.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -14,8 +14,7 @@
 */
```
papafe, line 14: This is effectively a beahviour breaking change. Should we keep this for next major version?

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -50,60 +49,47 @@ public EnumRepresentationConvention(BsonType representation)
...
-                Nullable.GetUnderlyingType(type).GetTypeInfo().IsEnum;
+            if (serializer is IChildSerializerConfigurable childSerializerConfigurable)
+            {
+                var childSerializer = childSerializerConfigurable.ChildSerializer;
+                var reconfiguredChildSerializer = Reconfigure(childSerializer);
+                return reconfiguredChildSerializer is null ? null : childSerializerConfigurable.WithChildSerializer(reconfiguredChildSerializer);
+            }
+
+            var serializerType = serializer.GetType();
+            if (serializerType.IsGenericType && serializerType.GetGenericTypeDefinition() == typeof(EnumSerializer<>))
```
papafe, line 77: I'm not sure if there is an easier/more efficient way of doing this here.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -50,60 +49,47 @@ public EnumRepresentationConvention(BsonType representation)
...
-                    var reconfiguredSerializer = representationConfigurableSerializer.WithRepresentation(_representation);
-                    memberMap.SetSerializer(reconfiguredSerializer);
-                }
+                memberMap.SetSerializer(representationConfigurableSerializer.WithRepresentation(_representation));
                 return;
             }
 
-            if (IsNullableEnum(memberType))
+            var reconfiguredSerializer = Reconfigure(serializer);
+            if (reconfiguredSerializer is not null)
```
rstam, line 61: Why do you prefer `is not null` to `!= null`?

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: Searching for an `EnumSerializer` is a little stricter than what we used to do, which was search for a serializer for an enum value type that was representation configurable.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: How about?

```
if (serializer.ValueType.IsEnum && serializer is IRepresentationConfigurable representationConfigurable)
{                                                                                                       
    return representationConfigurable.WithRepresentation(_representation);                              
}                                                                                                       
```

rstam (changes requested):

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: I think sometimes it reads a little bit better, but in this case it probably does not make much of a difference. So I can definitely use `!=` if we prefer it

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: That's perfect! I did not think about this 🤦

papafe:
@rstam I think this is ready to be merged, but I'd keep it open until we want to make the next major release, given that this is a behaviour breaking change

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: I agree with your choice here.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: Probably.

rstam (approved):
Sounds reasonable to wait until the next major release to deploy this behavioral breaking change.

papafe:
Closing this PR for now as we should merge it when we want to release next major version, as this is a behaviour breaking change

PR #1540, opened 2024-11-12
CSHARP-5412: Performance: Avoid allocations of BsonDeserializationContext when deserialising a batch

obligaron:
See [CSHARP-5412](https://jira.mongodb.org/browse/CSHARP-5412) for details.

This PR reuses a `BsonDeserializationContext` so that only one object needs to be allocated for a batch instead of one object per document in the batch.


papafe (approved):
LGTM

PR #1541, opened 2024-11-14
CSHARP-5017: Retry KMS requests on transient errors

papafe:


src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -301,25 +300,45 @@ private void SendKmsRequest(KmsRequest request, CancellationToken cancellation)
...
+                            throw new IOException();
+                        }
+
+                        var responseBytes = new byte[count];
+                        Buffer.BlockCopy(buffer, 0, responseBytes, 0, count);
+                        request.Feed(responseBytes);
+                    }
                 }
             }
+            catch (Exception ex) when (ex is IOException or SocketException)
```
papafe, line 338: Not sure here which other kind of network-errors related exceptions we could get

papafe:
This PR is still in a rough-ish state, and it's here mostly to get feedback on the approach used. 
What is missing:
- fix sync version of the methods
- add sync version of the test
- make the tests work on evergreen
- some small cleanup

src/MongoDB.Driver.Encryption/Library.cs
```diff
@@ -210,6 +210,13 @@ static Library()
                 () => __loader.Value.GetFunction<Delegates.mongocrypt_ctx_destroy>(("mongocrypt_ctx_destroy")), true);
             _mongocrypt_kms_ctx_get_kms_provider = new Lazy<Delegates.mongocrypt_kms_ctx_get_kms_provider>(
                 () => __loader.Value.GetFunction<Delegates.mongocrypt_kms_ctx_get_kms_provider>(("mongocrypt_kms_ctx_get_kms_provider")), true);
+
```
papafe, line 213: For these and the following methods/variables, is there a specific ordering? It does not seem to be completely alphabetical so for now I've put the new things on the bottom

adelinowona: There isn't a specific order for the methods, but many of them are grouped together. For instance, you'll notice that most of the setopt methods are clustered. For now, I recommend placing the mongocrypt_setopt_retry_kms method alongside the other setopt methods and leaving the other new methods at the bottom. You can add a TODO comment above the static constructor to organize the methods later (so you don't have unrelated changes to your PR). I'll handle that when addressing technical debt in Libmongocrypt later this quarter.

src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
adelinowona: I saw some [comment](https://jira.mongodb.org/browse/DRIVERS-1541?focusedCommentId=6258309&focusedId=6258309&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6258309:~:text=Jeffrey%20Yemin%20this%20work%20covers%20only%20HTTP%20errors.%20Socket%2Dlevel%20errors%20have%20to%20be%20handled%20by%20each%20driver.) on the Drivers ticket for this feature that it only pertains to HTTP errors so for socket-level errors do we also retry on those? Other comments in the ticket seems to suggest we'll need to tell libmongocrypt to reset its state if we encounter socket errors while reading a kms response.

src/MongoDB.Driver.Encryption/CryptContext.cs
```diff
@@ -170,6 +170,13 @@ public KmsRequestCollection GetKmsMessageRequests()
             return new KmsRequestCollection(requests, this);
         }
 
+        //TODO I think we should remove the previous method and use this
+        public KmsRequest GetNextKmsMessageRequest()
```
adelinowona, line 174: Agreed!

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1428,6 +1429,135 @@ public void ExternalKeyVaultTest(
...
+
+            await SetFailure(failureType, 1);
+
+            var ex2 = await Record.ExceptionAsync(async () => await clientEncryption.EncryptAsync(new BsonInt32(123),
+                new EncryptOptions("AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic", keyId: dataKey)));
+            ex2.Should().BeNull();
+
+            if (failureType == "network")
+            {
+                await SetFailure("network", 4);
```
adelinowona, line 1493: minor: could probably just use `failureType` since we know it is `network` after the if statement

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -1428,6 +1429,135 @@ public void ExternalKeyVaultTest(
...
+
+            const string endpoint = "127.0.0.1:9003";
+
+            var masterKey = kmsProvider switch
+            {
+                "aws" => new BsonDocument
+                {
+                    { "region", "foo" },
+                    { "key", "bar" },
+                    { "endpoint", "127.0.0.1:9003" }
```
adelinowona, line 1449: minor: replace "127.0.0.1:9003" here and other uses below with `endpoint`

src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
papafe: Yes, I think that comment is a little bit misleading. So, as far as I have understood the situation is like this:
- The HTTP error are actually handled internally by libmongocrypt when we "feed" the response bytes to it. It reads the response and understands if it's an HTTP response with an error code. 
- The network errors instead need to be recognized by us. When we do, we need to call `mongocrypt_kms_ctx_fail` (that is used inside the new `Fail` method) to notify libmongocrypt of the network error.
- Internally, the handling for both errors is the same, as it retries the request a certain number of times

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
papafe: Definitely

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
papafe: I thought about it but I would prefer to keep it this way to make it easier to read

src/MongoDB.Driver.Encryption/CryptContext.cs
papafe: Done :)

papafe:
@adelinowona I've fixed also the sync part of the test and done some cleanup. 
For me this is mergeable after a new version of libmongocrypt is released, because we need this small fix: https://github.com/mongodb/libmongocrypt/pull/908

papafe:
@adelinowona I've rebased on main, and I've run the necessary tests on evergreen too and it all seems good. If you want to give a final look :)

adelinowona (approved):
LGTM but I thought you wanted a new release of libmongocrypt with the fix you mentioned before merging?

papafe:
> LGTM but I thought you wanted a new release of libmongocrypt with the fix you mentioned before merging?

Yes, but I realised that actually we're not going to encounter that problem since it's a small issue only if we don't enable the retry of the KMS requests, that we enable by default (at the moment developers can't disable it)

adelinowona (approved):

PR #1545, opened 2024-11-15
CSHARP-4495: Add conventions and attributes to configure ObjectSerial…

papafe:
…izer AllowedTypes

### DIFF
--- a/tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
+++ b/tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
@@ -324,7 +324,7 @@ namespace MongoDB.Bson.Tests.Serialization.Conventions
             var conventionName = Guid.NewGuid().ToString();
 
             var subject = new ObjectSerializerAllowedTypesConvention(Assembly.GetExecutingAssembly());
-            ConventionRegistry.Register(conventionName, new ConventionPack {subject}, _ => true);
+            ConventionRegistry.Register(conventionName, new ConventionPack {subject}, t => t == typeof(TestClass));
 
             var classMap = new BsonClassMap<TestClass>(cm => cm.AutoMap());
             var memberMap = classMap.GetMemberMap("ObjectProp");


src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="allowedDeserializationTypes"></param>
+        /// <param name="allowedSerializationTypes"></param>
+        public AllowedTypesConvention(IEnumerable<Type> allowedDeserializationTypes, IEnumerable<Type> allowedSerializationTypes)
+        {
+            var allowedDeserializationTypesArray = allowedDeserializationTypes as Type[] ?? allowedDeserializationTypes.ToArray();
```
rstam, line 71: Interesting optimization...

src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
```diff
@@ -0,0 +1,147 @@
...
+        // According to CA1044 setter only properties shouldn't exist, but I think it would make sense for this case
+        // but I'm open to other suggestions
+        /// <summary>
+        ///
+        /// </summary>
+#pragma warning disable CA1044
+        public bool AllowDefaultFrameworkTypes
+#pragma warning restore CA1044
+        {
+            set
```
rstam, line 109: Maybe `init` instead of `set`?

src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        // This is defined as a property, so it can be used together with all the other possible constructors.
+        // According to CA1044 setter only properties shouldn't exist, but I think it would make sense for this case
+        // but I'm open to other suggestions
+        /// <summary>
+        ///
+        /// </summary>
+#pragma warning disable CA1044
+        public bool AllowDefaultFrameworkTypes
```
rstam, line 106: I like the idea of giving them a way to opt-in the default framework types.

src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        /// <summary>
+        /// Applies a modification to the member map.
+        /// </summary>
+        /// <param name="memberMap">The member map.</param>
+        public void Apply(BsonMemberMap memberMap)
+        {
+            var memberType = memberMap.MemberType;
+
+            if (memberType == typeof(object) &&  memberMap.GetSerializer() is ObjectSerializer os)
```
rstam, line 139: I would avoid the abbreviation and use `objectSerializer`

src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
rstam: However... we should probably make a defensive copy anyway because the caller might alter the array they passed in later.

src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// </summary>
+        /// <param name="allowedDeserializationTypes"></param>
+        /// <param name="allowedSerializationTypes"></param>
+        public AllowedTypesConvention(IEnumerable<Type> allowedDeserializationTypes, IEnumerable<Type> allowedSerializationTypes)
+        {
+            var allowedDeserializationTypesArray = allowedDeserializationTypes as Type[] ?? allowedDeserializationTypes.ToArray();
+            var allowedSerializationTypesArray = allowedSerializationTypes as Type[] ?? allowedSerializationTypes.ToArray();
+
+            _allowedDeserializationTypes = allowedDeserializationTypesArray.Contains;
+            _allowedDeserializationTypes = allowedSerializationTypesArray.Contains;
```
rstam, line 75: `Contains` is a LINQ method that uses an enumerator, so it might be more efficient to use:

```type => Array.IndexOf(allowedDeserializationTypesArray, type) != -1```

Though I suppose it is possible the LINQ implementation of `Contains` has optimizations for arrays.

**Review by rstam — CHANGES_REQUESTED** [review]
Everything looks very reasonable!

File: src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs (hunk 4 above)
Comment (papafe) [reply src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs:139]: Definitely :)

File: src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs (hunk 2 above)
Comment (papafe) [reply src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs:109]: Yep, `init` would make more sense here

File: src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs (hunk 1 above)
Comment (papafe) [reply src/MongoDB.Bson/Serialization/Conventions/AllowedTypesConvention.cs:71]: Yes, we should just make a copy

**Comment by papafe** [comment]
@rstam This is ready for review now. Apart from all your notes, I've also allowed the convention to work with collections and nested collection (as we did with the `EnumConvention`).

File: src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs [hunk 6]
```diff
@@ -0,0 +1,151 @@
...
+using System.Linq;
+using System.Reflection;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Bson.Serialization.Conventions
+{
+    /// <summary>
+    /// A convention that allows to set the types that can be safely serialized and deserialized with the <see cref="ObjectSerializer"/>.
+    /// </summary>
+    public class ObjectSerializerAllowedTypesConvention
```
BorisDog, line 27: minor: sealed

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,151 @@
...
+        public ObjectSerializerAllowedTypesConvention() : this(Assembly.GetCallingAssembly())
+        {
+        }
+
+        /// <summary>
+        /// Sets a value indicating whether default framework types should be added to the list of types that
+        /// can be serialized and deserialized.
+        /// </summary>
+#pragma warning disable CA1044
+        public bool AllowDefaultFrameworkTypes
```
BorisDog, line 101: Should the default for this value match the default of `ObjectSerializer` for `DefaultFrameworkAllowedTypes`?

BorisDog (changes requested):
Looks good overall

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: Probably yes!

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,151 @@
...
+        public ObjectSerializerAllowedTypesConvention(params Assembly[] allowedAssemblies)
+        {
+            _allowedDeserializationTypes = _allowedSerializationTypes = t => Array.IndexOf(allowedAssemblies, t.Assembly) != -1;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ObjectSerializerAllowedTypesConvention"/> class
+        /// that allows all types contained in the calling assembly.
+        /// </summary>
+        public ObjectSerializerAllowedTypesConvention() : this(Assembly.GetCallingAssembly())
```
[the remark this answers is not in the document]
papafe: Makes sense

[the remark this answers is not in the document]
papafe: I have removed the automatic add of the types in the calling assembly and I have made `AllowDefaultFrameworkTypes ` default to true, similar to the behaviour of the `ObjectSerializer`. For this reason now the delegates that are passed to the `ObjectSerializer` are built lazily.

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,156 @@
...
+        public ObjectSerializerAllowedTypesConvention()
+        {
+            _lazyAllowedDeserializationTypes = new Lazy<Func<Type, bool>>(() =>
+            {
+                return AllowDefaultFrameworkTypes
+                    ? _allowedDeserializationTypes is null
+                        ? ObjectSerializer.DefaultAllowedTypes
+                        : t => _allowedDeserializationTypes(t) || ObjectSerializer.DefaultAllowedTypes(t)
+                    : _allowedDeserializationTypes ?? ObjectSerializer.NoAllowedTypes;
+            });
```
papafe, line 107: Hope that this is readable enough

tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            var subject = new ObjectSerializerAllowedTypesConvention();
+            subject.AllowDefaultFrameworkTypes.Should().BeTrue();
+
+            var memberMap = CreateMemberMap(c => c.ObjectProp);
+            subject.Apply(memberMap);
+
+            var serializer = (ObjectSerializer)memberMap.GetSerializer();
+
+            //Type in default framework types
```
BorisDog, line 141: nit: space after `//` : `// Type`

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Linq;
+using System.Reflection;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Bson.Serialization.Conventions
+{
+    /// <summary>
+    /// A convention that allows to set the types that can be safely serialized and deserialized with the <see cref="ObjectSerializer"/>.
+    /// </summary>
+    public sealed class ObjectSerializerAllowedTypesConvention
```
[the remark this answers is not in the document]
papafe: I think that would make sense

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>
+        /// Default <see cref="ObjectSerializerAllowedTypesConvention"/> where all calling assembly types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllCallingAssemblyTypes => new(Assembly.GetCallingAssembly())
+                { AllowDefaultFrameworkTypes = false };
+
+        /// <summary>
+        /// Default <see cref="ObjectSerializerAllowedTypesConvention"/> where all calling assembly types and default framework types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllCallingAssemblyAndDefaultFrameworkTypes => new(Assembly.GetCallingAssembly());
```
papafe, line 40: Not sure if it makes sense to have both `AllowAllCallingAssemblyAndDefaultFrameworkTypes` and `AllowAllCallingAssemblyTypes` or keep only the second one (but then we need to decide if the default framework types are included or not)

BorisDog: I think we need only need `AllowAllCallingAssemblyAndDefaultFrameworkTypes`

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ObjectSerializerAllowedTypesConvention"/> class.
+        /// </summary>
+        public ObjectSerializerAllowedTypesConvention()
+        {
+            _lazyAllowedDeserializationTypes = new Lazy<Func<Type, bool>>(() => Construct(_allowedDeserializationTypes));
+            _lazyAllowedSerializationTypes = new Lazy<Func<Type, bool>>(() => Construct(_allowedSerializationTypes));
+            return;
```
BorisDog, line 131: no need for `return`

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllCallingAssemblyAndDefaultFrameworkTypes => new(Assembly.GetCallingAssembly());
+
+        /// <summary>
+        /// Default <see cref="ObjectSerializerAllowedTypesConvention"/> where all types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllTypes => new(ObjectSerializer.AllAllowedTypes);
+
+        /// <summary>
+        /// Default <see cref="ObjectSerializerAllowedTypesConvention"/> where no types are allowed for both serialization and deserialization.
```
BorisDog, line 48: Maybe not need for the word "Default", as it might be confused for an actual default?

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: Should keep the shortest name then? I'd vote for that

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: I agree, it's mostly to separate "visually" the method body from the local functions. I'm not super sold either way to be honest

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: Mmmh, I agree

rstam (changes requested):

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        /// <summary>
+        /// A predefined <see cref="ObjectSerializerAllowedTypesConvention"/> where all calling assembly types and default framework types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllCallingAssemblyAndDefaultFrameworkTypes => new(Assembly.GetCallingAssembly());
+
+        /// <summary>
+        /// A predefined <see cref="ObjectSerializerAllowedTypesConvention"/> where all types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllTypes => new(ObjectSerializer.AllAllowedTypes);
```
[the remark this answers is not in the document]
papafe: Makes sense. 
Regarding `AllowAllCallingAssemblyAndDefaultFrameworkTypes`, I'd prefer to keep it as a property, for consistency.

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// Initializes a new instance of the <see cref="ObjectSerializerAllowedTypesConvention"/> class.
+        /// </summary>
+        /// <param name="allowedTypes">A collection of the allowed types for both serialization and deserialization.</param>
+        public ObjectSerializerAllowedTypesConvention(IEnumerable<Type> allowedTypes)
+            : this()
+        {
+            var allowedTypesArray = allowedTypes.ToArray();
+
+            _allowedDeserializationTypes = allowedTypesArray.Contains;
+            _allowedSerializationTypes = allowedTypesArray.Contains;
```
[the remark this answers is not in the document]
papafe: This should use only one delegate, so it should not make a difference

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,173 @@
...
+            _allowedDeserializationTypes = _allowedSerializationTypes = t => allowedAssemblies.Contains(t.Assembly);
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ObjectSerializerAllowedTypesConvention"/> class.
+        /// </summary>
+        public ObjectSerializerAllowedTypesConvention()
+        {
+            _lazyAllowedDeserializationTypes = new Lazy<Func<Type, bool>>(() => Construct(_allowedDeserializationTypes));
+            _lazyAllowedSerializationTypes = new Lazy<Func<Type, bool>>(() => Construct(_allowedSerializationTypes));
```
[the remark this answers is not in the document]
papafe: The `Func` is built when it's used the first time, so the variables will be set by then

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            Func<Type, bool> Construct(Func<Type, bool> allowedTypes)
+            {
+                return AllowDefaultFrameworkTypes
+                    ? allowedTypes is null
+                        ? ObjectSerializer.DefaultAllowedTypes
+                        : t => allowedTypes(t) || ObjectSerializer.DefaultAllowedTypes(t)
+                    : allowedTypes ?? ObjectSerializer.NoAllowedTypes;
```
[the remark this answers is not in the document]
papafe: Looks good!

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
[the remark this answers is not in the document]
rstam: The issue with it being a property is that one usually expects that a property will have the same value no matter where it is called from, whereas it is less unexpected for a method to return a different value each time it is called (especially if you name the method in a way that implies that).

I don't insist on this though, but I don't want to highlight that this property is behaving in an odd way.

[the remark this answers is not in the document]
papafe: I think I agree with you on this, I changed it to be a method

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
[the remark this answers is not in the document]
rstam: This class needs to implement `IMemberMapConvention`.

Follow the pattern from other similar classes:

```
public sealed class ObjectSerializerAllowedTypesConvention : ConventionBase, IMemberMapConvention
 ```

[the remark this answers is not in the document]
Comment (rstam) [reply src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs:27]: To verify why implementing `IMemberMapConvention` is required try writing a test that registers this convention and see if it gets used during automapping.

This test will have to be disabled and only run manually in isolation, because registering this convention globally would affect other tests.

File: src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs [hunk 19]
```diff
@@ -0,0 +1,183 @@
...
+        /// <summary>
+        /// A predefined <see cref="ObjectSerializerAllowedTypesConvention"/> where all types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowAllTypes { get; } = new(ObjectSerializer.AllAllowedTypes);
+
+        /// <summary>
+        /// A predefined <see cref="ObjectSerializerAllowedTypesConvention"/> where no types are allowed for both serialization and deserialization.
+        /// </summary>
+        public static ObjectSerializerAllowedTypesConvention AllowNoTypes { get; } = new(ObjectSerializer.NoAllowedTypes)
+            { AllowDefaultFrameworkTypes = false };
```
rstam, line 40: This formatting threw me off. I didn't realize at first that the line was continued on the next line.

Maybe this way:

```
public static ObjectSerializerAllowedTypesConvention AllowNoTypes { get; } =
    new(ObjectSerializer.NoAllowedTypes) { AllowDefaultFrameworkTypes = false };
```

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -0,0 +1,183 @@
...
+        public static ObjectSerializerAllowedTypesConvention GetAllowAllCallingAssemblyAndDefaultFrameworkTypesConvention() => new(Assembly.GetCallingAssembly());
+
+        // private fields
+        private readonly Func<Type, bool> _allowedDeserializationTypes;
+        private readonly Func<Type, bool> _allowedSerializationTypes;
+
+        private readonly Lazy<Func<Type, bool>> _effectiveAllowedDeserializationTypes;
+        private readonly Lazy<Func<Type, bool>> _effectiveAllowedSerializationTypes;
+
+        private readonly bool _allowDefaultFrameworkTypes = true;
```
rstam, line 61: Consider placing line 61 immediately after line 56.

tests/MongoDB.Bson.Tests/Serialization/Conventions/MemberDefaultValueConventionTests.cs
```diff
@@ -37,42 +38,51 @@ private class B
         [Fact]
         public void TestMappingUsesMemberDefaultValueConvention()
         {
+            var conventionName = Guid.NewGuid().ToString();
```
papafe, line 41: I found these tests that were registering conventions but never unregistering them

tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
```diff
@@ -0,0 +1,354 @@
...
+            serializer.AllowedDeserializationTypes(typeof(TestClass)).Should().BeTrue();
+            serializer.AllowedSerializationTypes(typeof(TestClass)).Should().BeTrue();
+            serializer.AllowedDeserializationTypes(typeof(long)).Should().BeTrue();
+            serializer.AllowedSerializationTypes(typeof(long)).Should().BeTrue();
+            serializer.AllowedDeserializationTypes(typeof(EnumSerializer)).Should().BeFalse();
+            serializer.AllowedSerializationTypes(typeof(EnumSerializer)).Should().BeFalse();
+        }
+
+        [Fact]
+        public void Convention_should_be_applied_during_automapping()
```
papafe, line 322: @rstam I've added this test to verify that the convention gets registered during automap. I know you've said we should only run it manually, but can't we register the convention and unregister it at the end of the test like here?

tests/MongoDB.Bson.Tests/Serialization/Conventions/MemberDefaultValueConventionTests.cs
```diff
@@ -37,42 +38,51 @@ private class B
...
-            ConventionRegistry.Register("test", pack, t => t == typeof(A));
+            ConventionRegistry.Register(conventionName, pack, t => t == typeof(A));
 
             var classMap = new BsonClassMap<A>(cm => cm.AutoMap());
 
             var defaultValue = classMap.GetMemberMap("Match").DefaultValue;
             Assert.IsType<int>(defaultValue);
             Assert.Equal(1, defaultValue);
+
+            ConventionRegistry.Remove(conventionName);
```
rstam, line 52: Wow I am surprised the `Remove` method even exists.

A general principle of our whole serialization configuration machinery is that you shouldn't be able to change how things are serialized in the middle of the program run.

Though I guess this `Remove` method is somewhat less harmful because it doesn't change how anything that is already configured is serialized, only how future classes that have not yet been configured will end up being configured after the convention has been removed.

Note also that this convention does not HAVE to be unregistered. It is already programmed to ONLY apply to this one class with `t => t == typeof(A)`.

tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
rstam: As i commented above I'm surprised we even have a method to unregister a convention. You shouldn't be able to alter conventions in the middle of a program run.

We can either unregister it as you discovered (even though I didn't think you should be able to do that), or you can simply program the convention to only apply to the class in this test and just leave it registered forever.

tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
```diff
@@ -0,0 +1,354 @@
...
+        {
+            var conventionName = Guid.NewGuid().ToString();
+
+            var subject = new ObjectSerializerAllowedTypesConvention(Assembly.GetExecutingAssembly());
+            ConventionRegistry.Register(conventionName, new ConventionPack {subject}, _ => true);
```
rstam, line 327: Can't we use something more restrictive that `t => true`?

If we think about a time in the future where we want to run our tests in parallel relying on unregistering will no longer work.

tests/MongoDB.Bson.Tests/Serialization/Conventions/MemberDefaultValueConventionTests.cs
rstam: This was not a problem because of `t => t == typeof(A)`

rstam (changes requested):

tests/MongoDB.Bson.Tests/Serialization/Conventions/MemberDefaultValueConventionTests.cs
papafe: I understand we could have a convention that applies to a certain type and leave it registered, and while this likely does not affect anything else, shouldn’t we aim to keep the tests as independent as possible?

rstam: Yes, tests should be as independent as possible. That's why we should use `t => t == typeof(A)` instead of `t => true` here.

It's fine if you want to unregister this convention at the end of the test. I just wanted to make the following points:

1. The test was not inherently wrong because the registered convention only applied to this test
2. I was surprised we even had a `Remove` method

The `Remove` method concerns me because the way serialization is supposed to work is that an application configures serialization at startup and then leaves it alone. Removing a convention at some later time sounds like it would destabilize the application because POCOs might be serialized differently depending on whether the POCO is mapped before or after the convention was removed.

papafe: I agree with all the points :) 
I modified the predicate to be more restrictive, but kept the `Remove` to have a clean state after each test. I can agree also that maybe that method should not be there, but for now I'd say it's better to use it

rstam (approved):
LGTM

PR #1547, opened 2024-11-16
CSHARP-5305: Improve error message when custom LINQ extension method is executed client side.

rstam:


src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -42,7 +42,7 @@ internal static HashSet<T> ToHashSet<T>(this IEnumerable<T> source)
         /// <returns>Only meant to be used in Update specifications.</returns>
         public static TSource AllElements<TSource>(this IEnumerable<TSource> source)
         {
-            throw new NotSupportedException("This method is not functional. It is only usable in conjunction with MongoDB.");
+            throw new NotSupportedException($"{nameof(AllElements)} can only be used in LINQ queries executed server-side.");
```
rstam, line 45: Is this error message good enough or can you think of even better?

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -862,7 +862,7 @@ public static IEnumerable<TResult> TopN<TSource, TKey, TResult>(
         /// <returns>The filtered results.</returns>
         public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate, int limit)
         {
-            throw new NotSupportedException("This method is not functional. It is only usable in conjunction with MongoDB.");
+            throw new NotSupportedException($"{nameof(Where)} can only be used in LINQ queries executed server-side.");
```
sanych-sun, line 865: Can move relace by single helper method to throw:
```
private void ThrowNotSupportedException([CallerMemberName]string callerName = null)
{
    throw new NotSupportedException($"{callerName} can only be used in LINQ queries executed server-side.");
}
```
So each method could be converted to simple expression body method:
```
public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate, int limit) 
=> ThrowNotSupportedException()
```

src/MongoDB.Driver/Linq/MongoEnumerable.cs
BorisDog: IMHO it's good enough.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
BorisDog: Good idea.

damieng: I think losing "MongoDB" is a step backward as no other LINQ providers will be able to translate this.

rstam: Good point. I will include "MongoDB" in the error message.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/CustomLinqExtensionMethodHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using System.Runtime.CompilerServices;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Misc
+{
+    internal static class CustomLinqExtensionMethodHelper
+    {
+        public static Exception CreateNotSupportedException([CallerMemberName] string methodName = null)
+        {
+            throw new NotSupportedException($"{methodName} can only be used in MongoDB LINQ queries executed server-side.");
```
sanych-sun, line 25: Should we return it instead of throwing?

sanych-sun (changes requested):

src/MongoDB.Driver/Linq/DateTimeExtensions.cs
```diff
@@ -32,7 +33,7 @@ public static class DateTimeExtensions
         /// <returns>The resulting DateTime.</returns>
         public static DateTime Add(this DateTime @this, long value, DateTimeUnit unit)
         {
-            throw new InvalidOperationException("This DateTime.Add method is only intended to be used in LINQ queries.");
+            throw CustomLinqExtensionMethodHelper.CreateNotSupportedException();
```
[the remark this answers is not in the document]
sanych-sun: Kind of "Yes", but being consistent here probably better. I think we should accept this change.

sanych-sun (approved):
LGTM

PR #1549, opened 2024-11-18
CSHARP-5403: Fix queries and projection on GridFSFileInfo Id property.

rstam:


src/MongoDB.Driver/GridFS/GridFSFileInfoSerializerCompat.cs
```diff
@@ -43,6 +43,7 @@ public GridFSFileInfoSerializer()
             RegisterMember("ChunkSizeBytes", "chunkSize", new Int32Serializer());
             RegisterMember("ContentType", "contentType", new StringSerializer());
             RegisterMember("Filename", "filename", new StringSerializer());
+            RegisterMember("Id", "_id", ObjectIdSerializer.Instance);
```
rstam, line 46: It's not normal to have two properties that alias to the same underlying element, but I can't think of any reason why this wouldn't work.

Note that we don't support serializing this class, only *deserializing*.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5403Tests.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public class CSharp5403Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Project_Id_should_work()
+        {
+            var collection = GetCollection();
+
+            var find = collection
+                .Find(Builders<GridFSFileInfo>.Filter.Where(x => x.Id == ObjectId.Parse("111111111111111111111111")))
```
rstam, line 31: I wanted to verify that we could filter on `Id` or `IdAsBsonValue` as well as project it.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5403Tests.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+        private IMongoCollection<GridFSFileInfo> GetCollection()
+        {
+            var collection = GetCollection<GridFSFileInfo>("test");
+            CreateCollection(
+                collection.Database.GetCollection<BsonDocument>(collection.CollectionNamespace.CollectionName),
+                BsonDocument.Parse("{ _id : { $oid : '111111111111111111111111' }, filename : 'One' }"),
+                BsonDocument.Parse("{ _id : { $oid : '222222222222222222222222' }, filename : 'Two' }"));
```
rstam, line 71: We have to insert these as `BsonDocument` because we don't support serializing `GridFSFileInfo` values.

BorisDog (approved):
LGTM!

PR #1555, opened 2024-11-23
Update ScalarDiscriminatorConvention.cs To fix bug NotSupportedExcept…

duc01226:
To fix bug NotSupportedException("OfType is not supported with the configured discriminator convention.")

### DIFF
--- a/src/MongoDB.Bson/Serialization/Conventions/ScalarDiscriminatorConvention.cs
+++ b/src/MongoDB.Bson/Serialization/Conventions/ScalarDiscriminatorConvention.cs
@@ -20,7 +20,7 @@ namespace MongoDB.Bson.Serialization.Conventions
     /// <summary>
     /// Represents a discriminator convention where the discriminator is provided by the class map of the actual type.
     /// </summary>
-    public class ScalarDiscriminatorConvention : StandardDiscriminatorConvention
+    public class ScalarDiscriminatorConvention : StandardDiscriminatorConvention, IScalarDiscriminatorConvention
     {
         // constructors
         /// <summary>


damieng:
Hi.

The IScalarDiscriminatorConvention interface is new and requires an additional method that the older ScalarDiscriminatorConvention does not implement so this code would not compile.

The new interface provides a list of the discriminator values not just for the specified type but for all it's sub-types in order that IsType, GetType and where...is queries correctly return sub-types.

ScalarDiscriminatorConvention has never supported sub-types before and so it is missing this - additionally we do not yet have any way a user can specify this data in convenient way either through convention or attributes which is why this class does not yet implement it.

If you wish to use scalar discriminators correctly I would recommend implementing IScalarDiscriminatorConvention in your own class and providing the correct hierarchy metadata that suits your needs directly until we have a baked-in solution.

duc01226:
But i am having a bug when upgrade to version 3 related to polymorphism
OfType function. I debug and it throw exception because it doesn’t
implement the interface. The switch case got to exception.

I have one Entity Email and second child entity CommonEmail extend Email. I
have register them both. It works in version 2.

duc01226:
Hello,

I am encountering a bug when upgrading to version 3 that is related to the
polymorphism of the `OfType` function. I have debugged the issue and found
that an exception is thrown because the interface is not implemented. The
switch case then proceeds to handle the exception.

In my code, I have an `Email` entity and a child entity called
`CommonEmail` that extends `Email`. I have registered both of these
entities. While this setup worked in version 2, it is causing problems in
version 3.

Best regards,
Duc

damieng:
While v2 of the driverdid not throw an error in the it did not work as intended or expected. Performing a query with `OfType<Email>()` generates a `"_t": "Email"` MQL statement which is incorrect. It should include sub-types such as `CommonEmail` which it did not. The same problem also affects `.Where(e => e is Email)`.

You can track progress on a correct implementation for the C# driver our JIRA board for this issue at https://jira.mongodb.org/browse/CSHARP-5356

PR #1559, opened 2024-11-27
CSHARP-5356: Support `OfType` and `is` with scalar discriminators when using class mapped serializers.

rstam:


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -61,6 +62,9 @@ public class BsonClassMap
         private int _extraElementsMemberIndex = -1;
         private List<Type> _knownTypes = new List<Type>();
 
+        private ConcurrentDictionary<BsonValue, Type> _discriminatorToTypeMap = new();
+        private ConcurrentDictionary<Type, BsonValue> _typeToDiscriminatorMap = new();
```
rstam, line 66: This is the new information we need to know to implement an `IScalarDiscriminatorConvention` for `BsonClassMapSerializer`.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -695,6 +703,11 @@ public BsonClassMap Freeze()
                         }
                         _frozen = true;
 
+                        if (_discriminator != null)
+                        {
+                            AddKnownDiscriminator(_discriminator, _classType);
+                        }
+
```
rstam, line 710: Seems like the moment we `Freeze` a class map is the safest time to inform the base class maps of our existence.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1140,6 +1153,37 @@ public void SetExtraElementsMember(BsonMemberMap memberMap)
             _extraElementsMemberMap = memberMap;
         }
 
+        internal void AddKnownDiscriminator(BsonValue discriminator, Type type)
+        {
+            if (!_classType.IsAssignableFrom(type))
```
rstam, line 1158: Sanity check. Shouldn't happen.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1140,6 +1153,37 @@ public void SetExtraElementsMember(BsonMemberMap memberMap)
...
+            {
+                throw new ArgumentException($"Type {type} is not assignable to {_classType}.", nameof(type));
+            }
+
+            if (_classType == typeof(object) || _classType == typeof(ValueType) || _classType.IsInterface)
```
rstam, line 1163: Somewhat arbitrary choice to not track discriminators for `object`.

I'm concerned about conflicts with the `ObjectSerializer`.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1331,31 +1375,37 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
 
             IDiscriminatorConvention LookupDiscriminatorConvention()
             {
-                var classMap = this;
-                while (classMap != null)
+                if (_discriminatorConvention != null)
```
rstam, line 1378: Changed implementation from a loop to recursive.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1331,31 +1375,37 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
+                }
+
+                if (_isRootClass)
+                {
+                    return StandardDiscriminatorConvention.Hierarchical;
+                }
+
+                if (_baseClassMap.ClassType == typeof(object))
+                {
+                    return new BsonClassMapScalarDiscriminatorConvention(this, "_t");
```
rstam, line 1395: This is where we decide to use a `BsonClassMapScalarDiscriminatorConvention`.

We decide this because we got here without seeing a root class, which would have resulted in choosing a `StandardDiscriminatorConvention.Hierarchical` before getting here.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1331,31 +1375,37 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
+                }
+                else
+                {
+                    var discriminatorConvention = _baseClassMap.GetDiscriminatorConvention();
+                    if (discriminatorConvention is BsonClassMapScalarDiscriminatorConvention)
                     {
-                        // in this case LookupDiscriminatorConvention below will find it
-                        break;
+                        return new BsonClassMapScalarDiscriminatorConvention(this, discriminatorConvention.ElementName);
```
rstam, line 1402: Normally we'd just use the inherited discriminator convention but to make this all work we need a new convention at each level that will have its own `_classMap` reference.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public BsonValue[] GetDiscriminatorsForTypeAndSubTypes(Type type)
+        {
+            return _typeToDiscriminatorsForTypeAndSubTypesMap.GetOrAdd(type, MapTypeToDiscriminatorsForTypeAndSubTypes);
```
rstam, line 112: We could compute the list from scratch every time but it seems worth caching.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,152 @@
...
+            var bookmark = bsonReader.GetBookmark();
+            try
+            {
+                bsonReader.ReadStartDocument();
+                if (bsonReader.FindElement(_elementName))
+                {
+                    var context = BsonDeserializationContext.CreateRoot(bsonReader);
+                    return BsonValueSerializer.Instance.Deserialize(context);
+                }
+                bsonReader.ReturnToBookmark(bookmark);
```
damieng, line 142: Unnecessary given it's also in finally.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Linq;
+using FluentAssertions;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Driver.Linq;
+using MongoDB.Bson.Serialization.Serializers;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp5356Tests : Linq3IntegrationTest
```
damieng, line 25: Is it worth adding some .Where(x => x is Type) tests here too? and perhaps .Where(x => !(x is Type)) ?

rstam: Done.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+        [Fact]
+        public void Where_is_Animal_should_work()
+        {
+            var collection = GetCollection();
+
+            var queryable = collection.AsQueryable()
+                .Where(x => x is Animal);
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { } }");
```
rstam, line 124: I tried optimizing this stage away but that broke a lot of other tests.

Might be worth discussing and if we want to do it create a new ticket just for that.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+                .GroupBy(x => 1)
+                .Select(x => new { A = x.ToArray() })
+                .Select(x => new { A = x.A.Where(x => x is Animal).ToArray() });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $group : { _id : 1, _elements : { $push : '$$ROOT' } } }",
+                "{ $project : { A : '$_elements', _id : 0 } }",
+                "{ $project : { A : { $filter : { input : '$A', as : 'x', cond : true } }, _id : 0 } }");
```
rstam, line 280: Potential MQL optimization here since `cond : true`.

Is it worth it?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { _t : { $ne : 'Snake' } } }");
+
+            var results = queryable.ToList();
+            results.Select(x => x.Id).Should().Equal(1, 2);
+        }
+
+        [Fact]
+        public void Array_Where_is_Animal_should_work()
```
rstam, line 266: I tried adding tests for array `OfType` but that is not currently supported.

I could implement support in this ticket or I could create a new ticket. @BorisDog?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+                .GroupBy(x => 1)
+                .Select(x => new { A = x.ToArray() })
+                .Select(x => new { A = x.A.Where(x => !(x is Animal)).ToArray() });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $group : { _id : 1, _elements : { $push : '$$ROOT' } } }",
+                "{ $project : { A : '$_elements', _id : 0 } }",
+                "{ $project : { A : { $filter : { input : '$A', as : 'x', cond : { $not : true } } }, _id : 0 } }");
```
rstam, line 385: I was surprised to see `{ $not : true }` instead of `false`.

Is it worth simplifying?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+                .GroupBy(x => 1)
+                .Select(x => new { A = x.ToArray() })
+                .Select(x => new { A = x.A.Where(x => !(x is Mammal)).ToArray() });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $group : { _id : 1, _elements : { $push : '$$ROOT' } } }",
+                "{ $project : { A : '$_elements', _id : 0 } }",
+                "{ $project : { A : { $filter : { input : '$A', as : 'x', cond : { $not : { $in : ['$$x._t', ['Cat', 'Dog', 'Mammal']] } } } }, _id : 0 } }");
```
rstam, line 406: I was expecting to see `$nin` but turns out the server doesn't have `$nin` in aggregation expressions (it does have it in filters).

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,516 @@
...
+                .GroupBy(x => 1)
+                .Select(x => new { A = x.ToArray() })
+                .Select(x => new { A = x.A.Where(x => !(x is Cat)).ToArray() });
+
+            var stages = Translate(collection, queryable);
+            AssertStages(
+                stages,
+                "{ $group : { _id : 1, _elements : { $push : '$$ROOT' } } }",
+                "{ $project : { A : '$_elements', _id : 0 } }",
+                "{ $project : { A : { $filter : { input : '$A', as : 'x', cond : { $not : { $eq : ['$$x._t', 'Cat'] } } } }, _id : 0 } }");
```
rstam, line 427: I think `$not/$eq` could be simplified to `$ne`.

We'd have to convince ourselves that there are no edge cases where the results would differ.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="bsonReader"></param>
+        /// <param name="nominalType"></param>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public Type GetActualType(IBsonReader bsonReader, Type nominalType)
+        {
+            BsonSerializer.EnsureKnownTypesAreRegistered(nominalType);
```
rstam, line 68: Added call to `EnsureKnownTypesAreRegistered`,

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                return nominalType;
+            }
+
+            if (_classMap.DiscriminatorToTypeMap.TryGetValue(discriminator, out Type actualType))
+            {
+                return actualType;
+            }
+
+            if (nominalType == typeof(object) && discriminator.IsString)
```
rstam, line 81: This `if` is to play nice with the `ObjectSerializer`.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            return discriminators.OrderBy(x => x).ToArray();
+        }
+
+        private BsonValue ReadDiscriminator(IBsonReader bsonReader)
+        {
+            // the BsonReader is sitting at the value whose actual type needs to be found
+            var bsonType = bsonReader.GetCurrentBsonType();
+            if (bsonType == BsonType.Document)
```
rstam, line 146: Added a check that we're sitting at a document and not something else.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstSimplifier.cs
```diff
@@ -277,6 +280,28 @@ elemMatchOperation.Filter is AstFieldOperationFilter elemFilter &&
             }
         }
 
+        public override AstNode VisitFilterExpression(AstFilterExpression node)
```
rstam, line 283: Added several new simplifications to the `AstSimplifier`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/OfTypeMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    internal static class OfTypeMethodToAggregationExpressionTranslator
+    {
+        private static MethodInfo[] __ofTypeMethods =
+        {
+            EnumerableMethod.OfType,
+            QueryableMethod.OfType
+        };
+
+        public static AggregationExpression Translate(TranslationContext context, MethodCallExpression expression)
```
rstam, line 36: Added support for `OfType` in aggregation expressions.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/OfTypeMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,92 @@
...
+            var method = expression.Method;
+            var arguments = expression.Arguments;
+
+            if (method.IsOneOf(__ofTypeMethods))
+            {
+                var sourceExpression = arguments[0];
+                var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
+                NestedAsQueryableHelper.EnsureQueryableMethodHasNestedAsQueryableSource(expression, sourceTranslation);
+
+                var sourceAst = sourceTranslation.Ast;
```
rstam, line 47: This block of code is to support wrapped arrays as input to the stage.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/WhereMethodToAggregationExpressionTranslator.cs
```diff
@@ -40,9 +40,17 @@ public static AggregationExpression Translate(TranslationContext context, Method
             {
                 var sourceExpression = arguments[0];
                 var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
-                var itemSerializer = ArraySerializerHelper.GetItemSerializer(sourceTranslation.Serializer);
                 NestedAsQueryableHelper.EnsureQueryableMethodHasNestedAsQueryableSource(expression, sourceTranslation);
 
+                var sourceAst = sourceTranslation.Ast;
```
rstam, line 45: This block of code is to support wrapped arrays as input to the stage.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
```diff
@@ -40,7 +40,7 @@ private class C : A
         {
         }
 
-        [BsonDiscriminator("D~", RootClass = true)]
+        [BsonDiscriminator("D~")]
```
rstam, line 43: `RootClass = true` in the middle of the class hierarchy makes no sense.

This must have been an oversight.

tests/MongoDB.Bson.Tests/Serialization/Serializers/KnownTypesTests.cs
```diff
@@ -34,7 +34,6 @@ private class B : A
         {
         }
 
-        [BsonDiscriminator(RootClass = true)]
```
rstam, line 37: `RootClass = true` in the middle of the class hierarchy makes no sense.

This must have been an oversight.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4100FilterTests.cs
```diff
@@ -47,7 +47,7 @@ public void Contains_with_string_constant_and_char_constant_should_work()
 
             var stages = Translate(collection, queryable);
 
-            AssertStages(stages, "{ $match : { } }");
+            AssertStages(stages);
```
rstam, line 50: As a result of a new pipeline simplification any `{ $match : { } }` stages are removed as they do nothing (they let all stages through).

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4100FilterTests.cs
```diff
@@ -301,7 +301,7 @@ public void Contains_with_string_field_and_string_constant_and_comparisonType_sh
 #if !NETFRAMEWORK
         [Theory]
         [InlineData(StringComparison.CurrentCulture, "{ $match : { _id : { $type : -1 } } }")]
-        [InlineData(StringComparison.CurrentCultureIgnoreCase, "{ $match : { } }")]
+        [InlineData(StringComparison.CurrentCultureIgnoreCase, null)]
```
rstam, line 304: A value of `null` for  `expectedStage` means "no stage expected".

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5356Tests.cs
```diff
@@ -0,0 +1,606 @@
...
+                .OfType<Snake>();
+
+            var stages = Translate(collection, queryable);
+            AssertStages(stages, "{ $match : { _t : 'Snake' } }");
+
+            var results = queryable.ToList();
+            results.Select(x => x.Id).Should().Equal(3);
+        }
+
+        [Fact]
```
rstam, line 115: Added a whole bunch of new tests, some of which required implementing support for missing features.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Also not sure about interfaces. Maybe we do want to store known discriminators for interface class maps?

papafe: Regarding this point, this ticket here seems to be possibly related: https://jira.mongodb.org/browse/CSHARP-1907

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
```diff
@@ -329,7 +329,7 @@ public void TestSerializeGAsObject()
         {
             G g = new G { P = "x" };
             var json = g.ToJson<object>();
-            var expected = ("{ '_t' : ['D~', 'G~'], 'P' : 'x' }").Replace("'", "\"");
+            var expected = ("{ '_t' : 'G~', 'P' : 'x' }").Replace("'", "\"");
```
papafe, line 332: I suppose this means that this will be a breaking change, won't it?

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Yes it does seem related.

aradalvand:
This literally used to work before???

aradalvand:
Any timeframe for when this is likely to be released? Direly needed in 3.x since this used to work in 2.x.

rstam:
> Any timeframe for when this is likely to be released?

We plan to release the implementation of this new feature in 3.2. It is currently in code review and didn't quite make it into the 3.1 release.

>Direly needed in 3.x since this used to work in 2.x.

Actually, it didn't work in 2.x. At least not reliably. You can read a full explanation of why here:

https://jira.mongodb.org/browse/CSHARP-5356?focusedCommentId=6913293&focusedId=6913293&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6913293

What did work in 2.x was OfType/is queries when using `HierarchicalScalarDiscriminatorConvention`. This has not changed and continues to work.

What did not work in 2.x was OfType/is queries when using `ScalarDiscriminatorConvention`. Depending on the scenario it might or might not work. In 3.x we made the decision that throwing an exception was preferable to silently returning incorrect results.

aradalvand:
> This has not changed and continues to work.

@rstam Actually, it doesn't seem to:

```cs
using MongoDB.Bson.Serialization.Serializers;
using MongoDB.Driver;

using var client = new MongoClient();
var db = client.GetDatabase("test");
var events = db.GetCollection<IEvent>("events");
var query = events.AsQueryable().OfType<Foo>();
Console.WriteLine(query.ToString());

var objectSerializer = new ObjectSerializer(type => true);
BsonSerializer.RegisterSerializer(objectSerializer);
var convention = new HierarchicalDiscriminatorConvention("_t");
BsonSerializer.RegisterDiscriminatorConvention(typeof(Foo), convention);
BsonClassMap.RegisterClassMap<Foo>(doc =>
{
    doc.AutoMap();
    doc.SetDiscriminatorConvention(convention);
    doc.SetIsRootClass(true);
});
BsonSerializer.RegisterDiscriminatorConvention(typeof(Bar), convention);
BsonClassMap.RegisterClassMap<Bar>(doc =>
{
    doc.AutoMap();
    doc.SetDiscriminatorConvention(convention);
    doc.SetIsRootClass(true);
});

public interface IEvent;

public record Foo(int X) : IEvent;
public record Bar(int Y) : IEvent;
```
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <RootNamespace>mongodb_driver_test</RootNamespace>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="MongoDB.Driver" Version="3.1.0" />
  </ItemGroup>

</Project>
```

This throws:
```
Expression not supported: test.events.Aggregate([]).OfType() because OfType is not supported with the configured discriminator convention.
```

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
papafe: Is this something for which we should/can throw?

rstam:
@aradalvand I think you are running into a completely different type of problem in your latest example, which is that the document type of your collection is an `interface` and not a class, and I think we have some issues with that.

For one, it doesn't matter what discriminator convention you register at a lower level (such as `Foo` or `Bar`), when `OfType` is translated it uses the discriminator convention it finds for `IEvent`, which won't be a `HierarchicalDiscriminatorConvention`.

Also, I don't know what it means, if anything, to have TWO roots, as you have configured both `Foo` and `Bar` to be roots. 

While your example is interesting and we would like to find a way to make it work, I don't think the work we are doing in this ticket will address your latest example because the work in this ticket is about the scenario where the collection document type is a class (and not an interface).

rstam:
@aradalvand I can get your latest example to pass with the latest version of the driver if I also add:

    BsonSerializer.RegisterDiscriminatorConvention(typeof(IEvent), convention);

to register the hierarchical discriminator convention for the interface also.

I'm still not entirely sure whether there will be any issues stemming from having TWO root classes, since we have always assumed there would be a single root class. Nevertheless, your example works if I add that discriminator convention registration for the interface.

aradalvand:
@rstam Still throws the same exception:
```cs
using MongoDB.Bson.Serialization.Serializers;
using MongoDB.Driver;

using var client = new MongoClient();
var db = client.GetDatabase("test");
var events = db.GetCollection<IEvent>("events");
var query = events.AsQueryable().OfType<Foo>();
Console.WriteLine(query.ToString());

var objectSerializer = new ObjectSerializer(type => true);
BsonSerializer.RegisterSerializer(objectSerializer);
var convention = new HierarchicalDiscriminatorConvention("_t");
BsonSerializer.RegisterDiscriminatorConvention(typeof(Foo), convention);
BsonClassMap.RegisterClassMap<Foo>(doc =>
{
    doc.AutoMap();
    doc.SetDiscriminatorConvention(convention);
    doc.SetIsRootClass(true);
});
BsonSerializer.RegisterDiscriminatorConvention(typeof(Bar), convention);
BsonClassMap.RegisterClassMap<Bar>(doc =>
{
    doc.AutoMap();
    doc.SetDiscriminatorConvention(convention);
    doc.SetIsRootClass(true);
});
// ADDED THIS:
BsonSerializer.RegisterDiscriminatorConvention(typeof(IEvent), convention);

public interface IEvent;

public record Foo(int X) : IEvent;
public record Bar(int Y) : IEvent;
```

aradalvand:
@rstam Ah, thanks. You're right. I wish this stuff was better documented though, really hard to find out this type of thing intuitively.

Would you want me to create an issue for the interface thing?

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public class BsonClassMapScalarDiscriminatorConvention : IScalarDiscriminatorConvention
```
papafe, line 29: Should this be deriving from `StandardDiscriminatorConvention`?

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+                if (type.IsAssignableFrom(discriminatedType))
+                {
+                    discriminators.Add(discriminator);
+                }
+            }
+
+            return discriminators.OrderBy(x => x).ToArray();
+        }
+
+        private BsonValue ReadDiscriminator(IBsonReader bsonReader)
```
papafe, line 142: If we make this class derive from `StandardDiscriminatorConvention`, this is something that could be put on the base class, given that this is almost the same that is done inside `StandardDiscriminatorConvention.GetActualType`.

papafe (commented):
Overall it makes sense for me, just added some comments.
Also, it seems that we use the current `ScalarDicriminatorConvention` as the discriminator convention of `EnumerableSerializerBase`. Do we need to worry about the new discriminator convention there too?

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public BsonValue GetDiscriminator(Type nominalType, Type actualType)
+        {
+            if (actualType == nominalType && !_classMap.DiscriminatorIsRequired)
+            {
+                return null;
+            }
+
+            if (_classMap.TypeToDiscriminatorMap.TryGetValue(actualType, out BsonValue discriminator))
```
BorisDog, line 107: Does it make sense to consider preserving the `ScalarDiscriminatorConvention` implementation here?  

```
var classMap = BsonClassMap.LookupClassMap(actualType);
return classMap.Discriminator;
```

rstam: This new discriminator convention depends on the `TypeToDiscriminatorMap` being initialized correctly, so we only want to return a value IF and only IF we find it in the map.

Also, we need to remove any reference to class maps in `ScalarDiscriminatorConvention`.

Our entire implementation of polymorphism has been wonky in the sense that we assumed that only class mapped serializers were involved. I suppose we might have to invest some more time in this PR in figuring out what happens now with classes that are NOT class mapped.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                var discriminatedType = entry.Key;
+                var discriminator = entry.Value;
+                if (type.IsAssignableFrom(discriminatedType))
+                {
+                    discriminators.Add(discriminator);
+                }
+            }
+
+            return discriminators.OrderBy(x => x).ToArray();
```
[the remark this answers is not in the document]
rstam: Good point. I was just thinking we wanted a well defined order instead of a random order.

Order doesn't actually matter for the queries this will be used in.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: Could be. I hadn't considered that. Will look into it.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: That sounds reasonable. I will look into refactoring `StandardDiscriminatorConvnetion` to factor out the code that reads the discriminator value.

rstam (commented):
I haven't re-requested your review yet because I still need to look into Ferdinando's suggestion to derive `BsonClassMapScalarDiscriminatorConvention` from `StandardDiscriminatorConvention` and reuse the code for reading the discriminator value.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: This was a good suggestion. This class now derives from `StandardDiscriminatorConvention`.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: Great suggestion. Done.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -96,35 +96,27 @@ obj is StandardDiscriminatorConvention other &&
...
-                        if (discriminator.IsBsonArray)
-                        {
-                            discriminator = discriminator.AsBsonArray.Last(); // last item is leaf class discriminator
-                        }
-                        actualType = BsonSerializer.LookupActualType(nominalType, discriminator);
-                    }
-                    bsonReader.ReturnToBookmark(bookmark);
-                    return actualType;
-                }
+            var discriminator = ReadDiscriminator(bsonReader);
```
rstam, line 108: The code to *read* the discriminator value has been factored out to a new method.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -137,5 +129,47 @@ public Type GetActualType(IBsonReader bsonReader, Type nominalType)
...
         public override int GetHashCode() => 0;
+
+        // protected methods
+        /// <summary>
+        /// Gets the actual type.
+        /// </summary>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <param name="discriminator">The discriminator.</param>
+        /// <returns>The actual type.</returns>
+        protected virtual Type GetActualType(Type nominalType, BsonValue discriminator)
```
rstam, line 140: This new method is virtual so that `BsonClassMapScalarDiscriminatorConvention` can override it.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -137,5 +129,47 @@ public Type GetActualType(IBsonReader bsonReader, Type nominalType)
...
+        {
+            return BsonSerializer.LookupActualType(nominalType, discriminator);
+        }
+
+        /// <summary>
+        /// Reads the discriminator.
+        /// </summary>
+        /// <param name="bsonReader">The bsonReader.</param>
+        /// <returns>The discriminator, or null if no discriminator was found.</returns>
+        protected BsonValue ReadDiscriminator(IBsonReader bsonReader)
```
rstam, line 150: Moved up from `BsonClassMapScalarDiscriminatorConvention` to the base class where it can be reused.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using MongoDB.Bson.Serialization.Conventions;
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// A scalar discriminator convention for class mapped types.
+    /// </summary>
+    public class BsonClassMapScalarDiscriminatorConvention : StandardDiscriminatorConvention, IScalarDiscriminatorConvention
```
rstam, line 27: Now derives from `StandardDiscriminatorConvention`.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,122 @@
...
+            _classMap = classMap ?? throw new ArgumentNullException(nameof(classMap));
+        }
+
+        /// <summary>
+        /// Gets the actual type.
+        /// </summary>
+        /// <param name="nominalType">The nominal type.</param>
+        /// <param name="discriminator">The discriminator value.</param>
+        /// <returns>The actual type.</returns>
+        protected override Type GetActualType(Type nominalType, BsonValue discriminator)
```
rstam, line 56: `BsonClassMapScalarDiscriminatorConvention` overrides this method to use the information it has about derived types.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1140,6 +1153,36 @@ public void SetExtraElementsMember(BsonMemberMap memberMap)
             _extraElementsMemberMap = memberMap;
         }
 
+        internal void AddKnownDiscriminator(BsonValue discriminator, Type type)
+        {
+            if (!_classType.IsAssignableFrom(type))
+            {
+                throw new ArgumentException($"Type {type} is not assignable to {_classType}.", nameof(type));
```
damieng, line 1160: Nit: Missing quotes round type name in order to be consistent with other exceptions added.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (nominalType == typeof(object) && discriminator.IsString)
+            {
+                actualType = TypeNameDiscriminator.GetActualType(discriminator.AsString);
+                if (actualType != null)
+                {
+                    return actualType;
+                }
+            }
+
+            throw new BsonSerializationException($"No type found for discriminator value: {discriminator}.");
```
damieng, line 72: Probably should quote this too.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                return null;
+            }
+
+            if (_classMap.TypeToDiscriminatorMap.TryGetValue(actualType, out BsonValue discriminator))
+            {
+                return discriminator;
+            }
+
+            throw new BsonSerializationException($"No discriminator value found for type: {actualType}.");
```
damieng, line 93: Quotes.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
damieng: Are we sure the list won't grow/be further discovered after being cached?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/OfTypeMethodToAggregationExpressionTranslator.cs
damieng: Is this support of OfType against the root object's type or do we support discriminators at every object in the document?

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
damieng: Do we want to detect that and throw in case some customers moving from 2.x incorrectly configured their attributes?

damieng (approved):
There are a few comments mostly around quotes in exceptions plus whether we want to throw if we detect the RootClass attribute misconfigured and one question about caching but none are blockers.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1331,31 +1374,37 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
+                {
+                    return BsonSerializer.LookupDiscriminatorConvention(_classType);
+                }
+
+                if (_isRootClass)
+                {
+                    return StandardDiscriminatorConvention.Hierarchical;
+                }
+
+                if (_baseClassMap.ClassType == typeof(object))
```
papafe, line 1392: Does this mean we're at the base class of the hierarchy?

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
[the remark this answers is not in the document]
papafe: And would it be a problem if the order was random?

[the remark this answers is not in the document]
papafe: I think those are good point to keep the order, thanks for clarifying.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
rstam: Done.

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: Discriminator value is a `BsonValue`. If the value is a string it will have quotes around it if not it won't.

I want to avoid the following:

```No type found for discriminator value: ""abc"".```

src/MongoDB.Bson/Serialization/BsonClassMapScalarDiscriminatorConvention.cs
rstam: Done. But I'm actually not sure when to use quotes and when not to...

rstam:
A completely different approach is in review here:

https://github.com/mongodb/mongo-csharp-driver/pull/1607

rstam:
This PR replaced by:

https://github.com/mongodb/mongo-csharp-driver/pull/1607

PR #1561, opened 2024-12-03
CSHARP-5405: Remove the BSON document size validation requirement for…

sanych-sun:
… the client bulk write operation


adelinowona (approved):
LGTM

PR #1565, opened 2024-12-04
CSHARP-4926: Support token field type with Equals operators

adelinowona:


src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -122,7 +122,7 @@ public SearchDefinition<TDocument> Equals<TField>(
             FieldDefinition<TDocument, TField> path,
             TField value,
             SearchScoreDefinition<TDocument> score = null)
-            where TField : struct, IComparable<TField> =>
+            where TField : IComparable<TField> =>
```
adelinowona, line 125: Considering removing these constraints.

papafe: If they don't impact anything, I'd be ok with removing them

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -139,8 +139,26 @@ public SearchDefinition<TDocument> Equals<TField>(
             Expression<Func<TDocument, TField>> path,
             TField value,
             SearchScoreDefinition<TDocument> score = null)
-            where TField : struct, IComparable<TField> =>
+            where TField : IComparable<TField> =>
                 Equals(new ExpressionFieldDefinition<TDocument, TField>(path), value, score);
+        
+        /// <summary>
+        /// Creates a search definition that queries for documents where an indexed field is equal
```
papafe, line 146: Maybe we should specify that in this case it's a collection field?

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -261,6 +261,16 @@ public void EmbeddedDocument_typed()
...
         }
+        
+        [Fact]
+        public void Equals_with_array_should_render_supported_type()
+        {
+            var subjectTyped = CreateSubject<Person>();
+            
+            AssertRendered(
+                subjectTyped.Equals(p => p.Hobbies, "soccer"),
+                $"{{ equals: {{ path: 'hobbies', value: \"soccer\" }} }}");
```
papafe, line 272:
```suggestion
                """{{ equals: {{ path: 'hobbies', value: "soccer" }} }}""");
```
Small suggestion here. You don't need the `$` for interpolation, and you could use the raw string literal to avoid the need to escape the quotes

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
adelinowona: Definitely

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
adelinowona: Thanks, I think I can just remove the dollar sign and double brackets and use single quotes.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -138,9 +137,25 @@ public SearchDefinition<TDocument> Equals<TField>(
...
+        /// Creates a search definition that queries for documents where at least one element in an indexed array field is equal
+        /// to the specified value.
+        /// Supported value types are boolean, numeric, ObjectId, date and string.
+        /// </summary>
+        /// <typeparam name="TField">The type of elements contained in the indexed array field.</typeparam>
+        /// <param name="path">The indexed array field to search.</param>
+        /// <param name="value">The value to query for.</param>
+        /// <param name="score">The score modifier.</param>
+        /// <returns>An equality search definition.</returns>
+        public SearchDefinition<TDocument> Equals<TField>(
```
BorisDog, line 153: We also  probably need ` FieldDefinition<TDocument, Enumerable<TField>>` overload as well?

BorisDog (changes requested):

PR #1569, opened 2024-12-07
CSHARP-4872: Add support for Append in aggregate expressions.

rstam:


rstam:
This adds support for Append in expressions.

It does not add support for top-level IQueryable Append. I'm not sure if there is a good way to support top-level Append in MQL. Perhaps with a `$unionWith` stage...

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/AppendOrPrependMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,82 @@
...
+                if (elementExpression is ConstantExpression elementConstantExpression)
+                {
+                    var value = elementConstantExpression.Value;
+                    var serializedValue = SerializationHelper.SerializeValue(itemSerializer, value);
+                    elementTranslation = new AggregationExpression(elementExpression, AstExpression.Constant(serializedValue), itemSerializer);
+                }
+                else
+                {
+                    elementTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, elementExpression);
+                    if (!elementTranslation.Serializer.Equals(itemSerializer))
```
sanych-sun, line 65: Should we check for compatible serializers instead? For example if collection is of `long` and appended item is `int`.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4872Tests.cs
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+using System.Linq;
+using FluentAssertions;
+using MongoDB.TestHelpers.XunitExtensions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp4872Tests : Linq3IntegrationTest
```
sanych-sun, line 23: Should I be able to Append `int` to collection of `long`s?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/AppendOrPrependMethodToAggregationExpressionTranslator.cs
rstam: That should never happen. By the time we get here both serializers will always have the exact same `ValueType`.

The only question is whether they are serialized the same way.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4872Tests.cs
rstam: Yes, but...

The compiler will insert a `Convert` from `int` to `long` so in the end you are actually appending a `long`.

sanych-sun (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1574, opened 2024-12-12
CSHARP-2096: Make EnumRepresentationConvention also affect collections of Enums

papafe:


src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -33,76 +34,91 @@ public class EnumRepresentationConvention : ConventionBase, IMemberMapConvention
...
+
+            if (serializer.ValueType.IsEnum && serializer is IRepresentationConfigurable representationConfigurable)
+            {
+                return representationConfigurable.WithRepresentation(_representation);
+            }
+
+            return null;
+        }
+
+        // private methods
```
rstam, line 105: Duplicate line

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -33,76 +34,91 @@ public class EnumRepresentationConvention : ConventionBase, IMemberMapConvention
...
         // private methods
         private bool IsNullableEnum(Type type)
         {
             return
                 type.GetTypeInfo().IsGenericType &&
                 type.GetGenericTypeDefinition() == typeof(Nullable<>) &&
-                Nullable.GetUnderlyingType(type).GetTypeInfo().IsEnum;
+                Nullable.GetUnderlyingType(type)!.GetTypeInfo().IsEnum;
```
rstam, line 112: There is no longer any need to call `GetTypeInfo` here (or on line 110). That was a .NET 2.0 thing.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: Removed

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: Removed

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -33,76 +34,91 @@ public class EnumRepresentationConvention : ConventionBase, IMemberMapConvention
...
 
         /// <summary>
         /// Gets the representation.
         /// </summary>
         public BsonType Representation => _representation;
 
+        /// <summary>
+        /// Gets a boolean indicating if this convention should be also applied to collections of enums.
+        /// </summary>
+        public bool ShouldApplyToCollections => _shouldApplyToCollections;
```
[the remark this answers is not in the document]
papafe: We have a discussion going on slack, but I think it would make sense to keep the name we have now, since we are special casing the `NullableSerializer` that is the only `IChildSerializerConfigurable` that is not a collection serializer, and makes this backwards compatible

src/MongoDB.Bson/Serialization/IKeyAndValueSerializerConfigurable.cs
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// Represents a serializer that has a key and a value serializer that configuration attributes can be forwarded to.
+    /// </summary>
+    public interface IKeyAndValueSerializerConfigurable : IBsonDictionarySerializer
```
papafe, line 21: Not sure if this should be deriving from `IBsonDictionarySerializer`, or it should be independent from it.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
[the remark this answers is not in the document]
papafe: No, but we can discuss it during the standup maybe. If we do it though, we need to special case the nullable serializer, otherwise it would be a breaking change.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
```diff
@@ -19,29 +19,40 @@ namespace MongoDB.Bson.Serialization
...
-        /// The reconfigured serializer, or <c>null</c> if no leaf serializer could be reconfigured.
-        /// </returns>
-        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure)
+        /// Reconfigures a serializer using the specified <paramref name="reconfigure"/> method if the result of <paramref name="testFunction"/> is true or the function is null.
+        /// If the serializer implements <see cref="IChildSerializerConfigurable"/> and either:
+        /// - is a collection serializer and <paramref name="shouldApplyToCollections"/> is true;
+        /// - or is a <see cref="Nullable"/> serializer;
+        /// the method traverses and applies the reconfiguration to its child serializers recursively.
+        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure,
+            Func<IBsonSerializer, bool> testFunction = null, bool shouldApplyToCollections = true)
```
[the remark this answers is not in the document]
papafe: I would prefer to keep `shouldApplyTo...`. If we do so, this method becomes a one-stop method for reconfiguring serializers, including when we want to apply `reconfigure` to nullable but not child serializers for example. If we keep it out, then we need to keep the code for nullable serializers out, leading to code duplication.

src/MongoDB.Bson/Serialization/IKeyAndValueSerializerConfigurable.cs
papafe: This is a good idea!

papafe:
@rstam I've done some changes:
- changed `IKeyAndValue...` to `IMultipleChildren...` so it has a broader use
- I've modified `Reconfigure` to work with `IChidSerializable..`, `IMultipleChildrenSerializable...` and Nullable serializer (always)
- I've changed the name to `TopLevelOnly`. Still open to changing the name, but at least I flipped all the booleans so it should not be a problem. 

I think the remaining contentious points are about the `Reconfigure` method: if it should have the `testFunction` parameter and if it should have the `topLevelOnly` boolean, but we can discuss in the threads.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
```diff
@@ -14,33 +14,45 @@
...
-        /// The reconfigured serializer, or <c>null</c> if no leaf serializer could be reconfigured.
-        /// </returns>
-        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure)
+        /// Reconfigures a serializer using the specified <paramref name="reconfigure"/> method if the result of <paramref name="testFunction"/> is true or the function is null.
+        /// If the serializer implements <see cref="IChildSerializerConfigurable"/> and either:
+        /// - <paramref name="topLevelOnly"/> is false;
+        /// - or is a <see cref="Nullable"/> serializer;
+        /// the method traverses and applies the reconfiguration to its child serializers recursively.
+        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure,
+            Func<IBsonSerializer, bool> testFunction = null, bool topLevelOnly = false)
```
rstam, line 29: This method has too many knobs. It has 4:

1. TSerializer generic type
2. reconfigure Func (which might return null)
3. testFunction
4. topLevelOnly

That's 3 too many.

All we need is the `reconfigure` `Func`, which centralizes in itself any decisions about whether to reconfigure or not. It either reconfigures the serializer if applicable or returns null.

As far as the rest are concerned:

`TSerializer` is redundant because the `reconfigure` `Func` can decide for itself

`testFunction` is redundant because the `reconfigure` `Func` can decide for itself

`topLevelOnly` is redundant because if you only want to configure the top level then don't call this function.

Here's what I think this method should look like:

    // Reconfigures a serializer recursively.
    // The reconfigure Func should return null if it does not apply to a given serializer
    internal static IBsonSerializer ReconfigureSerializerRecursively(
        IBsonSerializer serializer,
        Func<IBsonSerializer, IBsonSerializer> reconfigure)
    {
        switch (serializer)
        {
            // check IMultipleChildSerializersConfigurableSerializer first because some serializer implement both interfaces
            case IMultipleChildSerializersConfigurableSerializer multipleChildSerializersConfigurable:
                {
                    var newChildSerializers = new List<IBsonSerializer>();
                                                                          
                    foreach (var childSerializer in multipleChildSerializersConfigurable.ChildSerializers)
                    {
                        var reconfiguredChildSerializer = ReconfigureSerializerRecursively(childSerializer, reconfigure) ?? childSerializer;
                        newChildSerializers.Add(reconfiguredChildSerializer);
                    }
                     
                    return multipleChildSerializersConfigurable.WithChildSerializers(newChildSerializers.ToArray());
                }
                 
            case IChildSerializerConfigurable childSerializerConfigurable:
                {
                    var childSerializer = childSerializerConfigurable.ChildSerializer;
                    var reconfiguredChildSerializer = ReconfigureSerializerRecursively(childSerializer, reconfigure) ?? childSerializer;
                    return reconfiguredChildSerializer != null ? childSerializerConfigurable.WithChildSerializer(reconfiguredChildSerializer) : null;
                }                                                                                                                                    
                                                                                                                                                     
            default:                                                                                                                                 
                return reconfigure(serializer);                                                                                                      
        }                                                                                                                                            
    }

src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
```diff
@@ -61,7 +61,7 @@ public BsonDateOnlyOptionsAttribute(BsonType representation, DateOnlyDocumentFor
         /// <returns>A reconfigured serializer.</returns>
         protected override IBsonSerializer Apply(IBsonSerializer serializer)
         {
-            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer(serializer, (DateOnlySerializer s) => s.WithRepresentation(_representation, _documentFormat));
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializer(serializer, (DateOnlySerializer s) => s.WithRepresentation(_representation, _documentFormat), topLevelOnly: false);
```
rstam, line 64: With the other suggestions this method would look like this:

    protected override IBsonSerializer Apply(IBsonSerializer serializer)
    {
        var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializerRecursively(serializer, Reconfigure);
        return reconfiguredSerializer ?? base.Apply(serializer);
                                                                
        IBsonSerializer Reconfigure(IBsonSerializer serializer) =>
            (serializer is DateOnlySerializer dateOnlySerializer) ? dateOnlySerializer.WithRepresentation(_representation, _documentFormat) : null;
    }

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -33,76 +33,59 @@ public class EnumRepresentationConvention : ConventionBase, IMemberMapConvention
...
+
         /// <summary>
         /// Applies a modification to the member map.
         /// </summary>
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
-            var memberType = memberMap.MemberType;
-            var memberTypeInfo = memberType.GetTypeInfo();
+            var reconfiguredSerializer =
```
rstam, line 70: With the other suggestions this method would look like this:

    {
        var serializer = memberMap.GetSerializer();

        IBsonSerializer reconfiguredSerializer;    
        if (_topLevelOnly && !serializer.ValueType.IsNullableEnum())
        {
            reconfiguredSerializer = Reconfigure(serializer);
        }
        else
        {
            reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializerRecursively(serializer, Reconfigure);
        }                                                                                                             

        if (reconfiguredSerializer is not null)                                                                       
        {                                                                                                             
            memberMap.SetSerializer(reconfiguredSerializer);                                                          
        }                                                                                                             

        IBsonSerializer Reconfigure(IBsonSerializer serializer) =>                                                    
            (serializer.ValueType.IsEnum && serializer is IRepresentationConfigurable representationConfigurable) ?   
                representationConfigurable.WithRepresentation(_representation) : null;                                
    }

rstam: Note that we use `_topLevel` locally here to decide whether we want to reconfigure recursively or not (with an exception for nullable enums for backward compatibility).

There is no need to pass the `topLevel` argument to `ReconfigureSerializerRecursively`.

src/MongoDB.Bson/Serialization/IMultipleChildrenSerializerConfigurableSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public interface IMultipleChildrenSerializerConfigurableSerializer
+    {
+        /// <summary>
+        /// Gets the children serializers.
+        /// </summary>
+        /// <value>
+        /// The children serializers.
+        /// </value>
+        IBsonSerializer[] ChildrenSerializers { get; }
```
rstam, line 29: This property should be called `ChildSerializers`.

It is ungrammatical to say `ChildrenSerializers`.

`Child` is an adjective here and should not be plural.

Compare "child serializer" to "child seat".

The plural is "child seats", not "children seats".

src/MongoDB.Bson/Serialization/IMultipleChildrenSerializerConfigurableSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// Represents a serializer that has multiple children serializers that configuration attributes can be forwarded to.
+    /// </summary>
+    public interface IMultipleChildrenSerializerConfigurableSerializer
```
rstam, line 21: Class name (and file name) should be:

        MultipleChildSerializersConfigurableSerializer

"Child" to "Children"

See below.

src/MongoDB.Bson/Serialization/IMultipleChildrenSerializerConfigurableSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// The children serializers.
+        /// </value>
+        IBsonSerializer[] ChildrenSerializers { get; }
+
+        /// <summary>
+        /// Returns a serializer that has been reconfigured with the specified children serializers.
+        /// </summary>
+        /// <param name="childrenSerializers">The children serializers.</param>
+        /// <returns>The reconfigured serializer.</returns>
+        IBsonSerializer WithChildrenSerializers(IBsonSerializer[] childrenSerializers);
```
rstam, line 36: IBsonSerializer WithChildSerializers(IBsonSerializer[] childSerializers);

rstam (changes requested):
You can see my suggested changes prototyped here:

https://github.com/rstam/mongo-csharp-driver/tree/csharp2096-rstam

src/MongoDB.Bson/Serialization/IMultipleChildrenSerializerConfigurableSerializer.cs
papafe: Yeah, I was unsure about that, I'll correct it!

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: @rstam I think this the last point to agree upon.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
rstam: I still think this method is much simpler and easier to understand if we only have 2 parameters instead of 5.

The only rule the reader needs to understand is that the reconfigure Func should either return a reconfigured serializer or null if it does not apply.

All the decision making about whether to reconfigure or not is centralized in one place (in the caller).

You are concerned about possible duplication of code, which is a valid concern. But if you look at my branch the amount of duplications is close to zero, and in return we only have to look in one place to understand whether and how the serializer is reconfigured.

I've slightly refactored my branch to make the calling code even simpler:

https://github.com/rstam/mongo-csharp-driver/tree/csharp2096-rstam

Look at all the callers of `ReconfigureSerializerRecursively` to verify how little duplication there actually is.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
```diff
@@ -14,33 +14,45 @@
...
-        /// <returns>
-        /// The reconfigured serializer, or <c>null</c> if no leaf serializer could be reconfigured.
-        /// </returns>
-        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure)
+        /// Reconfigures a serializer using the specified <paramref name="reconfigure"/> method if the result of <paramref name="testFunction"/> is true or the function is null.
+        /// If the serializer implements <see cref="IChildSerializerConfigurable"/> and either:
+        /// - <paramref name="topLevelOnly"/> is false;
+        /// - or is a <see cref="Nullable"/> serializer;
+        /// the method traverses and applies the reconfiguration to its child serializers recursively.
+        internal static IBsonSerializer ReconfigureSerializer<TSerializer>(IBsonSerializer serializer, Func<TSerializer, IBsonSerializer> reconfigure,
```
rstam, line 28: I think the method should be called `ReconfigureSerializerRecursively` to call attention to the recursive nature of the reconfiguration.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
rstam: Please look at my `ReconfigureSerializerRecursively` implementation carefully, in particular the new `anyChildSerializerWasReconfigured` variable to return `null` if no child was reconfigured.

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
```diff
@@ -14,33 +14,45 @@
...
+                        var reconfiguredChildSerializer = ReconfigureSerializer(childSerializer, reconfigure, testFunction,
+                            false);
+
+                        newSerializers.Add(reconfiguredChildSerializer ?? childSerializer);
+                    }
+
+                    return multipleChildSerializerConfigurable.WithChildSerializers(newSerializers.ToArray());
+                }
+                case IChildSerializerConfigurable childSerializerConfigurable when
+                    !topLevelOnly || Nullable.GetUnderlyingType(serializer.ValueType) != null:
```
rstam, line 50: A nullable enum test doesn't belong here. It only applies to the `EnumRepresentationConvention`.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: See comments below.

rstam (changes requested):

src/MongoDB.Bson/Serialization/SerializerConfigurator.cs
papafe: I think I got convinced 😄 
Probably it's better a little code duplication to make the code clearer.

papafe: Just pushed the changes.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -33,76 +33,62 @@ public class EnumRepresentationConvention : ConventionBase, IMemberMapConvention
...
         /// <summary>
         /// Applies a modification to the member map.
         /// </summary>
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
-            var memberType = memberMap.MemberType;
-            var memberTypeInfo = memberType.GetTypeInfo();
+            var serializer = memberMap.GetSerializer();
+            var reconfiguredSerializer = _topLevelOnly && Nullable.GetUnderlyingType(serializer.ValueType) == null ?
```
rstam, line 71: Why did you not like:

```!serializer.ValueType.IsNullableEnum()```

instead of:

```Nullable.GetUnderlyingType(serializer.ValueType) == null```

The second is longer and it seems suspicious to call `Nullable.GetUnderlyingType` without even knowing if the `ValueType` is nullable or not.

Yes, after some research I now know `Nullable.GetUnderlyingType` returns `null` if the `ValueType` is not `Nullable` but still seems wrong to call a `Nullable` method when we don't know if the `ValueType` is `Nullable` or not. I'm surprised `Nullable.GetUnderlyingType` doesn't throw an exception when `ValueType` is not `Nullable`. I had to research the documentation to figure out what it does in that case.

Calling `IsNullableEnum` is not only shorter but it's also self-describing, without having to decipher the obscure behavior of `GetUnderlyingType` for non `Nullable` types.

rstam: Note: `IsNullableEnum` is a new extension method added to `TypeExtensions.cs` on the branch I shared with you.

rstam (changes requested):
Looks good.

I have one question for you before I LGTM. Let me know what you think.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
rstam: Also... isn't the exception for "nullable enums", not for ALL nullable types?

Better to be precise.

papafe: I agree it's better to have it as an extension method, so the intention is clearer. Just pushed it

rstam (approved):
LGTM

PR #1582, opened 2024-12-31
CSHARP-5450: Verify that all calls to Translate or TranslateEnumerable are correct.

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/AsQueryableMethodToAggregationExpressionTranslator.cs
```diff
@@ -32,11 +32,12 @@ public static AggregationExpression Translate(TranslationContext context, Method
             {
                 var sourceExpression = arguments[0];
                 var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
+                NestedAsQueryableHelper.EnsureQueryableMethodHasNestedAsQueryableSource(expression, sourceTranslation);
```
rstam, line 35: @damieng another example of the complications caused by nested AsQueryable in the driver.

rstam:
I'm looking into why the patch build failed.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/AsQueryableMethodToAggregationExpressionTranslator.cs
rstam: Oops... this is the one place where we do *not* want to call `EnsureQueryableMethodHasNestedAsQueryableSource`

rstam:
I'm also going to rebase on `main` to resolve the merge conflicts.

rstam:
> I'm looking into why the patch build failed.

I fixed the failing tests.

There are still some open questions:

1. Can the `EnsureQueryableMethodHasNestedAsQueryableSource` tests be moved to `TranslateEnumerable` so we don't forget to do them

2. What about the existing calls that are *not* followed by a call to `EnsureQueryableMethodHasNestedAsQueryableSource`? They aren't method translators so calling `EnsureQueryableMethodHasNestedAsQueryableSource` would currently be a no-op, but perhaps there is something else that would trigger the check?

sanych-sun (approved):
LGTM

adelinowona:
> 1. Can the `EnsureQueryableMethodHasNestedAsQueryableSource` tests be moved to `TranslateEnumerable` so we don't forget to do them

Do all calls to `TranslateEnumerable` entail that where it was called has to do the `EnsureQueryableMethodHasNestedAsQueryableSource` check? If so, then I think it is ok move the tests into that method.

rstam:
> Do all calls to TranslateEnumerable entail that where it was called has to do the EnsureQueryableMethodHasNestedAsQueryableSource check? If so, then I think it is ok move the tests into that method.

No, not all calls to `TranslateEnumerable` are followed by a call to `EnsureQueryableMethodHasNestedAsQueryableSource `. But I'm not sure if they should be...

PR #1583, opened 2024-12-31
CSHARP-5442: Fixed use of serializer by Atlas Search operators

papafe:
[JIRA ticket](https://jira.mongodb.org/browse/CSHARP-5442)


papafe:
@BorisDog I've added you as a reviewer since it seems you were the one that worked on search, mostly. 
A couple of notes:
- Probably we need to change the `in` and `range` search definitions as well to use the correct serializers.
- This is effectively a behaviour breaking change, so we need to decide if we want include it in next version or wait for next major

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +119,61 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
-                Guid v => new BsonBinaryData(v, GuidRepresentation.Standard),
-                string v => (BsonString)v,
-                null => BsonNull.Value,
-                _ => throw new InvalidCastException()
-            };
+                //var fieldRenderArgs = args with { PathRenderArgs = args.PathRenderArgs with { AllowScalarValueForArray = true } };
+
+                var renderedField = _arrayField.Render(args);
+
+                var serializer =
```
papafe, line 148: Incredibly ugly, but was just testing if it worked

papafe: @rstam this is what I was referring to. 
Do you think it would be worth adding a new property to `RenderArgs` so that we can get directly the serializer we need?

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +119,88 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
 
     internal sealed class EqualsSearchDefinition<TDocument, TField> : OperatorSearchDefinition<TDocument>
     {
-        private readonly BsonValue _value;
+        private readonly TField _value;
```
rstam, line 122: Rename `TField` to `TValue` because that is what it is.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +119,88 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
+            typeof(DateTimeOffset),
+            typeof(ObjectId),
+            typeof(Guid),
+            typeof(string)
+        ];
+
+        public EqualsSearchDefinition(FieldDefinition<TDocument, TField> path, TField value, SearchScoreDefinition<TDocument> score)
+            : base(OperatorType.Equals, new SingleSearchPathDefinition<TDocument>(path), score)
+        {
+            ValidateType();
```
rstam, line 148: I question whether we need to have `ValidateType` at all. Looks to me like this was just a client-side limitation related to what .NET types `ToBsonValue` knew how to convert to BSON, but now that we are using the proper serializer this limitation should no longer exist.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
rstam: I think this is on the right track but can use a little tweaking to how the right serializer is found and to reduce duplication between how scalar and array fields are handled.

Something like this:

    private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args)
    {
        IBsonSerializer<TValue> valueSerializer;
        if (_field != null)
        {
            var renderedField = _field.Render(args);
            valueSerializer = renderedField.FieldSerializer;
        }
        else
        {
            var renderedArrayField = _arrayField.Render(args);
            valueSerializer = (IBsonSerializer<TValue>)ArraySerializerHelper.GetItemSerializer(renderedArrayField.FieldSerializer);
        }                                                                                                                          
                                                                                                                                   
        var document = new BsonDocument();                                                                                         
        using var bsonWriter = new BsonDocumentWriter(document);                                                                   
        var context = BsonSerializationContext.CreateRoot(bsonWriter);                                                             
        bsonWriter.WriteStartDocument();                                                                                           
        bsonWriter.WriteName("value");                                                                                             
        valueSerializer.Serialize(context, _value);                                                                                
        bsonWriter.WriteEndDocument();                                                                                             
                                                                                                                                   
        return document;                                                                                                           
    }

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -317,6 +317,8 @@ internal class IEnumerableSerializer<TItem> : SerializerBase<IEnumerable<TItem>>
         {
             private readonly IBsonSerializer<TItem> _itemSerializer;
 
+            public IBsonSerializer<TItem> ItemSerializer => _itemSerializer;
+
```
rstam, line 321: Instead of adding this new public property have the class implement `IBsonArraySerializer` and implement the `TryGetItemSerializationIinfo` method:

    public bool TryGetItemSerializationInfo(out BsonSerializationInfo serialization
    {
        serializationInfo = new(elementName: null, _itemSerializer, typeof(TItem));
        return true;                                                               
    }                                                                              

All serializers for array-like values should implement `IBsonArraySerializer`. I don't know why this class did not already implement it.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +119,88 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
-            new("value", _value);
+        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args)
+        {
+            if (_field is null)
+            {
+                var renderedField = _arrayField.Render(args);
+
+                var serializer =
+                    (renderedField.ValueSerializer as FieldValueSerializerHelper.IEnumerableSerializer<TField>)
```
rstam, line 168: Instead of trying to cast to the very specific class `FieldValueSerializerHelper.IEnumerableSerializer<TField>` use the more general from (used everywhere else) to get the item serializer from an array-like serializer:

    valueSerializer = (IBsonSerializer<TValue>)ArraySerializerHelper.GetItemSerializer(renderedArrayField.FieldSerializer);

This is why `FieldValueSerializerHelper.IEnumerableSerializer<TItem>` should implement `IBsonArraySerializer` (as all array-like serializers should). Not sure why it didn't already...

rstam (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
rstam: I checked with Boris and he confirmed that it is a server limitation which types are supported.

But even so, we probably shouldn't be checking this client side. We should just send the value to the server and let it return an error if it doesn't support the data type.

That way our driver will work well with all supported versions of the server, which might have different limitations. And if future server versions support new value types we won't have to make any changes.

rstam (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
papafe: I also like this idea. I tried to keep it consistent with what was there before, but I agree that moving this check server side would allow us to be more free in what we do and don't support over time

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
papafe: Perfect, this solves my doubts. That makes sense, I was mostly focusing on how to get the correct serializer before trying to refactor. 
Regarding `IBsonArraySerializer`, it seems that the majority of times when we implement that interface we also implement `IChildSerializerConfigurable`, and that makes sense. It also seems that for this use case the result we'd get would be the same, so what do you think about using `IChildSerializerConfigurable` instead? 
To be honest this is more of a generic question, I'm curious about if there's any preference of one over the other, of course considering that `TryGetItemSerializationInfo` can potentially return more info (not in this case though).

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -290,6 +290,8 @@ public void Equals_should_render_supported_type<T>(
                 subject.Equals("x", value),
                 $"{{ equals: {{ path: 'x', value: {valueRendered} }} }}");
 
+            //For the Guid we get (correctly): GuidSerializer cannot serialize a Guid when GuidRepresentation is Unspecified.
```
papafe, line 293: @rstam This test here is now failing because now we're using the default serializer and that doesn't have the `GuidRepresentation` specified. 
Do you have any suggestion on how to fix this? The exception makes sense, and normally we'd probably suggest statically registering a serializer, but this would impact all the other tests here. 
Should we ignore the `Guid` field for this test and make a separable test here that can only be run manually?

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
rstam: `IBsonArraySerializer` and `IChildSerializerConfigurable` are orthogonal. We can't just implement `IChildSerializerConfigurable` *instead* of `IBsonArraySerializer`.

In this case we *must* implement `IBsonArraySerializer`. 

It is true that in general whenever implementing `IBsonArraySerializer` we would likely also want to implement `IChildSerializerConfigurable`. However, that applies mostly to `public` serializers that a user might want to configure. This is an `internal` serializer, so I think implementing `IBsonArraySerializer` is sufficient.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
papafe: Yeah, I would have loved to avoid making special cases, but it seems we need to. I'll make some additional tests that can be run manually.

rstam: I'm fine with just skipping this test for Guids, because:

1. This test is against `IMongoCollection<BsonDocument>` and that is probably not used often
2. Anyone using `BsonDocument` should either not be using `Guid` or should take control of the conversion themselves

A manual test would involve registering a properly configured global `GuidSerializer`, but we already know that works (in part because the strongly typed version uses a properly configured `GuidSerializer`, though it's true it doesn't get it from the global registry).

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -141,6 +141,7 @@ public static IBsonSerializer GetSerializerForValueType(IBsonSerializer fieldSer
             return ConvertIfPossibleSerializer.Create(valueType, fieldType, fieldSerializer, serializerRegistry);
         }
 
+        //TODO Never used
```
papafe, line 144: Should we remove this?

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
papafe: `TField` is the type of the field, that is also the type of the value, right? 
Also if we do it here we'd need to make this change also in the other operators, and probably should propagate it as well.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -519,9 +551,8 @@ public void In<T>(T[] fieldValues, string[] fieldsRendered)
              new object[] { new decimal[] { 1.5m, 2.5m }, new[] { "NumberDecimal(\"1.5\")", "NumberDecimal(\"2.5\")" } },
              new object[] { new[] { "str1", "str2" }, new[] { "'str1'", "'str2'" } },
              new object[] { new[] { DateTime.MinValue, DateTime.MaxValue }, new[] { "ISODate(\"0001-01-01T00:00:00Z\")", "ISODate(\"9999-12-31T23:59:59.999Z\")" } },
-             new object[] { new[] { DateTimeOffset.MinValue, DateTimeOffset.MaxValue }, new[] { "ISODate(\"0001-01-01T00:00:00Z\")", "ISODate(\"9999-12-31T23:59:59.999Z\")" } },
+             new object[] { new[] { DateTimeOffset.MinValue, DateTimeOffset.MaxValue }, new[] { """{ "DateTime" : { "$date" : { "$numberLong" : "-62135596800000" } }, "Ticks" : 0, "Offset" : 0 } """, """ { "DateTime" : { "$date" : "9999-12-31T23:59:59.999Z" }, "Ticks" : 3155378975999999999, "Offset" : 0 } """ } },
              new object[] { new[] { ObjectId.Empty, ObjectId.Parse("4d0ce088e447ad08b4721a37") }, new[] { "{ $oid: '000000000000000000000000' }", "{ $oid: '4d0ce088e447ad08b4721a37' }" } },
-             new object[] { new object[] { (byte)1, (short)2, (int)3 }, new[] { "1", "2", "3" } }
```
papafe, line 524: Do we want to support this? This does not work anymore because before we were using the serializer depending on the type of the value, while now we're trying to look at the serializer appropriate for the field.

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -123,9 +123,9 @@ private protected enum OperatorType
             QueryString,
             Range,
             Regex,
-            Search,
+            Search,  //TODO This is never used, should we remove them?
```
papafe, line 126: Should we/can we remove those?

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -119,7 +119,7 @@ public SearchDefinition<TDocument> EmbeddedDocument<TField>(
         /// <param name="score">The score modifier.</param>
         /// <returns>An equality search definition.</returns>
         public SearchDefinition<TDocument> Equals<TField>(
-            FieldDefinition<TDocument, TField> path,
+            FieldDefinition<TDocument, TField> path,  //TODO We should try to uniform this to the In operator for next major (this should be SearchPathDefinition<TDocument>)
```
papafe, line 122: I can create a ticket for this if we want to go this way

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
rstam: `TField` and `TValue` are only the same when the field is not an array.

In the case of an array `TField` is `IEnumerable<TValue>`.

rstam: Did you remove the code you had earlier for `_arrayField`?

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -322,6 +323,12 @@ public IEnumerableSerializer(IBsonSerializer<TItem> itemSerializer)
                 _itemSerializer = itemSerializer;
             }
 
+            public bool TryGetItemSerializationInfo(out BsonSerializationInfo serializationInfo)
```
rstam, line 326: Keep methods in alphabetical order?

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +121,35 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
-            new("value", _value);
-
-        private static BsonValue ToBsonValue(TField value) =>
-            value switch
+        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args)
+        {
+            var searchPathDefinition = (SingleSearchPathDefinition<TDocument>)_path;
+            var renderedField = searchPathDefinition.Field.Render(args);
+            var serializer = renderedField.FieldSerializer switch
```
rstam, line 136: Maybe rename `serializer` to `valueSerializer`.

To remove confusion about whether it is the field serializer or the value serializer (which are not the same in the case of array fields).

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +121,35 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
 
     internal sealed class EqualsSearchDefinition<TDocument, TField> : OperatorSearchDefinition<TDocument>
     {
-        private readonly BsonValue _value;
+        private readonly TField _value;
 
         public EqualsSearchDefinition(FieldDefinition<TDocument> path, TField value, SearchScoreDefinition<TDocument> score)
```
rstam, line 126: It is unfortunate that `path` is of type `FieldDefinition<TDocument>` instead of `FieldDefinition<TDocument, TField>`.

This is not type safe because it allows mismatches between the field type and the value type that are not caught at compile time.

In an earlier commit you had changed this but now it's removed. Can you explain why?

rstam (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
papafe: I tried to uniform the code between the `Equals`, `In` and `Range` search definition classes. 
Unfortunately for `In` and `Range` we can't have the same constructor with `FieldDefinition<TDocument, Field>` because of the current public API of the operators in `SearchDefinitionBuilder`. 
I agree that this is not type safe, but at least `EqualsSearchDefinition` is an internal class, and the operators in `SearchDefinitionBuilder` that use it have type safety (either through expressions or field definitions).

If this was a major version, I'd suggest to modify `SingleSearchPathDefinition` (the subclass of `SearchPathDefinition` for single fields) from using one generic parameter `TDocument` to two `TDocument, TField`, so we could be on the safer side. 
Probably it would be worth actually removing `SearchPathDefinition<TDocument>` as an input parameter of the operators, and create multiple overrides for the various subclasses of `SearchPathDefinition`, so that we can use as many generic parameters as necessary. 

I think this is something that would be worth doing for next major in order to improve the API, but I don't think we can do much at the moment without making breaking changes.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -588,34 +611,70 @@ public void In_should_throw_when_values_are_invalid()
...
         {
             var subjectTyped = CreateSubject<Person>();
-            
+
             AssertRendered(
                 subjectTyped.In(p => p.Hobbies, ["dance", "ski"]),
                 "{ in: { path: 'hobbies', value: ['dance', 'ski'] } }");
         }
 
+        [Fact]
```
papafe, line 667: According to the mongo docs `In` should support wildcard paths, but we had no tests for that.

rstam (approved):
I'm not too familiar with the Search API but as far as I can tell this looks OK.

You should get one more LGTM before merging to main though.

papafe:
> I'm not too familiar with the Search API but as far as I can tell this looks OK.
> 
> You should get one more LGTM before merging to main though.

@sanych-sun can you take a look too?

src/MongoDB.Driver/Search/SearchDefinition.cs
sanych-sun: Cannot find any reference to this or anything on Documentation. @BorisDog  can we remove this?

papafe:
@BorisDog another thing to consider is that this is a behaviour breaking change, so we need to consider if we want to merge this now or wait. I think it could also be seen as a bug fix to be honest, but we should take a decision. 
The difference is that now the correct registered serializers are used to serialize the values for the `equals`, `in` and `range` operators, while before default conversions were used, ignoring the registered serializers.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -118,38 +121,35 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
         }
 
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
-            new("value", _value);
-
-        private static BsonValue ToBsonValue(TField value) =>
-            value switch
+        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args)
+        {
+            var searchPathDefinition = (SingleSearchPathDefinition<TDocument>)_path;
```
BorisDog, line 134: It's a bit inefficient that we need to render the path twice.
Also it would be good to extract this to a single helper method.

What do you and @rstam  think about having new internal `Render` method in `SearchPathDefinition`, something like

`(BsonValue, RenderedFieldDefinition) Render(...)`

And passing field serializer in `OperatorSearchDefinition.RenderArguments() `?

BorisDog (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
papafe: I am a little confused, where are we rendering the path again apart from here?

papafe: I see your point. I also think it's a little complicated to extract a method like that to `SearchPathDefinition`, since `(BsonValue, RenderedFieldDefinition) Render(...)` would make sense for a `SingleSearchPathDefinition`, but not for `MultiSearchPathDefinition` or `WildCardSearchPathDefinition`.

BorisDog: Yeah I meant that the returned `RenderedFieldDefinition` would be null in some cases. I agree that it's not the greatest design, but I don't have any better suggestion as of now. I does have the upside of not rendering the path twice.
As long as all the API changes are internal, we can evolve them in the future if we come up with a better idea (the new suggested Render method should be interna).

papafe: In [60c1d57](https://github.com/mongodb/mongo-csharp-driver/pull/1583/commits/60c1d57c437759784c0df48a6888e8a8650261a6), I've added a fix to avoid rendering the path twice. It's not super nice, and I'm not sure it's worth it given that probably the network call of the atlas search would be much greater than the time needed to render the path twice.

papafe: I think that if we want to have special cases depending on the path we could also specialize `OperatorSearchDefinition.RenderArguments` and produce a different method for each path type, for example `OperatorSearchDefinition.RenderArgumentsForSingleSearchPath` where we can pass additional arguments. This would require quite some changes, but could also allow to control which kind of path can be used with certain kinds of operators.

papafe:
@BorisDog I've done some corrections following your suggestions, and it looks much better than before, thank you 😄 
I'm a little bit unsure about the names though, they feel a little redundant

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -286,14 +288,18 @@ public void Equals_should_render_supported_type<T>(
             var subject = CreateSubject<BsonDocument>();
             var subjectTyped = CreateSubject<Person>();
 
-            AssertRendered(
-                subject.Equals("x", value),
-                $"{{ equals: {{ path: 'x', value: {valueRendered} }} }}");
+            //There is no property called "x" where to pick up a properly configured GuidSerializer for the tests
```
BorisDog, line 291: I'm not sure I fully understand this exclusion. Is the default guid serializer different from the one used by serialization?

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -313,29 +319,55 @@ public void Equals_should_render_supported_type<T>(
             new object[] { (float)1, "1", Exp(p => p.Float), nameof(Person.Float) },
             new object[] { (double)1, "1", Exp(p => p.Double), nameof(Person.Double) },
             new object[] { DateTime.MinValue, "ISODate(\"0001-01-01T00:00:00Z\")", Exp(p => p.Birthday), "dob" },
-            new object[] { DateTimeOffset.MaxValue, "ISODate(\"9999-12-31T23:59:59.999Z\")", Exp(p => p.DateTimeOffset), nameof(Person.DateTimeOffset) },
+            new object[] { DateTimeOffset.MaxValue, """{ "DateTime" : { "$date" : "9999-12-31T23:59:59.999Z" }, "Ticks" : 3155378975999999999, "Offset" : 0 }""", Exp(p => p.DateTimeOffset), nameof(Person.DateTimeOffset) },
```
BorisDog, line 322: So does this mean that we are changing the defaults also for non typed queries (`SearchDefinitionBuilder<BsonDocument>`)?

If so, that might be considered a bug and a breaking change.

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
papafe: I think I need to improve the comment, I agree it's not super clear. 
The issue here is that when we use a string path, we're going to get the default serializer for the value, so `GuidSerializer`. In v3.0 we decided to have `GuidRepresentation.Unspecified` as the default Guid representation, and this will throw an exception when trying to serialize/deserialize. Normally a user would go over this by registering a default `GuidSerializer` with the proper representation (`BsonSerializer.Register...(GuidSerializer)`), but we can't do it because the registered serializers are global and this would be a problem for all the other tests.

For the typed path on line 305 it's not an issue since we get the default field serializer from the class, and the `Guid` property has the `BsonGuidRepresentation` attribute on it. 

This is more of a testing issue and should be solved when we can pass down the serialization domain 😁

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
papafe: Yes it will change also for non typed queries. They would need to convert the `DateTime` themselves if they want to keep the same behaviour.

src/MongoDB.Driver/Search/SearchPathDefinition.cs
```diff
@@ -104,11 +105,17 @@ public static implicit operator SearchPathDefinition<TDocument>(List<string> fie
         /// <param name="args">The render arguments.</param>
         /// <returns>The rendered field.</returns>
         protected string RenderField(FieldDefinition<TDocument> fieldDefinition, RenderArgs<TDocument> args)
+            => RenderFieldWithSerializer(fieldDefinition, args).renderedPath;
+
+        internal virtual (BsonValue, IBsonSerializer) RenderAndGetFieldSerializer(RenderArgs<TDocument> args) => (Render(args), null);
+
+        internal (string renderedPath, IBsonSerializer fieldSerializer) RenderFieldWithSerializer(FieldDefinition<TDocument> fieldDefinition, RenderArgs<TDocument> args)
```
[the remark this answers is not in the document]
papafe: Yes, I think it makes sense

papafe:
@BorisDog, as an additional note. I've tried making an untyped query (using `BsonDocument`) and passing a `DateTime` value for the `Equal` operator. No error returned but zero results as expected.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
papafe: It's probably a breaking change, but I'm not sure this can be considered a bug. If you're working directly with Bson documents then I assume that you would like to control how to the values are serialized in any case.

rstam (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -116,40 +122,36 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
 
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
-            new("value", _value);
-
-        private static BsonValue ToBsonValue(TField value) =>
-            value switch
+        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args,
+            IBsonSerializer fieldSerializer = null)
+        {
+            var valueSerializer = fieldSerializer switch
```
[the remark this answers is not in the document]
papafe: It's actually a little different between the various operators, so I'd prefer to keep it here

papafe:
@rstam @BorisDog I've followed Robert's suggestions about the code style.
I've also come up with an extension method to enable the new behaviour, so this won't be a breaking change. It's all in the last commit, so it can be reverted if necessary

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -54,6 +56,31 @@ public static implicit operator SearchDefinition<TDocument>(string json) =>
...
+        /// Sets the use of default serialization for the specified <see cref="SearchDefinition{TDocument}"/>.
+        /// When set to true, the default serializers will be used to serialize the values of certain Atlas Search operators, such as "Equals", "In" and "Range". This will become the default behaviour in version 4.0 of the library.
+        /// If not enabled, then a default conversion will be used.
+        /// </summary>
+        /// <typeparam name="TDocument">The type of the document.</typeparam>
+        /// <param name="searchDefinition">The search definition instance.</param>
+        /// <param name="useDefaultSerialization">Whether to use the default serialization or not.</param>
+        /// <returns>The same <see cref="SearchDefinition{TDocument}"/> instance with default serialization enabled.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if <paramref name="searchDefinition"/> is not of a valid type/>.</exception>
+        public static SearchDefinition<TDocument> WithDefaultSerialization<TDocument>(this SearchDefinition<TDocument> searchDefinition, bool useDefaultSerialization)
```
sanych-sun, line 74: The method name is a little confusing: why I as a user should ask driver to do something that is sounds as default. May be it should be named like `WithConfiguredSerializer` or something like that?

papafe: Yeah, the naming is confusing to be honest! I had some difficulty coming up with a reasonable name, but your suggestion makes sense.

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -135,6 +160,8 @@ private protected enum OperatorType
         protected readonly SearchPathDefinition<TDocument> _path;
         protected readonly SearchScoreDefinition<TDocument> _score;
 
+        protected bool _useConfiguredSerialization = false;
```
rstam, line 163: I'm not convinced this is a good idea.

Might as well call this `_useTheWrongSerializer`.

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinition.cs
papafe: You're not convinced about the naming or the actual API/solution?

rstam: I'm not convinced this `bool` should even exist.

I somewhat tongue in cheek suggested a name that illustrates why this makes no sense to me.

We have a bug: we are using the wrong serializer.

We are fixing the bug: we now use the correct serializer.

But somehow we think a user would like to use the wrong serializer???

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinition.cs
papafe: I understand your point, but I suppose that some developers could already use this feature as is, and this would be considered a breaking change (even though it was not correct).  The idea here would be to support this behaviour until next major and then get rid of it. 
@BorisDog am I right?

rstam: While this sounds like a valid concern, I can't figure out any scenario in which it would actually be valid.

In essence we are replacing are "hand rolled" serializer (the `ToBsonValue` method) with using the correct serializer.

There are only two scenarios:

1. ToBsonValue and the configured serializer return the same serialized values
2. Or they return different values

In the first case there is no problem.

In the second case the value returned by `ToBsonValue` is wrong and I don't see how the query could have worked, and therefore why we would keep the existing behavior.

Can you point me to a test that passes when `_useConfiguredSerialization` is `false` and fails when it is `true`?

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinition.cs
BorisDog: My concern is with pipelines using `BsonDocument` (unfortunately this happens more often than we'd like).
For example `DateTimeOffset` is serialized by default as document, while users with `BsonDocument` pipelines might be relying on it to be serialized as `BsonDate`.

I think in case we should introduce the "correct" behavior in the next major version.

rstam: @papafe can you point me to a test that passes when `_useConfiguredSerialization` is `false` and fails when it is `true`?

I want to confirm that this is not just a hypothetical concern, and if real, whether there isn't a better way to address than turning off the fix entirely by default.

papafe: @rstam sorry for the late reply. 
If you look at the tests `Equals_should_render_supported_type` (it's already in main) and `Equals_with_configured_serializer_should_render_supported_type`, they are essentially the same test but with different test data (`EqualsSupportedTypesTestData` and `EqualsWithConfiguredSerializerSupportedTypesTestData`). The only difference between the two sets of data is the line with `DateTimeOffset`, and that's because in the current implementation that is serialized as a `BsonDate` while the default `DateTimeOffsetSerializer` uses a document representation.

papafe: @rstam did you have the change to look at this?

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -219,51 +233,53 @@ public GeoWithinSearchDefinition(
...
-            var array = new BsonArray(values.Select(ToBsonValue));
-
-            var bsonType = array[0].GetType();
-            _values = Ensure.That(array, arr => arr.All(v => v.GetType() == bsonType), nameof(values), "All values must be of the same type.");
+            _values = values.ToArray();
         }
 
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
-            new("value", _values);
+        private protected override BsonDocument RenderArguments(
```
rstam, line 256: Refactor this method to emphasize that the only difference `_useConfiguredSerializers` makes is how the values are serialized:

    private protected override BsonDocument RenderArguments(
        RenderArgs<TDocument> args,
        IBsonSerializer fieldSerializer)
    {
        BsonValue serializedValues;
        if (_useConfiguredSerializers)
        {
            var arraySerializer = fieldSerializer switch
            {
                null => new ArraySerializer<TField>(args.SerializerRegistry.GetSerializer<TField>()),
                IBsonArraySerializer => fieldSerializer,                                             
                _ => new ArraySerializer<TField>((IBsonSerializer<TField>)fieldSerializer)           
            };                                                                                       
                                                                                                     
            serializedValues = SerializationHelper.SerializeValue(arraySerializer, _values);         
        }                                                                                            
        else                                                                                         
        {                                                                                            
            serializedValues = new BsonArray(_values.Select(ToBsonValue));                           
        }                                                                                            
                                                                                                     
        return new BsonDocument("value", serializedValues);                                          
    }

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -349,69 +369,81 @@ public QueryStringSearchDefinition(FieldDefinition<TDocument> defaultPath, strin
...
+            RenderArgs<TDocument> args,
+            IBsonSerializer fieldSerializer) => new()
             {
                 { "defaultPath", _defaultPath.Render(args) },
                 { "query", _query }
             };
     }
 
-    internal sealed class RangeSearchDefinition<TDocument, TField> : OperatorSearchDefinition<TDocument>
+    internal sealed class RangeSearchDefinition<TDocument, TValue> : OperatorSearchDefinition<TDocument>
```
rstam, line 381: As mentioned above keep `TField`.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -349,69 +369,81 @@ public QueryStringSearchDefinition(FieldDefinition<TDocument> defaultPath, strin
...
-            SearchRangeV2<TField> range,
+            SearchRangeV2<TValue> range,
             SearchScoreDefinition<TDocument> score)
                 : base(OperatorType.Range, path, score)
         {
             _range = range;
         }
 
-        private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args)
+        private protected override BsonDocument RenderArguments(
```
rstam, line 394: Refactor this method to emphasize that the only difference `_useConfiguredSerializers` makes is how the min and max values are serialized:

    private protected override BsonDocument RenderArguments(
        RenderArgs<TDocument> args,
        IBsonSerializer fieldSerializer)
    {
        BsonValue serializedMin;
        BsonValue serializedMax;
        if (_useConfiguredSerializers)
        {
            var valueSerializer = fieldSerializer switch
            {
                null => args.SerializerRegistry.GetSerializer<TField>(),
                IBsonArraySerializer => ArraySerializerHelper.GetItemSerializer(fieldSerializer),
                _ => fieldSerializer
            };
              
            serializedMin = _range.Min == null ? null : SerializationHelper.SerializeValue(valueSerializer, _range.Min.Value);
            serializedMax = _range.Max == null ? null : SerializationHelper.SerializeValue(valueSerializer, _range.Max.Value);
        }                                                                                                                     
        else                                                                                                                  
        {                                                                                                                     
            serializedMin = _range.Min == null ? null : ToBsonValue(_range.Min.Value);                                        
            serializedMax = _range.Max == null ? null : ToBsonValue(_range.Max.Value);                                        
        }                                                                                                                     
                                                                                                                              
        var minInclusive = _range.Min?.Inclusive ?? false;                                                                    
        var maxInclusive = _range.Max?.Inclusive ?? false;                                                                    
                                                                                                                              
        return new()                                                                                                          
        {                                                                                                                     
            { minInclusive ? "gte" : "gt", serializedMin, serializedMin != null },                                            
            { maxInclusive ? "lte" : "lt", serializedMax, serializedMax != null }                                             
        };                                                                                                                    
    }

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -54,6 +56,31 @@ public static implicit operator SearchDefinition<TDocument>(string json) =>
             json != null ? new JsonSearchDefinition<TDocument>(json) : null;
     }
 
+    /// <summary>
+    /// Extensions for SearchDefinition
```
rstam, line 60: /// Extension methods for SearchDefinition.

rstam (changes requested):

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -156,13 +183,53 @@ private protected OperatorSearchDefinition(OperatorType operatorType, SearchPath
...
+            renderedArgs.Add("score", () => _score.Render(args), _score != null);
             return new(_operatorType.ToCamelCase(), renderedArgs);
         }
 
-        private protected virtual BsonDocument RenderArguments(RenderArgs<TDocument> args) => new();
+        private protected virtual BsonDocument RenderArguments(
+            RenderArgs<TDocument> args,
+            IBsonSerializer fieldSerializer) => new();
+
+        internal void SetUseConfiguredSerialization(bool useDefaultSerialization)
```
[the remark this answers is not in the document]
papafe: I've renamed it to `useConfiguredSerializers` as in the other places. I had to change the name of the method to `SetUseConfiguredSerializersInternal` though, otherwise it would cause confusion (internally) between this and the extension method.

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -54,6 +56,31 @@ public static implicit operator SearchDefinition<TDocument>(string json) =>
...
+        /// Sets the use of the configured serialization for the specified <see cref="SearchDefinition{TDocument}"/>.
+        /// When set to true, the configured serializers will be used to serialize the values of certain Atlas Search operators, such as "Equals", "In" and "Range". This will become the default behaviour in version 4.0 of the library.
+        /// If not enabled, then a default conversion will be used.
+        /// </summary>
+        /// <typeparam name="TDocument">The type of the document.</typeparam>
+        /// <param name="searchDefinition">The search definition instance.</param>
+        /// <param name="useDefaultSerialization">Whether to use the configured serialization or not.</param>
+        /// <returns>The same <see cref="SearchDefinition{TDocument}"/> instance with default serialization enabled.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if <paramref name="searchDefinition"/> is not of a valid type/>.</exception>
+        public static SearchDefinition<TDocument> WithConfiguredSerialization<TDocument>(this SearchDefinition<TDocument> searchDefinition, bool useDefaultSerialization)
```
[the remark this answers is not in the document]
papafe: I see your point with `With` and I changed it to `Set`, even though I'm not super convinced by this. I think that in general `With` at least gives the impression of having an object (freshly created or not) returned, while `Set` feel more like a setter method that just changes state and returns void. I don't have strong opinions on either ways though.

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -532,20 +640,44 @@ public void In_typed<T>(
...
 
             AssertRendered(
-                subject.In(fieldExpression, fieldValues),
+                subjectTyped.In(fieldExpression, fieldValues),
+                $"{{ in: {{path: '{fieldNameRendered}', value: {fieldValuesArray} }} }}");
+        }
+
+        [Theory]
+        [MemberData(nameof(InTypedWithConfiguredSerializersTestData))]
+        public void InWithConfiguredSerialization_typed<T>(
```
rstam, line 657: Find remaining places where "configuredserialization" or "configured serialization" is used...

sanych-sun:
Please add appropriate label to the PR.

src/MongoDB.Driver/Search/SearchDefinition.cs
```diff
@@ -54,6 +56,33 @@ public static implicit operator SearchDefinition<TDocument>(string json) =>
...
+        /// Sets the use of the configured serialization for the specified <see cref="SearchDefinition{TDocument}"/>.
+        /// When set to true, the configured serializers will be used to serialize the values of certain Atlas Search operators, such as "Equals", "In" and "Range". This will become the default behaviour in version 4.0 of the library.
+        /// If not enabled, then a default conversion will be used.
+        /// </summary>
+        /// <typeparam name="TDocument">The type of the document.</typeparam>
+        /// <param name="searchDefinition">The search definition instance.</param>
+        /// <param name="useConfiguredSerializers">Whether to use the configured serializers or not.</param>
+        /// <returns>The same <see cref="SearchDefinition{TDocument}"/> instance with default serialization enabled.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if <paramref name="searchDefinition"/> is not of a valid type/>.</exception>
+        public static SearchDefinition<TDocument> SetUseConfiguredSerializers<TDocument>(this SearchDefinition<TDocument> searchDefinition, bool useConfiguredSerializers)
```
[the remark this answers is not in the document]
papafe: Sounds good!

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
papafe: It slipped through the cracks, it was the last one 🤦

rstam (approved):
LGTM

papafe:
@rstam @BorisDog I've inverted the boolean controlling the behaviour, so now by default the registered serializers will be used. 
I kept the same name for the method controlling this behaviour, as it seems it made the most sense.

BorisDog (approved):
LGTM!

rstam (approved):
LGTM

papafe:
> We should reword the title of the PR, so it would sounds like a feature. For release notes.

@sanych-sun Your comment made me realize that maybe the bug tag is more appropriate for this PR in the current state. We are actually correcting the behaviour of something that is already there AND also adding the possibility of having the previous behaviour. If this had to go in the release notes under only one section, I think it should be "bug".

PR #1588, opened 2025-01-08
CSHARP-5429: Support strings with range operator for atlas search

adelinowona:


src/MongoDB.Driver/Search/StringSearchRange.cs
```diff
@@ -0,0 +1,129 @@
...
+            => StringSearchRange.Empty.Lt(value);
+        
+        /// <summary>
+        /// Adds a less than value to a string search range.
+        /// </summary>
+        /// <param name="stringSearchRange">The string search range.</param>
+        /// <param name="value">The value.</param>
+        /// <returns>StringSearchRange.</returns>
+        public static StringSearchRange Lt(this StringSearchRange stringSearchRange, string value)
+            => new(stringSearchRange.Min, max: value, stringSearchRange.IsMinInclusive, false);
```
papafe, line 110: Better add the name of the last parameter, so it's easier to understand what's happening here.

src/MongoDB.Driver/Search/StringSearchRange.cs
```diff
@@ -0,0 +1,129 @@
...
+            => StringSearchRange.Empty.Lte(value);
+        
+        /// <summary>
+        /// Adds a less than or equal to value to a string search range.
+        /// </summary>
+        /// <param name="stringSearchRange">The string search range.</param>
+        /// <param name="value">The value.</param>
+        /// <returns>StringSearchRange.</returns>
+        public static StringSearchRange Lte(this StringSearchRange stringSearchRange, string value)
+            => new(stringSearchRange.Min, max: value, stringSearchRange.IsMinInclusive, true);
```
papafe, line 127: Same comment as before

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -611,6 +611,33 @@ public SearchDefinition<TDocument> Range<TField>(
             SearchScoreDefinition<TDocument> score = null)
             where TField : struct, IComparable<TField> =>
                 new RangeSearchDefinition<TDocument, TField>(path, range, score);
+
```
papafe, line 614: You're missing a constructor that takes `Expression<Func<TDocument, IEnumerable<TField>>> ` for `path`, and a relative test

papafe (changes requested):
Overall it looks good! Some smaller comments

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
adelinowona: I already added this in another PR that was merged. I just need to rebase my current PR to show that.

papafe: Maybe I'm missing something but I mean another constructor with this signature:

```
        public SearchDefinition<TDocument> Range<TField>(
            Expression<Func<TDocument, IEnumerable<TField>>> path,
            StringSearchRange range,
            SearchScoreDefinition<TDocument> score = null) =>
            Range(new ExpressionFieldDefinition<TDocument>(path), range, score);
```

adelinowona: Oh I thought you were mentioning it for other Range methods that use `SearchRange`. So for the new Range overloads using `StringSearchRange`, we don't need another overload with `IEnumerable<TField>` since `TField` isn't used anywhere else for the parameters. So if a user uses a field that is an `IEnumerable`, `TField` would be an `IEnumerable`.

papafe: My bad! You're right, in this case we got no parameters.

papafe (commented):
Looks good to me, let's just maybe talk about it during next standup to be sure we all agree this should be merged now

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -358,31 +358,42 @@ private protected override BsonDocument RenderArguments(RenderArgs<TDocument> ar
...
+            {
+                _min = ToBsonValue(_range.Min.Value);
+                _minInclusive = range.Min.Inclusive;
+            }
+            
+            if (range.Max != null)
+            {
+                _max = ToBsonValue(_range.Max.Value);
+                _maxInclusive = range.Max.Inclusive;
+            }
```
adelinowona, line 386: Just some changes to make some tests pass for now. Will be removed when synced with Ferdinando's PR

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -856,30 +856,6 @@ public void QueryString_typed()
...
-        [Fact]
-        public void RangeDouble()
-        {
-            var subject = CreateSubject<Person>();
-
-            AssertRendered(
-                subject.Range(p => p.Age, SearchRangeBuilder.Gt(1.5).Lt(2.5)),
-                "{ range: { path: 'age', gt: 1.5, lt: 2.5 } }");
-        }
-
```
adelinowona, line 882: I think we can just test this in the `Range_should_render_supported_types` test method so I removed them.

tests/MongoDB.Driver.Tests/Search/SearchRangeBuilderTests.cs
```diff
@@ -18,6 +18,8 @@
 using MongoDB.Driver.Search;
 using Xunit;
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
adelinowona, line 21: I added some obsoletions but now sure if we want to do that now. If we want obsoletions now, not sure if we should keep these tests.

src/MongoDB.Driver/Search/SearchRangeV2.cs
```diff
@@ -0,0 +1,155 @@
...
+ * limitations under the License.
+ */
+
+namespace MongoDB.Driver.Search
+{
+    /// <summary>
+    /// Represents a bound value.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the bound value.</typeparam>
+    public class Bound<TValue>
```
papafe, line 22: This could be a struct as well

papafe: And maybe I'd put this under the definition of SearchRangeV2

papafe (changes requested):
LGTM, only some minor comments

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -604,6 +604,7 @@ public SearchDefinition<TDocument> QueryString<TField>(
         /// <param name="range">The field range.</param>
         /// <param name="score">The score modifier.</param>
         /// <returns>A range search definition.</returns>
+        [Obsolete("Use Range method with SearchRangeV2 instead.")]
```
BorisDog, line 607: Eventually (4.0?) it would be ideal to have the new `SearchRangeV2 `to be named `SearchRange`. Wondering of what is the best path for that.

Will we just remove old `SearchRange` and rename `SearchRangeV2` to `SearchRange` ? That would be breaking, but maybe acceptable.

cc @rstam

BorisDog (changes requested):
Looks great overall!

src/MongoDB.Driver/Search/SearchRangeV2.cs
adelinowona: Good point I'll move the definition under SearchRangeV2. @papafe Regarding making Bound a struct, I didn't do it because then I'll have to make the parameters to SearchRangeV2 nullable and retrieving the Min or Max from SearchRangeV2 will not look too nice. 

With Bound as a struct, retrieving looks like this: `searchRangeV2.Min.Value.Value`
Currently with Bound as a class, retrieving looks nicer: `searchRangeV2.Min.Value`

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
adelinowona: So I think we should plan to consolidate our search range implementations in version 4.0 by renaming `SearchRangeV2` to `SearchRange` and removing the original `SearchRange`. Until then, users can choose the implementation that best suits their needs: the current `SearchRange` for basic functionality or `SearchRangeV2` for reference type support. While this will be a breaking change in 4.0, it's appropriate for a major version release and maybe doesn't require deprecation warnings in the current version. Users of either implementation will need to update their code in 4.0 - those using the original `SearchRange` will need to adapt to the new implementation, and those using `SearchRangeV2` will simply need to update their class references to `SearchRange`. So using deprecation warnings will only benefit users in that they will have an easier time updating their in 4.0

src/MongoDB.Driver/Search/SearchRangeV2.cs
papafe: Reasonable!

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -438,7 +451,7 @@ public SpanSearchDefinition(SearchSpanDefinition<TDocument> clause)
         private protected override BsonDocument RenderArguments(RenderArgs<TDocument> args) =>
             _clause.Render(args);
     }
-
+
```
BorisDog, line 454: Space.

src/MongoDB.Driver/Search/SearchRangeV2.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>Gets the lower bound of the range.</summary>
+        public Bound<TValue> Min { get; }
+    }
+    
+    /// <summary>
+    /// Represents a bound value.
+    /// </summary>
+    /// <typeparam name="TValue">The type of the bound value.</typeparam>
+    public sealed class Bound<TValue>
```
BorisDog, line 51: Nest under `SearchRangeV2`.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -639,7 +639,53 @@ public SearchDefinition<TDocument> Range<TField>(
             SearchRange<TField> range,
             SearchScoreDefinition<TDocument> score = null)
             where TField : struct, IComparable<TField> =>
-                new RangeSearchDefinition<TDocument, TField>(path, range, score);
+            Range(
+                path,
+                new SearchRangeV2<TField?>(
```
BorisDog, line 644: My understanding is that 
`SearchRange<T>.Min == null` should translate to `SearchRangeV2<T>.Min == null` and not to
`SearchRangeV2<T?>.Min.Value == null`. 

Is this correct?

If so we need to add test that SearchDefinition with both search ranges renders the same in this case.

src/MongoDB.Driver/Search/SearchRangeV2.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>
+        /// Adds a greater or equal than value to a search range.
+        /// </summary>
+        /// <typeparam name="TValue">The type of the value.</typeparam>
+        /// <param name="searchRange">Search range.</param>
+        /// <param name="value">The value.</param>
+        /// <returns>Search range.</returns>
+        public static SearchRangeV2<TValue> Gte<TValue>(this SearchRangeV2<TValue> searchRange, TValue value)
+            => new(min: new Bound<TValue>(value, inclusive: true), searchRange.Max);
```
BorisDog, line 115: Does
`=> new(min: new(value, inclusive: true), searchRange.Max);` work here?

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -639,7 +639,53 @@ public SearchDefinition<TDocument> Range<TField>(
...
+        /// <returns>A range search definition.</returns>
+        public SearchDefinition<TDocument> Range<TField>(
+            Expression<Func<TDocument, IEnumerable<TField>>> path,
+            SearchRangeV2<TField> range,
+            SearchScoreDefinition<TDocument> score = null) =>
+            Range(new ExpressionFieldDefinition<TDocument>(path), range, score);
+        
+        /// <summary>
+        /// Creates a search definition that queries for documents where a field is in the specified range.
+        /// </summary>
```
rstam, line 679: Missing:

        /// <typeparam name="TField">The type of the field.</typeparam>

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
rstam: Good catch! So something like:

     Range(
         path,
         new SearchRangeV2<TField>(
             range.Min.HasValue ? new(range.Min.Value, range.IsMinInclusive) : null,
             range.Max.HasValue ? new(range.Max.Value, range.IsMaxInclusive) : null),
         score);

rstam (changes requested):
LGTM

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -639,7 +639,54 @@ public SearchDefinition<TDocument> Range<TField>(
             SearchRange<TField> range,
             SearchScoreDefinition<TDocument> score = null)
             where TField : struct, IComparable<TField> =>
-                new RangeSearchDefinition<TDocument, TField>(path, range, score);
+            Range(
+                path,
+                new SearchRangeV2<TField>(
+                    range.Min.HasValue ? new (range.Min.Value, range.IsMinInclusive) : null,
```
BorisDog, line 645: now need for space after new.
new(...)

src/MongoDB.Driver/Search/SearchRangeV2.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        /// <summary>
+        /// Adds a greater or equal than value to a search range.
+        /// </summary>
+        /// <typeparam name="TValue">The type of the value.</typeparam>
+        /// <param name="searchRange">Search range.</param>
+        /// <param name="value">The value.</param>
+        /// <returns>Search range.</returns>
+        public static SearchRangeV2<TValue> Gte<TValue>(this SearchRangeV2<TValue> searchRange, TValue value)
+            => new(min: new (value, inclusive: true), searchRange.Max);
```
BorisDog, line 115: No need for space after `new`, here and elsewhere.

BorisDog (approved):
LGTM!

PR #1590, opened 2025-01-10
CSHARP-5435: NotSupportedException when using Set with sub-documents in an Update with Aggregation Pipeline.

rstam:


src/MongoDB.Bson/Serialization/Serializers/BsonClassMapSerializer.cs
```diff
@@ -23,11 +23,22 @@
 
 namespace MongoDB.Bson.Serialization
 {
+    /// <summary>
+    /// An interface implemented by BsonClassMapSerializer.
+    /// </summary>
+    public interface IBsonClassMapSerializer
```
rstam, line 29: This non-generic interface makes it possible for the LINQ translator to get the `ClassMap` without using reflection.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ExpressionToAggregationExpressionTranslator.cs
```diff
@@ -27,7 +27,7 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggreg
     internal static class ExpressionToAggregationExpressionTranslator
     {
         // public static methods
-        public static AggregationExpression Translate(TranslationContext context, Expression expression)
+        public static AggregationExpression Translate(TranslationContext context, Expression expression, IBsonSerializer resultSerializer = null)
```
rstam, line 30: If the caller passes in a `resultSerializer` it means that the translation MUST use that translator.

If the caller passes in `null` then the result can use whatever serializer comes out from the translation.

rstam: Making the `resultSerializer` argument optional reduced the number of changes required.

Let me know if you think it should be mandatory, but then every single call of this method would need to be touched.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -28,22 +28,28 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggreg
...
 
         public static AggregationExpression Translate(
             TranslationContext context,
             Expression expression,
             NewExpression newExpression,
-            IReadOnlyList<MemberBinding> bindings)
+            IReadOnlyList<MemberBinding> bindings,
+            IBsonSerializer resultSerializer)
         {
+            if (resultSerializer != null)
```
rstam, line 48: Only use the new code if the caller told us we had to use a particular `resultSerializer`.

Using the existing code if not preserves maximum backward compatibility.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -100,6 +106,71 @@ public static AggregationExpression Translate(
...
+                }
+                var creatorMapArguments = creatorMap?.Arguments?.ToArray();
+
+                for (var i = 0; i < constructorParameters.Length; i++)
+                {
+                    var parameterName = constructorParameters[i].Name;
+                    var argumentExpression = constructorArguments[i];
+                    var memberName = creatorMapArguments?[i].Name; // null if there is no classMap
+
+                    var (elementName, memberSerializer) = FindMemberElementNameAndSerializer(argumentExpression, classMap, memberName, documentSerializer, parameterName);
```
rstam, line 144: `FindMemberElementNameAndSerializer` will prioritize using the `classMap` and `memberName` if they are available, otherwise it will fallback to trying to match constructor parameter names to member names.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -100,6 +106,71 @@ public static AggregationExpression Translate(
...
+                    if (elementName == null)
+                    {
+                        throw new ExpressionNotSupportedException(expression, because: $"couldn't find matching class member for constructor parameter {parameterName}");
+                    }
+
+                    var argumentTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, argumentExpression, memberSerializer);
```
rstam, line 150: Note that we are telling `Translate` that the `resultSerializer` must be `memberSerializer`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -100,6 +106,71 @@ public static AggregationExpression Translate(
...
+                var memberAssignment = (MemberAssignment)binding;
+                var member = memberAssignment.Member;
+                var valueExpression = memberAssignment.Expression;
+                if (!documentSerializer.TryGetMemberSerializationInfo(member.Name, out var memberSerializationInfo))
+                {
+                    throw new ExpressionNotSupportedException(valueExpression, expression, because: $"couldn't find member {member.Name}");
+                }
+                var memberSerializer = memberSerializationInfo.Serializer;
+
+                var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression, memberSerializer);
```
rstam, line 166: Note that we are telling `Translate` that the `resultSerializer` must be `memberSerializer`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -190,6 +261,36 @@ private static void EnsureDefaultValue(BsonMemberMap memberMap)
...
+            {
+                var memberMap = FindMemberMap(expression, classMap, memberName);
+                return (memberMap.ElementName, memberMap.GetSerializer());
+            }
+
+            // otherwise fall back to calling TryGetMemberSerializationInfo on potential matches
+            var bindingFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.FlattenHierarchy | BindingFlags.IgnoreCase;
+            foreach (var memberInfo in documentSerializer.ValueType.GetMember(constructorParameterName, bindingFlags))
+            {
+                if (documentSerializer.TryGetMemberSerializationInfo(memberInfo.Name, out var serializationInfo))
```
rstam, line 285: While using a case insensitive match could give us some false hits, `TryGetMemberSerializationInfo` will return `false` for the false hits.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -24,15 +24,34 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggreg
...
+            }
+
+            var items = new List<AstExpression>();
             foreach (var itemExpression in expression.Expressions)
             {
-                var itemTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, itemExpression);
+                var itemTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, itemExpression, itemSerializer);
                 items.Add(itemTranslation.Ast);
-                itemSerializer ??= itemTranslation.Serializer;
+                if (itemSerializer == null)
```
rstam, line 51: If the caller did not provide a `resultSerializer` then `itemSerializer` will be initialized to the serializer of the first translated item in the array.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewArrayInitExpressionToAggregationExpressionTranslator.cs
```diff
@@ -42,12 +61,14 @@ public static AggregationExpression Translate(TranslationContext context, NewArr
             }
 
             var ast = AstExpression.ComputedArray(items);
-
-            var arrayType = expression.Type;
-            var itemType = arrayType.GetElementType();
-            itemSerializer ??= BsonSerializer.LookupSerializer(itemType); // if the array is empty itemSerializer will be null
-            var arraySerializerType = typeof(ArraySerializer<>).MakeGenericType(itemType);
-            var arraySerializer = (IBsonSerializer)Activator.CreateInstance(arraySerializerType, itemSerializer);
+            if (arraySerializer == null)
```
rstam, line 64: If the caller did not provide a `resultSerializer` we create one now.

Note: `arraySerializer` and `resultSerializer` are the same thing just that `arraySerializer` is declared `IBsonArraySerializer` instead of just `IBsonSerializer`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ExpressionToAggregationExpressionTranslator.cs
```diff
@@ -67,21 +67,21 @@ public static AggregationExpression Translate(TranslationContext context, Expres
...
-                    return MemberInitExpressionToAggregationExpressionTranslator.Translate(context, (MemberInitExpression)expression);
+                    return MemberInitExpressionToAggregationExpressionTranslator.Translate(context, (MemberInitExpression)expression, resultSerializer);
                 case ExpressionType.Negate:
                     return NegateExpressionToAggregationExpressionTranslator.Translate(context, (UnaryExpression)expression);
                 case ExpressionType.New:
-                    return NewExpressionToAggregationExpressionTranslator.Translate(context, (NewExpression)expression);
+                    return NewExpressionToAggregationExpressionTranslator.Translate(context, (NewExpression)expression, resultSerializer);
                 case ExpressionType.NewArrayInit:
-                    return NewArrayInitExpressionToAggregationExpressionTranslator.Translate(context, (NewArrayExpression)expression);
+                    return NewArrayInitExpressionToAggregationExpressionTranslator.Translate(context, (NewArrayExpression)expression, resultSerializer);
```
rstam, line 84: We need to do some more investigation to see where else we might need to pass the `resultSerializer`.

For example, we currently handle `T[]`, but what about `List<T>` or other enumerable types?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/TranslationContext.cs
```diff
@@ -13,26 +13,27 @@
...
 using MongoDB.Driver.Linq.Linq3Implementation.Misc;
-using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
 
 namespace MongoDB.Driver.Linq.Linq3Implementation.Translators
 {
     internal class TranslationContext
     {
         #region static
         public static TranslationContext Create(
-            Expression expression,
```
rstam, line 29: Apparently we just forgot to remove this unused parameter before now.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToSetStageTranslators/ExpressionToSetStageTranslator.cs
```diff
@@ -153,7 +153,8 @@ private static AstComputedField CreateComputedField(TranslationContext context,
                 }
                 else
                 {
-                    var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression);
+                    var resultSerializer = serializationInfo.Serializer;
+                    var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression, resultSerializer);
```
rstam, line 157: This is the entry point at which translation of `Update.Set` tells the `Translate` method what `resultSerializer` to use.

From here the `resultSerializer` is propagated down to where it is needed.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5435Tests.cs
```diff
@@ -0,0 +1,226 @@
...
+                updateError.Render(new(coll.DocumentSerializer, BsonSerializer.SerializerRegistry))
+                    .AsBsonArray
+                    .Cast<BsonDocument>();
+            AssertStages(updateStages, "{ $set : { ValueObject : { Value : { $cond : { if : { $eq : ['$ValueObject', null] }, then : 1, else : { $add : ['$ValueObject.Value', 1] } } } } } }");
+
+            coll.UpdateOne(filter, updateError, new() { IsUpsert = true });
+        }
+
+        [Fact]
+        public void Test_set_ValueObject_to_derived_value_using_property_setter()
```
rstam, line 79: This test still fails. Maybe we could discuss solutions.

I suspect we need some combination of the following to make this work:

1. A way of asking a serializer for `T` what serializer to use for a base class `B` or a derived class `D`
2. A way of checking that the serializer of the translated value is "compatible" with the result serializer

The second almost surely requires the first.

rstam:
All the variants are failing because we still need to figure out how to handle setting the property to a derived value.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ExpressionToAggregationExpressionTranslator.cs
rstam: Alex and I are discussing whether to make a new ticket for this work, something like "add `targetSerializer` parameter to `Translate` methods to pass down expected serializer for the result".

A separate ticket makes sense to me because this is a much broader issue than just `Update.Set`.

I also think we should make the `targetSerializer` parameter mandatory, even though it will require a lot of work. That way we can find all the places where it needs to be passed down (you mentioned in another comment that we might have missed some enumerable cases, so that's an example of something we might find by making the parameter mandatory).

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ExpressionToAggregationExpressionTranslator.cs
rstam: Agreed. I think we are missing  more scenarios where the `targetSerializer` need to passed down.

Note: I'm renaming what I used to call `resultSerializer` to `targetSerializer` because "result" sounds more like an output than an input.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToSetStageTranslators/ExpressionToSetStageTranslator.cs
```diff
@@ -153,7 +153,8 @@ private static AstComputedField CreateComputedField(TranslationContext context,
                 }
                 else
                 {
-                    var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression);
+                    var resultSerializer = serializationInfo.Serializer;
```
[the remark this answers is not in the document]
rstam: True.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5435Tests.cs
rstam: This issue is still pending. I haven't solved it yet. I'm not sure it can be solved without adding new functionality to polymorphic serializers to ask for a new serializer for a base or derived type.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/TranslationContext.cs
```diff
@@ -43,14 +44,14 @@ public static TranslationContext Create(
         private readonly ExpressionTranslationOptions _translationOptions;
 
         private TranslationContext(
+            ExpressionTranslationOptions translationOptions,
             SymbolTable symbolTable,
             NameGenerator nameGenerator,
-            ExpressionTranslationOptions translationOptions,
             TranslationContextData data = null)
```
adelinowona, line 50: Is there a special reason for moving the translationOptions parameter here?

rstam:
Because the work is broader than just Update.Set a new JIRA ticket CSHARP-5461 has been created. Update.Set is just one manifestation of the broader problem. 

This PR is replaced by this one: https://github.com/mongodb/mongo-csharp-driver/pull/1596

PR #1591, opened 2025-01-13
CSHARP-5456: BsonGuidRepresentation does not work with Nullable

papafe:


src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
```diff
@@ -50,13 +50,15 @@ public GuidRepresentation GuidRepresentation
         /// <inheritdoc/>
         public void Apply(BsonMemberMap memberMap)
         {
-            var guidSerializer = memberMap.GetSerializer() as GuidSerializer;
-            if (guidSerializer == null)
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializerRecursively(memberMap.GetSerializer(), Reconfigure);
+            if (reconfiguredSerializer == null)
             {
                 throw new InvalidOperationException("[BsonGuidRepresentationAttribute] can only be used when the serializer is a GuidSerializer.");
```
papafe, line 56: Should we also change the text here?

src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
```diff
@@ -61,10 +61,12 @@ public BsonDateOnlyOptionsAttribute(BsonType representation, DateOnlyDocumentFor
...
         {
-            return SerializerConfigurator.ReconfigureSerializerRecursively(serializer, Reconfigure) ?? base.Apply(serializer);
+            if (serializer is DateOnlySerializer dateOnlySerializer)
+            {
+                return dateOnlySerializer.WithRepresentation(_representation, _documentFormat);
+            }
 
-            IBsonSerializer Reconfigure(IBsonSerializer s)
-                => s is DateOnlySerializer dos ? dos.WithRepresentation(_representation, _documentFormat) : null;
+            return base.Apply(serializer);
```
papafe, line 69: This is unrelated, but I thought it would be ok to do it inside this PR since this is a very small change. I can create another ticket otherwise. 

The reason for this change is that `base.Apply` does the same job as `ReconfigureSerializerRecursively`. Before we were trying to do the same reconfiguration twice, if not successful.

src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
```diff
@@ -50,13 +50,15 @@ public GuidRepresentation GuidRepresentation
         /// <inheritdoc/>
         public void Apply(BsonMemberMap memberMap)
         {
-            var guidSerializer = memberMap.GetSerializer() as GuidSerializer;
-            if (guidSerializer == null)
+            var reconfiguredSerializer = SerializerConfigurator.ReconfigureSerializerRecursively(memberMap.GetSerializer(), Reconfigure);
```
rstam, line 53: Rename variable to `reconfiguredSerializer` (it's missing the letter `d`)

src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
rstam: Should `BsonSerializationOptionsAttribute.Apply` be calling `ReconfigureSerializerRecursively`?

Or should `BsonSerializationOptionsAttribute.Apply` *not* be configuring child serializers at all? By doing this in the base `Apply` we are essentially saying that ALL attributes apply to child serializers also, which may or may not correct.

Thoughts?

src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
rstam: Maybe refactor this to match `BsonDateOnlyOptionsAttribute`? 

By which I mean let `base.Apply` do the recursive application and only check for `is GuidSerializer` here.

Is ``ReconfigureSerializerRecursively` even necessary any more if move the logic to`base.Apply`?

src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
rstam: Change it how?

src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
rstam: I guess one question might be whether we need to throw an exception when this attribute is applied to non-Guid serializers or whether that is a no-op.

rstam (changes requested):

src/MongoDB.Bson/Serialization/Attributes/BsonGuidRepresentationAttribute.cs
papafe: Regarding the first question, let's discuss on slack as I fear it would be cumbersome to go back on forth on comments here 😁 

Regarding the exception, I think we should throw. I think it would be better to throw, so developers can easily see that they've applied the attribute to the wrong property.

rstam (changes requested):

papafe:
@rstam Just requested your review again after our discussion on slack. 
I think the final contention point was if `BsonSerializationOptionsAttribute.Apply` should be calling `ReconfigureSerializerRecursively`, in order to work also with `IMultipleChildSerializersConfigurable`. I would prefer to keep this as is, as it would be a behaviour change otherwise, and do this change for next major.

src/MongoDB.Bson/Serialization/Attributes/BsonDateOnlyOptionsAttribute.cs
rstam: Aren't we no longer calling `ReconfigureSerializerRecursively` with this change?

Is that a breaking behavior change since we were calling it before?

rstam (changes requested):

rstam:
> I would prefer to keep this as is, as it would be a behaviour change otherwise, and do this change for next major.

Why is handling multiple child serializer any more breaking a behavior change than the behavior changes you are already making in this PR?

src/MongoDB.Bson/Serialization/Attributes/BsonSerializationOptionsAttribute.cs
```diff
@@ -52,18 +52,16 @@ public virtual void Apply(BsonMemberMap memberMap)
         protected virtual IBsonSerializer Apply(IBsonSerializer serializer)
         {
             // if none of the overrides applied the attribute to the serializer see if it can be applied to a child serializer
-            var childSerializerConfigurable = serializer as IChildSerializerConfigurable;
-            if (childSerializerConfigurable != null)
+            if (serializer is IChildSerializerConfigurable childSerializerConfigurable)
```
rstam, line 55: We need to try to apply the attribute to all the children using `IMultiChildSerializerConfigurable`.

But then we also need to figure out what to do if the attribute applies to one child and not another.

Here's an example  to motivate thinking about this:

```
public Dictionary<Guid, int> { get; set; }
```

Say we want the `Guid` serialized as a string (because document element names must be strings) but we want `int` to be serialized as an int. How would we do that?

It currently can be done when configuring in code, but not with attributes.

I understand you don't want to tackle things like that in this PR and that's fine, but you should create a JIRA ticket for it.

tests/MongoDB.Driver.Tests/Specifications/uuid/prose-tests/ImplicitEncodingTests.cs
```diff
@@ -114,6 +113,27 @@ public void Implicit_encoding_with_unspecified_representation_should_work_as_exp
...
+            DropCollection(collection);
+            collection.InsertOne(document);
+
+            var insertedDocument = FindSingleDocument(collection);
+            insertedDocument.Id.Should().Be(guid);
+            insertedDocument.GuidArray[0].Should().Be(guid);
+
+            var insertedDocumentAsBsonDocument = FindSingleDocumentAsBsonDocument(collection);
+            var binaryData = (BsonBinaryData)insertedDocumentAsBsonDocument["_id"];
+            binaryData.SubType.Should().Be(BsonBinarySubType.UuidStandard);
```
rstam, line 132: Consider adding:

    binaryData.Bytes.Should().Equal(0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff);

tests/MongoDB.Driver.Tests/Specifications/uuid/prose-tests/ImplicitEncodingTests.cs
```diff
@@ -114,6 +113,27 @@ public void Implicit_encoding_with_unspecified_representation_should_work_as_exp
...
+            var binaryData2 = (BsonBinaryData)((BsonArray)insertedDocumentAsBsonDocument["GuidArray"])[0];
```
rstam, line 133: Consider adding:
    binaryData2.Bytes.Should().Equal(0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff);

src/MongoDB.Bson/Serialization/Attributes/BsonSerializationOptionsAttribute.cs
rstam: Thinking about this a bit more... maybe this means that we *can't* apply attributes using `IMultiChildSerializerConfigurable`.

How do we know which child serializer the user intended the attribute for?

rstam (approved):
LGTM with comments

src/MongoDB.Bson/Serialization/Attributes/BsonSerializationOptionsAttribute.cs
papafe: I agree we should try to make a system that allows a more fine grained configuration of where the attribute needs to be applied. 
Regarding the second point.... I also agree with this, but I think it's more of an issue for certain kinds of attributes/conventions (given that at the moment we use it for `EnumRepresentationConvention`). 
For instance, with a more "specific" attirbute/convention, like `EnumRepresentationConvention`, I think it would be fine to apply it to all the children, as it would not affect types that are not enum in this case.
For more generic attributes/conventions, like `BsonRepresentationAttribute`, then it's a big issue, like in your example. 
I think this is something we should probably consider when we decide how to tackle this with attributes in another ticket

tests/MongoDB.Driver.Tests/Specifications/uuid/prose-tests/ImplicitEncodingTests.cs
papafe: Added

tests/MongoDB.Driver.Tests/Specifications/uuid/prose-tests/ImplicitEncodingTests.cs
papafe: Added

PR #1597, opened 2025-01-16
CSHARP-5359: Double check that all client BulkWrite tests are impleme…

sanych-sun:
…nted

### DIFF
--- a/tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
+++ b/tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
@@ -173,6 +173,11 @@ namespace MongoDB.Driver.Tests.Specifications
                 RequireServer.Check().VersionGreaterThanOrEqualTo(SemanticVersion.Parse("4.4"));
             }
 
+            if (testCase.Name.Contains("client-bulkWrite.json:client bulkWrite with writeConcern in a transaction causes a transaction error"))
+            {
+                throw new SkipException("CSHARP Driver does not comply with the requirement to throw in case explicit writeConcern were used");
+            }
+
             Run(testCase);
         }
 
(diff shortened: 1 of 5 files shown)


tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -173,6 +173,11 @@ public void Transactions(JsonDrivenTestCase testCase)
                 RequireServer.Check().VersionGreaterThanOrEqualTo(SemanticVersion.Parse("4.4"));
             }
 
+            if (testCase.Name.Contains("client-bulkWrite.json:client bulkWrite with writeConcern in a transaction causes a transaction error"))
```
sanych-sun, line 176: It seems like we missed or intentionally skipped the following changes to requirements: https://github.com/mongodb/specifications/pull/358

CSharp driver instead of throwing is selecting appropriate write concern and ignore provided in method call. This logic is applicable for any transactional operation, not just clientBulkWrite. Should we discuss what to do next here?

sanych-sun: Here: https://github.com/mongodb/mongo-csharp-driver/blob/660f33d01fe9604044e2877e4564ce10dbfc2593/src/MongoDB.Driver/Core/Operations/WriteConcernHelper.cs#L23

sanych-sun: Jira ticket was created as it was discussed and scheduled to be fixed in next major release, to avoid behavioral breaking change in minor release.

BorisDog, line 176: Add this to `__ignoredTests`?

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -260,7 +255,13 @@ private static void RequireKmsMock() =>
             "hello with speculativeAuthenticate",
             "hello without speculativeAuthenticate is always observed",
             "legacy hello with speculativeAuthenticate",
-            "legacy hello without speculativeAuthenticate is always observed"
+            "legacy hello without speculativeAuthenticate is always observed",
+
+            // Skipped because CSharp Driver has an issue with handling read timeout for sync code-path. CSHARP-3662
```
BorisDog, line 260: Please add category comment  "// transactions".

BorisDog (approved):
LGTM + comment

PR #1601, opened 2025-01-25
CSHARP-5473: Provide API to turn LINQ expression into MQL.

rstam:
The BIG question for this PR is whether we want to do this at all, and if so what the public API should be.

### DIFF
--- a/src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
+++ b/src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
@@ -40,6 +40,15 @@ namespace MongoDB.Driver.Linq
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>The value that results from executing the specified query.</returns>
         Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Translates an IQueryable to MQL.
+        /// </summary>
+        /// <param name="queryable">The IQueryable.</param>
+        /// <param name="outputSerializer">The output serializer.</param>
+        /// <typeparam name="TResult">The type of the result.</typeparam>
+        /// <returns>An array of MQL pipeline stages represented as BsonDocuments.</returns>
+        BsonDocument[] Translate<TResult>(IQueryable<TResult> queryable, out IBsonSerializer<TResult> outputSerializer);
     }
 
     /// <summary>
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
@@ -55,6 +55,7 @@ namespace MongoDB.Driver.Linq.Linq3Implementation
         public abstract TResult Execute<TResult>(Expression expression);
         public abstract Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken);
         public abstract ExpressionTranslationOptions GetTranslationOptions();
+        public abstract BsonDocument[] Translate<TResult>(IQueryable<TResult> queryable, out IBsonSerializer<TResult> outputSerializer);
     }
 
     internal sealed class MongoQueryProvider<TDocument> : MongoQueryProvider
@@ -151,6 +152,15 @@ namespace MongoDB.Driver.Linq.Linq3Implementation
             var translationOptions = _options?.TranslationOptions;
             var database = _database ?? _collection?.Database;
             return translationOptions.AddMissingOptionsFrom(database?.Client.Settings.TranslationOptions);
+        }
+
+        public override BsonDocument[] Translate<TResult>(IQueryable<TResult> queryable, out IBsonSerializer<TResult> outputSerializer)
+        {
+            var translationOptions = GetTranslationOptions();
+            var executableQuery = ExpressionToExecutableQueryTranslator.Translate<TDocument, TResult>(provider: this, queryable.Expression, translationOptions);
+            var stages = executableQuery.Pipeline.Ast.Stages;
+            outputSerializer = (IBsonSerializer<TResult>)executableQuery.Pipeline.OutputSerializer;
+            return stages.Select(s => s.Render().AsBsonDocument).ToArray();
         }
     }
 }
--- a/tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
+++ b/tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
@@ -1,0 +1,56 @@
+﻿/* Copyright 2010-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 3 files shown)


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
```diff
@@ -0,0 +1,56 @@
...
+using FluentAssertions;
+using MongoDB.Driver.Linq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp5473Tests : Linq3IntegrationTest
+    {
+        [Fact]
+        public void Select_decimal_divide_should_work()
```
sanych-sun, line 26: _Minor_ : Test method name seems to be wrong.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
```diff
@@ -0,0 +1,56 @@
...
+            var collection = GetCollection();
+
+            var queryable = collection.AsQueryable()
+                .Select(x => x.X + 1);
+
+            var provider = (IMongoQueryProvider)queryable.Provider;
+            var stages = provider.Translate(queryable, out var outputSerializer);
+            AssertStages(stages, "{ $project : { _v : { $add : ['$X', 1] }, _id : 0 } }");
+
+            var result = queryable.First();
```
sanych-sun, line 37: Should we try to execute translated MQL in some way instead to make sure the translation was done properly?

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
rstam: Fixed.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5473Tests.cs
rstam: Yes, but not necessarily to verify that the translation was done correctly.

I'm more interested in showing HOW you could execute the result of `Translate`.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -152,5 +153,14 @@ public override ExpressionTranslationOptions GetTranslationOptions()
             var database = _database ?? _collection?.Database;
             return translationOptions.AddMissingOptionsFrom(database?.Client.Settings.TranslationOptions);
         }
+
+        public override BsonDocument[] Translate<TResult>(IQueryable<TResult> queryable, out IBsonSerializer<TResult> outputSerializer)
```
sanych-sun, line 157: Should we check if provided queryable is MongoDB queryable?

sanych-sun (dismissed):
LGTM + minor question.

damieng:
I think if we want to ship this for now we should use [ExperimentalAttribute].

Alternatively I could use reflection to do the same sort of thing on the already-released version and make sure it meets our use cases on the EF provider before shipping.

rstam:
> I think if we want to ship this for now we should use the `[Experimental]` attribute.
> 
> Alternatively I could use reflection to do the same sort of thing on the already-released version and make sure it meets our use cases on the EF provider before shipping.

You are the one that asked for this. If you are not sure it does what you need then we shouldn't ship it yet.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
BorisDog: Or maybe receiving an expression is sufficient?

rstam: It would be sufficient.

I chose `IQueryable` to emphasize that it's the `Expression` associated with an `IQueryable` that is being translated, not any old expression.

But the .NET method takes an `Expression`, so we could also here.

@damieng what do you think?

BorisDog (approved):
LGTM

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
adelinowona: The translate method seems to only care about the expression of the queryable so I think we should just take `Expression` here as well. We can add a note in the comments of methods to emphasize the expression should be associated with an `IQueryable` and maybe throw an exception if we catch such as a case?

src/MongoDB.Driver/Linq/IMongoQueryProvider.cs
```diff
@@ -40,6 +40,15 @@ public interface IMongoQueryProvider : IQueryProvider
...
         Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Translates an Expression to MQL.
+        /// </summary>
+        /// <param name="expression">The expression.</param>
+        /// <param name="outputSerializer">The output serializer.</param>
+        /// <typeparam name="TResult">The type of the result.</typeparam>
+        /// <returns>An array of MQL pipeline stages represented as BsonDocuments.</returns>
+        BsonDocument[] Translate<TResult>(Expression expression, out IBsonSerializer<TResult> outputSerializer);
```
rstam, line 51: Adelin commented:

> The translate method seems to only care about the expression of the queryable so I think we should just take Expression here as well.

The latest commit implements his suggestion.

@BorisDog and @damieng will this work for you two in the Analyzer and the EF Core Provider?

BorisDog: I think receiving expression is more consistent with `ExecuteAsync`.
Analyzer will not use this functionality at all.

BorisDog (approved):
If this is still needed for EF, LGTM.

src/MongoDB.Driver/Linq/Linq3Implementation/MongoQueryProvider.cs
```diff
@@ -152,5 +153,14 @@ public override ExpressionTranslationOptions GetTranslationOptions()
...
             return translationOptions.AddMissingOptionsFrom(database?.Client.Settings.TranslationOptions);
         }
+
+        public override BsonDocument[] Translate<TResult>(Expression expression, out IBsonSerializer<TResult> outputSerializer)
+        {
+            var translationOptions = GetTranslationOptions();
+            var executableQuery = ExpressionToExecutableQueryTranslator.Translate<TDocument, TResult>(provider: this, expression, translationOptions);
+            var stages = executableQuery.Pipeline.Ast.Stages;
+            outputSerializer = (IBsonSerializer<TResult>)executableQuery.Pipeline.OutputSerializer;
+            return stages.Select(s => s.Render().AsBsonDocument).ToArray();
```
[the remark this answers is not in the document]
adelinowona: @rstam this doesn't seem worth it to me but I defer the decision to you. Typical pipelines have about 2-10 stages and this also is not in a hot loop. We will increase verbosity for minor performance gains.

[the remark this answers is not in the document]
rstam: Agreed. I don't think this micro optimization is worth it.

Let's keep the code simple.

PR #1602, opened 2025-01-26
CSHARP-1585: Add support for `"@<elem>"` special field name in ElementMatchFilterDefinition.

rstam:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
```diff
@@ -13,42 +13,43 @@
...
-            var serializerRegistry = BsonSerializer.SerializerRegistry;
-            var documentSerializer = serializerRegistry.GetSerializer<Document>();
-            var context = TranslationContext.Create(translationOptions: null);
-            var symbol = context.CreateSymbol(parameter, documentSerializer, isCurrent: true);
-            context = context.WithSymbol(symbol);
-            var filter = ExpressionToFilterTranslator.Translate(context, expression.Body, exprOk: false);
-            var simplifiedFilter = AstSimplifier.Simplify(filter);
+            var collection = GetCollection<Document>();
+            var filter = Builders<Document>.Filter.Where(
+                document => document.Details.A.Any(x => x.Any(y => Regex.IsMatch(y.DeviceName, @".Name0."))));
```
rstam, line 33: Note that LINQ already supported a way to write the query the user wanted to run.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
```diff
@@ -13,42 +13,43 @@
...
+            translatedFilter.Should().Be("{ 'Details.A' : { $elemMatch : { $elemMatch : { DeviceName : /.Name0./ } } } }");
+        }
+
+        [Fact]
+        public void Filter_Builder_ElemMatch_ElemMatch_should_translate_correctly()
+        {
+            var collection = GetCollection<Document>();
+            var deviceFilter = Builders<Device>.Filter.Regex(x => x.DeviceName, new BsonRegularExpression(".Name0."));
+            var deviceArrayFilter = Builders<Device[]>.Filter.ElemMatch("@<elem>", deviceFilter);
+            var filter = Builders<Document>.Filter.ElemMatch(x => x.Details.A, deviceArrayFilter);
```
rstam, line 47: Writing the same query using filter builders was very hard. It took me multiple tries.

It also requires adding support for the special `"@<elem>"` field name.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
```diff
@@ -13,42 +13,43 @@
...
+            var translatedFilter = TranslateFindFilter(collection, find);
+            translatedFilter.Should().Be("{ 'Details.A' : { $elemMatch : { $elemMatch : { DeviceName : /.Name0./ } } } }");
+        }
+
+        [Fact]
+        public void Filter_Builder_ElemMatch_ElemMatch_should_translate_correctly()
+        {
+            var collection = GetCollection<Document>();
+            var deviceFilter = Builders<Device>.Filter.Regex(x => x.DeviceName, new BsonRegularExpression(".Name0."));
+            var deviceArrayFilter = Builders<Device[]>.Filter.ElemMatch("@<elem>", deviceFilter);
```
sanych-sun, line 46: I'm OK with using "@<elem>" constant internally, but can we create additional overload for `ElemMatch` method without name parameter at all (and internally it could default to use "@<elem>")

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -473,6 +472,35 @@ public FilterDefinition<TDocument> BitsAnySet(Expression<Func<TDocument, object>
...
         }
 
+        /// <summary>
+        /// Creates an element match filter for an array value.
+        /// </summary>
+        /// <remarks>TDocument must implement IEnumerable{TITem} when using this overload of ElemMatch.</remarks>
+        /// <param name="filter">The item filter.</param>
+        /// <returns>An element match filter.</returns>
+        public FilterDefinition<TDocument> ElemMatch<TItem>(FilterDefinition<TItem> filter)
+            // where TDocument : IEnumerable<TItem> (can only be checked at runtime)
```
rstam, line 482: This is one of the weird parts of adding this overload.

This constraint must hold true but it can't be checked at compile time.

rstam: I'm not worried about this new overload resulting in ambiguities with the existing overloads because the number of parameters is different.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
```diff
@@ -13,42 +13,43 @@
...
+            var translatedFilter = TranslateFindFilter(collection, find);
+            translatedFilter.Should().Be("{ 'Details.A' : { $elemMatch : { $elemMatch : { DeviceName : /.Name0./ } } } }");
+        }
+
+        [Fact]
+        public void Filter_Builder_ElemMatch_ElemMatch_should_translate_correctly()
+        {
+            var collection = GetCollection<Document>();
+            var deviceFilter = Builders<Device>.Filter.Regex(x => x.DeviceName, new BsonRegularExpression(".Name0."));
+            var deviceArrayFilter = Builders<Device[]>.Filter.ElemMatch(deviceFilter);
```
rstam, line 46: This line was changed in the latest commit to use the new overload.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
rstam: Does the sequence of builders make sense to you?

```
Builders<Device>.Filter
Builders<Device[]>.Filter
Builders<Document>.Filter
```

It's subtle and it took me awhile to figure it out.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1880,8 +1880,11 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
             }
 
             var renderedFilter = _filter.Render(args.WithNewDocumentType(itemSerializer) with { RenderForElemMatch = true });
+            var renderedElemMatch = new BsonDocument("$elemMatch", renderedFilter);
 
-            return new BsonDocument(renderedField.FieldName, new BsonDocument("$elemMatch", renderedFilter));
+            return renderedField.FieldName == "@<elem>" ?
```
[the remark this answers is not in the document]
BorisDog: Is it possible to indicate the absence of field by `_field == null` ? 
Additional ctor:  `public ElementMatchFilterDefinition(FilterDefinition<TItem> filter)`

Alternatively: `class NullFieldDefinition : FieldDefinition`

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp1585Tests.cs
BorisDog: Makes sense to me.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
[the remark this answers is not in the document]
rstam: I considered using `null` but in almost all cases `null` is an error, so I thought it would be better to use a special name.

A new class is theoretically an option. The current approach mirrors what we didn't in the Ast classes (which are internal). Do we really want a new public class? Maybe... not sure.

Another way to think about it is that the field is not really "missing". It's just a quirk of `$elemMatch` that there is no name for the elements being matched. So a special name also makes a lot of sense to me, more than `null`.

I'm also concerned that if we did introduce a new class that it would be open to being used anywhere a `FieldDefinition` is expected, even where this new class makes no sense.

So all in all, I'm leaning towards sticking with the special name.

src/MongoDB.Driver/FilterDefinitionBuilder.cs
BorisDog: Let's add a test for the runtime check?

BorisDog (changes requested):

sanych-sun (approved):
LGTM

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1854,22 +1882,36 @@ internal sealed class ElementMatchFilterDefinition<TDocument, TItem> : FilterDef
...
         public override BsonDocument Render(RenderArgs<TDocument> args)
         {
-            var renderedField = _field.Render(args);
+            string fieldName = null;
+            IBsonSerializer fieldSerializer;
+
+            if (_field == null)
+            {
+                fieldSerializer = args.DocumentSerializer;
+            }
```
BorisDog, line 1897: Do we need to 'fake' the `fieldSerializer` here?
Do we need to try deducing the `itemSerializer` from `DocumentSerializer`, isn't 
`itemSerializer = args.SerializerRegistry.GetSerializer<TItem>();` sufficient?

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1852,24 +1877,61 @@ internal sealed class ElementMatchFilterDefinition<TDocument, TItem> : FilterDef
...
             _field = Ensure.IsNotNull(field, nameof(field));
             _filter = filter;
+
+            var fieldType = field.GetFieldType();
+            if (fieldType != null) // for some implementations of FieldDefinition TField is not known
+            {
+                var ienumerableItem = typeof(IEnumerable<>).MakeGenericType(typeof(TItem));
+                if (!fieldType.Implements(ienumerableItem))
+                {
+                    throw new ArgumentException($"ElemMatch requires that {fieldType} implement IEnumerable<{typeof(TItem)}>.");
```
BorisDog, line 1904: Is this change safe? Could it be breaking?
For example for users that are using BsonDocument/BsonValue for most properties:

```
 var field = new StringFieldDefinition<BsonDocument, BsonValue>("a");
 FilterDefinitionBuilder<BsonDocument>.ElemMatch<BsonDocument>(field, "{b: 1}")
```

src/MongoDB.Driver/FilterDefinitionBuilder.cs
BorisDog: That makes sense, thanks for the explanation. The refactored code is more clear.

BorisDog (changes requested):

src/MongoDB.Driver/FieldDefinition.cs
```diff
@@ -454,4 +454,23 @@ public override RenderedFieldDefinition Render(RenderArgs<TDocument> args)
...
     }
+
+    internal static class FieldDefinitionExtensions
+    {
+        public static Type GetFieldType<TDocument>(this FieldDefinition<TDocument> fieldDefinition)
+        {
+            var fieldDefinitionType = fieldDefinition.GetType();
+            if (fieldDefinitionType.IsConstructedGenericType)
+            {
+                if (fieldDefinitionType.GetGenericTypeDefinition() == typeof(ExpressionFieldDefinition<,>) ||
```
adelinowona, line 465: Not significant but probably good to introduce a variable for `fieldDefinitionType.GetGenericTypeDefinition()` rather than calling it multiple times here

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1852,24 +1877,61 @@ internal sealed class ElementMatchFilterDefinition<TDocument, TItem> : FilterDef
         private readonly FieldDefinition<TDocument> _field;
         private readonly FilterDefinition<TItem> _filter;
 
+        public ElementMatchFilterDefinition(FilterDefinition<TItem> filter)
+            // where TDocument : IEnumerable<TItem> (can only be checked at runtime)
+        {
+            _filter = filter;
+
+            var ienumerableItem = typeof(IEnumerable<>).MakeGenericType(typeof(TItem));
```
sanych-sun, line 1885: What the benefit to do `MakeGenericType`, instead of `typeof(IEnumerable<TItem>)`?

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1852,24 +1877,61 @@ internal sealed class ElementMatchFilterDefinition<TDocument, TItem> : FilterDef
         private readonly FieldDefinition<TDocument> _field;
         private readonly FilterDefinition<TItem> _filter;
 
+        public ElementMatchFilterDefinition(FilterDefinition<TItem> filter)
+            // where TDocument : IEnumerable<TItem> (can only be checked at runtime)
+        {
+            _filter = filter;
+
+            var ienumerableItem = typeof(IEnumerable<>).MakeGenericType(typeof(TItem));
+            if (!typeof(TDocument).Implements(ienumerableItem))
```
sanych-sun, line 1886: Why not simply `TDocument is IEnumerable<TItem>`?

src/MongoDB.Driver/FilterDefinitionBuilder.cs
```diff
@@ -1852,24 +1877,61 @@ internal sealed class ElementMatchFilterDefinition<TDocument, TItem> : FilterDef
...
+            if (enumerableSerializer != null)
             {
-                var arraySerializer = renderedField.FieldSerializer as IBsonArraySerializer;
+                var arraySerializer = enumerableSerializer as IBsonArraySerializer;
                 BsonSerializationInfo itemSerializationInfo;
                 if (arraySerializer == null || !arraySerializer.TryGetItemSerializationInfo(out itemSerializationInfo))
                 {
-                    var message = string.Format("The serializer for field '{0}' must implement IBsonArraySerializer and provide item serialization info.", renderedField.FieldName);
+                    var message = fieldName == null ?
+                        string.Format($"The serializer '{enumerableSerializer.GetType()}' must implement IBsonArraySerializer and provide item serialization info.") :
```
sanych-sun, line 1933: Why do we need `string.Format` here? It seems like string interpolation did all the work.

sanych-sun (changes requested):

src/MongoDB.Driver/FilterDefinitionBuilder.cs
BorisDog: As there might be more cases like this, I don't see much value in spending time to identify all of them now.
The proposed new check is not related to the new functionality, so it's not strictly necessary in this PR, so I propose to extract this change to a 4.0 ticket.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/FieldDefinition.cs
```diff
@@ -454,4 +454,24 @@ public override RenderedFieldDefinition Render(RenderArgs<TDocument> args)
             return new RenderedFieldDefinition(rendered.FieldName, rendered.UnderlyingSerializer);
         }
     }
+
+    internal static class FieldDefinitionExtensions
```
BorisDog, line 458: Now that the validation was removed, do we need this class?

BorisDog (changes requested):

src/MongoDB.Driver/FieldDefinition.cs
rstam: You are right, we don't.

BorisDog (approved):
LGTM!

adelinowona (approved):

PR #1606, opened 2025-01-30
CSHARP-5479: Fix AWS lambda tests

adelinowona:


evergreen/evergreen.yml
```diff
@@ -2067,8 +2071,6 @@ task_groups:
             - "AWS_ACCESS_KEY_ID"
             - "AWS_SECRET_ACCESS_KEY"
             - "AWS_SESSION_TOKEN"
-          env:
```
sanych-sun, line 2070: Are you sure we can remove this?

evergreen/evergreen.yml
```diff
@@ -2115,8 +2117,6 @@ task_groups:
...
-            VAULT_NAME: "gcpkms"
```
sanych-sun, line 2119: Are you sure we can remove this?

evergreen/evergreen.yml
```diff
@@ -2161,10 +2161,9 @@ task_groups:
             - "AWS_ACCESS_KEY_ID"
             - "AWS_SECRET_ACCESS_KEY"
             - "AWS_SESSION_TOKEN"
-          env:
-            VAULT_NAME: "atlas"
```
sanych-sun, line 2165: Are you sure we can remove this?

sanych-sun (changes requested):

evergreen/evergreen.yml
adelinowona: I reviewed the script in [driver-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/csfle/azurekms/setup-secrets.sh), and it appears that the `VAULT_NAME` variable is not utilized. Therefore, there's no benefit in setting it.

evergreen/evergreen.yml
adelinowona: same thing, VAULT_NAME is not used in the setup script

sanych-sun (approved):
LGTM

evergreen/evergreen.yml
adelinowona: In this case, the setup script does utilize the `VAULT_NAME` variable. However, setting it here won't affect its value in the script due to the way it is [declared](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/atlas/setup-secrets.sh#L8).

tests/FaasTests/LambdaTests/template.yaml
```diff
@@ -33,20 +33,6 @@ Resources:
...
-        Fn::Sub: ApplicationInsights-SAM-${AWS::StackName}
-      ResourceQuery:
-        Type: CLOUDFORMATION_STACK_1_0
-  ApplicationInsightsMonitoring:
-    Type: AWS::ApplicationInsights::Application
-    Properties:
-      ResourceGroupName:
-        Ref: ApplicationResourceGroup
-      AutoConfigurationEnabled: 'true'
 Outputs:
```
adelinowona, line 36: Removing this as it is not needed and was causing errors

PR #1608, opened 2025-02-05
Add LinqIntegrationTest file template

sanych-sun:
This changes is hard to review, but it adds LinqIntegrationTest file template for Rider/ReSharper.


rstam (approved):
LGTM

PR #1615, opened 2025-02-18
CSHARP-5494: Deprecate Hedged Reads options

papafe:


adelinowona (changes requested):
I think we expose the hedged options in the readPreference options using the `ReadPreferenceHedge` class. Should we deprecate that as well?

papafe:
> I think we expose the hedged options in the readPreference options using the `ReadPreferenceHedge` class. Should we deprecate that as well?

🤦 Thanks a lot for noticing, I'll take a look

adelinowona (commented):
Hmmm seeing all the warnings suppression is a bit ugly but I am not sure if there's another way to do this. We could add the suppression at the top of the file but I doubt we want to do that since it's best to know exactly what in a file you are suppressing than suppressing the whole file. cc @BorisDog

src/MongoDB.Driver/Core/ReadPreference.cs
```diff
@@ -174,7 +180,9 @@ public ReadPreference(
             ReadPreferenceMode mode,
             IEnumerable<TagSet> tagSets = null,
             TimeSpan? maxStaleness = null,
+#pragma warning disable CS0618 // Type or member is obsolete
```
BorisDog, line 183: Will using this constructor without the hedge parameter, generate an obsolete warning?
If so, we should probably provide a new ctor.

BorisDog (changes requested):
I think we should stick with the current approach for warnings suppression. It's not the most convenient option, but is the safest. As we don't want to ignore things we don't want to ignore :)

PR #1616, opened 2025-02-18
CSHARP-5493: Fix stackoverflow in EnumRepresentationConvention

papafe:


BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Jira/CSharp5493Tests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [Fact]
+        private void Test()
+        {
+            var pack = new ConventionPack { new EnumRepresentationConvention(BsonType.String) };
+            ConventionRegistry.Register("enumRecursive", pack, t => t == typeof(RecursiveClass));
+
+            var client = DriverTestConfiguration.Client;
+            var database = client.GetDatabase("test");
+            database.DropCollection("recursiveClass");
+            var collection = database.GetCollection<RecursiveClass>("recursiveClass");
```
BorisDog, line 38: Do we need to use an actual DB here?
Can this be tested by getting the `BsonClassMap` of the class?

BorisDog (approved):
The fix LGTM, few questions about the test.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -67,6 +69,14 @@ public EnumRepresentationConvention(BsonType representation, bool topLevelOnly)
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
+            var memberType = memberMap.MemberType;
+
+            if (!memberType.IsEnum && Nullable.GetUnderlyingType(memberType) == null && !memberType.IsArray &&
```
BorisDog, line 74: Please double check other conventions. `ObjectSerializerAllowedTypesConvention` for example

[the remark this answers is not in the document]
sanych-sun: Does it mean we will get some class map with a member that is not very well configured?

sanych-sun (changes requested):

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -67,6 +68,14 @@ public EnumRepresentationConvention(BsonType representation, bool topLevelOnly)
...
```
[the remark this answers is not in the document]
papafe: Yes, we apply it for collections of enum. Actually this issue is coming out from a PR of mine where we added support for collections of enum: https://github.com/mongodb/mongo-csharp-driver/pull/1574
Before we were retrieving the serializers only after checking if it was an enum or a nullable enum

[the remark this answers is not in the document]
papafe: I have inverted the condition though, I agree it's probably more readable like this.

sanych-sun (approved):
LGTM

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -67,6 +68,13 @@ public EnumRepresentationConvention(BsonType representation, bool topLevelOnly)
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
+            var memberType = memberMap.MemberType;
+
+            if (!(memberType.IsEnum || memberType.IsNullableEnum() || memberType.IsArray || typeof(IEnumerable).IsAssignableFrom(memberType)))
```
rstam, line 73: I had a hard time reading this. It seemed like it was completely backwards, until... I finally saw that tiny little `!` at the beginning.

This is fine for now, but a helper method would improve readability. Something like:

    if (!CouldAppy()) 
    { 
        return;
    }

or perhaps:

    if (CouldNotApply()) 
    { 
        return: 
    }

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -152,6 +153,14 @@ public bool AllowDefaultFrameworkTypes
         /// <param name="memberMap">The member map.</param>
         public void Apply(BsonMemberMap memberMap)
         {
+            var memberType = memberMap.MemberType;
+
+            if (memberType != typeof(object) && Nullable.GetUnderlyingType(memberType) == null && !memberType.IsArray &&
```
rstam, line 158: Nullable.GetUnderlyingType(memberType) == null

is hard to read... and it relies on subtleties (for example, why doesn't it throw an exception when `memberType` is not `Nullable`???).

We have a helper method for this:

    !memberType.IsNullable()

PR #1622, opened 2025-02-24
CSHARP-5507: Improve readability of some conventions code

papafe:


src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -90,6 +90,11 @@ IBsonSerializer Reconfigure(IBsonSerializer s)
         }
 
         // private methods
+        private bool CouldApply(Type memberType)
```
rstam, line 93: This could be a local method inside `Apply`.

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -170,6 +169,13 @@ public void Apply(BsonMemberMap memberMap)
             }
         }
 
+        // private methods
+        private bool CouldApply(Type memberType)
```
rstam, line 173: This could be a local method inside `Apply`.

rstam (changes requested):
Depending on what you think of my comments I could also LGTM as is.

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: Makes sense, I've moved this and the same method in the other convention.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -87,6 +87,9 @@ public void Apply(BsonMemberMap memberMap)
 
             IBsonSerializer Reconfigure(IBsonSerializer s)
                 => s.ValueType.IsEnum ? (s as IRepresentationConfigurable)?.WithRepresentation(_representation) : null;
+
+            bool CouldApply(Type type)
```
rstam, line 91: Could be `static`.

But it's not important.

rstam (approved):
LGTM

PR #1626, opened 2025-02-26
CSHARP-5469: Support $lookup in CSFLE and QE

papafe:


tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2535,7 +2752,7 @@ public void ViewAreProhibitedTest([Values(false, true)] bool async)
                         view,
                         async,
                         documents: new BsonDocument("test", 1)));
-                exception.Message.Should().Be("Encryption related exception: cannot auto encrypt a view.");
+                exception.Message.Contains("cannot auto encrypt a view");
```
papafe, line 2755: This was failing because the new error message was reporting also the view name. I've modified the test to be adherent to the prose test, that prescribe only that the exception message string must contain "cannot auto encrypt a view").

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2685,7 +2902,7 @@ private void AssertInnerEncryptionException<TInnerException>(Exception ex, strin
             where TInnerException : Exception
             => AssertInnerEncryptionException<TInnerException>(ex, e => e.Message.Should().Contain(exceptionMessageContains));
 
-        private IMongoClient ConfigureClient(
+        private IMongoClient ConfigureClient( //TODO Am I wrong or both this method and ConfigureClientEncrypted can create an encrypted client? It's confusing
```
papafe, line 2905: Just a comment

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2511,6 +2510,224 @@ public void RewrapManyDataKeyOptions_with_should_validate_provider_is_set()
...
+
+            PipelineDefinition<BsonDocument, BsonDocument> CreatePipeline(string pipelineJson)
+            {
+                return Bson.Serialization.BsonSerializer.Deserialize<List<BsonDocument>>(pipelineJson);
+            }
+        }
+
+        // 25. Test $lookup (case 9)
+        [Fact]
+        public void TestLookupUnsupported()
```
papafe, line 2686: Unfortunately this case cannot be done together with the others, as this prescribes that there is an exception thrown when using an older version of mongocryptid

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
```diff
@@ -13,7 +13,52 @@
...
+    <PropertyGroup>
+        <AnyBinaryMissing Condition="!Exists('$(MSBuildProjectDirectory)/runtimes/osx/native/libmongocrypt.dylib')
+        or !Exists('$(MSBuildProjectDirectory)/runtimes/linux/native/x64/libmongocrypt.so')
+        or !Exists('$(MSBuildProjectDirectory)/runtimes/linux/native/arm64/libmongocrypt.so')
+        or !Exists('$(MSBuildProjectDirectory)/runtimes/linux/native/alpine/libmongocrypt.so')
+        or !Exists('$(MSBuildProjectDirectory)/runtimes/win/native/mongocrypt.dll')">true</AnyBinaryMissing>
+        <LibMongoCryptAllSourceUrl>https://mciuploads.s3.amazonaws.com/libmongocrypt/all/master/33fdf65cce5a0c0cdd293c64ed40e4a8205c3ce0/libmongocrypt-all.tar.gz</LibMongoCryptAllSourceUrl>
+    </PropertyGroup>
+
+    <Target Name="DownloadAllNativeBinaries" BeforeTargets="BeforeBuild" Condition="'$(AnyBinaryMissing)' == 'true' and '$(LibMongoCryptAllSourceUrl)' != ''">
```
papafe, line 25: This is a new target that gets all the binaries from the same zip file. The target won't run if `LibMongoCryptAllSourceUrl` is empty, and then the usual targets for downloading binaries will do it.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -873,10 +872,10 @@ void AssertResult(Exception ex, string expectedExceptionInfo)
                 }
             }
 
-            Guid CreateDataKeyTestCaseStep(ClientEncryption testCaseClientEncription, BsonDocument masterKey, bool async)
+            Guid CreateDataKeyTestCaseStep(ClientEncryption testCaseClientEncryption, BsonDocument masterKey, bool async)
             {
                 var dataKeyOptions = new DataKeyOptions(masterKey: masterKey);
```
damieng, line 877: Is this going to support joining where different data keys are used for two different collections in the same query?

papafe: I'm not sure, but I don't think this is the use case for this PR. The main issue is that previous to this change collection that use automatic encryption could not be used in $lookup operations

damieng: Right but if each collection uses a different data key then $lookup is going to fail.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
papafe: The single zip file is being created by evergreen for the libmongocrpyt repo. 
For libmongocrypt releases, instead, there are as many zip files as targets and this is what is used by the `DownloadNativeBinary` target that is already in the csproj file currently. 

This new target has been added mostly to test the unreleased version of libmongocrypt, so we're not even sure we're going to keep this.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
papafe: Ah yes, I think you're correct. I don't think this is going to be supported.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
adelinowona: The newest version of libmongocrypt has just been released so I don't think we'll need this

adelinowona (commented):
The newest version of libmongocrypt has just been released as a heads up. However, I think there is still some value in exploring ways to make testing unreleased libmongocrypt versions easier. My biggest concern is that the `LibMongoCryptAllSourceUrl` will just download all the binaries which is a big download. Not sure if there is a way to also download specific binaries in the same way we do with the released libmongocrypt version. Something to investigate.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2511,6 +2510,223 @@ public void RewrapManyDataKeyOptions_with_should_validate_provider_is_set()
...
+                return Bson.Serialization.BsonSerializer.Deserialize<List<BsonDocument>>(pipelineJson);
+            }
+        }
+
+        // 25. Test $lookup (case 9)
+        [Fact]
+        public void TestLookupUnsupported()
+        {
+            RequireServer.Check().Supports(Feature.Csfle2QEv2).ClusterTypes(ClusterType.ReplicaSet);
+            RequireServer.Check().VersionLessThan(new SemanticVersion(8, 1, 0, "alpha2-67-gd329906"));
```
adelinowona, line 2690: same as my other comments above

adelinowona (changes requested):

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2511,6 +2510,223 @@ public void RewrapManyDataKeyOptions_with_should_validate_provider_is_set()
             exception.Should().BeOfType<ArgumentException>().Subject.ParamName.Should().Be("provider");
         }
 
+        // 25. Test $lookup (cases 1-8)
+        [Fact]
+        public void TestLookup()
+        {
+            RequireServer.Check().Supports(Feature.Csfle2QEv2).ClusterTypes(ClusterType.ReplicaSet);
+            RequireServer.Check().VersionGreaterThanOrEqualTo(new SemanticVersion(8, 1, 0, "alpha2-67-gd329906"));
```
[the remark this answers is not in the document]
BorisDog: We should make it a `Feature` then. So it can be used instead of `Feature.Csfle2QEv2`.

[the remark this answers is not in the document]
papafe: The issue here is that version 8.1.0 is not out yet, so RequireServer.Check().VersionGreaterThanOrEqualTo("8.1.0") will never be true (yet), so the test would not run. 

And I didn't think about the feature, it makes sense. For now I'll use the alpha version in the feature, then we can correct it later.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2511,6 +2510,223 @@ public void RewrapManyDataKeyOptions_with_should_validate_provider_is_set()
             exception.Should().BeOfType<ArgumentException>().Subject.ParamName.Should().Be("provider");
         }
 
+        // 25. Test $lookup (cases 1-8)
+        [Fact]
+        public void TestLookup()
+        {
+            RequireServer.Check().Supports(Feature.Csfle2QEv2).ClusterTypes(ClusterType.ReplicaSet);
```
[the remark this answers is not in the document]
papafe: You're right!

adelinowona (approved):
LGTM

PR #1631, opened 2025-03-07
CSHARP-5453: Improve field encryption usability with attributes/API

papafe:


papafe:
@sanych-sun @damieng Can you give a look at the current state of the API? It's mostly functioning too, but I would like to get some feedback before writing tests. 
Some notes:
- At the moment the naming is derived from the name the fields have in the json schema, as presented [in the documentation](https://www.mongodb.com/docs/manual/core/csfle/fundamentals/create-schema/#std-label-csfle-fundamentals-pattern-properties). 
- I'll take care of the xml docs only at the end.

papafe:
@sanych-sun @damieng A couple of notes for the review:
- I have decided to reuse `EncryptionAlgorithm` and not create a new enum, even though `EncryptionAlgorithm` contains some values that are not valid for the schema, but will throw an exception.
- `bsonType` is actually optional, differently from what we discussed together. For this reason I slightly modified the overloads to avoid ambiguities when calling the methods.
- The whole `encrypt` section of a field could be empty. This is supported in the PR

damieng (dismissed):
LGTM

sanych-sun (approved):
LGTM

src/MongoDB.Bson/Serialization/BsonTypeExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <summary>
+    /// A static class containing extension methods for <see cref="BsonType"/>.
+    /// </summary>
+    public static class BsonTypeExtensions
+    {
+        /// <summary>
+        /// Maps a <see cref="BsonType"/> to its corresponding string representation.
+        /// </summary>
+        /// <param name="type">The input type to map.</param>
+        public static string ToStringRepresentation(this BsonType type)
```
papafe: @rstam this is the new method

rstam: I'm not sure we should add this to the public API just so EF Core can use it. EF Core can easily duplicate this very small method.

Unless there is a clear scenario where a USER would find this useful, I wouldn't add it to the public API.

Definitely a good idea to centralize this conversion in an `internal` method for our own consumption.

If we do make this public I suggest it be done as a separate JIRA ticket so that users can see it.

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonTypeExtensions.cs
damieng: It's a useful method with a clear and defined purpose and limited API surface. If I need it in the EF provider chances are other people need it too and probably also had to reproduce it. They'd need it right now if they're doing their own queryable encryption or CSFLE work and are trying to leverage BsonType in order to generate the schema.

papafe: @rstam What do you think?

rstam: > They'd need it right now if they're doing their own queryable encryption

Not really. They would just use the string constants directly.

rstam: `ToStringRepresentation` sounds a little vague.

This is not just ANY string representation, this is the string representation the SERVER uses.

If we are going to add this to the public API (which I think is questionable) maybe a better name is called for, something like `ToServerString`.

I don't know why you chose `ToStringRepresentation` instead of `ToString`. I'm assuming it's because of conflicts with `Object.ToString`?

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonTypeExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+using System;
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// A static class containing extension methods for <see cref="BsonType"/>.
+    /// </summary>
+    public static class BsonTypeExtensions
```
rstam: While I'm not convinced we should add this class, if we do add it, I think this is the wrong location.

I would put it in the `ObjectModel` folder right next to the `BsonType.cs` file.

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonTypeExtensions.cs
papafe: I agree with this, moved.

src/MongoDB.Bson/Serialization/BsonTypeExtensions.cs
papafe: Yes, I didn't use `ToString` otherwise it gets shadowed by `Object.ToString()`. 
I would be fine to change the name to something more representative if we're ok with putting this in the public API.

PR #1632, opened 2025-03-11
CSHARP-5519: In a filter Any with array constant and predicate comparing item to field should translate to $in.

rstam:
There are test failing but I want to review as is first and discuss.

The following tests fail:

1. Tests with a filter with nested AsQueryable
2. Tests with Select

Let's plan a Zoom meeting to discuss why they are failing. Potentially we might just accept these as limitations and open new JIRA tickets for them.

Other than those issues I think the rest of the PR is in good shape.


rstam:
Closing this PR as it will be fixed by CSHARP-5572.

PR #1633, opened 2025-03-11
CSHARP-5514: Revert skip to non-lb-connection-establishment

adelinowona:
Revert test skip from https://github.com/mongodb/mongo-csharp-driver/pull/1618

The server problem that initiated the test skip has been resolved so we can revert the skip and run the test on the latest server build.


BorisDog (approved):
LGTM. Just double check the EG.

PR #1634, opened 2025-03-12
CSHARP-5472: Investigate changes in SERVER-97071: Ensure stability of resumeTokens across versions

sanych-sun:


src/MongoDB.Driver/Core/ChangeStreamNamespaceTypeSerializer.cs
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Serializers;
+
+namespace MongoDB.Driver.Core
+{
+    internal static class ChangeStreamNamespaceTypeSerializer
```
sanych-sun, line 21: This class needed only to hold the instance of `EnumSerializer<ChangeStreamNamespaceType>`. Cannot have that as static variable in `ChangeStreamDocumentSerializer` because it's generic. Open for any suggestion for a better place to have this instance variable

PR #1637, opened 2025-03-18
CSHARP-5515: Implement $convert binData <-> int, long

papafe:


src/MongoDB.Driver/Mql.cs
```diff
@@ -152,5 +152,221 @@ public static bool IsNullOrMissing<TField>(TField field)
         {
             throw CustomLinqExtensionMethodHelper.CreateNotSupportedException();
         }
+
+        /// <summary>
```
papafe, line 156: I will add the docs at the end

papafe:
@rstam This is a draft, as there are a couple of things missing, but I wanted to get some initial feedback.
Regarding the `ConvertToBinDataFromType(Type value, BsonBinarySubType subtype, string format, BsonBinaryData onError, BsonBinaryData onNull)` methods, I'm unsure if we should allow `onError` and `onNull` to be nulls.
What do you think?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -0,0 +1,551 @@
...
+                new TestClass {Id = 2, BinaryProperty = new BsonBinaryData(Guid.Parse("867dee52-c331-484e-92d1-c56479b8e67e"), GuidRepresentation.Standard), StringProperty = "867dee52-c331-484e-92d1-c56479b8e67e"},
+                new TestClass {Id = 3, BinaryProperty = new BsonBinaryData(Convert.FromBase64String("AAAAAAAA4L8=")), DoubleProperty = -0.5},  //LittleEndian
+                new TestClass {Id = 4, BinaryProperty = new BsonBinaryData(Convert.FromBase64String("ogIAAA==")), IntProperty = 674, LongProperty = 674}, //LittleEndian
+                new TestClass {Id = 5, BinaryProperty = new BsonBinaryData(Convert.FromBase64String("wAQAAAAAAAA=")), DoubleProperty = -2.5},  //BigEndian
+                new TestClass {Id = 6, BinaryProperty = new BsonBinaryData(Convert.FromBase64String("AAAAKg==")), IntProperty = 42, LongProperty = 42}, //BigEndian
+            ];
+
+            protected override void InitializeTestCase()
+            {
+                base.InitializeTestCase();
```
papafe, line 524: In another ticket we could extend `MongoCollectionFixture` to accept both typed and untyped `InitialData`

sanych-sun: I suppose for now you can use ClassFixture that accepts 2 type-parameter and initialize collection from BsonDocuments:
https://github.com/mongodb/mongo-csharp-driver/blob/cf2892b1a84f8ad13d61b901cac2a08a929ccde7/tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4880Tests.cs#L58

papafe: Seems like a good plan!

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -22,27 +22,39 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
     internal sealed class AstConvertExpression : AstExpression
     {
         private readonly AstExpression _input;
+        private readonly string _format;
         private readonly AstExpression _onError;
         private readonly AstExpression _onNull;
         private readonly AstExpression _to;
+        private readonly AstExpression _subType;
```
rstam, line 29: Data type should be `BsonBinarySubType?` because the server requires this to be a constant.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -22,27 +22,39 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
...
+        private readonly Mql.ByteOrder? _byteOrder;
```
rstam, line 30: Nit: keep fields in alphabetical order.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -22,27 +22,39 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
...
 
         public AstConvertExpression(
             AstExpression input,
             AstExpression to,
             AstExpression onError = null,
-            AstExpression onNull = null)
+            AstExpression onNull = null,
+            AstExpression subType = null,
+            string format = null,
+            Mql.ByteOrder? byteOrder = null)
```
rstam, line 39: I would propose the following order for the parameters:

```
 public AstConvertExpression(
     AstExpression input,
     AstExpression to,
     BsonBinarySubType? subType = null,
     Mql.ByteOrder? byteOrder = null,  
     string format = null,             
     AstExpression onError = null,     
     AstExpression onNull = null)      
```

The thinking is that `to/subType/byteOrder/format` are tightly related and should be together.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -22,27 +22,39 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
...
         }
 
+        public Mql.ByteOrder? ByteOrder => _byteOrder;
         public AstExpression Input => _input;
+        public string Format => _format;
         public override AstNodeType NodeType => AstNodeType.ConvertExpression;
         public AstExpression OnError => _onError;
         public AstExpression OnNull => _onNull;
         public AstExpression To => _to;
+        public AstExpression SubType => _subType;
```
rstam, line 57: Nit: keep properties in alphabetical order.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -68,14 +88,18 @@ public AstConvertExpression Update(
             AstExpression input,
             AstExpression to,
             AstExpression onError,
-            AstExpression onNull)
+            AstExpression onNull,
+            AstExpression subType,
+            string format,
+            Mql.ByteOrder? byteOrder)
```
rstam, line 94: Proposed the following order for the parameters:

```
public AstConvertExpression Update(
    AstExpression input,           
    AstExpression to,              
    AstExpression onError,         
    AstExpression onNull)          
```

Note that any fields that are not `AstExpressions` are not visited and therefore don't need to be parameters to the `Update` method.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -255,16 +255,15 @@ public static AstExpression Constant(BsonValue value)
             return new AstConstantExpression(value);
         }
 
-        public static AstExpression Convert(AstExpression input, AstExpression to, AstExpression onError = null, AstExpression onNull = null)
+        public static AstExpression Convert(AstExpression input, AstExpression to, AstExpression onError = null, AstExpression onNull = null,
+            AstExpression subType = null, string format = null, Mql.ByteOrder? byteOrder = null)
```
rstam, line 259: Proposed order of parameters:

```
public static AstExpression Conv
    AstExpression input,
    AstExpression to,
    BsonBinarySubType? subType =
    Mql.ByteOrder? byteOrder = n
    string format = null,
    AstExpression onError = null
    AstExpression onNull = null)
```

to match the proposed order of the `AstConvertExpression` constructor parameters.

src/MongoDB.Driver/Mql.cs
```diff
@@ -49,6 +49,213 @@ public static TValue Constant<TValue>(TValue value, BsonType representaion)
...
+        /// <param name="value">The value.</param>
+        /// <param name="byteOrder">The byte ordering of BsonBinaryData.</param>
+        /// <param name="onError">The onError value.</param>
+        /// <param name="onNull">The onNull value.</param>
+        /// <returns>The converted value.</returns>
+        public static double? ConvertToNullableDouble(BsonBinaryData value, ByteOrder byteOrder, double? onError, double? onNull)
+        {
+            throw CustomLinqExtensionMethodHelper.CreateNotSupportedException();
+        }
+
```
rstam, line 258: I was surprised to see all these method names prefixed with `Convert`. I was expecting `ToBsonBinaryData`, `ToInt32`, `ToString`, etc...

What do others think?

rstam: I don't think this set of methods is quite right. I propose the following set:

```
// BsonBinaryData -> double
public static double ToDouble(BsonBinaryData value, ByteOrder byteOrder)
public static double ToDouble(BsonBinaryData value, ByteOrder byteOrder, double onError, double onNull)

// BsonBinaryData -> double?
public static double? ToNullableDouble(BsonBinaryData value, ByteOrder byteOrder)
public static double? ToNullableDouble(BsonBinaryData value, ByteOrder byteOrder, double? onError, double? onNull)

// BsonBinaryData -> int
public static int ToInt32(BsonBinaryData value, ByteOrder byteOrder)
public static int ToInt32(BsonBinaryData value, ByteOrder byteOrder, int onError, int onNull)

// BsonBinaryData -> int?
public static int? ToNullableToInt32(BsonBinaryData value, ByteOrder byteOrder)
public static int? ToNullableToInt32(BsonBinaryData value, ByteOrder byteOrder, int? onError, int? onNull)

// BsonBinaryData -> long
public static long ToInt64(BsonBinaryData value, ByteOrder byteOrder)
public static long ToInt64(BsonBinaryData value, ByteOrder byteOrder, long onError, long onNull)

// BsonBinaryData -> long?
public static long? ToNullableToInt64(BsonBinaryData value, ByteOrder byteOrder)
public static long? ToNullableToInt64(BsonBinaryData value, ByteOrder byteOrder, long? onError, long? onNull)

// BsonBinaryData -> string
public static string ToString(BsonBinaryData value, string format)
public static string ToString(BsonBinaryData value, string format, string onError, string onNull)

// double -> BsonBinaryData
public static BsonValue ToBsonBinaryData(double value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(double value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// double? -> BsonBinaryData
public static BsonValue ToBsonBinaryData(double? value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(double? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// int -> BsonBinaryData
public static BsonValue ToBsonBinaryData(int value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(int value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// int? -> BsonBinaryData
public static BsonValue ToBsonBinaryData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// long -> BsonBinaryData
public static BsonValue ToBsonBinaryData(long value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(long value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// long? -> BsonBinaryData
public static BsonValue ToBsonBinaryData(long? value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonValue ToBsonBinaryData(long? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// string -> BsonBinaryData
public static BsonValue ToBsonBinaryData(string value, BsonBinarySubType subType, string format)
public static BsonValue ToBsonBinaryData(string value, BsonBinarySubType subType, string format, BsonValue onError, BsonValue onNull)
```

The `ToBsonBinaryData` methods are tricky. The result could be *either* a `BsonBinaryData` instance or a `BsonNull` instance. The only common subtype is `BsonValue`.

Also, we could consider adding overloads using `byte[]` for every existing overload using `BsonBinaryData`. A `byte[]` could be considered to be the same thing as a `BsonBinaryData` value with a `subType` of `0`.

rstam (changes requested):

src/MongoDB.Driver/Mql.cs
```diff
@@ -152,5 +359,20 @@ public static bool IsNullOrMissing<TField>(TField field)
         {
             throw CustomLinqExtensionMethodHelper.CreateNotSupportedException();
         }
+
+        /// <summary>
+        ///
+        /// </summary>
+        public enum ByteOrder
```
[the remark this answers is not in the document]
papafe: I'm not sure either, but it's only used here, so I thought it made sense to keep it close to the methods where it is used.

rstam (changes requested):

src/MongoDB.Driver/Mql.cs
papafe: Regarding the naming, I've added a thread on slack to discuss this, so we can reach a common decision. 

I agree we could add the "non-nullable" versions of the `ToBsonBinaryData` methods. Regarding the methods in the other direction, should we also have both versions? E.g. `int? ToNullableInt(data)/int ToInt(data)`?

Regarding the "nullability" of `BsonBinaryData`, I was thinking about it as well. I'm not sure what we should do here as we can't even serialize a null value for `BsonBinaryData`.  I can see that when serializing a POCO with a BsonBinaryData property with a value of null we use a `{ _csharpnull: true }` document to signal it.  We could do something similar here, but it seems tricky and unintuitive. 

Regarding the overloads, I think we could add them on request, if developers ask for them.

src/MongoDB.Driver/Mql.cs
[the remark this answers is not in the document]
rstam: I understand your thinking and it does make sense, but... it makes the enum more awkward for the user to use. They have to type `Mql.ByteOrder.LittleEndian` instead of `ByteOrder.LittleEndian`.

There are lots of similar places where we have consistently chosen to *not* nest the enums for this reason.

rstam (changes requested):

src/MongoDB.Driver/Mql.cs
rstam: I would stay away from `{ _csharpnull : true }`. It may or may not have been a good idea in the first place.

It was introduced a very long time ago as a way to distinguish between a `BsonValue` property value of C# null vs `BsonNull.Instance`.

I'm not sure we even handle `_csharpnull` correctly anywhere outside of serialization. For example, a LINQ query may or may not work. There's a chance it would work because the correct serializer would be invoked, but I wouldn't be surprised if there are edge cases that might not be handled correctly.

In any case, on the SERVER side there is no such thing as `{ _csharpnull : true }`. Server side, the result of `{ $convert : { input: <value>,  to : "binData", ... } }` will NEVER be `{ _csharpnull : true }`. It will either be an instance of `BsonBinaryData` or `BsonNull` or whatever `onError` or `onNull` returned.

rstam (changes requested):

src/MongoDB.Driver/Mql.cs
papafe: Yes, I agree it would be confusing. 

So, now the question is if we allow at all to have `onNull` and `onError` to have a null value in those conversion methods. Apart from rendering the operator, I think the issue here would also be deserializing a null value in `BsonBinaryData`, right?

If we don't allow `onNull` and `onError` to be null, then there will be no possible runtime errors, even though we're limiting the conversion.

rstam: The arguments for `onNull` and `onError` *cannot* be `null`.

They need to be expressions that can be translated to MQL.

If you want the result to be "null" on the server you use `BsonNull.Value` as the value for `onError` or `onNull`.

A "null" on the server is not the same thing as `null` in C#.

rstam (changes requested):

src/MongoDB.Driver/Mql.cs
rstam: Let me rephrase that:

A value of `null` for either the `onError` or `onNull` arguments CANNOT be translated to MQL.

It COULD be interpreted (by the LINQ translator) to mean you DID NOT provide a value for `onError` or `onNull` and therefore do not want that field in the MQL.

So you could write something like this:

```Mql.ToInt32(x.BinaryField, ByteOrder.LittleEndian, onError: null, onNull: 0)```

Which would be the way of saying you do NOT want to specify an `onError` argument. You want errors to result in the query failing.

It would translate to:

```{ $convert : { input : "$BinaryField", to : "int", byteOrder : "little", onNull : 0 } }```

Notice there is no `onError` field in that MQL.

**Review by rstam — CHANGES_REQUESTED** [review]

File: src/MongoDB.Driver/Mql.cs (hunk 9 above)
Comment (papafe) [reply src/MongoDB.Driver/Mql.cs:258]: Ok, I see your point. So far, I'm allowing `onError` and `onNull` to be null values, since all the parameters have been nullable. In this case they get considered the same as specifying `BsonNull`. 

If we don't allow null values for `onError` and `onNull`, then how do we allow developers to specify `BsonNull` values for `onError` and `onNull`?

Comment (rstam) [reply src/MongoDB.Driver/Mql.cs:258]: > So far, I'm allowing onError and onNull to be null values, since all the parameters have been nullable. In this case they get considered the same as specifying BsonNull.

That's a possible way to set it up. But then you can't just serialize a C# `null` using the proper serializer, you probably have to handle it yourself.

This is difficult, because the server has NO concept of `null` similar to C#, where `null` is a special pointer (typically all zeroes in the underlying representation) that points to nothing.

On the server, "null" is just one more `BsonValue` just like all the other ones.

Comment (rstam) [reply src/MongoDB.Driver/Mql.cs:258]: By the way, in the comment before the previous one (with the 3 scenarios), the `Optional<T>` class was designed precisely to tell the difference between an explicit `null` value and not passing a value at all.

When you have a method like `foo(int? arg = null)` you can't tell the difference between the following two:

```
foo(); // arg omitted and defaults to `null`
foo(null); // explicit null provided
```

But it the method is declared `foo(Optional<int?> arg)` you can:

```
foo(); // arg omitted and defaults to `default(Optional<int?>)` (which is not the same as `null`)
foo(null); // explicit null provided
```

But... Boris doesn't like `Optional<T>` so that might be a non-starter.

**Review by rstam — CHANGES_REQUESTED** [review]

File: src/MongoDB.Driver/Mql.cs (hunk 9 above)
Comment (rstam) [reply src/MongoDB.Driver/Mql.cs:258]: Actually, I should take that all back (mostly). While `Optional<T>` does work as I described, you can't omit optional arguments in a LINQ expression, so we should not design any API for use in LINQ that relies on optional arguments.

Comment (papafe) [reply src/MongoDB.Driver/Mql.cs:258]: Ok, so we go back to the "problem-specific" `Optional`-like class I was thinking 😄 
Something like
```
var onNull = OnValue.Create<int?>(23); //with value
var onNull = OnValue.Create<int?>(null); //with BsonNull
var onNull = OnValue.Default<int?>(); //with default value (not included in the MQL)
```
What do you think? This way we never use optional values.

Comment (rstam) [reply src/MongoDB.Driver/Mql.cs:258]: I think the easiest way is to simply have overloads with and without `onError` and `onNull` in every combination.

If the overload includes `onError` then the MQL will also. If not not, and same for `onNull`.

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs (hunk 3 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs]: Fixed

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs (hunk 4 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs]: Fixed

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs (hunk 5 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs]: Fixed

File: src/MongoDB.Driver/Mql.cs (hunk 10 above)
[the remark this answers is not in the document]
Comment (papafe) [reply src/MongoDB.Driver/Mql.cs]: Done, moved to outer scope but kept it in the same file.

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs (hunk 7 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs]: Fixed

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs (hunk 8 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs]: Fixed

File: tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs (hunk 2 above)
Comment (papafe) [reply tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs:524]: Done

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs (hunk 6 above)
Comment (papafe) [reply src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs]: Done

File: src/MongoDB.Driver/Mql.cs (hunk 9 above)
Comment (papafe) [reply src/MongoDB.Driver/Mql.cs:258]: Renaming done.

Comment (papafe) [reply src/MongoDB.Driver/Mql.cs:258]: Regarding the API method to include...

#### ToNullableInt (same for other nullable methods including ToString)
```csharp
// To include
public static int? ToNullableInt(BsonBinaryData value, ByteOrder byteOrder)
public static int? ToNullableInt(BsonBinaryData value, ByteOrder byteOrder, int? onError)
public static int? ToNullableInt(BsonBinaryData value, ByteOrder byteOrder, int? onError, int? onNull)

// To NOT include
public static int? ToNullableInt(BsonBinaryData value, ByteOrder byteOrder, int? onNull) //Not included as it is the same as previous one with onNull = null
```
I think we can skip the last method as it is the same as the method with both `onError` and `onNull` and `onNull = null` (if `onNull` is not specified then it will get a value of `null`). This way we also avoid having two methods with just `onNull`/`onError` for which we'd need to have different names.

#### ToInt
```
// To include
public static int ToInt(BsonBinaryData value, ByteOrder byteOrder, int onNull)
public static int ToInt(BsonBinaryData value, ByteOrder byteOrder, int onError, int onNull)

// To NOT include
public static int ToInt(BsonBinaryData value, ByteOrder byteOrder)
public static int ToInt(BsonBinaryData value, ByteOrder byteOrder, int onError)
```
We can only add methods with `onNull`, because if not specified then it will have a default value of `null`, and then we'll have issues as we've defined a return value of `int`, not `int?`

Talking about the non-nullable `To` methods (`ToInt`, `ToDouble`, `ToLong`), do we really need those? Everything that can be done with those can be done also with the equivalent nullable version, even though slightly less convenient. 

#### ToBinData
```csharp
// To include
public static BsonBinaryData ToBinData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder)
public static BsonBinaryData ToBinData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError)
public static BsonBinaryData ToBinData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError, BsonValue onNull)

// To NOT include
public static BsonBinaryData ToBinData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder, BsonValue onError) 
```
The reasons not to include the method that only has `onError` is the same as before with `ToNullableInt`. 

For the `ToBinData` methods, also, do we really need to have the non-nullable first parameter version? For example:
```csharp
public static BsonBinaryData ToBinData(int value, BsonBinarySubType subtype, ByteOrder byteOrder)
```
To me having this does not seem particularly useful, as we have the implicit conversion between `int` and `int?`, and the result value is independent from the nullability. Am I missing something?

rstam: The result of `ToBsonBinaryData` is NOT the same for `int` or `int?` values.

For `int` the result is known to be `BsonBinaryData` (because the value can't be `null`).

For `int?` the result is either `BsonBinaryData` or `BsonNull` depending on whether the `int?` value is `null` or not.

src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -193,6 +195,16 @@ public class Feature
...
+        /// <summary>
+        /// Gets the conversion of binary data to/from numeric types feature.
+        /// </summary>
+        public static Feature ConvertBinDataToFromNumeric => __convertBinDataToFromNumeric;
+
+        /// <summary>
+        /// Gets the conversion of binary data to/from string feature.
+        /// </summary>
+        public static Feature ConvertBinDataToFromString => __convertBinDataToFromString;
+
```
rstam, line 207: A feature representing an `$xyz` operator is named `XyzOperator`.

In this case the `$convert` operator is already supported, so we need to qualify `ConvertOperator` with what is newly supported.

Suggested names:

```
ConvertOperatorBinDataToFromNumeric
ConvertOperatorBinDataToFromString
```

Same for backing fields.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -21,27 +21,39 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
...
             AstExpression onError = null,
             AstExpression onNull = null)
         {
             _input = Ensure.IsNotNull(input, nameof(input));
             _to = Ensure.IsNotNull(to, nameof(to));
             _onError = onError;
             _onNull = onNull;
+            _subType = subType;
+            _format = format;
+            _byteOrder = byteOrder;
```
rstam, line 47: Maybe put assignments in same order as the parameters?

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -255,16 +255,16 @@ public static AstExpression Constant(BsonValue value)
             return new AstConstantExpression(value);
         }
 
-        public static AstExpression Convert(AstExpression input, AstExpression to, AstExpression onError = null, AstExpression onNull = null)
+        public static AstExpression Convert(AstExpression input, AstExpression to,
+            BsonBinarySubType? subType = null, ByteOrder? byteOrder = null, string format = null,
+            AstExpression onError = null, AstExpression onNull = null)
```
rstam, line 260: I recommend that if you can't fit all parameters on one line that you switch to a separate line for each parameter. I find this more readable:

```
 public static AstExpression Convert(
     AstExpression input,
     AstExpression to,
     BsonBinarySubType? subType = null,
     ByteOrder? byteOrder = null,      
     string format = null,             
     AstExpression onError = null,     
     AstExpression onNull = null)      
```

src/MongoDB.Driver/Mql.cs
```diff
@@ -49,6 +49,213 @@ public static TValue Constant<TValue>(TValue value, BsonType representaion)
...
         }
 
+        /// <summary>
+        /// Converts a string to a BsonBinaryData using the $convert aggregation operator.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <param name="subtype">The BsonBinaryData subtype of the result value.</param>
+        /// <param name="format">The format string.</param>
+        /// <returns>The converted value.</returns>
+        public static BsonBinaryData ToBinData(string value, BsonBinarySubType subtype, string format)
```
rstam, line 59: I've been calling this method `ToBsonBinaryData` but you keep using `ToBinData`.

I understand it's shorter, but it's not accurate. There is no data type called `BinData`.

src/MongoDB.Driver/Mql.cs
```diff
@@ -49,6 +49,213 @@ public static TValue Constant<TValue>(TValue value, BsonType representaion)
...
+        }
+
+        /// <summary>
+        /// Converts an int to a BsonBinaryData using the $convert aggregation operator.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <param name="subtype">The BsonBinaryData subtype of the result value.</param>
+        /// <param name="byteOrder">The byte order of BsonBinaryData.</param>
+        /// <returns>The converted value.</returns>
+        public static BsonBinaryData ToBinData(int? value, BsonBinarySubType subtype, ByteOrder byteOrder)
```
rstam, line 86: We need separate overloads for `int` and `int?`.

A property in a user's POCO might be an `int` or it might be an `int?`.

I realize there is an automatic conversion from `int` to `int?` but I don't know if that's how we would want to handle that.

rstam: Same for `long` and `double`.

src/MongoDB.Driver/Mql.cs
```diff
@@ -49,6 +49,213 @@ public static TValue Constant<TValue>(TValue value, BsonType representaion)
...
+            throw CustomLinqExtensionMethodHelper.CreateNotSupportedException();
+        }
+
+        /// <summary>
+        /// Converts a BsonBinaryData to int? using the $convert aggregation operator.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <param name="byteOrder">The byte ordering of BsonBinaryData.</param>
+        /// <returns>The converted value.</returns>
+        public static int? ToNullableInt(BsonBinaryData value, ByteOrder byteOrder)
```
rstam, line 193: We need separate `ToInt32` and `ToNullableInt32` methods so that the user can control the type of the result.

They might not want the result to be nullable.

src/MongoDB.Driver/Mql.cs
rstam: Ferdinando and I discussed in Slack an approach with an options class that would allow the user to write something like the following:

```
Mql.ToInt(x.BinaryField, ByteOrder.LittleEndian))
Mql.ToInt(x.BinaryField, ByteOrder.LittleEndian, new() { OnError. = 0 })
Mql.ToInt(x.BinaryField, ByteOrder.LittleEndian, new() { OnNull = 0 })
Mql.ToInt(x.BinaryField, ByteOrder.LittleEndian, new() { OnError = 0, OnNull = 0 }))
```
This would require only two overloads. One with `options` and one without.

The options class can keep track whether `OnError` or `OnNull` were set or not.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -0,0 +1,532 @@
...
+            ];
+        }
+
+        public class TestClass
+        {
+            public int Id { get; set; }
+            public BsonBinaryData BinaryProperty { get; set; }
+            public double? DoubleProperty { get; set; }
+            public int? IntProperty { get; set; }
+            public long? LongProperty { get; set; }
```
rstam, line 528: These properties are named incorrectly because they are nullable. They should be named:

```
public double? NullableDoubleProperty { get; set; }     
public int? NullableIntProperty { get; set; }
public long? NullableLongProperty { get; set; }
```

At the same time, we need to add the following three NON-nullable properties to test against:

```
public double DoubleProperty { get; set; }     
public int IntProperty { get; set; }
public long LongProperty { get; set; }
```

rstam (changes requested):

src/MongoDB.Driver/Core/Misc/Feature.cs
papafe: Done

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
papafe: Fixed

src/MongoDB.Driver/Mql.cs
papafe: Fixed

src/MongoDB.Driver/Mql.cs
papafe: Fixed

src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -44,6 +44,8 @@ public class Feature
         private static readonly Feature __clientBulkWrite = new Feature("ClientBulkWrite", WireVersion.Server80);
         private static readonly Feature __clientSideEncryption = new Feature("ClientSideEncryption", WireVersion.Server42);
         private static readonly Feature __clusteredIndexes = new Feature("ClusteredIndexes", WireVersion.Server53);
+        private static readonly Feature __convertOperatorBinDataToFromNumeric = new Feature("ConvertBinDataToFromNumeric", WireVersion.Server81);
+        private static readonly Feature __convertOperatorBinDataToFromString= new Feature("ConvertBinDataToFromString", WireVersion.Server80);
```
rstam, line 48: Nit: change string constants to match feature name.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -21,27 +21,35 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions
 {
     internal sealed class AstConvertExpression : AstExpression
     {
+        private readonly ByteOrder? _byteOrder;
         private readonly AstExpression _input;
-        private readonly AstExpression _onError;
-        private readonly AstExpression _onNull;
+        private readonly string _format;
+        private readonly ConvertOptions _options;
```
rstam, line 27: Undo all these changes at the AST level.

These changes are not necessary and the AST level should be self contained and not depend on any classes like `ConvertOptions` defined at a higher level in the public API.

rstam (changes requested):
Quick initial pass. More thorough review later today.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
papafe: I see your point, I restored it to the older version. I wasn't sure of at what level we wanted to make the conversion.

src/MongoDB.Driver/Core/Misc/Feature.cs
papafe: Fixed

PR #1648, opened 2025-03-27
CSHARP-5541: move rankFusion methods to IAggregateFluentExtensions

adelinowona:


rstam:
I'm thinking about this before reviewing further. It feels so wrong.

BorisDog (approved):
LGTM

rstam (approved):
This feels wrong, but sounds like our hand is forced. We should consider switching from interfaces to classes for our public API in 4.0.

PR #1654, opened 2025-03-28
CSHARP-5545: Fix flaky test .TransactionsConvenientApi_testCase__transaction-options.json_withTransaction_explicit_transaction_options_override_client_options

sanych-sun:


src/MongoDB.Driver/Core/Clusters/ServerSelectors/WritableServerSelector.cs
```diff
@@ -79,6 +79,7 @@ public IEnumerable<ServerDescription> SelectServers(ClusterDescription cluster,
             var serversList = servers.ToList(); // avoid multiple enumeration
             if (CanUseSecondaries(cluster, serversList))
             {
+                _mayUseSecondary.EffectiveReadPreference = _mayUseSecondary.ReadPreference;
```
sanych-sun, line 82: This code is making sure that we use proper ReadPreference in case previous server selection attempt felled back to use Primary (for the corner case when the first attempt happening when cluster was still of Unknown type).

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -132,14 +132,10 @@ public void Run(
                 KillOpenTransactions(DriverTestConfiguration.Client);
             }
 
-            _entityMap = UnifiedEntityMap.Create(_eventFormatters, _loggingService.LoggingSettings, async);
+            BsonDocument lastKnownClusterTime = AddInitialData(DriverTestConfiguration.Client, initialData);
+            _entityMap = UnifiedEntityMap.Create(_eventFormatters, _loggingService.LoggingSettings, async, lastKnownClusterTime);
```
sanych-sun, line 136: Changing order of operations help to reduce chance of "catalog changed" problem. Double-checked with Java, they also process initial data prior to create entities.

BorisDog (approved):
LGTM

PR #1659, opened 2025-04-01
CSHARP-5552: Add support for $convert in LINQ

papafe:


papafe:
@rstam This is the new PR for the $convert operator support. It's still a draft because I'm trying to understand some details.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -56,9 +68,17 @@ public override BsonValue Render()
                 { "$convert", new BsonDocument
                     {
                         { "input", _input.Render() },
-                        { "to", _to.Render() },
+                        { "to", _to.Render(), _subType == null },
+                        { "to", () => new BsonDocument
+                            {
+                                {"type", _to.Render() },
+                                {"subtype", (int)_subType!.Value},
+                            }, _subType != null
```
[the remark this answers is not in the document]
rstam: In other words, like this:

     { "to", () => new BsonDocument
         {
             { "type", _to.Render() },
             { "subtype", (int)_subType!.Value },
         },                                      
         _subType != null                        
     },

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -56,9 +68,17 @@ public override BsonValue Render()
...
+                        },
                         { "onError", () => _onError.Render(), _onError != null },
-                        { "onNull", () => _onNull.Render(), _onNull != null }
+                        { "onNull", () => _onNull.Render(), _onNull != null },
+                        { "format", () => _format, _format != null},
+                        { "byteOrder", () => _byteOrder!.Value.Render(), _byteOrder != null}
```
rstam, line 81: nit: space after { and before }

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
```diff
@@ -255,16 +255,21 @@ public static AstExpression Constant(BsonValue value)
...
         {
             Ensure.IsNotNull(input, nameof(input));
             Ensure.IsNotNull(to, nameof(to));
 
-            if (to is AstConstantExpression toConstantExpression &&
-                (toConstantExpression.Value as BsonString)?.Value is string toValue &&
-                toValue != null &&
-                onError == null &&
-                onNull == null)
+            if (onError == null && onNull == null && subType == null && format == null && byteOrder == null &&
```
rstam, line 270: TODO: this simplification seems like it belongs in `AstSimplifier.cs`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/AstEnumExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            };
+        }
+
+        public static string Render(this ByteOrder byteOrder)
+        {
+            return byteOrder switch
+            {
+                ByteOrder.BigEndian => "big",
+                ByteOrder.LittleEndian => "little",
+                _ => throw new ArgumentException($"Unexpected Mql.ByteOrder: {byteOrder}.", nameof(byteOrder))
```
rstam, line 57: It's just `ByteOrder`, not `Mql.ByteOrder`. Consider using `nameof`:

    _ => throw new ArgumentException($"Unexpected {nameof(ByteOrder)}: {byteOrder}.", nameof(byteOrder))

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            var method = expression.Method;
+            var arguments = expression.Arguments;
+
+            if (!method.Is(MqlMethod.Convert))
+            {
+                throw new ExpressionNotSupportedException(expression);
+            }
+
+            AstExpression onErrorAst = null;
```
rstam, line 28: Generally declare variables at the last minute just before they are needed. Move to just before line 40.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,85 @@
...
+            AstExpression onNullAst = null;
```
rstam, line 29: Move this to just before line 45.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var fieldAst = ExpressionToAggregationExpressionTranslator.Translate(context, arguments[0]).Ast;
+
+            if (arguments[1] is not ConstantExpression constantExpression)
+            {
+                throw new InvalidOperationException("The 'options' argument must be a constant expression");
```
rstam, line 35: I don't think `options` can be restricted to a constant.

The `OnError` and `OnNull` values need to be computed expressions in some cases.

So `options` *might* be a constant, but not necessarily.

This complicates things quite a bit...

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            var options = (ConvertOptions)constantExpression.Value;
+
+            if (options.OnErrorWasSet)
+            {
+                onErrorAst = options.GetOnError();
```
rstam, line 42: We need to `Translate` the `OnError` expression.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            if (options.OnNullWasSet)
+            {
+                onNullAst = options.GetOnNull();
```
rstam, line 47: We need to `Translate` the `OnNull` expression.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodCallExpressionToAggregationExpressionTranslator.cs
```diff
@@ -138,6 +138,9 @@ public static TranslatedExpression Translate(TranslationContext context, MethodC
                 case "LongCount":
                     return CountMethodToAggregationExpressionTranslator.Translate(context, expression);
 
+                case "Convert":
+                    return ConvertMethodToAggregationExpressionTranslator.Translate(context, expression);
+
```
rstam, line 143: The main section above is for translators for a single method. Move this to after line 41.

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,111 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+            set => _subType = value;
+        }
+
+        internal abstract bool OnErrorWasSet { get; }
+
+        internal abstract bool OnNullWasSet { get; }
+
+        internal abstract BsonValue GetOnError();
+
+        internal abstract BsonValue GetOnNull();
```
rstam, line 229: This isn't quite right. The `OnError` and `OnNull` values need to be `Translated` because they might be computations.

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,111 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+    }
+
+    /// <summary>
+    /// Represents the typed options parameter for the conversion methods in the Mql static class.
+    /// This class allows to set 'onError' and 'onNull'.
+    /// </summary>
+    /// <typeparam name="TResult"> The type of 'onError' and 'onNull'.</typeparam>
+    public class ConvertOptions<TResult> : ConvertOptions
```
rstam, line 237: Us `TTo` instead of `TResult` to be consistent with `Convert<TFrom, TTo`.

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,111 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+                _onNullWasSet = true;
+            }
+        }
+
+        internal override bool OnErrorWasSet => _onErrorWasSet;
+        internal override bool OnNullWasSet => _onNullWasSet;
+
+        internal override BsonValue GetOnError() => BsonValue.Create(_onError);
+
+        internal override BsonValue GetOnNull() => BsonValue.Create(_onNull);
```
rstam, line 275: `OnError` and `OnNull` need to be translated to handle computed `OnError` and `OnNull` values.

Also, in general never call `BsonValue.Create(object)`. Instead, always use the correct serializer to convert a value to a `BsonValue`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
rstam: Example of computed `OnNull`:

```Mql.Convert<int, string>(x.S, new Options<int> { OnNull = x.T })```

File: src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs [hunk 14]
```diff
@@ -255,36 +255,19 @@ public static AstExpression Constant(BsonValue value)
...
-                    "string" => AstUnaryOperator.ToString,
-                    _ => (AstUnaryOperator?)null
-                };
-                if (unaryOperator.HasValue)
-                {
-                    return AstExpression.Unary(unaryOperator.Value, input);
-                }
-            }
-
-            return new AstConvertExpression(input, to, onError, onNull);
```
rstam, line 287: Sorry, I thought YOU had added this code and that's why I was suggesting that it be moved to the AstSimplifer.

I now see that this code was here all along.

I understand why the other tests are now failing. It has to do with the exact order the simplifications happen in.

You could add these lines towards the end of the `AstSimplifier` `VisitCondExpression` method:

```
// { $cond : [{ $eq : [expr, null] }, null, { $convert : { input: expr, ... } }] } => { $convert : { input: expr, ... } }
if (expr2 is AstConvertExpression convertExpression &&                                                                   
    convertExpression.Input == expr1)                                                                                    
{                                                                                                                        
    return Visit(expr2);                                                                                                 
}                                                                                                                        ```
```
This applies the same simplification for `$convert` as we already do for `$toDouble` etc...

rstam: Or you could just leave this alone since that's the way it currently is...

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstSimplifier.cs
```diff
@@ -107,6 +107,36 @@ static bool OperatorMapsNullToNull(AstUnaryOperator @operator)
             }
         }
 
+        public override AstNode VisitConvertExpression(AstConvertExpression node)
+        {
+            var optimizedNode = (AstConvertExpression)base.VisitConvertExpression(node);
+
+            // If possible, uses $toType ($toInt, $toDouble, ...) operators instead of $convert
+            if (optimizedNode is { OnError: null, OnNull: null, SubType: null, Format: null, ByteOrder: null, To: AstConstantExpression toConstantExpression } &&
+                (toConstantExpression.Value as BsonString)?.Value is { } toValue)
```
rstam, line 116: Yes this is 2 lines instead of the original 5 but it sure is tricky to read...

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,145 @@
+using System;
```
rstam, line 1: Copyright missing.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,145 @@
...
+            string format = null;
+            AstExpression onErrorAst = null;
+            AstExpression onNullAst = null;
+            BsonBinarySubType? subType = null;
+
+            var fieldAst = ExpressionToAggregationExpressionTranslator.Translate(context, arguments[0]).Ast;
+
+            var optionExpression = arguments[1];
+
+            if (optionExpression is ConstantExpression constantExpression)
```
rstam, line 34: This all looks like it's on the right track, but this method has gotten really long.

Consider moving the handling/translation of the `optionsExpression` to one or more helper methods.

rstam (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
papafe: Good catch! Not sure why it wasn't added by the IDE as usual-

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
papafe: Makes sense. I'll keep it as it was before then

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,170 @@
...
+        private static T GetConstantValue<T>(Expression expression, string fieldName)
+        {
+            if (expression is not ConstantExpression constantExpression)
+            {
+                throw new ExpressionNotSupportedException($"The {fieldName} field must be a constant expression.");
+            }
+            return (T)constantExpression.Value;
+        }
+
+        private static BsonType GetBsonType(Type type)  //TODO Do we have this kind of info somewhere else...?
```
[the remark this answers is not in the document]
rstam: I can't find any.

There probably is nowhere else in the code where we need to map a C# type to a `BsonType` like this. Everywhere else it is up to the serializer to determine how the C# type is represented.

This is an unusual case because we decide to let the user express the C# return type using the `TTo` generic type parameter, and we let the C# type imply the `BsonType`.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstExpression.cs
rstam, line 270: Consider formatting with one condition per line to make it easier to read:

    if (to is AstConstantExpression toConstantExpression &&
        (toConstantExpression.Value as BsonString)?.Value is { } toValue &&
        subType == null &&                                                 
        byteOrder == null &&                                               
        format == null &&                                                  
        onError == null &&                                                 
        onNull == null)

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConstantExpressionToAggregationExpressionTranslator.cs
```diff
@@ -14,7 +14,6 @@
 */
 
 using System.Linq.Expressions;
-using MongoDB.Bson.IO;
```
rstam, line 17: Unnecessary diff.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberExpressionToAggregationExpressionTranslator.cs
```diff
@@ -43,7 +43,6 @@ public static TranslatedExpression Translate(TranslationContext context, MemberE
                 {
                     case "HasValue": return HasValuePropertyToAggregationExpressionTranslator.Translate(context, expression);
                     case "Value": return ValuePropertyToAggregationExpressionTranslator.Translate(context, expression);
-                    default: break;
```
rstam, line 46: Diff is unrelated to the PR.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MemberExpressionToAggregationExpressionTranslator.cs
```diff
@@ -66,7 +65,7 @@ public static TranslatedExpression Translate(TranslationContext context, MemberE
 
             if (!DocumentSerializerHelper.AreMembersRepresentedAsFields(containerTranslation.Serializer, out _))
             {
-                if (member is PropertyInfo propertyInfo  && propertyInfo.Name == "Length")
+                if (member is PropertyInfo propertyInfo && propertyInfo.Name == "Length")
```
rstam, line 68: Diff is unrelated to the PR.

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,121 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+        /// <summary>
+        /// Little endian order.
+        /// </summary>
+        LittleEndian,
+    }
+
+    /// <summary>
+    /// Represents the options parameter for <see cref="Mql.Convert{TFrom, TTo}(TFrom, ConvertOptions{TTo})"/>.
+    /// </summary>
+    public abstract class ConvertOptions
```
rstam, line 190: This should probably be in a its own file called `ConvertOptions.cs`.

Along with `ByteOrder` (unless you think `ByteOrder` should be in its own file).

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,121 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+    /// This class allows to set 'onError' and 'onNull'.
+    /// </summary>
+    /// <typeparam name="TTo"> The type of 'onError' and 'onNull'.</typeparam>
+    public class ConvertOptions<TTo> : ConvertOptions
+    {
+        private TTo _onError;
+        private bool _onErrorWasSet;
+        private TTo _onNull;
+        private bool _onNullWasSet;
+        private readonly IBsonSerializer _serializer;
```
rstam, line 243: This class should not have a serializer. Anything serializer related belongs in the translator.

src/MongoDB.Driver/Mql.cs
```diff
@@ -153,4 +168,121 @@ public static bool IsNullOrMissing<TField>(TField field)
...
+                _onNullWasSet = true;
+            }
+        }
+
+        internal override bool OnErrorWasSet => _onErrorWasSet;
+        internal override bool OnNullWasSet => _onNullWasSet;
+
+        internal override BsonValue GetOnError() => _serializer.ToBsonValue(_onError);
+
+        internal override BsonValue GetOnNull() => _serializer.ToBsonValue(_onNull);
```
rstam, line 286: Replace the previous two methods with:

    internal override bool OnErrorWasSet(out object onError)
    {                                                       
        onError = _onError;                                 
        return _onErrorWasSet;                              
    }                                                       
                                                            
    internal override bool OnNullWasSet(out object onNull)  
    {                                                       
        onNull = _onNull;                                   
        return _onNullWasSet;                               
    }                                                       

See other related suggestions.

rstam (changes requested):

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/DateFromStringMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -21,7 +21,6 @@
 using MongoDB.Bson;
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Core.TestHelpers.XunitExtensions;
-using MongoDB.Driver.Linq;
```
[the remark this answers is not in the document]
papafe: Yes, I was looking at the file and I noticed it. I think it's a very small change, can't we do it here without opening a new ticket?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConstantExpressionToAggregationExpressionTranslator.cs
papafe: I noticed it while I was looking at the file. It's a very small change, do we really need to open another PR to fix this?

src/MongoDB.Driver/Mql.cs
papafe: Makes sense.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,171 @@
...
+        private static T GetConstantValue<T>(Expression expression, string fieldName)
+        {
+            if (expression is not ConstantExpression constantExpression)
+            {
+                throw new ExpressionNotSupportedException($"The {fieldName} field must be a constant expression.");
+            }
+            return (T)constantExpression.Value;
+        }
+
+        private static BsonType GetBsonType(Type type)
```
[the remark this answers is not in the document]
papafe: Wouldn't it be convenient here to use the serializers registered in `StandardSerializers`? It seems that that's the perfect class for what we're trying to do here. Of course we'd need to specify the serializers that are not in that list.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,171 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.Ast;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators.MethodTranslators
+{
+    internal class ConvertMethodToAggregationExpressionTranslator
+    {
+        public static TranslatedExpression Translate(TranslationContext context, MethodCallExpression expression)
```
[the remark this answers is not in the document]
papafe: I have a question regarding the use of `EnsureSerializerIsCompatible'. This seems more an exception that could be raised because we made some error in selecting the correct expected serializer than something that could be raised from developers code, am I right?

src/MongoDB.Bson/Serialization/Serializers/ByteSerializer.cs
```diff
@@ -23,6 +23,16 @@ namespace MongoDB.Bson.Serialization.Serializers
     /// </summary>
     public sealed class ByteSerializer : StructSerializerBase<byte>, IRepresentationConfigurable<ByteSerializer>
     {
+        #region static
```
papafe, line 26: Even if we decided not to use them, it's a good idea to add the `Instance`, to uniform the serializers.

src/MongoDB.Bson/Serialization/Serializers/UInt16Serializer.cs
```diff
@@ -25,6 +25,15 @@ namespace MongoDB.Bson.Serialization.Serializers
     [CLSCompliant(false)]
     public sealed class UInt16Serializer : StructSerializerBase<ushort>, IRepresentationConfigurable<UInt16Serializer>, IRepresentationConverterConfigurable<UInt16Serializer>
     {
+        #region static
+        private static readonly UInt16Serializer __instance = new();
```
rstam, line 29: private static readonly UInt16Serializer __instance = new UInt16Serializer(representation: BsonType.Int32);

src/MongoDB.Bson/Serialization/Serializers/UInt64Serializer.cs
```diff
@@ -25,6 +25,15 @@ namespace MongoDB.Bson.Serialization.Serializers
     [CLSCompliant(false)]
     public sealed class UInt64Serializer : StructSerializerBase<ulong>, IRepresentationConfigurable<UInt64Serializer>, IRepresentationConverterConfigurable<UInt64Serializer>
     {
+        #region static
+        private static readonly UInt64Serializer __instance = new();
```
rstam, line 29: private static readonly UInt64Serializer __instance = new UInt64Serializer(representation: BsonType.Int32);

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/SerializationHelper.cs
```diff
@@ -77,6 +77,14 @@ public static void EnsureRepresentationIsNumeric(Expression expression, Expressi
             }
         }
 
+        public static void EnsureSerializerIsCompatible(Expression expression, IBsonSerializer actualSerializer, IBsonSerializer expectedSerializer)
```
rstam, line 80: Maybe it would be helpful to the user to provide the `containingExpression` also for context:

    public static void EnsureSerializerIsCompatible(Expression expression, Expression containingExpression, IBsonSerializer actualSerializer, IBsonSerializer expectedSerializer)
    {                                                                                                                                                                            
        if (!actualSerializer.Equals(expectedSerializer))                                                                                                                        
        {                                                                                                                                                                        
            throw new ExpressionNotSupportedException(expression, containingExpression, because: "the result serializer is not compatible with the expected serializer");        
        }                                                                                                                                                                        
    }

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/SerializationHelper.cs
```diff
@@ -238,6 +246,5 @@ public static BsonArray SerializeValues(IBsonSerializer itemSerializer, IEnumera
                 writer.WriteEndDocument();
             }
             return document["_v"].AsBsonArray;
-        }
-    }
+        } }
```
rstam, line 249: What happened here? :)

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,201 @@
...
+                TranslationContext context,
+                Expression expression,
+                Expression optionsExpression,
+                IBsonSerializer toSerializer)
+        {
+            return optionsExpression switch
+            {
+                ConstantExpression constantExpression => TranslateOptions(constantExpression, toSerializer),
+                MemberInitExpression memberInitExpressionExpression => TranslateOptions(context, expression, memberInitExpressionExpression, toSerializer),
+                _ => throw new ExpressionNotSupportedException(optionsExpression, containingExpression: expression, because: "the Options argument must be either a constant or a member initialization expression.")
```
rstam, line 65: Why is `Options` capitalized? The name of the parameter is `options`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,201 @@
...
+            };
+        }
+
+
```
rstam, line 69: Extra blank line.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,201 @@
...
+                {
+                    case nameof(ConvertOptions.ByteOrder):
+                        byteOrder = memberExpression.GetConstantValue<ByteOrder?>(expression);
+                        break;
+                    case nameof(ConvertOptions.Format):
+                        format = memberExpression.GetConstantValue<string>(expression);
+                        break;
+                    case nameof(ConvertOptions<object>.OnError):
+                        onErrorTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, memberExpression);
+                        SerializationHelper.EnsureSerializerIsCompatible(memberExpression, onErrorTranslation.Serializer, expectedSerializer: toSerializer);
```
rstam, line 131: Let's include the `containingExpression` in the exception:

       SerializationHelper.EnsureSerializerIsCompatible(memberExpression, containingExpression: expression, onErrorTranslation.Serializer, expectedSerializer: toSerializer);

The immediate containing expression is the `optionsExpression`, but it might be more useful to include the call to `Convert` expression in the exception message.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,201 @@
...
+                        break;
+                    case nameof(ConvertOptions<object>.OnNull):
+                        onNullTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, memberExpression);
+                        SerializationHelper.EnsureSerializerIsCompatible(memberExpression, onNullTranslation.Serializer, expectedSerializer: toSerializer);
```
rstam, line 135: Let's include the `containingExpression` in the exception:

    SerializationHelper.EnsureSerializerIsCompatible(memberExpression, containingExpression: expression, onNullTranslation.Serializer, expectedSerializer: toSerializer);

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,201 @@
...
+                TypeCode.SByte => BsonType.Int32,
+                TypeCode.Single => BsonType.Double,
+                TypeCode.UInt16 => BsonType.Int32,
+                TypeCode.UInt32 => BsonType.Int64,
+                TypeCode.UInt64 => BsonType.Decimal128,
+
+                _ => throw new ExpressionNotSupportedException(expression, because: $"{toType} is not a valid TTo for Convert")
+            };
+
+            return (bsonType, StandardSerializers.GetSerializer(toType));
```
rstam, line 198: I don't like the idea of using the `StandardSerializers` here.

We are working with a defined set of of `toType` serializers that may or may not be the same as the `StandardSerializers`. In fact they are not, as you discovered by having to handle some cases specially.

I liked your original switch where you handled all cases in one switch, just modified to return the serializer also:

    private static (BsonType ToBsonType, IBsonSerializer ToSerialzier) TranslateToType(Expression expression, Type toType)
    {
        return Type.GetTypeCode(Nullable.GetUnderlyingType(toType) ?? toType) switch
        {
            TypeCode.Boolean => (BsonType.Boolean, BooleanSerializer.Instance),
            TypeCode.Byte => (BsonType.Int32, ByteSerializer.Instance),
            TypeCode.Char => (BsonType.String, StringSerializer.Instance),
            TypeCode.DateTime => (BsonType.DateTime, DateTimeSerializer.Instance),
            TypeCode.Decimal => (BsonType.Decimal128, DecimalSerializer.Instance),
            TypeCode.Double => (BsonType.Double, DoubleSerializer.Instance),
            TypeCode.Int16 => (BsonType.Int32, Int16Serializer.Instance),
            TypeCode.Int32 => (BsonType.Int32, Int32Serializer.Instance),
            TypeCode.Int64 => (BsonType.Int64, Int64Serializer.Instance),
            TypeCode.SByte => (BsonType.Int32, SByteSerializer.Instance),
            TypeCode.Single => (BsonType.Double, DoubleSerializer.Instance),
            TypeCode.String => (BsonType.String, StringSerializer.Instance),
            TypeCode.UInt16 => (BsonType.Int32, UInt16Serializer.Instance),
            TypeCode.UInt32 => (BsonType.Int64, Int32Serializer.Instance),
            TypeCode.UInt64 => (BsonType.Decimal128, UInt64Serializer.Instance),
                                                                                
            _ when toType == typeof(byte[]) => (BsonType.Binary, ByteArraySerializer.Instance),
            _ when toType == typeof(BsonBinaryData) => (BsonType.Binary, BsonBinaryDataSerializer.Instance),
            _ when toType == typeof(Decimal128) => (BsonType.Decimal128, Decimal128Serializer.Instance),
            _ when toType == typeof(Guid) => (BsonType.Binary, GuidSerializer.StandardInstance),
            _ when toType == typeof(ObjectId) => (BsonType.ObjectId, ObjectIdSerializer.Instance),
                                                                                                  
            _ => throw new ExpressionNotSupportedException(expression, because: $"{toType} is not a valid TTo for Convert")
        };                                                                                                                 
    }

src/MongoDB.Driver/ConvertOptions.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public BsonBinarySubType? SubType
+        {
+            get => _subType;
+            set => _subType = value;
+        }
+
+        internal abstract bool OnErrorWasSet(out object onError);
+
+        internal abstract bool OnNullWasSet(out object onError);
```
rstam, line 60: internal abstract bool OnNullWasSet(out object onNull);

rstam (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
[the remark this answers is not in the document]
papafe: Aah, got it.

src/MongoDB.Bson/Serialization/Serializers/ByteSerializer.cs
```diff
@@ -23,6 +23,16 @@ namespace MongoDB.Bson.Serialization.Serializers
...
+        private static readonly ByteSerializer __instance = new();
```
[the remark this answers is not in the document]
papafe: It seems that we never explictly state the representation for any of the other serializers that have a static `Instance` property. Besides, if we don't specify the representation we can be sure that the static instance we get is the same as the one obtained by just calling the constructor. (Even though I can imagine we are not going to change the default representation of serializers)

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/SerializationHelper.cs
papafe: The IDE must have done something funky 😁

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
papafe: I'll go back to this, but we still need to take care of nullable types.

rstam: I forgot that `TTo` could be nullable. Does that change what you want to do?

My initial concern with using `StandardSerializers` was that I wasn't 100% sure what "Standard" meant and whether it would always result in the right thing being done (for Convert).

I was also concerned that some of the serializers needed by Convert might not belong in the "Standard" set.

papafe: It's true that some of the serializers need by convert are not in the "Standard" set, but it seems that it covered most of the "basic" types. It seems that `StandardSerializers`' methods are used by `ConstantExpressionToAggregationExpressionTranslator` to serialize the constant value, so I thought this could be a similar thing.
Anyways, I think the current solution makes sense, the only small annoyance is that we don't have static "Instances" of the nullable serializers, but I think it's a small thing.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            return (subType, byteOrder, format, onErrorTranslation?.Ast, onNullTranslation?.Ast);
+        }
+
+        private static (BsonType ToBsonType, IBsonSerializer ToSerializer) TranslateToType(Expression expression, Type toType)
+        {
+            var underlyingType = Nullable.GetUnderlyingType(toType);
+            var isNullable = underlyingType != null;
+            var effectiveType = underlyingType ?? toType;
```
rstam, line 150: I very much dislike calling `Nullable.GetUnderlyingType` when we do NOT know if the type is nullable or not. It seems like this method *should* throw an exception when called with a non-nullable type. I know it does not, but that seems more like a quirk and an implementation error on Microsoft's part, one that we should not rely on.

I suggest:

    var isNullable = toType.IsNullable();
    var valueType = isNullable ? Nullable.GetUnderlyingType(toType) : toType;

`effectiveType` is not a terrible name, but we tend to use `valueType` when talking about nullable values.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            (BsonType, IBsonSerializer) result = Type.GetTypeCode(effectiveType) switch
```
rstam, line 152: I was trying to figure out why you had to use this odd `result` variable here...

The reason is that the `switch` expression doesn't have a "best" result type that applies to all cases, and the compiler needs some guidance. The declaration of the `result` variable is one way to provide that guidance.

Consider this alternative that doesn't require the intermediate `result` variable:

    var (bsonType, valueSerializer) = (ValueTuple<BsonType, IBsonSerializer>)(Type.GetTypeCode(valueType) switch
    {
        ....
    });

where casting the result of the `switch` expression to the `ValueTuple` provides the guidance to the compiler.

Alternatively, if you were to cast even ONE of the results to `IBsonSerializer` that would work also, but it is a bit esoteric to do it that way:

    TypeCode.Boolean => (BsonType.Boolean, (IBsonSerializer)BooleanSerializer.Instance),

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/ConstantExpressionToAggregationExpressionTranslator.cs
rstam: What is your current thinking on this unnecessary diff? Can it be removed from this PR?

rstam (changes requested):

sanych-sun (changes requested):

sanych-sun (changes requested):
Please check evergreen also. There are some test failures related to the changes.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslator.cs
papafe: That's cool, I didn't think about it.

src/MongoDB.Driver/ConvertOptions.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        public BsonBinarySubType? SubType
+        {
+            get => _subType;
+            set => _subType = value;
+        }
+
+        internal abstract bool TryGetOnError(out object onError);
+
+        internal abstract bool TryGetOnNull(out object onNull);
```
rstam, line 58: I'm not convinced this name change is a good one.

Not every method that returns a `bool` should be named `TryXyz`.

In my opinion methods should only be named `TryXyz` when they have a chance of *failure*, but we want to return `false` instead of throwing an exception.

In this case there is no chance of failure. We merely want to know whether `OnError` or `OnNull` was set, and if so to what.

papafe: I see your point.

rstam (approved):
LGTM

PR #1665, opened 2025-04-08
CSHARP-5204: Test Heartbeats don't gossip cluster time

adelinowona:
It turns out we were already not gossiping cluster times in our monitoring connections. Our monitors send hello requests  which use a `NoCoreSession` instance when the wire protocol command is created for a hello request. A `NoCoreSession` instance has a null `ClusterTime` which cannot be changed and calling `session.AdvanceClusterTime` is also a no-op for a `NoCoreSession`. So this PR just adds the new prose test described in the Drivers spec update: https://github.com/mongodb/specifications/commit/b76cbf96807d96aa8ae4337a3f2b3c9be39d3487

### DIFF
--- a/tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
+++ b/tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
@@ -21,6 +21,7 @@ using FluentAssertions;
 using MongoDB.Bson;
 using MongoDB.Bson.TestHelpers;
 using MongoDB.Driver.Core;
+using MongoDB.Driver.Core.Clusters;
 using MongoDB.Driver.Core.Events;
 using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Core.TestHelpers.Logging;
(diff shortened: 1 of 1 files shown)


tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
```diff
@@ -302,6 +303,54 @@ public async Task Ensure_server_session_are_allocated_only_on_connection_checkou
...
+                .Capture<ServerHeartbeatStartedEvent>()
+                .Capture<ServerHeartbeatSucceededEvent>()
+                .Capture<CommandStartedEvent>();
+
+            using var c1 = DriverTestConfiguration.CreateMongoClient(
+                settings =>
+                {
+                    settings.ClusterConfigurator = c => c.Subscribe(eventCapturer);
+                    settings.DirectConnection = true;
+                    settings.HeartbeatInterval = TimeSpan.FromSeconds(2000);
```
BorisDog, line 321: Should this be a much shorter interval, like `10ms` in the spec?

tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
```diff
@@ -302,6 +303,54 @@ public async Task Ensure_server_session_are_allocated_only_on_connection_checkou
...
+                events =>
+                {
+                    var capturedEvents = events.ToArray();
+                    return capturedEvents.Count(e => e is ServerHeartbeatSucceededEvent) > 1 &&
+                           capturedEvents
+                               .Where((e, i) =>
+                                   e is ServerHeartbeatStartedEvent &&
+                                   capturedEvents[i + 1] is ServerHeartbeatSucceededEvent)
+                               .Any();
+                }, TimeSpan.FromSeconds(1), "Didn't get any server heartbeat pairs");
```
BorisDog, line 344: Seems like a race, with the larger HeartbeatInterval setting in `c1`

BorisDog (changes requested):
Looks good overall. Some question regarding the timings + failed test on EG.

sanych-sun (changes requested):
Please check Evergreen failures, as new test is failing for some variants.

tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
adelinowona: Yeah I was testing something locally and I think I accidentally included this change in my final commit.

tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
adelinowona: fixed the heartbeatInterval setting in `c1` so this should be good.

BorisDog (approved):
LGTM.
I've scheduled additional EG tasks to test stability of this test. Please double check that.

tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
```diff
@@ -302,6 +303,59 @@ public async Task Ensure_server_session_are_allocated_only_on_connection_checkou
...
+            c2.GetDatabase("test").GetCollection<BsonDocument>("test").InsertOne(new BsonDocument("advance", "$clusterTime"));
+
+            eventCapturer.Clear();
+
+            eventCapturer.WaitForOrThrowIfTimeout(
+                events =>
+                {
+                    var capturedEvents = events.ToArray();
+                    return capturedEvents.Count(e => e is ServerHeartbeatSucceededEvent) > 1 &&
+                           capturedEvents
```
sanych-sun, line 344: I think this code could be simplified a little by using the following approach:
```
capturedEvents
    .SkipUntil(e => <first event condition>)
    .Any(e => <second event condition>)
```

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/Specifications/sessions/SessionsProseTests.cs
adelinowona: Done. thanks!

sanych-sun (approved):
LGTM

PR #1673, opened 2025-04-21
CSHARP-4535: Support queries after casting IQueryable<Derived> to IQueryable<Base>.

rstam:
If this approach looks good then I should still look for all places where a new `Symbol` is created and see if any other places need to wrap a `DowncastingSerializer` around a serializer for a derived type.

### DIFF
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Misc/Symbol.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Misc/Symbol.cs
@@ -13,6 +13,7 @@
 * limitations under the License.
 */
 
+using System;
 using System.Linq.Expressions;
 using MongoDB.Bson.Serialization;
 using MongoDB.Driver.Core.Misc;
@@ -37,6 +38,11 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Misc
             _ast = Ensure.IsNotNull(ast, nameof(ast));
             _serializer = Ensure.IsNotNull(serializer, nameof(serializer));
             _isCurrent = isCurrent;
+
+            if (_serializer.ValueType != _parameter.Type)
+            {
+                throw new ArgumentException($"Value type {_serializer.ValueType} of serializer type {_serializer.GetType()} does not match parameter type {_parameter.Type}.");
+            }
         }
 
         // public properties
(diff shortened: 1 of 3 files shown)


src/MongoDB.Driver/Linq/Linq3Implementation/Misc/Symbol.cs
```diff
@@ -37,6 +38,11 @@ public Symbol(ParameterExpression parameter, string name, AstExpression ast, IBs
             _ast = Ensure.IsNotNull(ast, nameof(ast));
             _serializer = Ensure.IsNotNull(serializer, nameof(serializer));
             _isCurrent = isCurrent;
+
+            if (_serializer.ValueType != _parameter.Type)
+            {
+                throw new ArgumentException($"Value type {_serializer.ValueType} of serializer type {_serializer.GetType()} does not match parameter type {_parameter.Type}.");
+            }
```
rstam, line 45: Centralized this validation in the constructor for `Symbol`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/ExpressionToFilterTranslator.cs
```diff
@@ -56,9 +56,10 @@ public static AstFilter TranslateLambda(
             bool asRoot = false)
         {
             var parameterExpression = lambdaExpression.Parameters.Single();
-            if (parameterSerializer.ValueType != parameterExpression.Type)
+
+            if (parameterSerializer.ValueType != parameterExpression.Type && parameterExpression.Type.IsAssignableFrom(parameterSerializer.ValueType))
             {
-                throw new ArgumentException($"ValueType '{parameterSerializer.ValueType.FullName}' of parameterSerializer does not match parameter type '{parameterExpression.Type.FullName}'.", nameof(parameterSerializer));
+                parameterSerializer = DowncastingSerializer.Create(parameterExpression.Type, parameterSerializer.ValueType, parameterSerializer);
```
rstam, line 62: The validation has been moved to the `Symbol` constructor. Instead here, if we don't have the exactly right serializer but it can be adapted using a `DowncastingSerializer` then do that.

If we like this approach I still need to search whether there are any other places where the same thing needs to be done.

sanych-sun (approved):
LGTM

PR #1674, opened 2025-04-22
CSHARP-5574: Fix EFCore encryption tests failure in C# Driver CI

sanych-sun:


damieng (approved):
LGTM but there are two platforms that had test failures. Re-running.

PR #1675, opened 2025-04-23
Bring back support for netstandard2.0

RogerBarreto:
When attempting to update to the latest driver we stumbled into a limitation that was not before until major 3.0 dropping support for netstandard 2.0.

- https://github.com/microsoft/semantic-kernel/pull/11663

In the good willingness to bring this support back I'm proposing this PR.

Thanks

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -4,7 +4,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;net472;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
     <LangVersion>12.0</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CodeAnalysisRuleSet>..\..\MongoDB.ruleset</CodeAnalysisRuleSet>
--- a/src/MongoDB.Bson/TargetFramework.cs
+++ b/src/MongoDB.Bson/TargetFramework.cs
@@ -22,6 +22,8 @@ namespace MongoDB.Bson
             "net472";
 #elif NETSTANDARD2_1
             "netstandard21";
+#elif NETSTANDARD2_0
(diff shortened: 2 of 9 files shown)


RogerBarreto:
@damieng I'm work in Semantic Kernel, a SDK for AI maintained by Microsoft and we do have a connector there, and is a very good standard to have projects that support `netstandard2.0` which allow a huge code base including .Net Framework from 4.0 up to 4.8, basically in our MongoDB connector we are stuck to the latest version of `2.3.0` where `netstandard 2.0` was locked.

damieng:
At the time when we looked at usage statistics there was not enough usage below 4.7.2 to justify any extra effort there. Given the Semantic Kernel connector for MongoDB is a separate package to SK itself would it not just make sense for that project to target netstandard 2.1?

damieng:
If you want to talk offline you can message me directly on the Microsoft corporate Teams (I'm a Guest so I can't initiate the conversation there).

JamesKovacs:
There is an additional problem with using `netstandard2.0` assemblies with .NET Framework 4.7+. `ObjectId.GenerateNewId()` relies on `System.Random`. On .NET Framework, `System.Random` instances are initialized with the current system ticks. On .NET Core, `System.Random` instances are initialized via a cryptographically-secure RNG. Consider this code:

```
var rnd1 = new Random();
var rnd2 = new Random();
```

On .NET Core, these two instances will output different PRNG sequences but on .NET Framework the two instances will generate the same PRNG sequence.

While we could consider fixing .NET Framework to initialize `Random` (used for `ObjectId` generation) to also use cryptographically-secure seed, there are other places in the code where `netstandard2.0` differs from `net472`, such as our networking stack.

Thus it wouldn't be as simple as re-introducing `netstandard2.0` support, but carefully testing all our `netstandard2.0` code to make sure it is compatible with .NET Framework - much of which would require runtime checks.

Our current direction is to use `net472` for .NET Framework 4.7+ and `netstandard2.1`/`net60`/etc. for .NET Core (which includes .NET 5+). Hopefully that explanation helps you understand our decision.

PR #1676, opened 2025-04-24
CSHARP-5560: CSOT: Refactor IOperationExecutor to use operation context

sanych-sun:


src/MongoDB.Driver/OperationExecutor.cs
```diff
@@ -13,50 +13,146 @@
...
 
-        public TResult ExecuteReadOperation<TResult>(IReadBinding binding, IReadOperation<TResult> operation, CancellationToken cancellationToken)
+        public TResult ExecuteReadOperation<TResult>(
+            IReadOperation<TResult> operation,
+            ReadOperationOptions options,
+            IClientSessionHandle session,
+            CancellationToken cancellationToken)
         {
-            return operation.Execute(binding, cancellationToken);
+            bool isOwnSession = session == null;
```
adelinowona, line 40: nit: use var? 

here and other methods in this file

src/MongoDB.Driver/OperationExecutor.cs
```diff
@@ -13,50 +13,146 @@
...
+        {
+            // TODO: CreateReadWriteBinding from MongoClient did not used ChannelPinningHelper, double-check if it's OK to start using it
+            return ChannelPinningHelper.CreateReadWriteBinding(_client.GetClusterInternal(), session.WrappedCoreSession.Fork());
+        }
+
+        private IClientSessionHandle StartImplicitSession()
         {
-            return _client.StartImplicitSessionAsync(cancellationToken);
+            var options = new ClientSessionOptions { CausalConsistency = false, Snapshot = false };
+            ICoreSessionHandle coreSession = _client.GetClusterInternal().StartSession(options.ToCore(isImplicit: true));
```
adelinowona, line 154: nit: use var?

src/MongoDB.Driver/ReadOperationOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.Driver
+{
+    internal record class ReadOperationOptions(ReadPreference ExplicitReadPreference = null, ReadPreference DefaultReadPreference = null);
```
adelinowona, line 18: no need for specifying `class`? I believe `record` is equivalent to `record class`

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +98,117 @@ public override MongoCollectionSettings Settings
...
-
-            foreach (var request in requestsArray)
-            {
-                request.ThrowIfNotValid();
-            }
-
-            options = options ?? new BulkWriteOptions();
 
-            var renderArgs = GetRenderArgs();
-            var operation = CreateBulkWriteOperation(session, requestsArray, options, renderArgs);
```
adelinowona, line 247: Why did we remove all the checks here? Are they not needed anymore? Can requests not be null?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +98,117 @@ public override MongoCollectionSettings Settings
...
-            await ExecuteWriteOperationAsync(session, aggregateOperation, cancellationToken).ConfigureAwait(false);
+            await OperationExecutor.ExecuteWriteOperationAsync(aggregateOperation, _writeOperationOptions, session, cancellationToken).ConfigureAwait(false);
         }
 
-        public override BulkWriteResult<TDocument> BulkWrite(IEnumerable<WriteModel<TDocument>> requests, BulkWriteOptions options, CancellationToken cancellationToken = default(CancellationToken))
+        public override BulkWriteResult<TDocument> BulkWrite(IEnumerable<WriteModel<TDocument>> requests, BulkWriteOptions options, CancellationToken cancellationToken = default)
         {
-            return UsingImplicitSession(session => BulkWrite(session, requests, options, cancellationToken), cancellationToken);
+            using var session = OperationExecutor.StartImplicitSession(cancellationToken);
+            return BulkWrite(session, requests, options, cancellationToken);
```
adelinowona, line 200: I noticed that for some of the other methods above (like `AggregateToCollection`) you did not use `OperationExecutor` to create the implicit session and just used the existing `UsingImplicitSession` mechanism. Should we update all those use cases as well? or am I missing something?

adelinowona (changes requested):
Looks good! left some minor comments

tests/MongoDB.Driver.Tests/OperationExecutorTests.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                subject.StartImplicitSession(CancellationToken.None);
+            }
+
+            clusterMock.Verify(c => c.StartSession(It.Is<CoreSessionOptions>(v => v.IsImplicit && v.IsCausallyConsistent == false && v.IsSnapshot == false)));
+        }
+
+        [Theory]
+        [MemberData(nameof(ImplicitSessionTestCases))]
+        public async Task ExecuteReadOperationShouldStartAndDisposeImplicitSessionIfNeeded(bool shouldCreateSession, bool isAsync, IClientSessionHandle session)
```
BorisDog, line 49: Don't we use this format:
`ExecuteReadOperation_should_start_and..` ?

BorisDog (changes requested):

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: The idea was to hide the implicit session creation inside of the OperationExecutor. But for some corner cases (like Aggregate method) we should have access to the created session, so sometimes we have to create the implicit session from outside.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -66,6 +67,9 @@ public MongoClient(MongoClientSettings settings)
 
             _cluster = _settings.ClusterSource.Get(_settings.ToClusterKey());
             _operationExecutor = new OperationExecutor(this);
+            _readOperationOptions = new ReadOperationOptions(DefaultReadPreference: _settings.ReadPreference);
+            _writeOperationOptions = new WriteOperationOptions();
+
```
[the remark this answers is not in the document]
sanych-sun: Done

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -117,8 +121,6 @@ internal MongoClient(IOperationExecutor operationExecutor, MongoClientSettings s
         // internal methods
         internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettings messageEncoderSettings)
         {
-            ThrowIfDisposed();
```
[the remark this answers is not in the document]
sanych-sun: Reverted the change.

src/MongoDB.Driver/AggregateHelper.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    internal static class AggregateHelper
+    {
+        public static RenderedPipelineDefinition<TResult> RenderAggregatePipeline<TDocument, TResult>(PipelineDefinition<TDocument, TResult> pipeline, RenderArgs<TDocument> renderArgs, out bool isAggregateToCollection)
+        {
+            var renderedPipeline = pipeline.Render(renderArgs);
+
+            var lastStage = renderedPipeline.Documents.LastOrDefault();
+            var lastStageName = lastStage?.GetElement(0).Name;
+            isAggregateToCollection = lastStage != null && (lastStageName == "$out" || lastStageName == "$merge");
```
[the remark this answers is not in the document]
sanych-sun: Done

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -345,167 +323,160 @@ public IMongoCollection<TDocument> GetCollection<TDocument>(string name, MongoCo
             return new MongoCollectionImpl<TDocument>(this, new CollectionNamespace(_databaseNamespace, name), settings, _cluster, _operationExecutor);
         }
 
-        public IAsyncCursor<string> ListCollectionNames(ListCollectionNamesOptions options = null, CancellationToken cancellationToken = default(CancellationToken))
+        public IAsyncCursor<string> ListCollectionNames(ListCollectionNamesOptions options = null, CancellationToken cancellationToken = default)
         {
-            return UsingImplicitSession(session => ListCollectionNames(session, options, cancellationToken), cancellationToken);
+            var cursor = OperationExecutor.ExecuteReadOperation(CreateListCollectionNamesOperation(options),
+                _readOperationOptions with { DefaultReadPreference = ReadPreference.Primary },
+                null, cancellationToken);
```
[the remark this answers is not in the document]
sanych-sun: Done

src/MongoDB.Driver/ReadOperationOptions.cs
sanych-sun: I'm OK with removing the class here. I've left it only to stress on the fact that it should be class, not struct. As changing to struct might affect its functionality (because of the copy-by-value nature of structs).

tests/MongoDB.Driver.Tests/OperationExecutorTests.cs
sanych-sun: Fixed.

src/MongoDB.Driver/OperationExecutor.cs
sanych-sun: Fixed.

src/MongoDB.Driver/OperationExecutor.cs
sanych-sun: Fixed.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Thank you for reporting this. I've converted rest of the methods and removed `UsingImplicitSession`.

src/MongoDB.Driver/OperationExecutor.cs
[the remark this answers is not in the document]
sanych-sun: Fixed.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +98,117 @@ public override MongoCollectionSettings Settings
         }
 
         // public methods
-        public override IAsyncCursor<TResult> Aggregate<TResult>(PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
+        public override IAsyncCursor<TResult> Aggregate<TResult>(PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default)
         {
             return UsingImplicitSession(session => Aggregate(session, pipeline, options, cancellationToken), cancellationToken);
```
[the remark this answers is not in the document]
sanych-sun: Fixed.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -83,6 +85,8 @@ public override IMongoIndexManager<TDocument> Indexes
             get { return new MongoIndexManager(this); }
         }
 
+        private IOperationExecutor OperationExecutor => _operationExecutor;
```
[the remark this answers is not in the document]
sanych-sun: Removed as it was discussed.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +98,117 @@ public override MongoCollectionSettings Settings
...
-            options = options ?? new AggregateOptions();
+            options ??= new AggregateOptions();
 
             var renderArgs = GetRenderArgs(options.TranslationOptions);
-            var renderedPipeline = pipeline.Render(renderArgs);
-
-            var lastStage = renderedPipeline.Documents.LastOrDefault();
-            var lastStageName = lastStage?.GetElement(0).Name;
-            if (lastStage != null && (lastStageName == "$out" || lastStageName == "$merge"))
+            var renderedPipeline = AggregateHelper.RenderAggregatePipeline(pipeline, renderArgs, out bool isAggregateToCollection);
```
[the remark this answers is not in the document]
sanych-sun: Even though it's possible, I do not think we should do that. Mostly because:
- such method will have too many parameters (have to pass: namespace, write concern, read concern, encoder settings, read preference in addition to the current existing parameters)
- there should be 2 overloads of such methods: to accept collection namespace and database namespace
- there is additional pattern of the current method usage in `AggregateToCollection` method, that should throw in case when rendered pipeline's last stage is not `out` stage.

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -778,9 +717,10 @@ private async Task<IWriteOperation<BsonDocument>> CreateDropCollectionOperationA
                 });
         }
 
-        private ListCollectionsOperation CreateListCollectionNamesOperation(ListCollectionNamesOptions options, RenderArgs<BsonDocument> renderArgs)
+        private ListCollectionsOperation CreateListCollectionNamesOperation(ListCollectionNamesOptions options)
```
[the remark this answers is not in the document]
sanych-sun: Done.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1809,6 +1544,12 @@ private DropIndexOperation CreateDropAllOperation(DropIndexOptions options)
 
             private DropIndexOperation CreateDropOneOperation(string name, DropIndexOptions options)
             {
+                Ensure.IsNotNullOrEmpty(name, nameof(name));
+                if (name == "*")
+                {
+                    throw new ArgumentException("Cannot specify '*' for the index name. Use DropAllAsync to drop all indexes.", "name");
```
BorisDog, line 1550: minor: mistake in the original code, should be `DropAll` and `DropAllAsync`.

BorisDog (changes requested):
partial review: minor comments so far.

src/MongoDB.Driver/OperationExecutor.cs
```diff
@@ -13,50 +13,170 @@
...
+                throw new InvalidOperationException("Read preference in a transaction must be primary.");
+            }
+
+            // TODO: CreateReadBinding from MongoClient did not used ChannelPinningHelper, double-check if it's OK to start using it
+            return ChannelPinningHelper.CreateReadBinding(_client.GetClusterInternal(), session.WrappedCoreSession.Fork(), readPreference);
+        }
+
+        private IReadWriteBindingHandle CreateReadWriteBinding(IClientSessionHandle session)
+        {
+            // TODO: CreateReadWriteBinding from MongoClient did not used ChannelPinningHelper, double-check if it's OK to start using it
```
BorisDog, line 163: Need to address this TODO.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -100,7 +104,7 @@ public MongoClient(string connectionString)
         internal MongoClient(IOperationExecutor operationExecutor, MongoClientSettings settings)
             : this(settings)
         {
-            _operationExecutor = operationExecutor;
+            _operationExecutor = new OperationExecutorWrapper(operationExecutor);
```
BorisDog, line 107: Consider using the `Fork` approach.

src/MongoDB.Driver/WriteOperationOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB.Driver
+{
+    internal record class WriteOperationOptions();
```
adelinowona, line 18: remove class here too

adelinowona (changes requested):

src/MongoDB.Driver/MongoClient.cs
sanych-sun: The problem here is: MongoClient implements IDisposable, and disposed mongo client throws ObjectDisposedException. MongoDatabase and MongoCollection are not disposable on other hand. However most of work regarding commands execution is implemented in OperationExecutor, that is passed into Database and Collection. OperationExecutor holds reference to mongo client, but not used to implement IDisposable - which means that operation executor passed to database would be half-operational after the mongo client would be disposed. By half-operational I mean it will throw on any code path which require accessing mongo client: for example any implicit session creation will require that.

This is what I've tried to address here - I've made OperationExecutor disposable and move all disposed checks there. So now we will have consistent behaviour: once mongo client will be disposed - all databases and collections created from the client will throw. So this is a good part.

The bed part: mongo client has a number of methods to clone itself, by adjusting some settings (like WithReadConcern). Such method should create a new copy of mongo client with separate operation executor - so disposal of the original mongo client will not affect the created one. Plus we have a tests that wants to substitute implementation of the OperationExcecutor with the mock.

The only option to achieve all of that is to introduce the OperationExecutorFactory parameter - the factory method that accepts IMongoClient and returns OperationExecutor. So tests could substitute it, and clonning methods could ask for the fresh OperationExecutor.

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -50,6 +51,9 @@ public MongoDatabase(IMongoClient client, DatabaseNamespace databaseNamespace, M
             _settings = Ensure.IsNotNull(settings, nameof(settings)).Freeze();
             _cluster = Ensure.IsNotNull(cluster, nameof(cluster));
             _operationExecutor = Ensure.IsNotNull(operationExecutor, nameof(operationExecutor));
+
+            _readOperationOptions = new ReadOperationOptions(DefaultReadPreference: _settings.ReadPreference);
+            _writeOperationOptions = new WriteOperationOptions();
```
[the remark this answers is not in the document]
sanych-sun: Done.

src/MongoDB.Driver/ReadOperationOptions.cs
sanych-sun: Done

src/MongoDB.Driver/WriteOperationOptions.cs
sanych-sun: Done

tests/MongoDB.Driver.Tests/OperationExecutorTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using Moq;
+using Xunit;
+
+namespace MongoDB.Driver.Tests
+{
+    public class OperationExecutorTests
+    {
+        [Theory]
+        [ParameterAttributeData]
+        public async Task StartImplicitSession_should_call_cluster([Values(true, false)]bool isAsync)
```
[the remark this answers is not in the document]
sanych-sun: Done

src/MongoDB.Driver/OperationExecutor.cs
sanych-sun: Add parameter to disable channel pinning + ticket to investigate why do we need that: https://jira.mongodb.org/browse/CSHARP-5600

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -58,6 +58,8 @@ private MongoCollectionImpl(IMongoDatabase database, CollectionNamespace collect
             _documentSerializer = Ensure.IsNotNull(documentSerializer, nameof(documentSerializer));
 
             _messageEncoderSettings = GetMessageEncoderSettings();
+            _readOperationOptions = new ReadOperationOptions(DefaultReadPreference: _settings.ReadPreference);
+            _writeOperationOptions = new WriteOperationOptions();
```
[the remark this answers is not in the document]
sanych-sun: Done

BorisDog (approved):
LGTM!

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -345,167 +328,162 @@ public IMongoCollection<TDocument> GetCollection<TDocument>(string name, MongoCo
             return new MongoCollectionImpl<TDocument>(this, new CollectionNamespace(_databaseNamespace, name), settings, _cluster, _operationExecutor);
         }
 
-        public IAsyncCursor<string> ListCollectionNames(ListCollectionNamesOptions options = null, CancellationToken cancellationToken = default(CancellationToken))
+        public IAsyncCursor<string> ListCollectionNames(ListCollectionNamesOptions options = null, CancellationToken cancellationToken = default)
         {
-            return UsingImplicitSession(session => ListCollectionNames(session, options, cancellationToken), cancellationToken);
+            var cursor = _operationExecutor.ExecuteReadOperation(
+                CreateListCollectionNamesOperation(options),
+                _readOperationOptions with { DefaultReadPreference = ReadPreference.Primary },
```
rstam: Why `DefaultReadPreference` instead of `ExplicitReadPreference`?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -947,15 +756,28 @@ private AggregateToCollectionOperation CreateAggregateToCollectionOperation<TRes
 
         private BulkMixedWriteOperation CreateBulkWriteOperation(
             IClientSessionHandle session,
-            IEnumerable<WriteModel<TDocument>> requests,
-            BulkWriteOptions options,
-            RenderArgs<TDocument> renderArgs)
+            IReadOnlyList<WriteModel<TDocument>> requests,
+            BulkWriteOptions options)
         {
+            Ensure.IsNotNull((object)requests, nameof(requests));
```
rstam: Usually arguments are validated by public methods only, to alert the user that they passed an invalid value.

When WE call private methods WE should never pass an invalid argument, so it is odd to be calling `Ensure.IsNotNull` here in a private method.

sanych-sun: The idea was to convert most of the public methods to expression-body methods (see other methods, for example `FindOneAndDelete`). Therefore I've decided to move some parameters validation inside of `CreateXXXOperation` methods.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: This came from ListCollectionNames method, see line 358 before the changes.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -293,314 +243,295 @@ public override MongoCollectionSettings Settings
...
-        {
-            return UsingImplicitSession(session => Count(session, filter, options, cancellationToken), cancellationToken);
-        }
+        public override long Count(FilterDefinition<TDocument> filter, CountOptions options, CancellationToken cancellationToken = default)
+            => _operationExecutor.ExecuteReadOperation(
+                CreateCountOperation(filter, options),
+                _readOperationOptions,
+                session: null,
+                allowChannelPinning: true,
+                cancellationToken: cancellationToken);
```
rstam: Why did you suddenly switch from `using var session ...` to this?

Wouldn't it be better to do:

```
public override long Count(FilterDefinition<TDocument> filter, CountOptions options, CancellationToken cancellationToken = default)
{                                                                                                                                  
    using var session = _operationExecutor.StartImplicitSession(cancellationToken: cancellationToken);                             
    return CountDocuments(session, filter, options, cancellationToken);                                                            
}                                                                                                                                  
```

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,33 +136,39 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
         // public methods
         /// <inheritdoc/>
         public ClientBulkWriteResult BulkWrite(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
-            => UsingImplicitSession(session => BulkWrite(session, models, options, cancellationToken), cancellationToken);
+            => _operationExecutor.ExecuteWriteOperation<ClientBulkWriteResult>(
+                CreateClientBulkWriteOperation(models, options),
+                _writeOperationOptions,
+                session: null,
+                allowChannelPinning: false,
+                cancellationToken: cancellationToken);
```
rstam: Why not use the `using var session ...` approach you used in parts of `MongoClientImpl.cs`?

```
public ClientBulkWriteResult BulkWrite(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
{                                                                                                                                                                 
    using var session = _operationExecutor.StartImplicitSession(cancellationToken);                                                                               
    return BulkWrite(session, models, options, cancellationToken);                                                                                                
}                                                                                                                                                                 
```

It's less code duplication (specifically not duplicating the tricky part of whether to pass `null` or not for `session` and what value to pass for `allowChannelPinning`.

It very clearly emphasizes that the only difference between this method and the next is that an implicit session is being used.

src/MongoDB.Driver/OperationExecutor.cs
```diff
@@ -13,50 +13,180 @@
...
+        public TResult ExecuteReadOperation<TResult>(IReadOperation<TResult> operation,
+            ReadOperationOptions options,
+            IClientSessionHandle session,
+            bool allowChannelPinning,
+            CancellationToken cancellationToken)
         {
-            return await operation.ExecuteAsync(binding, cancellationToken).ConfigureAwait(false);
+            ThrowIfDisposed();
+            var isOwnSession = session == null;
+            session ??= StartImplicitSession(cancellationToken);
```
rstam: Change implementation to NOT allow passing `null` for `session`.

Let callers create implicit session instead (see other comments).

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,33 +136,35 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
         /// <inheritdoc/>
-        public Task<ClientBulkWriteResult> BulkWriteAsync(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
-            => UsingImplicitSession(session => BulkWriteAsync(session, models, options, cancellationToken), cancellationToken);
+        public ClientBulkWriteResult BulkWrite(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
+            => _operationExecutor.ExecuteWriteOperation<ClientBulkWriteResult>(
+                CreateClientBulkWriteOperation(models, options),
+                _writeOperationOptions,
+                Ensure.IsNotNull(session, nameof(session)),
+                allowChannelPinning: false,
+                cancellationToken: cancellationToken);
```
rstam: I think converting this to an expression valued body is a mistake. The result is a complicated expression with lots of things happening in one single expression.

I propose the following:

```
public ClientBulkWriteResult BulkWrite(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
{                                                                                                                                                                                               
    var operation = CreateClientBulkWriteOperation(models, options);                                                                                                                            
    return ExecuteWriteOperation<ClientBulkWriteResult>(session, operation, cancellationToken);                                                                                                 
}                                                                                                                                                                                               
```

and keeping the `ExecuteWriteOperation` helper method but refactoring to:

```
private TResult ExecuteWriteOperation<TResult>(IClientSessionHandle session, IWriteOperation<TResult> operation, CancellationToken cancellationToken = default(CancellationToken))
{                                                                                                                                                                                 
    return _operationExecutor.ExecuteWriteOperation(session, operation, _writeOperationOptions, allowChannelPinning: false, cancellationToken);                                   
}                                                                                                                                                                                 
```

This allows the decomposing the various things going on into separate statements.

It also simplifies the code because the `options` and the `allowChannelPinning` values are the same for the entire file, and can be supplied once in the `ExecuteWrite` helper method.

src/MongoDB.Driver/IOperationExecutor.cs
```diff
@@ -16,20 +16,36 @@
...
-    internal interface IOperationExecutor
+    internal interface IOperationExecutor : IDisposable
     {
-        TResult ExecuteReadOperation<TResult>(IReadBinding binding, IReadOperation<TResult> operation, CancellationToken cancellationToken);
-        Task<TResult> ExecuteReadOperationAsync<TResult>(IReadBinding binding, IReadOperation<TResult> operation, CancellationToken cancellationToken);
+        TResult ExecuteReadOperation<TResult>(IReadOperation<TResult> operation,
+            ReadOperationOptions options,
+            IClientSessionHandle session,
+            bool allowChannelPinning,
+            CancellationToken cancellationToken);
```
rstam, line 30: Consider having `session` be the FIRST parameter for consistency with other methods where it is usually the first parameter.

Also consider formatting like this:

```
TResult ExecuteReadOperation<TResult>(
    IClientSessionHandle session,
    IReadOperation<TResult> operation,
    ReadOperationOptions options,
    bool allowChannelPinning,
    CancellationToken cancellationToken);
```

so that every parameter is on a line by itself. The idea is that either all parameters are on one line or if not each on its own line.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -97,10 +85,27 @@ public MongoClient(string connectionString)
         {
         }
 
-        internal MongoClient(IOperationExecutor operationExecutor, MongoClientSettings settings)
-            : this(settings)
+        internal MongoClient(MongoClientSettings settings, Func<IMongoClient, IOperationExecutor> operationExecutorFactory)
         {
-            _operationExecutor = operationExecutor;
+            _operationExecutorFactory = Ensure.IsNotNull(operationExecutorFactory, nameof(operationExecutorFactory));
+            _settings = Ensure.IsNotNull(settings, nameof(settings)).FrozenCopy();
```
rstam, line 90: Consider swapping 90-91 to initialize fields in the same order as they were provided in the parameter list.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,33 +136,35 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
+                cancellationToken: cancellationToken);
 
         /// <inheritdoc/>
-        public Task<ClientBulkWriteResult> BulkWriteAsync(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
+        public async Task<ClientBulkWriteResult> BulkWriteAsync(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
         {
-            var operation = CreateClientBulkWriteOperation(models, options);
-            return ExecuteWriteOperationAsync<ClientBulkWriteResult>(session, operation, cancellationToken);
+            using var session = _operationExecutor.StartImplicitSession();
+            return await BulkWriteAsync(session, models, options, cancellationToken).ConfigureAwait(false);
```
rstam, line 159: Refactor a bit to:

```
public Task<ClientBulkWriteResult> BulkWriteAsync(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
{                                                                                                                                                                            
    using var session = _operationExecutor.StartImplicitSession();                                                                                                           
    return BulkWriteAsync(session, models, options, cancellationToken);                                                                                                      
}                                                                                                                                                                            
```

because we don't need to use async/await in this case. It is slightly more efficient not to.

src/MongoDB.Driver/MongoClient.cs
rstam: Same for many other methods.

src/MongoDB.Driver/MongoClient.cs
rstam: Same for many other methods.

rstam (changes requested):
I'm going to stop my review here to see what you think of the suggestions so far. If you agree I will wait for a further commit.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: We DO need async/await, otherwise session will be disposed too early.

rstam: OK that's subtle. 

Either we should reconsider this code or add a comment.

rstam: Maybe even consider NOT using the `using var` shortcut to make the code structure emphasize a bit that the call to `await` is INSIDE a `using` scope:

```
using (var session = _operationExecutor.StartImplicitSession())
{
    return await BulkWriteAsync(session, models, options, cancellationToken).ConfigureAwait(false);
}                                                                                                 
```

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Done

src/MongoDB.Driver/IOperationExecutor.cs
sanych-sun: Done

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Done

src/MongoDB.Driver/MongoClient.cs
sanych-sun: I think we should be aware of this pitfall. Not sure if `using` block makes it easier to avoid the problem. I've run into similar "optimizations" well before `using` declaration was widely used.

rstam: True, we *should* be aware of this pitfall, but it is all too easy to miss when the tiny `using` keyword is hidden somewhere earlier in the method (it won't always be on the immediate line before it).

I suggested the refactoring because the indentation of the call to `await` provides a visual cue that this code is INSIDE a `using` scope.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,32 +136,36 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
+        }
 
         /// <inheritdoc/>
         public ClientBulkWriteResult BulkWrite(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
         {
+            ThrowIfDisposed();
+            Ensure.IsNotNull(session, nameof(session));
             var operation = CreateClientBulkWriteOperation(models, options);
-            return ExecuteWriteOperation<ClientBulkWriteResult>(session, operation, cancellationToken);
+            return ExecuteWriteOperation<ClientBulkWriteResult>(session, operation, _writeOperationOptions, cancellationToken);
```
rstam: `_writeOperationOptions` argument is unnecessary. It's always the same.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,32 +136,36 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
+            ThrowIfDisposed();
+            using var session = _operationExecutor.StartImplicitSession();
+            return BulkWrite(session, models, options, cancellationToken);
+        }
 
         /// <inheritdoc/>
         public ClientBulkWriteResult BulkWrite(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
         {
+            ThrowIfDisposed();
+            Ensure.IsNotNull(session, nameof(session));
```
rstam, line 148: I can't find it right now but I'm remembering a Microsoft guideline that arguments should be validated *before* throwing an `ObjectDisposedException`.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,32 +136,36 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
-
-        /// <inheritdoc/>
-        public Task<ClientBulkWriteResult> BulkWriteAsync(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
+        public async Task<ClientBulkWriteResult> BulkWriteAsync(IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
         {
-            var operation = CreateClientBulkWriteOperation(models, options);
-            return ExecuteWriteOperationAsync<ClientBulkWriteResult>(session, operation, cancellationToken);
+            ThrowIfDisposed();
+            using var session = _operationExecutor.StartImplicitSession();
+            return await BulkWriteAsync(session, models, options, cancellationToken).ConfigureAwait(false);
```
rstam, line 159: This way of calling the other overload also means that arguments are NOT validated when this overload of BulkWriteAsync is called, but any exceptions related to argument validation are postponed until we await the Task.

rstam: Correct implementation of `async` interfaces is not easy!

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -241,213 +247,163 @@ public IMongoDatabase GetDatabase(string name, MongoDatabaseSettings settings =
...
         public IAsyncCursor<string> ListDatabaseNames(
             ListDatabaseNamesOptions options,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             ThrowIfDisposed();
-
-            return UsingImplicitSession(session => ListDatabaseNames(session, options, cancellationToken), cancellationToken);
+            var listDatabasesOptions = CreateListDatabasesOptionsFromListDatabaseNamesOptions(options);
+            var databases = ListDatabases(listDatabasesOptions, cancellationToken);
```
rstam: Where did the implicit session go?

For consistency you should create the implicit session and call the other `ListDatabaseNames` overload.

I think you are trying to short circuit something here but don't think you should.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -241,213 +247,163 @@ public IMongoDatabase GetDatabase(string name, MongoDatabaseSettings settings =
...
+        public async Task<IAsyncCursor<string>> ListDatabaseNamesAsync(
             ListDatabaseNamesOptions options,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             ThrowIfDisposed();
-
-            return UsingImplicitSessionAsync(session => ListDatabaseNamesAsync(session, options, cancellationToken), cancellationToken);
+            var listDatabasesOptions = CreateListDatabasesOptionsFromListDatabaseNamesOptions(options);
+            var databases = await ListDatabasesAsync(listDatabasesOptions, cancellationToken).ConfigureAwait(false);
```
rstam: What happened to the implicit session here?

I think you are trying to short circuit something but it would be better to create the implicit session and call the other overlaod of `ListDatabaseNamesAsync`.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -241,213 +247,163 @@ public IMongoDatabase GetDatabase(string name, MongoDatabaseSettings settings =
...
             IClientSessionHandle session,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             ThrowIfDisposed();
-
-            return ListDatabases(session, null, cancellationToken);
+            Ensure.IsNotNull(session, nameof(session));
+            var operation = CreateListDatabaseOperation(null);
+            return ExecuteReadOperation(session, operation, _readOperationOptions, cancellationToken);
```
rstam: Just call the other overload passing `null` for `options`.

It's not a good idea to copy/paste the body of the other method here as an "optimization". If we ever change something in the other overload we could forget to make the same change here.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -241,213 +247,163 @@ public IMongoDatabase GetDatabase(string name, MongoDatabaseSettings settings =
...
             IClientSessionHandle session,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             ThrowIfDisposed();
-
-            return ListDatabasesAsync(session, null, cancellationToken);
+            Ensure.IsNotNull(session, nameof(session));
+            var operation = CreateListDatabaseOperation(null);
+            return ExecuteReadOperationAsync(session, operation, _readOperationOptions, cancellationToken);
```
rstam: Just call the other overload of `ListDatabasesAsync` like before. No need to copy/paste the implementation from that method to this one.

rstam (changes requested):

src/MongoDB.Driver/MongoClient.cs
sanych-sun: I suppose it was the same in previous version as well. Because `UsingImplicitSessionAsync` was async.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: I'll double-check the difference.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: I'll double-check the difference.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -458,95 +402,85 @@ internal Task<IClientSessionHandle> StartImplicitSessionAsync(CancellationToken
...
         /// <inheritdoc/>
         public IChangeStreamCursor<TResult> Watch<TResult>(
             IClientSessionHandle session,
             PipelineDefinition<ChangeStreamDocument<BsonDocument>, TResult> pipeline,
             ChangeStreamOptions options = null,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             Ensure.IsNotNull(session, nameof(session));
-            Ensure.IsNotNull(pipeline, nameof(pipeline));
```
rstam, line 476: Why was this line removed?

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -458,95 +402,85 @@ internal Task<IClientSessionHandle> StartImplicitSessionAsync(CancellationToken
...
         /// <inheritdoc/>
         public Task<IChangeStreamCursor<TResult>> WatchAsync<TResult>(
             IClientSessionHandle session,
             PipelineDefinition<ChangeStreamDocument<BsonDocument>, TResult> pipeline,
             ChangeStreamOptions options = null,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             Ensure.IsNotNull(session, nameof(session));
-            Ensure.IsNotNull(pipeline, nameof(pipeline));
```
rstam, line 503: Why was this line removed?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1298,22 +1154,30 @@ private FindOperation<TResult> CreateMapReduceOutputToCollectionFindOperation<TR
...
-            if (session.IsInTransaction && readPreference.ReadPreferenceMode != ReadPreferenceMode.Primary)
-            {
-                throw new InvalidOperationException("Read preference in a transaction must be primary.");
-            }
+        private TResult ExecuteReadOperation<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, ReadOperationOptions options, CancellationToken cancellationToken)
+            => _operationExecutor.ExecuteReadOperation(session, operation, options, true, cancellationToken);
 
-            return ChannelPinningHelper.CreateReadBinding(_cluster, session.WrappedCoreSession.Fork(), readPreference);
-        }
+        private Task<TResult> ExecuteReadOperationAsync<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, ReadOperationOptions options, CancellationToken cancellationToken)
```
rstam, line 1217: All but 2 callers pass _readOperationOptions.

Consider having two overloads and having the first call the second with _readOperationOptions.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1298,22 +1154,30 @@ private FindOperation<TResult> CreateMapReduceOutputToCollectionFindOperation<TR
...
+            return deferredCursor;
         }
 
-        private IReadBindingHandle CreateReadBinding(IClientSessionHandle session, ReadPreference readPreference)
-        {
-            if (session.IsInTransaction && readPreference.ReadPreferenceMode != ReadPreferenceMode.Primary)
-            {
-                throw new InvalidOperationException("Read preference in a transaction must be primary.");
-            }
+        private TResult ExecuteReadOperation<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, ReadOperationOptions options, CancellationToken cancellationToken)
```
rstam, line 1211: All but 2 callers pass _readOperationOptions.

Consider having two overloads and having the first call the second with _readOperationOptions.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1298,22 +1154,30 @@ private FindOperation<TResult> CreateMapReduceOutputToCollectionFindOperation<TR
...
+            => _operationExecutor.ExecuteReadOperation(session, operation, options, true, cancellationToken);
 
-            return ChannelPinningHelper.CreateReadBinding(_cluster, session.WrappedCoreSession.Fork(), readPreference);
-        }
+        private Task<TResult> ExecuteReadOperationAsync<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, ReadOperationOptions options, CancellationToken cancellationToken)
+            => _operationExecutor.ExecuteReadOperationAsync(session, operation, options, true, cancellationToken);
+
+        private TResult ExecuteWriteOperation<TResult>(IClientSessionHandle session, IWriteOperation<TResult> operation, WriteOperationOptions options, CancellationToken cancellationToken)
```
rstam: Remove `options` parameter?

Every single caller passes `_writeOperationOptions`.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1298,22 +1154,30 @@ private FindOperation<TResult> CreateMapReduceOutputToCollectionFindOperation<TR
...
+            => _operationExecutor.ExecuteWriteOperation(session, operation, options, true, cancellationToken);
+
+        private Task<TResult> ExecuteWriteOperationAsync<TResult>(IClientSessionHandle session, IWriteOperation<TResult> operation, WriteOperationOptions options, CancellationToken cancellationToken)
```
rstam: Remove `options` parameter?

Every single caller passes `_writeOperationOptions`.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -133,32 +136,36 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
+            ThrowIfDisposed();
+            using var session = _operationExecutor.StartImplicitSession();
+            return await BulkWriteAsync(session, models, options, cancellationToken).ConfigureAwait(false);
         }
 
         /// <inheritdoc/>
-        public void DropDatabase(string name, CancellationToken cancellationToken = default(CancellationToken))
+        public Task<ClientBulkWriteResult> BulkWriteAsync(IClientSessionHandle session, IReadOnlyList<BulkWriteModel> models, ClientBulkWriteOptions options = null, CancellationToken cancellationToken = default)
         {
+            Ensure.IsNotNull(session, nameof(session));
```
rstam, line 165: Should we use the `BulkWriteAsyncCore` approach to throw argument validation exceptions and `ObjectDisposedException` BEFORE entering the async portion of the code?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +96,121 @@ public override MongoCollectionSettings Settings
...
+            return await AggregateAsync(session, pipeline, options, cancellationToken).ConfigureAwait(false);
         }
 
-        public override async Task<IAsyncCursor<TResult>> AggregateAsync<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default(CancellationToken))
+        public override async Task<IAsyncCursor<TResult>> AggregateAsync<TResult>(IClientSessionHandle session, PipelineDefinition<TDocument, TResult> pipeline, AggregateOptions options, CancellationToken cancellationToken = default)
         {
             Ensure.IsNotNull(session, nameof(session));
             Ensure.IsNotNull(pipeline, nameof(pipeline));
-            options = options ?? new AggregateOptions();
+            options ??= new AggregateOptions();
```
rstam, line 136: Should we use the `AggregateAsync` approach to throw argument validation exceptions and `ObjectDisposedException` BEFORE entering the async portion of the code?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -94,160 +96,121 @@ public override MongoCollectionSettings Settings
...
-        public override BulkWriteResult<TDocument> BulkWrite(IClientSessionHandle session, IEnumerable<WriteModel<TDocument>> requests, BulkWriteOptions options, CancellationToken cancellationToken = default(CancellationToken))
+        public override BulkWriteResult<TDocument> BulkWrite(IClientSessionHandle session, IEnumerable<WriteModel<TDocument>> requests, BulkWriteOptions options, CancellationToken cancellationToken = default)
         {
             Ensure.IsNotNull(session, nameof(session));
-            Ensure.IsNotNull((object)requests, nameof(requests));
-
             var requestsArray = requests.ToArray();
-            if (requestsArray.Length == 0)
-            {
-                throw new ArgumentException("Must contain at least 1 request.", nameof(requests));
```
rstam, line 236: What happened to these argument validations?

rstam: Looks like you moved them to a PRIVATE helper method.

I don't recommend that.

Microsoft guidelines are to throw argument validation exceptions from the PUBLIC method. The idea is that the application writer should see the method THEY called at the top of the call stack, not some more deeply nested helper method.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -293,314 +243,245 @@ public override MongoCollectionSettings Settings
...
+            using var session = _operationExecutor.StartImplicitSession();
+            return Count(session, filter, options, cancellationToken);
         }
 
         [Obsolete("Use CountDocuments or EstimatedDocumentCount instead.")]
-        public override long Count(IClientSessionHandle session, FilterDefinition<TDocument> filter, CountOptions options, CancellationToken cancellationToken = default(CancellationToken))
+        public override long Count(IClientSessionHandle session, FilterDefinition<TDocument> filter, CountOptions options, CancellationToken cancellationToken = default)
         {
             Ensure.IsNotNull(session, nameof(session));
-            Ensure.IsNotNull(filter, nameof(filter));
```
rstam, line 305: Same comment here about missing argument validation in the PUBLIC method.

rstam: And in other places also.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1526,231 +1314,182 @@ public override MongoCollectionSettings Settings
...
+                using var session = _collection._operationExecutor.StartImplicitSession();
+                await DropOneAsync(session, name, options, cancellationToken).ConfigureAwait(false);
             }
 
-            public override Task DropOneAsync(IClientSessionHandle session, string name, CancellationToken cancellationToken = default(CancellationToken))
-            {
-                return DropOneAsync(session, name, null, cancellationToken);
-            }
+            public override Task DropOneAsync(IClientSessionHandle session, string name, CancellationToken cancellationToken = default)
+                => DropOneAsync(session, name, null, cancellationToken);
```
rstam, line 1493: Just a comment that I've seen many diffs like this method that are just noise. Nothing really changed, just syntactic changes.

It's not wrong to do this, but it adds considerable overhead to the review because of the additional diffs it creates.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Moved parameter validation back into public methods.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Moved parameter validation back into public methods.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Moved parameter validation back into public methods.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Moved parameter validation back into public methods.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Agree, but a far as we used to have this behavior before lets implement your suggestion in another PR. I'll create a ticket for that.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1138,9 +1020,16 @@ private FindOneAndReplaceOperation<TProjection> CreateFindOneAndReplaceOperation
...
-            FindOneAndUpdateOptions<TDocument, TProjection> options,
-            RenderArgs<TDocument> renderArgs)
+            FindOneAndUpdateOptions<TDocument, TProjection> options)
         {
+            options = options ?? new FindOneAndUpdateOptions<TDocument, TProjection>();
+
+            if (update is PipelineUpdateDefinition<TDocument> && (options.ArrayFilters != null && options.ArrayFilters.Any()))
+            {
+                throw new NotSupportedException("An arrayfilter is not supported in the pipeline-style update.");
+            }
```
rstam: Does this argument validation need to be moved back to the public method?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -947,15 +812,22 @@ private AggregateToCollectionOperation CreateAggregateToCollectionOperation<TRes
...
+            IReadOnlyList<WriteModel<TDocument>> requests,
+            BulkWriteOptions options)
         {
+            options ??= new BulkWriteOptions();
+            var renderArgs = GetRenderArgs();
             var effectiveWriteConcern = session.IsInTransaction ? WriteConcern.Acknowledged : _settings.WriteConcern;
 
+            var writeModels = requests.Select((model, index) =>
+            {
+                model.ThrowIfNotValid();
```
rstam, line 835: Does the validation need to be moved back to the public method?

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1526,231 +1355,195 @@ public override MongoCollectionSettings Settings
...
-                DropIndexOptions options,
-                CancellationToken cancellationToken)
+            public override void DropOne(IClientSessionHandle session, string name, DropIndexOptions options, CancellationToken cancellationToken)
             {
                 Ensure.IsNotNull(session, nameof(session));
                 Ensure.IsNotNullOrEmpty(name, nameof(name));
                 if (name == "*")
                 {
-                    throw new ArgumentException("Cannot specify '*' for the index name. Use DropAllAsync to drop all indexes.", "name");
+                    throw new ArgumentException("Cannot specify '*' for the index name. Use DropAll and DropAllAsync to drop all indexes.", "name");
```
rstam: Do we really want to mention `DropAllAsync` here?

They called `DropOne`, so the correct alternative method is `DropAll`.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1526,231 +1355,195 @@ public override MongoCollectionSettings Settings
...
-                DropIndexOptions options,
-                CancellationToken cancellationToken)
+            public override Task DropOneAsync(IClientSessionHandle session, string name, DropIndexOptions options, CancellationToken cancellationToken)
             {
                 Ensure.IsNotNull(session, nameof(session));
                 Ensure.IsNotNullOrEmpty(name, nameof(name));
                 if (name == "*")
                 {
-                    throw new ArgumentException("Cannot specify '*' for the index name. Use DropAllAsync to drop all indexes.", "name");
+                    throw new ArgumentException("Cannot specify '*' for the index name. Use DropAll and DropAllAsync to drop all indexes.", "name");
```
rstam: Do we really want to mention `DropAll` here?

They called `DropOneAsync`, so the correct alternative method is `DropAllAsync`.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1526,231 +1355,195 @@ public override MongoCollectionSettings Settings
...
             public override IEnumerable<string> CreateMany(
                 IClientSessionHandle session,
                 IEnumerable<CreateIndexModel<TDocument>> models,
                 CreateManyIndexesOptions options,
-                CancellationToken cancellationToken = default(CancellationToken))
+                CancellationToken cancellationToken = default)
             {
                 Ensure.IsNotNull(session, nameof(session));
-                Ensure.IsNotNull(models, nameof(models));
+                Ensure.IsNotNull((object)models, nameof(models));
```
rstam: Took me awhile to figure out why this cast to `(object)` was added.

At first glance it makes no sense!

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -305,30 +289,33 @@ public void DropCollection(IClientSessionHandle session, string name, DropCollec
         {
             Ensure.IsNotNull(session, nameof(session));
             Ensure.IsNotNullOrEmpty(name, nameof(name));
-            var operation = CreateDropCollectionOperation(name, options, session, cancellationToken);
+
+            var collectionNamespace = new CollectionNamespace(_databaseNamespace, name);
+            var encryptedFields = GetEncryptedFields(session, collectionNamespace, options, cancellationToken);
```
rstam: How is this related to this PR?

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -578,17 +514,22 @@ private ClientBulkWriteOperation CreateClientBulkWriteOperation(IReadOnlyList<Bu
...
-            ListDatabasesOptions options,
-            MessageEncoderSettings messageEncoderSettings,
-            ExpressionTranslationOptions translationOptions)
+        private DropDatabaseOperation CreateDropDatabaseOperation(string name)
+            => new(new DatabaseNamespace(name), GetMessageEncoderSettings())
+            {
+                WriteConcern = _settings.WriteConcern
+            };
+
+        private ListDatabasesOperation CreateListDatabaseOperation(ListDatabasesOptions options)
```
rstam: Might be a good opportunity to correct the spelling of this method, should be plural:

```CreateListDatabasesOperation```

File: src/MongoDB.Driver/MongoCollectionImpl.cs [hunk 56]
```diff
@@ -94,36 +96,25 @@ public override MongoCollectionSettings Settings
...
-            options = options ?? new AggregateOptions();
+            options ??= new AggregateOptions();
 
             var renderArgs = GetRenderArgs(options.TranslationOptions);
-            var renderedPipeline = pipeline.Render(renderArgs);
-
-            var lastStage = renderedPipeline.Documents.LastOrDefault();
-            var lastStageName = lastStage?.GetElement(0).Name;
-            if (lastStage != null && (lastStageName == "$out" || lastStageName == "$merge"))
+            var renderedPipeline = AggregateHelper.RenderAggregatePipeline(pipeline, renderArgs, out bool isAggregateToCollection);
```
rstam: Consider `out var isAggregateToCollection` instead of `out bool`

rstam: Same for other places.

src/MongoDB.Driver/MongoCollectionImpl.cs
rstam: Should this call the other overload instead of `_operationExecutor` directly?

    => ExecuteReadOperation(session, operation, _readOperationOptions, cancellationToken);

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -856,53 +745,68 @@ private ChangeStreamOperation<TResult> CreateChangeStreamOperation<TResult>(
...
         {
-            using (var binding = CreateReadBinding(session, readPreference))
-            {
-                return await _operationExecutor.ExecuteReadOperationAsync(binding, operation, cancellationToken).ConfigureAwait(false);
-            }
+            return collections.Select(collection => collection["name"].AsString);
         }
 
-        private T ExecuteWriteOperation<T>(IClientSessionHandle session, IWriteOperation<T> operation, CancellationToken cancellationToken)
+        private BsonDocument GetEncryptedFields(IClientSessionHandle session, CollectionNamespace collectionNamespace, DropCollectionOptions options, CancellationToken cancellationToken)
```
rstam: I don't know the difference between "encrypted fields" and "effective encrypted fields", but

    return effectiveEncryptedFields;

makes me think this method should be called `GetEffectiveEncryptedField`.

Same for async version.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -1298,22 +1189,36 @@ private FindOperation<TResult> CreateMapReduceOutputToCollectionFindOperation<TR
...
+            return deferredCursor;
         }
 
-        private IReadBindingHandle CreateReadBinding(IClientSessionHandle session, ReadPreference readPreference)
-        {
-            if (session.IsInTransaction && readPreference.ReadPreferenceMode != ReadPreferenceMode.Primary)
-            {
-                throw new InvalidOperationException("Read preference in a transaction must be primary.");
-            }
+        private TResult ExecuteReadOperation<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, CancellationToken cancellationToken)
```
rstam, line 1208: Maybe fix the order of these so the sync overloads are before the async overloads.

Instead of sync/async/sync/async.

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -856,53 +745,68 @@ private ChangeStreamOperation<TResult> CreateChangeStreamOperation<TResult>(
                 translationOptions);
         }
 
-        private IEnumerable<string> ExtractCollectionNames(IEnumerable<BsonDocument> collections)
-        {
-            return collections.Select(collection => collection["name"].AsString);
-        }
+        private TResult ExecuteReadOperation<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, CancellationToken cancellationToken)
```
rstam, line 748: Maybe fix the order of these so the sync overloads are before the async overloads.

Instead of sync/async/sync/async.

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -856,53 +745,68 @@ private ChangeStreamOperation<TResult> CreateChangeStreamOperation<TResult>(
...
+            => _operationExecutor.ExecuteReadOperation(session, operation, _readOperationOptions, true, cancellationToken);
```
rstam: Maybe call the other overload instead of `_operationExecutor` directly?

    => ExecuteReadOperation(session, operation, _readOperationOptions, cancellationToken);

src/MongoDB.Driver/MongoDatabase.cs
```diff
@@ -856,53 +745,68 @@ private ChangeStreamOperation<TResult> CreateChangeStreamOperation<TResult>(
...
 
-        private T ExecuteReadOperation<T>(IClientSessionHandle session, IReadOperation<T> operation, CancellationToken cancellationToken)
-        {
-            var readPreference = ReadPreferenceResolver.GetEffectiveReadPreference(session, null, _settings.ReadPreference);
-            return ExecuteReadOperation(session, operation, readPreference, cancellationToken);
-        }
+        private Task<TResult> ExecuteReadOperationAsync<TResult>(IClientSessionHandle session, IReadOperation<TResult> operation, CancellationToken cancellationToken)
+            => _operationExecutor.ExecuteReadOperationAsync(session, operation, _readOperationOptions, true, cancellationToken);
```
rstam: Maybe call the other overload instead of _operationExecutor directly?

    => ExecuteReadOperationAsync(session, operation, _readOperationOptions, cancellationToken);

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: @BorisDog as far as I remember you've asked to change the error message. Could you please confirm?

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: @BorisDog as far as I remember you've asked to change the error message. Could you please confirm?

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: As it was discussed, this prevents warning about possible multiple enumeration. Will revert.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: As it was discussed - this is refactoring of `CreateDropCollectionOperation` and `CreateDropCollectionOperationAsync` operations.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: I've merged validation with mapping to reduce number of enumerating over the collection. I keen to keep that optimization.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoClient.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: Done

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoDatabase.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: As it was discussed with Boris, I'll update wording to mention DropAll in DropOne, and DropAllAsync in DropOneAsync.

PR #1682, opened 2025-05-05
Add Big Endian Support for Float32 in BinaryVectorWriter.WriteToBytes<T>()

medhatiwari:
## Description
This PR adds Big Endian support for System.Single (Float32) to the BinaryVectorWriter.WriteToBytes<T>() method.

## Background
While running the MongoDB.Bson.Tests test suite on a Big Endian (s390x) system, we encountered 34 consistent test failures within the BinaryVectorSerializerTests class.
Each failure was caused by a System.NotSupportedException indicating that binary vector data of float32 type is not yet supported on Big Endian architectures.

##  Exception Observed
```
System.NotSupportedException: Binary vector data is not supported on Big Endian architecture yet.
```

### Sample Failing Tests
Some of the test cases that failed due to this limitation include:

`BinaryVectorSerializerTests.BinaryVectorSerializer_should_deserialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.BinaryVectorSerializer_should_serialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.ArrayAsBinaryVectorSerializer_should_deserialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.ArrayAsBinaryVectorSerializer_should_serialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.MemoryAsBinaryVectorSerializer_should_serialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.MemoryAsBinaryVectorSerializer_should_deserialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.ReadOnlyMemoryAsBinaryVectorSerializer_should_serialize_bson_vector<Float32>`

`BinaryVectorSerializerTests.ReadOnlyMemoryAsBinaryVectorSerializer_should_deserialize_bson_vector<Float32>`

### Why This Fix Is Necessary
This limitation was blocking test pass status on Big Endian platforms such as s390x. Adding support for float32 serialization in Big Endian format:

Enables consistent behavior across architectures

Completes existing deserialization support added earlier in BinaryVectorReader.cs

### Changes Introduced
Added Big Endian branch to BinaryVectorWriter.WriteToBytes<T>() for T == float.

Used `BinaryPrimitives.WriteSingleBigEndian()` to write bytes in the correct order.

Left existing Little Endian logic untouched to preserve behavior.

### DIFF
--- a/src/MongoDB.Bson/Serialization/BinaryVectorReader.cs
+++ b/src/MongoDB.Bson/Serialization/BinaryVectorReader.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
@@ -26,7 +27,6 @@ namespace MongoDB.Bson.Serialization
             where TItem : struct
         {
             var (items, padding, vectorDataType) = ReadBinaryVectorAsArray<TItem>(vectorData);
-
             return CreateBinaryVector(items, padding, vectorDataType);
         }
 
@@ -41,29 +41,38 @@ namespace MongoDB.Bson.Serialization
             switch (vectorDataType)
             {
                 case BinaryVectorDataType.Float32:
-
                     if ((vectorDataBytes.Span.Length & 3) != 0)
                     {
                         throw new FormatException("Data length of binary vector of type Float32 must be a multiple of 4 bytes.");
                     }
 
-                    if (BitConverter.IsLittleEndian)
+                    if (typeof(TItem) != typeof(float))
                     {
-                        var singles = MemoryMarshal.Cast<byte, float>(vectorDataBytes.Span);
-                        items = (TItem[])(object)singles.ToArray();
+                        throw new NotSupportedException($"Expected float for Float32 vector type, but found {typeof(TItem)}.");
                     }
-                    else
+
+                    int count = vectorDataBytes.Length / 4;
+                    float[] floatArray = new float[count];
+
+                    for (int i = 0; i < count; i++)
                     {
-                        throw new NotSupportedException("Binary vector data is not supported on Big Endian architecture yet.");
+                        floatArray[i] = BitConverter.IsLittleEndian
+                            ? MemoryMarshal.Read<float>(vectorDataBytes.Span.Slice(i * 4, 4))
+                            : BinaryPrimitives.ReadSingleBigEndian(vectorDataBytes.Span.Slice(i * 4, 4));
                     }
+
+                    items = (TItem[])(object)floatArray;
                     break;
+
                 case BinaryVectorDataType.Int8:
                     var itemsSpan = MemoryMarshal.Cast<byte, TItem>(vectorDataBytes.Span);
-                    items = (TItem[])(object)itemsSpan.ToArray();
+                    items = itemsSpan.ToArray();
                     break;
+
                 case BinaryVectorDataType.PackedBit:
                     items = (TItem[])(object)vectorDataBytes.ToArray();
                     break;
+
                 default:
                     throw new NotSupportedException($"Binary vector data type {vectorDataType} is not supported.");
             }
--- a/src/MongoDB.Bson/Serialization/BinaryVectorWriter.cs
+++ b/src/MongoDB.Bson/Serialization/BinaryVectorWriter.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Buffers.Binary;
 using System.Runtime.InteropServices;
 
 namespace MongoDB.Bson.Serialization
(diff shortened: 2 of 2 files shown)


medhatiwari:
Hi @BorisDog, if everything if fine, can this be merged?

medhatiwari:
Hi @BorisDog, just following up to check if there's any update on this PR. Please let me know if any further changes are needed.

src/MongoDB.Bson/Serialization/BinaryVectorReader.cs
```diff
@@ -41,29 +41,39 @@ public static (TItem[] Items, byte Padding, BinaryVectorDataType VectorDataType)
...
+                    int count = vectorDataBytes.Length / 4; // 4 bytes per float
+                    float[] floatArray = new float[count];
+
+                    for (int i = 0; i < count; i++)
                     {
-                        throw new NotSupportedException("Binary vector data is not supported on Big Endian architecture yet.");
+                        // Each float32 is 4 bytes. So to extract the i-th float, we slice 4 bytes from offset i * 4. Use little-endian or big-endian decoding based on platform.
+                        floatArray[i] = BitConverter.IsLittleEndian
+                            ? MemoryMarshal.Read<float>(vectorDataBytes.Span.Slice(i * 4, 4))   // fast, unaligned read on little endian
+                            : BinaryPrimitives.ReadSingleBigEndian(vectorDataBytes.Span.Slice(i * 4, 4));   // correctly reassemble 4 bytes as big-endian float
```
BorisDog, line 62: I think we need to read little endian here. BSON is always little endian, so we need to swap the bytes when reading on big endian.
If I am not mistaken, if we do `ReadSingleBigEndian` on a big endian machine, the bytes would not be swapped.

Same logic applies to Write as well.

In any case, we need to validate this solution with a real server, for example comparing the binary vector values read from Atlas deployment and the actual values.

BorisDog (changes requested):

BorisDog (changes requested):
Review is pending on requested changes.

src/MongoDB.Bson/IO/BinaryPrimitivesCompat.cs
```diff
@@ -31,5 +32,79 @@ public static void WriteDoubleLittleEndian(Span<byte> destination, double value)
...
+            int intValue = new FloatIntUnion { FloatValue = value }.IntValue;
+
+            destination[0] = (byte)(intValue);
+            destination[1] = (byte)(intValue >> 8);
+            destination[2] = (byte)(intValue >> 16);
+            destination[3] = (byte)(intValue >> 24);
+#endif
+        }
+
+        // This layout trick allows safely reinterpreting float as int and vice versa.
```
BorisDog, line 100: No need for the comment.

tests/MongoDB.Bson.Tests/IO/BinaryPrimitivesCompatTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Theory]
+        [InlineData(0f)]
+        [InlineData(1.0f)]
+        [InlineData(-1.5f)]
+        [InlineData(float.MaxValue)]
+        [InlineData(float.MinValue)]
+        [InlineData(float.NaN)]
+        [InlineData(float.PositiveInfinity)]
+        [InlineData(float.NegativeInfinity)]
+        public void WriteAndReadSingleLittleEndian_should_roundtrip_correctly(float value)
```
BorisDog, line 18: Organize methods in alphabetical order.

src/MongoDB.Bson/Serialization/BinaryVectorReader.cs
```diff
@@ -123,6 +111,30 @@ TExpectedItem[] AsTypedArrayOrThrow<TExpectedItem>()
...
+            float[] result;
+            if (BitConverter.IsLittleEndian)
+            {
+                result = MemoryMarshal.Cast<byte, float>(span).ToArray();
+            }
+            else
+            {
+                var count = span.Length / 4;
+		 result = new float[count];
+		 for (int i = 0; i < count; i++)
```
BorisDog, line 131: Please use 4 whitespaces instead of tab.

tests/MongoDB.Bson.Tests/IO/BinaryPrimitivesCompatTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [Fact]
+        public void ReadSingleLittleEndian_should_throw_on_insufficient_length()
+        {
+            var shortBuffer = new byte[3];
+            var exception = Record.Exception(() =>
+                BinaryPrimitivesCompat.ReadSingleLittleEndian(shortBuffer));
+
+            exception.Should().BeOfType<ArgumentOutOfRangeException>();
+            exception.Message.Should().Contain("length");
```
BorisDog, line 41: Please use the following pattern:
```
var e = exception.Should().BeOfType<ArgumentOutOfRangeException>().Subject;
e.ParamName.Should().Be("length");
```

and in `WriteSingleLittleEndian_should_throw_on_insufficient_length` as well.

src/MongoDB.Bson/IO/BinaryPrimitivesCompat.cs
```diff
@@ -31,5 +32,55 @@ public static void WriteDoubleLittleEndian(Span<byte> destination, double value)
...
         }
+
+        public static float ReadSingleLittleEndian(ReadOnlySpan<byte> source)
+        {
+#if NET6_0_OR_GREATER
+            return BinaryPrimitives.ReadSingleLittleEndian(source);
+#else
+            if (source.Length < 4)
+            {
+                throw new ArgumentOutOfRangeException(nameof(source), "Source span is too small to contain a float.");
```
BorisDog, line 43: `nameof(source.Length)`?

BorisDog (changes requested):
Few more minor comments.

BorisDog:
@medhatiwari  So would passing `nameof(source.Length)`  and expecting `Length` solve the issue?

medhatiwari:
> @medhatiwari I think `nameof(source.Length)` returns `Length` on all platforms. The problem is probably with `BinaryPrimitives` exception, which does not use `nameof(source.Length)`.
> 
> You can just move the validation to the begging of the method.

ah, moving the validation outside the #else block and using `nameof(source.Length)` did the trick — it now consistently returns "Length", and the tests pass. I’m still a bit curious though: when the same check was inside the #else, it was returning "length" instead. Any idea why that subtle difference in placement affects the casing?

BorisDog:
> I’m still a bit curious though: when the same check was inside the #else, it was returning "length" instead.
I couldn't reproduce that, but `BinaryPrimitives` does return low case "length".

BorisDog (approved):
LGTM!

PR #1684, opened 2025-05-06
CSHARP4040: Fix bug when using field with same element name as discriminator

papafe:


tests/MongoDB.Driver.Tests/Jira/Csharp4040Tests.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            [BsonId] public ObjectId Id { get; set; } = ObjectId.GenerateNewId();
+
+            [BsonElement("_t")]
+            public string Field1 { get; set; }
+        }
+
+        private class DerivedDocument : BaseDocument {}
+
+        [Fact]
```
papafe, line 37: These tests can eventually be moved to somewhere more appropriate.

tests/MongoDB.Bson.Tests/Serialization/Conventions/StandardDiscriminatorConventionTests.cs
```diff
@@ -28,6 +28,12 @@ public void TestConstructorThrowsWhenElementNameContainsNulls()
             Assert.Throws<ArgumentException>(() => new ScalarDiscriminatorConvention("a\0b"));
         }
 
+        [Fact]
```
papafe, line 31: This seems to be a reasonable limitation, doesn't it?

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1317,16 +1317,31 @@ public void UnmapProperty(string propertyName)
         /// Gets the discriminator convention for the class.
         /// </summary>
         /// <returns>The discriminator convention for the class.</returns>
-        internal IDiscriminatorConvention GetDiscriminatorConvention()
+        internal IDiscriminatorConvention GetDiscriminatorConvention(bool checkConflicts = false)
```
papafe, line 1320: I tried to find a place where to check that there are no conflicts with the other fields element names. I decided to do it here so it's done once and not continuously when serializing objects. 
For now `checkConflicts` is `true` only when `GetDiscriminatorConvention` is called in `BsonClassMapSerializer.SerializeDiscriminator` (that is used only when serializing classes). 
I'm not sure this is the optimal place for this, and if it would be worth throwing also when deserializing, for instance.

rstam: Why would we ever NOT check for conflicts? Why do we even need the `checkConflicts` parameter?

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -39,14 +39,19 @@ public abstract class StandardDiscriminatorConvention : IDiscriminatorConvention
         /// <param name="elementName">The element name.</param>
         protected StandardDiscriminatorConvention(string elementName)
         {
-            if (elementName == null)
+            if (string.IsNullOrEmpty(elementName))
             {
-                throw new ArgumentNullException("elementName");
+                throw new ArgumentNullException(nameof(elementName));
```
rstam, line 44: throw new ArgumentException("Element names cannot be null or empty.", nameof(elementName));

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -39,14 +39,19 @@ public abstract class StandardDiscriminatorConvention : IDiscriminatorConvention
...
             }
             if (elementName.IndexOf('\0') != -1)
             {
-                throw new ArgumentException("Element names cannot contain nulls.", "elementName");
+                throw new ArgumentException("Element names cannot be null or empty.", nameof(elementName));
```
rstam, line 48: throw new ArgumentException("Element names cannot contain nulls.", "elementName");

Should not have been changed.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -39,14 +39,19 @@ public abstract class StandardDiscriminatorConvention : IDiscriminatorConvention
...
             }
+            if (elementName == "_id")
+            {
+                throw new ArgumentException("Element names cannot be '_id'.", nameof(elementName));
```
rstam, line 52: Of course element names can be `_id`. I think what you wanted to say was that the discriminator element name could not be `_id`.

    throw new ArgumentException("The discriminator element name cannot be '_id'.", nameof(elementName));

rstam: But why? The discriminator element name can be anything that doesn't conflict with an element name that is already used.

If the class doesn't have an `_id` I don't see any reason why it couldn't be used as the discriminator element name.

I wouldn't recommend anyone doing that, but it seems legal.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1317,16 +1317,31 @@ public void UnmapProperty(string propertyName)
...
+            {
+                var conflictingMemberMap = _allMemberMaps.FirstOrDefault(memberMap => memberMap.ElementName == discriminatorConvention.ElementName);
+
+                if (conflictingMemberMap != null)
+                {
+                    var fieldOrProperty = conflictingMemberMap.MemberInfo is FieldInfo ? "field" : "property";
+
+                    throw new BsonSerializationException(
+                        $"The {fieldOrProperty} {conflictingMemberMap.MemberName} of type '{_classType.FullName}' cannot use element name '{discriminatorConvention.ElementName}' " +
+                        $"because it is already being used by the discriminator convention '{discriminatorConvention.GetType().Name}'.");
```
rstam, line 1341: I would turn the error message around:

    throw new BsonSerializationException(
        $"The discriminator element name cannot be {discriminatorConvention.ElementName} " +
        $"because it is already being used by the {fieldOrProperty} {conflictingMemberMap.MemberName} of type {_classType.FullName}");

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
rstam: Also, we should probably consider that even if the discriminator element name doesn't conflict that the base class level, what if a derived class introduces a conflicting element name?

rstam (changes requested):

tests/MongoDB.Driver.Tests/Jira/Csharp4040Tests.cs
```diff
@@ -0,0 +1,50 @@
...
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Jira
+{
+    public class Csharp4040Tests
```
rstam, line 25: Fix file name and class name:

`Csharp4040Tests` => `CSharp4040Tests`

rstam (changes requested):

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
papafe: Yes, this should have gone for the previous condition.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
papafe: We discussed on slack, but I've removed the parameter and moved the conflict checking inside so that it is run once and not all the time the method is called.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
rstam: Would you consider slightly more precise wording?

    throw new ArgumentException("Discriminator element name cannot be null or empty.", nameof(elementName));

I think `element name` (singular) is better than plural.

Also, I prefer `Discriminator element name` to `Element name` because element names CAN be *empty*. It is only *discriminator* element names that cannot be empty (but that's an arbitrary limitation that we are imposing, in theory they could be empty because `""` is an element name just as much as `"_t"` is).

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -39,14 +39,15 @@ public abstract class StandardDiscriminatorConvention : IDiscriminatorConvention
...
-            if (elementName == null)
+            if (string.IsNullOrEmpty(elementName))
             {
-                throw new ArgumentNullException("elementName");
+                throw new ArgumentException("Element names cannot be null or empty.", nameof(elementName));
             }
             if (elementName.IndexOf('\0') != -1)
             {
-                throw new ArgumentException("Element names cannot contain nulls.", "elementName");
+                throw new ArgumentException("Element names cannot contain nulls.", nameof(elementName));
```
rstam, line 48: Consider `Element name` (singular)?

rstam (changes requested):

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
papafe: I think it makes sense, the wording was to be uniform with the other exception. I've changed that too, I think the singular is more appropriate here.

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
papafe: I also think it's more appropriate, I will change the wording also on the other exception.

rstam (approved):
LGTM

papafe:
Hi @dorki, sorry to hear about that. Could you possibly create a ticket following the description [here](https://www.mongodb.com/docs/drivers/csharp/current/issues-and-help/#bugs---feature-requests)? 
It would also be useful to have a reproducible code snippet or an example of the expressions that are failing now, so we can have a better idea of the issue and possibly recommend some alternatives. You can add that to the ticket itself.

dorki:
will do, but may i ask what was the idea behind this change? since the element selected for the discriminator, is serialized, i should be able to query it, and therefore it must exists in the class map ?

PR #1685, opened 2025-05-06
CSHARP-3494: Fix discriminator for generic types

papafe:


tests/MongoDB.Driver.Tests/Jira/CSharp3494Tests.cs
```diff
@@ -0,0 +1,66 @@
...
+using FluentAssertions;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Driver.Tests.Linq.Linq3Implementation;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Jira
+{
+    public class CSharp3494Tests : Linq3IntegrationTest
```
papafe, line 25: We can decide to move these tests somewhere else if it makes more sense. Also this test does not need to derive from `Linq3IntegrationTest`.

tests/MongoDB.Driver.Tests/Jira/CSharp3494Tests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            public T Value { get; set; }
+        }
+
+        [Fact]
+        public void Correct_discriminator_should_be_used_for_generic_type()
+        {
+            var document = new DerivedDocument<int> { Id = 1, Value = 42 };
+            var serialized = document.ToJson(typeof(BaseDocument));
+            serialized.Should().Be("""{ "_t" : "DerivedDocument<Int32>", "_id" : 1, "Value" : 42 }""");
```
rstam, line 42: I thought the `_id` was always supposed to be first.

But maybe we only do that during serialization (and only in `BsonClassMapSerializer`).

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1057,7 +1057,7 @@ public void Reset()
             _creatorMaps.Clear();
             _creator = null;
             _declaredMemberMaps = new List<BsonMemberMap>();
-            _discriminator = _classType.Name;
+            _discriminator = BsonUtils.GetFriendlyTypeName(_classType);
```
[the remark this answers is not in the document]
papafe: We had an offline discussion, and we decided it would make sense to keep this fix limited to `BsonClassMap` based serializers.

tests/MongoDB.Driver.Tests/Jira/CSharp3494Tests.cs
papafe: Actually I investigated a little... It seems that for `BsonClassMapSerializer` the `_id` field gets serialized first only if `BsonSerializationArgs.SerializeIdFirst` is true, but only set it to true on tests.

tests/MongoDB.Driver.Tests/Jira/CSharp3494Tests.cs
```diff
@@ -0,0 +1,66 @@
...
+        public void Correct_type_should_be_instantiated_with_discriminator_for_generic_type()
+        {
+            //This test needs to use a different set of classes than the previous one, otherwise the discriminators could have been already
+            //registered, depending on the order of the tests. We need BsonKnownTypes for this to work.
+            var serialized = """{ "_t" : "DerivedDocument2<Int32>", "_id" : 1, "Value" : 42 }""";
+            var rehydrated = BsonSerializer.Deserialize<BaseDocument2>(serialized);
+            rehydrated.Should().BeOfType<DerivedDocument2<int>>();
+        }
+    }
+}
```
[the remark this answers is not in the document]
papafe: Done

tests/MongoDB.Driver.Tests/Jira/CSharp3494Tests.cs
papafe: I've moved these tests to the `DiscriminatorTests`. It seems more appropriate.

src/MongoDB.Bson/Serialization/BsonClassMap.cs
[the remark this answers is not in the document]
rstam: Do we need a new JIRA ticket for supporting discriminators for generic types when NOT using `BsonClassMap`?

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
```diff
@@ -65,6 +67,101 @@ private class H : G
...
+            [BsonId]
+            public int Id { get; set; }
+
+            public T1 Value1 { get; set; }
+
+            public T2 Value2 { get; set; }
+        }
+
+        //The deserialization tests for generic types needs to use a different set of classes than the serialization ones,
+        //otherwise the discriminators could have been already registered, depending on the order of the tests.
```
rstam, line 92: Why is this dependent on the order of the tests?

Configuration of classes/discriminators should work no matter what order the tests are run in.

rstam (changes requested):

src/MongoDB.Bson/Serialization/BsonClassMap.cs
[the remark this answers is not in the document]
papafe: I have created [this ticket here](https://jira.mongodb.org/browse/CSHARP-5605) to try to make the discriminator conventions completely separated from `BsonClassMap`.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
papafe: I think this here at this stage was probably unnecessary and I've uniformed the tests.

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
```diff
@@ -65,6 +67,79 @@ private class H : G
         {
         }
 
+        // BaseDocument, BasedDocument and derived classes are used for tests with generic types
```
rstam: This comment references a type `BasedDocument` that does not exist.

Maybe change to:

    // BaseDocument and derived classes are used for tests with generic types

rstam (approved):
LGTM with suggested change to one comment

tests/MongoDB.Bson.Tests/Serialization/Serializers/DiscriminatorTests.cs
papafe:
```suggestion
        // BaseDocument and derived classes are used for tests with generic types
```

papafe: Definitely. I'll fix it.

PR #1687, opened 2025-05-06
CSHARP-3268: Investigate Military Time Failures

rstam:


src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1373,7 +1373,7 @@ private DateTime ParseJavaScriptDateTimeString(string dateTimeString)
         {
             // if DateTime.TryParse succeeds we're done, otherwise assume it's an RFC 822 formatted DateTime string
             DateTime dateTime;
-            if (DateTime.TryParse(dateTimeString, out dateTime))
+            if (!dateTimeString.EndsWith("A") && DateTime.TryParse(dateTimeString, out dateTime))
```
rstam, line 1376: For some reason `DateTime.TryParse` returns `true` for time zone `A` but who knows what it thinks the `A` means.

If the string ends in `A` then we should use our own parsing method.

src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1479,30 +1479,30 @@ private DateTime ParseJavaScriptDateTimeString(string dateTimeString)
...
-                        case "Q": offset = TimeSpan.FromHours(4); break;
-                        case "R": offset = TimeSpan.FromHours(5); break;
-                        case "S": offset = TimeSpan.FromHours(6); break;
-                        case "T": offset = TimeSpan.FromHours(7); break;
-                        case "U": offset = TimeSpan.FromHours(8); break;
-                        case "V": offset = TimeSpan.FromHours(9); break;
-                        case "W": offset = TimeSpan.FromHours(10); break;
-                        case "X": offset = TimeSpan.FromHours(11); break;
-                        case "Y": offset = TimeSpan.FromHours(12); break;
+                        case "A": offset = TimeSpan.FromHours(1); break;
```
rstam, line 1482: The sign of all the offsets for `A` to `Y` was backwards.

tests/MongoDB.Bson.Tests/Jira/CSharp275Tests.cs
```diff
@@ -22,125 +22,110 @@ namespace MongoDB.Bson.Tests.Jira
...
-            public Test(string json, string iso)
-            {
-                this.Json = json;
-                this.Iso = iso;
-            }
-        }
-
-        private Test[] _tests = new Test[]
-        {
+        public static object[][] TestParseDatesData => [
```
rstam, line 25: Convert to supply data to a `Theory`.

BorisDog (commented):
Looks good, just a question about "A" and "Z" abbreviations.

src/MongoDB.Bson/IO/JsonReader.cs
rstam: In the existing test data nothing else ends with "A". But you are right, there might be a timezone abbreviation we don't know about that ends in A.

We could use a regular expression:

    if (!Regex.Matches(dateString, "[0-9 ]A$") && ...)

rstam: We have a test case for `Z`. I think it is a standard abbreviation for UTC and is handled by `DateTime.TryParse`.

src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1479,30 +1479,30 @@ private DateTime ParseJavaScriptDateTimeString(string dateTimeString)
...
+                        case "P": offset = TimeSpan.FromHours(-3); break;
+                        case "Q": offset = TimeSpan.FromHours(-4); break;
+                        case "R": offset = TimeSpan.FromHours(-5); break;
+                        case "S": offset = TimeSpan.FromHours(-6); break;
+                        case "T": offset = TimeSpan.FromHours(-7); break;
+                        case "U": offset = TimeSpan.FromHours(-8); break;
+                        case "V": offset = TimeSpan.FromHours(-9); break;
+                        case "W": offset = TimeSpan.FromHours(-10); break;
+                        case "X": offset = TimeSpan.FromHours(-11); break;
+                        case "Y": offset = TimeSpan.FromHours(-12); break;
```
rstam, line 1505: Did you mean that we should add a case for "Z" here?

I don't see how we would ever reach here with a "Z" because `DateTime.TryParse` would have already handled it.

src/MongoDB.Bson/IO/JsonReader.cs
BorisDog: Thanks, just wanted to confirm that Z is for UTC.
Looking at .net code, seems that A is related to Era (AD is also valid).

What worries me is that "Mon, 10 Oct 2011 11:22:33 A" is parsed to a certain value in .NET, and we'll be overriding that. If going with this change we need to convince ourselves that this is the right thing to do.

src/MongoDB.Bson/IO/JsonReader.cs
```diff
@@ -1373,7 +1373,9 @@ private DateTime ParseJavaScriptDateTimeString(string dateTimeString)
         {
             // if DateTime.TryParse succeeds we're done, otherwise assume it's an RFC 822 formatted DateTime string
             DateTime dateTime;
-            if (DateTime.TryParse(dateTimeString, out dateTime))
+
+            // DateTime.Parse doesn't understand military time zones, so don't call it when a military time zone is present
+            if (!Regex.IsMatch(dateTimeString, " [A-Y]$") && DateTime.TryParse(dateTimeString, out dateTime))
```
rstam, line 1378: Here's my suggestion: 

Either we support military time zones or we don't. And if we do support them, then we need to support ALL military time zones reliably.

Since `DateTime.TryParse` does *not* understand military time zones, use a regex to detect the presence of a military time zone and do NOT call `DateTime.TryParse` in that case. The reason to NOT call `DateTime.TryParse` is that `DateTime.TryParse` gets confused by SOME military time zones and interprets them as something else.

I'm not sure why we call `DateTime.TryParse` at all here... since we have our own parsing for RFC 822 format date strings. I guess the idea here is to be "lenient" in what we accept, and that we will parse anything that `DateTime.TryParse` understands and fall back to our own RFC 822 datetime parser for anything `DateTime.TryParse` does not understand.

rstam:
Blocked on CSHARP-5604.

PR #1688, opened 2025-05-07
CSHARP1907: Fix discriminator for interface-based collections

papafe:


papafe:
I'm still investigating what is the best solution for this. I've opened this as a draft to have a starting point for discussion and to see the CI results.

PR #1689, opened 2025-05-07
DEVPROD-17449 - Replace perf.send with new direct end point for performance data submission

MAhmadShah:
The perf.send functionality in evergreen is no longer maintained and is not the preferred method of sending performance data to the signal processing service. This PR updates the evergreen yaml file to instead send the data down stream using the preferred performance-monitoring-api.corp.mongodb.com/raw_perf_results/cedar_report end point.

Changes were tested in [this patch](https://spruce.mongodb.com/version/681b77b8ddc9c60007e74f7c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


adelinowona (approved):
LGTM!

PR #1697, opened 2025-05-19
CSHARP-5593: Add .git-blame-ignore-revs file

adelinowona:
This is adding a file in which we can list the commit hashes of commits we want git blame to ignore. This file is picked up by Github and used when looking at the Blame tab on the Github web UI.


BorisDog (approved):
LGTM

PR #1700, opened 2025-05-27
CSHARP-5572: Implement new SerializerFinder

rstam:


src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerMap.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                serializer = UpcastingSerializer.Create(baseType: serializer.ValueType, derivedType: node.Type, baseTypeSerializer: serializer);
+            }
+            else
+            {
+                throw new ArgumentException($"Serializer value type {serializer.ValueType} does not match expression value type {node.Type}", nameof(serializer));
+            }
+        }
+
+        if (_map.TryGetValue(node, out var existingSerializer))
```
adelinowona, line 79: I think we should move the duplicate check to the beginning of the `AddSerializer` method to implement fail-fast behavior. This would avoid unnecessary serializer transformation work when a serializer already exists for the node.

However, I notice the current implementation places this check at the end, likely to preserve information about the conflicting serializer for the exception. While this debugging information could be helpful, I question just how useful it is. Knowing where a duplicate attempt happened seems more useful knowing what the conflicting serializer would have been.

I lean toward the fail-fast approach since the conflicting serializer details seem like a "nice-to-have" rather than essential debugging information.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitor.cs
```diff
@@ -0,0 +1,74 @@
...
+using ExpressionVisitor = System.Linq.Expressions.ExpressionVisitor;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.KnownSerializerFinders;
+
+internal partial class KnownSerializerFinderVisitor : ExpressionVisitor
+{
+    private bool _isMakingProgress = true;
+    private readonly KnownSerializerMap _knownSerializers;
+    private int _oldKnownSerializersCount = 0;
+    private int _pass = 0;
```
adelinowona: Seems like this is never used?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitor.cs
```diff
@@ -0,0 +1,74 @@
...
+    public int Pass => _pass;
+
+    public bool IsMakingProgress => _isMakingProgress;
+
+    public void EndPass()
+    {
+        var newKnownSerializersCount = _knownSerializers.Count;
+        if (newKnownSerializersCount == _oldKnownSerializersCount)
+        {
+            if (_useDefaultSerializerForConstants)
```
adelinowona, line 43: Could we maybe make use of a state enum as such:
```
private enum DiscoveryPhase
{
    InferringFromContext,      // Not using defaults
    UsingDefaultSerializers,   // Last attempt with defaults
    Complete                    // No more progress possible
}
```

to replace the use of all these booleans used in keeping track of the state of the visitor?

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitor.cs
rstam: Removed.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitor.cs
rstam: It's somewhat reasonable, but I think the booleans work better here.

There are only 2 booleans.

One issue is that if we ever had more states the order of the enum values could get very confusing and maybe even no order is possible.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerMap.cs
rstam: If we move the error check further up then we have to remove information about the redundant serializer from the error message.

Also, I don't think we need to worry about failing fast because we actually never expect to hit this error. This check is here to help us find bugs. In the absence of bugs this exception will never be thrown.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinder.cs
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Linq.Expressions;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.KnownSerializerFinders;
+
+internal static class KnownSerializerFinder
```
sanych-sun: Term `Known` in the class name suggests there might be unknown serializers. But as far as I remember from our meetings currently we make sure each expression MUST have serializer assigned. May be we should rename it to `SerializerFinder`?

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerMap.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public bool IsNotKnown(Expression node)
+    {
+        return !IsKnown(node);
+    }
+
+    public bool IsKnown(Expression node)
+    {
+        return node != null && _map.ContainsKey(node);
```
sanych-sun: Is there any real use cases when we pass node=null? Sounds like we should throw instead.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3431 @@
...
+            {
+                DeduceCollectionAndItemSerializers(collectionExpression, itemExpression);
+                DeduceReturnsBooleanSerializer();
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+
+            bool IsCollectionContainsMethod(out Expression collectionExpression, out Expression itemExpression)
```
sanych-sun: We already have similar detection in at least 2 places:
ContainsMethodToAggregationExpressionTranslator,
ContainsMethodToFilterTranslator

Now it's going to be 3rd. Should we consolidate this into some helper method?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3431 @@
...
+                }
+
+                DeduceReturnsBooleanSerializer();
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+
+            bool IsContainsValueInstanceMethod(out Expression collectionExpression, out Expression valueExpression)
```
sanych-sun, line 868: We have very similar logic in:
ContainsValueMethodToFilterTranslator,
ContainsValueMethodToAggregationExpressionTranslator

Now it's going to be 3rd. Should we consolidate this into some helper method?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3431 @@
...
+
+                DeduceCollectionAndCollectionSerializers(objectExpression, otherExpression);
+                DeduceReturnsBooleanSerializer();
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+
+            static bool IsSetEqualsMethod(MethodInfo method)
```
sanych-sun: Similar checks are done in:
SetEqualsMethodToAggregationExpressionTranslator,

Should we consolidate into some helper method?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3431 @@
...
+                    AddKnownSerializer(node, resultSerializer);
+                }
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+        }
+
+        bool IsDictionaryContainsKeyMethod(out Expression keyExpression)
```
sanych-sun: Almost exact logic could be find in:
ContainsKeyMethodToAggregationExpressionTranslator,
ContainsKeyMethodToFilterTranslator

This is 3rd copy of this. Also this version of check is more strict (DeclaringType should mention "Dictionary"). Is it OK to have a diverged logic here and in Translators? I suppose we have to consolidate all 3 usages into helper method.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
```diff
@@ -27,7 +27,6 @@ namespace MongoDB.Bson.Serialization.Serializers
     /// <typeparam name="TDictionary">The type of the dictionary.</typeparam>
     public abstract class DictionarySerializerBase<TDictionary> :
         ClassSerializerBase<TDictionary>,
-        IBsonDocumentSerializer,
```
sanych-sun, line 30: Removing of the interface is a breaking change. Do we really need to remove it?

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
```diff
@@ -132,22 +131,6 @@ obj is DictionarySerializerBase<TDictionary> other &&
         /// <inheritdoc/>
         public override int GetHashCode() => 0;
 
-        /// <inheritdoc/>
-        public bool TryGetMemberSerializationInfo(string memberName, out BsonSerializationInfo serializationInfo)
```
sanych-sun, line 136: Removing of public method is the breaking change. Do we really need to remove it?

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -50,6 +50,48 @@ public static TValue Deserialize<TValue>(this IBsonSerializer<TValue> serializer
...
+            {
+                return derivedTypeSerializer;
+            }
+
+            if (!baseType.IsAssignableFrom(derivedTypeSerializer.ValueType))
+            {
+                throw new ArgumentException($"{baseType} is not assignable from {derivedTypeSerializer.ValueType}.");
+            }
+
+            return BsonSerializer.LookupSerializer(baseType); // TODO: should be able to navigate from serializer
```
sanych-sun: Is there any way to implement the logic without using of static API?

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -50,6 +50,48 @@ public static TValue Deserialize<TValue>(this IBsonSerializer<TValue> serializer
...
+            {
+                return baseTypeSerializer;
+            }
+
+            if (!baseTypeSerializer.ValueType.IsAssignableFrom(derivedType))
+            {
+                throw new ArgumentException($"{baseTypeSerializer.ValueType} is not assignable from {derivedType}.");
+            }
+
+            return BsonSerializer.LookupSerializer(derivedType); // TODO: should be able to navigate from serializer
```
sanych-sun: Is there any way to implement the logic without using of static API?

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderHelperMethods.cs
```diff
@@ -0,0 +1,251 @@
...
+                knownSerializers = null;
+                return false;
+            }
+        }
+
+        knownSerializers = knownSerializersList;
+        return true;
+    }
+
+    private bool AnyIsKnown(IEnumerable<Expression> nodes, out IBsonSerializer knownSerializer)
```
sanych-sun: _Minor_: May be `IsAnyKnown`? Here and more in this class.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
```diff
@@ -0,0 +1,251 @@
...
+        if (itemSerializer is IUnknowableSerializer)
+        {
+            return UnknowableSerializer.Create(collectionType);
+        }
+
+        return collectionType switch
+        {
+            _ when collectionType.IsArray => ArraySerializer.Create(itemSerializer),
+            _ when collectionType.IsConstructedGenericType && collectionType.GetGenericTypeDefinition() == typeof(IEnumerable<>) => IEnumerableSerializer.Create(itemSerializer),
+            _ when collectionType.IsConstructedGenericType && collectionType.GetGenericTypeDefinition() == typeof(IOrderedEnumerable<>) => IOrderedEnumerableSerializer.Create(itemSerializer),
```
sanych-sun, line 99: Is this code reachable? As far as I understood `IOrderedEnumerable` is inhereted from  `IEnumerable`, so it will choose the previous branch.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
```diff
@@ -0,0 +1,251 @@
...
+            _ when collectionType.IsConstructedGenericType && collectionType.GetGenericTypeDefinition() == typeof(IQueryable<>) => IQueryableSerializer.Create(itemSerializer),
```
sanych-sun, line 100: Same for `IQueryable`: it implements `IEnumerable`

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitConditional.cs
```diff
@@ -0,0 +1,40 @@
...
+        var ifTrueExpression = node.IfTrue;
+        var ifFalseExpression = node.IfFalse;
+
+        DeduceConditionaSerializers();
+        base.VisitConditional(node);
+        DeduceConditionaSerializers();
+
+        return node;
+
+        void DeduceConditionaSerializers()
```
sanych-sun: Typo: "ConditionaL"

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitIndex.cs
```diff
@@ -0,0 +1,86 @@
...
+                (declaringType == typeof(BsonValue) || declaringType.IsSubclassOf(typeof(BsonValue))) &&
+                arguments.Count == 1 &&
+                arguments[0] is var index &&
+                (index.Type == typeof(int) || index.Type == typeof(string));
+        }
+
+        bool IsDictionaryIndexer()
+        {
+            return
+                indexer.DeclaringType.Name.Contains("Dictionary") &&
```
sanych-sun: Are we really need to validate the class name? Is there any better way to do that? For example `Hashtable` or `SortedList` implements IDictionary, but it has no Dictionary in the name.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            return
+                (declaringType.ImplementsInterface(typeof(IEnumerable)) || declaringType == typeof(BitArray)) &&
+                node.Type == typeof(int) &&
+                (member.Name == "Count" || member.Name == "Length");
+        }
+
+        bool IsTupleOrValueTuple(Type type)
+        {
+            return
```
sanych-sun: For `net6` and higher there is a better way to check: both `Tuple` and `ValueTuple` implements `ITuple` interface. I suppose we can have #if def here and use interface approach for modern targets, when have something like the proposed code for net472.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitUnary.cs
```diff
@@ -0,0 +1,312 @@
...
+                if (!sourceType.IsEnum)
+                {
+                    throw new ExpressionNotSupportedException(expression, because: "source type is not an enum");
+                }
+                if (!targetType.IsEnum)
+                {
+                    throw new ExpressionNotSupportedException(expression, because: "target type is not an enum");
+                }
+
+                // if (sourceSerializer is IHasRepresentationSerializer sourceHasRepresentationSerializer &&
```
sanych-sun: Do we need this commented code?

sanych-sun (changes requested):
Partial review.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
rstam: Yes we need to do this. This interface was not supposed to be implemented and it makes no sense to do so.

I believe Adelin has done the same thing in his Dictionary support branch.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
rstam: See above.

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
rstam: Not until Ferdinando finishes his work on serialization domains.

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
rstam: Not until Ferdinando finishes his work on serialization domains.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinder.cs
rstam: We have been calling this the KnownSerializerFinder since the beginning of LINQ3.

But I will rename it at your request. A shorter name is nice.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderHelperMethods.cs
rstam: Done.

Same for several other similarly named method in this file.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
rstam: Should be reachable.

Note that we are checking for `==` not "implements".

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
rstam: Same as above.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitConditional.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitIndex.cs
rstam: Good question. I will investigate.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitListInit.cs
```diff
@@ -0,0 +1,38 @@
...
+        var newExpression = node.NewExpression;
+        var initializers = node.Initializers;
+
+        DeduceSerialiers();
+        base.VisitListInit(node);
+        DeduceSerialiers();
+
+        return node;
+
+        void DeduceSerialiers()
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Nice to know!

Is there any danger of false positives if unexpected classes also happen to implement `ITuple`?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: I will consolidate in future commit.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitUnary.cs
```diff
@@ -0,0 +1,312 @@
...
+            return null;
+
+            static IBsonSerializer GetTargetSerializer(UnaryExpression node, Type sourceType, Type targetType, IBsonSerializer sourceSerializer)
+            {
+                if (targetType == sourceType)
+                {
+                    return sourceSerializer;
+                }
+
+                // handle conversionsn to BsonValue before any others
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitUnary.cs
rstam: I will investigate.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerMap.cs
rstam: I removed the null check and nothing failed.

I don't think we need to throw on `null`. We can just let `ContainsKey` throw for us.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/UnknownSerializerFinder.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private Expression _expressionWithUnknownSerializer = null;
+    private readonly KnownSerializerMap _knownSerializers;
+
+    public UnknownSerializerFinder(KnownSerializerMap knownSerializers)
+    {
+        _knownSerializers = knownSerializers;
+    }
+
+    public Expression ExpressionWithUnknownSerialier => _expressionWithUnknownSerializer;
```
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitConditional.cs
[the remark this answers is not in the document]
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitIndex.cs
rstam: Note: `Hashtable` implements `IDctionary` but not `IDictionary<TKey, TValue>`.

Same for `SortedList`.

Not sure we intended to support those, and if we do maybe in a different ticket?

We don't currently have any LINQ tests for `Hashtable` or `SortedList`.

rstam: Refactored as discussed on Slack.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Not doing just yet until we have a discussion about false positives.

src/MongoDB.Driver/Linq/Linq3Implementation/KnownSerializerFinders/KnownSerializerFinderVisitMethodCall.cs
rstam: Refactored all cases to use existing `EnumerableMethod.IsContainsMethod` method.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Use existing `EnumerableMethod.IsConstaintMethod` method.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Use existing `EnumerableMethod.IsConstainsMethod` method.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitUnary.cs
rstam: If I uncomment some tests involving client-side projections fail.

Enum to enum server-side requires that the representation be numeric, but client-side does not (because the conversion happens after deserialization of the source enum).

I will figure out how to support this.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5532Tests.cs
```diff
@@ -0,0 +1,199 @@
...
+                                    0
+                                ]
+                            }
+                        },
+                        in : "$$this._id"
+                    }
+                }
+            }
+            """;
+        if (!Feature.FilterLimit.IsSupported(CoreTestConfiguration.MaxWireVersion))
```
rstam, line 138: We usually create two strings depending on whether a feature is supported or not but, in this case, the MQL is so long it's easier to just remove `limit : 1` if it is not supported.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitUnary.cs
rstam: I researched it.

It is OK to ignore this problem in the `SerializerFinder`.

It will be correctly detected in the translator.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Refactored into a single shared method.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
sanych-sun: I do agree that it looks suspicious for DictionarySerializer to implement IBsonDocumentSerializer. But we trying to avoid breaking changes in minor releases. Is there any way to separate this into additional ticket and do it later in scope of v4?

rstam: I would not call this a breaking change.

The IBsonDocumentSerializer interface is for POCOs where each property of the POCO is mapped to an element of the BsonDocument. 

That simply is not true of a Dictionary, even if the representation of Dictionary is a document.

The properties of the Dictionary class do NOT have matching elements in any BsonDocument, so implementation of this interface is a mistake.

I can't think of a single thing that might have worked before that removal of this interface would break.

That being said, I did revert this change to see if this PR depended on this in any way, and all tests still pass anyway, so if you insist, we could factor this out to a separate JIRA ticket.

@adelinowona weren't you going to do this same thing in your PR for full Dictionary support?

sanych-sun: I do agree that this interface looks redundant here, but `DictionarySerializerBase` is a public class and we are about to remove an interface implementation from it - technically this is a breaking change. IMHO we'd better to postpone this change until v4 if it's possible.

rstam: I thought it was a required change for this PR, but apparently not since all tests pass without out it.

I have reverted this change.

I will create a new JIRA ticket for this.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitBinary.cs
```diff
@@ -0,0 +1,182 @@
...
+        var rightExpression = node.Right;
+
+        if (node.NodeType == ExpressionType.Add && node.Type == typeof(string))
+        {
+            DeduceStringSerializer(node);
+            return node;
+        }
+
+        if (IsSymmetricalBinaryOperator(@operator) &&
+            CanDeduceSerializer(leftExpression, rightExpression, out var unknownNode, out var otherNodeSerializer))
```
sanych-sun: _Minor_: I found only 2 usages of `CanDeduceSerializer`, both looks similar (TODO in the another usage says we have to check value types). I think we can simplify by refactoring to `TryDeduceSerializer` and do the type check and `AddNodeSerializer` inside it.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitBinary.cs
```diff
@@ -0,0 +1,182 @@
...
+                else if (
+                    leftSerializer is INullableSerializer nullableSerializer &&
+                    nullableSerializer.ValueSerializer is var nullableSerializerValueSerializer &&
+                    nullableSerializerValueSerializer.ValueType == node.Type)
+                {
+                    AddNodeSerializer(node, nullableSerializerValueSerializer);
+                }
+                else
+                {
+                    DeduceUnknowableSerializer(node);
```
sanych-sun: It worth to add a comment why we decided stop looking for the serializer for the node.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/IgnoreNodeSerializer.cs
```diff
@@ -0,0 +1,33 @@
...
+internal static class IgnoreNodeSerializer
+{
+    public static IBsonSerializer Create(Type valueType)
+    {
+        var serializerType = typeof(IgnoreNodeSerializer<>).MakeGenericType(valueType);
+        return (IBsonSerializer)Activator.CreateInstance(serializerType);
+    }
+}
+
+internal class IgnoreNodeSerializer<TValue> : SerializerBase<TValue>
```
sanych-sun, line 39: What the difference between `IgnoreNodeSerializer` and `UnknowableSerializer`?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitListInit.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Linq.Expressions;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
+
+internal partial class SerializerFinderVisitor
+{
+    protected override Expression VisitListInit(ListInitExpression node)
+    {
+        var newExpression = node.NewExpression;
+        var initializers = node.Initializers;
```
sanych-sun: `initializers` variable seems to be unused. Should we handle initializers somehow?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
```diff
@@ -0,0 +1,234 @@
...
+        var memberName = member.Name;
+
+        base.VisitMember(node);
+
+        if (IsNotKnown(node))
+        {
+            var containingExpression = node.Expression;
+            if (IsKnown(containingExpression, out containingSerializer))
+            {
+                // TODO: handle special cases like DateTime.Year etc.
```
sanych-sun: Is this TODO still actual?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                var resultSerializer = node.Member switch
+                {
+                    _ when declaringType == typeof(BsonValue) => GetBsonValuePropertySerializer(),
+                    _ when IsCollectionCountOrLengthProperty() => GetCollectionCountOrLengthPropertySerializer(),
+                    _ when declaringType == typeof(DateTime) => GetDateTimePropertySerializer(),
```
sanych-sun, line 51: Should we do something similar for `DateTimeOffset`?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                throw new ExpressionNotSupportedException(node, because: "IDictionarySerializer does not implement IBsonDictionarySerializer");
+            }
+
+            var keySerializer =  dictionarySerializer.KeySerializer;
+            var valueSerializer = dictionarySerializer.ValueSerializer;
+
+            return memberName switch
+            {
+                "Keys" => ICollectionSerializer.Create(keySerializer),
```
sanych-sun, line 170: I was expecting `GetDictionaryPropertySerializer` and `GetIDictionaryPropertySerializer` to be the exact copy. Why do we use `ICollectionSerializer` here instead of `DictionaryKeyCollectionSerializer` and `DictionaryValueCollectionSerializer`.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.ExtensionMethods;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
+
+internal partial class SerializerFinderVisitor
+{
+    private static readonly HashSet<MethodInfo> __absMethods =
```
sanych-sun: We have exactly the same collection in `AbsMethodToAggregationExpressionTranslator`. Should we move it to some shared place?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+        MathMethod.AbsDecimal,
+        MathMethod.AbsDouble,
+        MathMethod.AbsInt16,
+        MathMethod.AbsInt32,
+        MathMethod.AbsInt64,
+        MathMethod.AbsSByte,
+        MathMethod.AbsSingle
+    ];
+
+    private static readonly HashSet<MethodInfo> __aggregateMethods =
```
sanych-sun: We have exactly the same collection in `AggregateMethodToAggregationExpressionTranslator`. Should we move it to some shared place?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+        WindowMethod.PercentileWithInt64,
+        WindowMethod.PercentileWithNullableDecimal,
+        WindowMethod.PercentileWithNullableDouble,
+        WindowMethod.PercentileWithNullableInt32,
+        WindowMethod.PercentileWithNullableInt64,
+        WindowMethod.PercentileWithNullableSingle,
+        WindowMethod.PercentileWithSingle
+    ];
+
+    private static readonly HashSet<MethodInfo> __countMethods =
```
sanych-sun: This collection present already in 2 other places:
`CountMethodToAggregationExpressionTranslator` and `CountMethodToAggregationExpressionTranslator`. Should we move it to some shared place?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+        EnumerableMethod.CountWithPredicate,
+        EnumerableMethod.LongCount,
+        EnumerableMethod.LongCountWithPredicate,
+        QueryableMethod.Count,
+        QueryableMethod.CountWithPredicate,
+        QueryableMethod.LongCount,
+        QueryableMethod.LongCountWithPredicate
+    ];
+
+    private static readonly HashSet<MethodInfo> __countWithPredicateMethods =
```
sanych-sun: This class contains many static collections that are in common with translators. Can we consolidate these collections in some helper classes to reduce the code duplication?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+        }
+
+        bool IsDictionaryContainsKeyMethod(out Expression keyExpression)
+        {
+            if (method.DeclaringType.Name.Contains("Dictionary") &&
```
sanych-sun: Let's move this method to `DictionaryMethod` helper class and avoid Contains("Dictionary") by checking for interface or base class.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+        }
+
+        void DeduceAggregateMethodSerializers()
+        {
+            if (method.IsOneOf(__aggregateMethods))
```
sanych-sun: I cannot find any usage of results of `IsItemSerializerKnown` method call results, so it probably could be removed. And if it's removed then outer `if` is redundant.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+                    AddNodeSerializer(node, resultSerializer);
+                }
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+        }
+
+        void DeduceAsinhMethodSerializers()
```
sanych-sun: We have number of exactly the same methods in this class to process all math trigonometry methods, should we group then into a single collection `__trigonometricMathMethods` and process in the single method like:
```   
     void DeduceTrigonometryMethodSerializers()
        {
            if (method.Is(__trigonometricMathMethods))
            {
                DeduceReturnsDoubleSerializer();
            }
            else
            {
                DeduceUnknownMethodSerializer();
            }
        }
```

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,3386 @@
...
+            {
+                DeduceUnknownMethodSerializer();
+            }
+
+            static IBsonSerializer GetResultSerializer(Expression expression, Type toType)
+            {
+                var isNullable = toType.IsNullable();
+                var valueType = isNullable ? Nullable.GetUnderlyingType(toType) : toType;
+
+                var valueSerializer = (IBsonSerializer)(Type.GetTypeCode(valueType) switch
```
sanych-sun, line 911: This logic looks very similar to `StandardSerializers.GetSerializer`. Can we leverage it here?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitNewArray.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            switch (node.NodeType)
+            {
+                case ExpressionType.NewArrayBounds:
+                    DeduceNewArrayBoundsSerializers();
+                    break;
+
+                case ExpressionType.NewArrayInit:
+                    DeduceNewArrayInitSerializers();
+                    break;
```
sanych-sun, line 47: Should we throw in `default` case?

sanych-sun (changes requested):
Please also review all TODOs (remove, address or create a ticket for future if we can postpone the changes) and failing unit-test.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitBinary.cs
rstam: I've removed the `CanDeduceSerializer` method entirely becase there was another existing method that could be used.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Probably not until we add support for `DateTimeOffset` in general.

No tests are failing because we didn't do this.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Yes, but I don't know the full list.

I've edited the comment to not mention `DateTime.Year` (since that one is already done)

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Yes this is surprising.

It's because `Dictionary<TKey, TValue>.Keys` and `IDictionary<TKey, TValue>.Keys` have different types. Same for `Values`.

An unexpected glitch in Microsoft's design of these types. (Constrained by having to maintain backward compatibility).

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Moved to `MathMethod.cs` from where it can be shared.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Added to `EnumerableOrQueryableMethod.cs`.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitBinary.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitListInit.cs
rstam: Not sure yet. Added a TODO comment for later investigation.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: I've removed this one line.

I don't understand why you think the outer if is redundant.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Good idea. Done.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Added a TODO for future investigation.

There are 11 types handled here that `StandardSerializers` doesn't know about and a few of the ones it does know about are configured differentlly here, so it's tricky.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Done.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitNewArray.cs
rstam: We could, but I don't think it's necessary.

First of all, there are no other cases here AFAIK.

And if there are, it will result in a missing serializer exception a bit later on.

I don't think we normally throw exceptions when deducing things in the SerializerFinder. We either deduce something or we do nothing.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/IgnoreNodeSerializer.cs
rstam: `IgnoreSerializer` is when we know what the node means but there is no serializer needed for the node. But since we throw an exception if a node has no serializer we assign this dummy serializer to such nodes.

`UnknowableSerializer` is when we know what the node means in general, but not in this particular case. An example would be a method call expression with an unknown method (we know about method call expressions in general but can encounter methods we don't know anything about).

`UnknowableSerializer` comes into play with client-side projections. If we didn't assign the `UnknowableSerializer` dummy serializer to method call nodes we didn't know about we would throw a missing serializer exception which would prevent us from getting to the client-side projection.

sanych-sun: Thank you for the explanation. Should we add this as a short summary on that classes?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: The redundant code is still there:
```
_ = IsItemSerializerKnown(sourceExpression, out var sourceItemSerializer);
```

Regarding the "outer if". as far as I understood each condition inside is exclusive, so we can convert them into series of:
if -> elseif -> elseif ->  DeduceUnknownMethodSerializer. But this is optional and total up to you.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,2548 @@
...
+using MongoDB.Driver.Linq.Linq3Implementation.ExtensionMethods;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
+
+namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
+
+internal partial class SerializerFinderVisitor
+{
+    private static HashSet<MethodInfo>[] __averageOrMedianOrPercentileOverloads =
```
sanych-sun: I do like the fact that you found a way not-to-copy-paste the list of methods. But on other hand it means we always have to iterate the array and check each Hashset if it has the method or not... What if instead of having array of hashsets we simply create a single hashset will all methods via UnionWith method? Also we can create a helper method to merge the hashsets so the final code would looks like:
so instead of
```
private static HashSet<MethodInfo>[] __averageOrMedianOrPercentileOverloads =
[
    EnumerableOrQueryableMethod.AverageOverloads,
    MongoEnumerableMethod.MedianOverloads,
    MongoEnumerableMethod.PercentileOverloads,
    WindowMethod.PercentileOverloads
];
```
we will have something like:
```
private static IReadOnlyHashset<MethodInfo> __averageOrMedianOrPercentileOverloads = MergeHashsets(EnumerableOrQueryableMethod.AverageOverloads, MongoEnumerableMethod.MedianOverloads, MongoEnumerableMethod.PercentileOverloads, WindowMethod.PercentileOverloads);

private static HashSet<MethodInfo> MergeHashSets(params Hashset<MethodInfo>[] list)
{
    var result = new Hashset<MethodInfo>();
    foreach(itm in list)
    {
        result.UnionWith(itm);
    }
    return result;
}
```

Also most likely there will be no needs in the new overload of IsOneOf.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/MethodInfoExtensions.cs
```diff
@@ -110,6 +111,37 @@ public static bool IsInstanceCompareToMethod(this MethodInfo method)
...
+                else
+                {
+                    return comparands.Contains(method);
+                }
+            }
+
+            return false;
+        }
+
+        public static bool IsOneOf(this MethodInfo method, params HashSet<MethodInfo>[] comparands)
```
sanych-sun: If we implement suggestion to create a unions of hashsets instead of arrays of hashsets this `IsOneOf` overload could be removed.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/MathMethod.cs
```diff
@@ -103,6 +106,42 @@ static MathMethod()
...
+            [
+                __log,
+                __log10, // it's convenient to treat Log10 as if it was an overload
+                __logWithNewBase
+            ];
+
+            __trigonometricMethods =
+            [
+                __acos,
+                __acosh,
```
sanych-sun, line 148: `__acosh`, `__asinh` and `__atanh` could be `null`. Is it OK if the hashset contains the `null` entries? I think it would be safer not to let nulls into the collection.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/StringMethod.cs
```diff
@@ -173,13 +196,148 @@ static StringMethod()
...
+                __concatWith3Objects,
+                __concatWith3Strings,
+                __concatWith4Strings,
+                __concatWithObjectArray,
+                __concatWithStringArray
+            ];
+
+            __containsOverloads =
+            [
+                __containsWithChar,
```
sanych-sun, line 249: Is it OK to have posible `null` in the collection?

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/StringMethod.cs
```diff
@@ -173,13 +196,148 @@ static StringMethod()
...
+                __containsWithCharAndComparisonType,
+                __containsWithString,
+                __containsWithStringAndComparisonType
+            ];
+
+            __endsWithOverloads =
+            [
+                __endsWithWithChar,
```
sanych-sun, line 257: Possible `null`.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/StringMethod.cs
```diff
@@ -173,13 +196,148 @@ static StringMethod()
...
+                __splitWithCharsAndCount,
+                __splitWithCharsAndCountAndOptions,
+                __splitWithCharsAndOptions,
+                __splitWithStringsAndCountAndOptions,
+                __splitWithStringsAndOptions
+            ];
+
+            __startsWithOverloads =
+            [
+                __startsWithWithChar,
```
sanych-sun, line 391: Possible `null`.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -0,0 +1,2548 @@
...
+            {
+                DeduceReturnsDateTimeSerializer();
+            }
+            else
+            {
+                DeduceUnknownMethodSerializer();
+            }
+        }
+
+        void DeduceAddDaysMethodSerializers()
```
sanych-sun, line 350: Should we do the same trick for all `DateTime` methods as we did for trigonometrical functions?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewKeyValuePairExpressionToAggregationExpressionTranslator.cs
```diff
@@ -13,13 +13,12 @@
...
 */
 
-using System;
 using System.Collections.Generic;
 using System.Linq.Expressions;
 using MongoDB.Bson;
-using MongoDB.Bson.Serialization;
 using MongoDB.Bson.Serialization.Serializers;
 using MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
```
sanych-sun: This using seems to be redundant.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/ConcatMethodToPipelineTranslator.cs
```diff
@@ -20,6 +20,7 @@
 using MongoDB.Driver.Linq.Linq3Implementation.ExtensionMethods;
 using MongoDB.Driver.Linq.Linq3Implementation.Misc;
 using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
```
sanych-sun: This using seems to be redundant.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslator.cs
```diff
@@ -22,6 +22,7 @@
 using MongoDB.Driver.Linq.Linq3Implementation.ExtensionMethods;
 using MongoDB.Driver.Linq.Linq3Implementation.Misc;
 using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
+using MongoDB.Driver.Linq.Linq3Implementation.Serializers;
```
sanych-sun: This using seems to be redundant.

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/TypeExtensions.cs
```diff
@@ -294,23 +341,21 @@ public static bool IsNullableOf(this Type type, Type valueType)
...
+                TypeCode.Decimal or
+                TypeCode.Double or
+                TypeCode.Int16 or
+                TypeCode.Int32 or
+                TypeCode.Int64 or
+                TypeCode.SByte or
+                TypeCode.Single or
+                TypeCode.UInt16 or
+                TypeCode.UInt32 or
+                TypeCode.UInt64;
```
rstam: The failing tests were because changing from `type == typeof(int)` to `Type.GetTyepCode(type) == TypeCode.In32` was unexpectedly returning true for `enum` as well as for `int`.

Will revert to `type == typeof(int)`.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/MathMethod.cs
rstam: Good point.

In the new factory method you are proposing to create method sets we can filter out nulls.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Great idea.

I will add a new `MethodSet` class and `IReadOnlyMethodSet` interface to encapsulate these ideas.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewKeyValuePairExpressionToAggregationExpressionTranslator.cs
rstam: Removed

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/ConcatMethodToPipelineTranslator.cs
rstam: Removed

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/UnionMethodToPipelineTranslator.cs
rstam: Removed

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
rstam: Removed. Did not change the if structure.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/IgnoreNodeSerializer.cs
sanych-sun: Can we add the information provided above to the classes summary so it will not get lost?

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/MethodInfoSet.cs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public abstract bool Contains(MethodInfo method);
+
+    public abstract IEnumerator<MethodInfo> GetEnumerator();
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+}
+
+internal sealed class ArrayBasedMethodInfoSet : MethodInfoSet
```
sanych-sun: Both `Array` and `HashSet<>` implementes `Collection<T>`. I think we can have a single implementation of the `MethodInfoSet` class, but pass the underlying collection to the constructor. (which could be private and used from static Create method only)

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
```diff
@@ -192,13 +194,29 @@ internal static class EnumerableMethod
...
+        private static readonly IReadOnlyMethodInfoSet __pickWithComputedNOverloads;
+        private static readonly IReadOnlyMethodInfoSet __pickWithNOverloads;
+        private static readonly IReadOnlyMethodInfoSet __pickWithSortByOverloads;
+        private static readonly IReadOnlyMethodInfoSet __reverseOverloads;
+
         // static constructor
         static EnumerableMethod()
         {
+            // initialize methods before sets of methods
+#if NET10_OR_GREATER
```
sanych-sun: We do not have such target yet. Need to discuss if we want to have such upfront optimizations.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        do
+        {
+            visitor.StartPass();
+            visitor.Visit(expression);
+            visitor.EndPass();
+        }
+        while (visitor.IsMakingProgress);
+
+        //#if DEBUG
```
sanych-sun: Should we remove this commented #ifdef?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitListInit.cs
sanych-sun: Still need to be investigated.

sanych-sun (changes requested):
Nothing major, small suggestions. + please review TODOs created in the PR as it was discussed.

src/MongoDB.Driver/Linq/Linq3Implementation/Serializers/IgnoreNodeSerializer.cs
rstam: Done

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4593Tests.cs
```diff
@@ -0,0 +1,144 @@
...
+        public IMongoCollection<Order> Orders { get; private set; }
+        public IMongoCollection<Entity> Entities { get; private set; }
+
+        protected override void InitializeFixture()
+        {
+            Orders = CreateCollection<Order>("orders");
+            Orders.InsertMany(
+            [
+                    new Order { Id = "a", RateBasisHistoryId = "abc" }
+                ]);
```
sanych-sun: _Minor_: please fix the indent.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp2472Tests.cs
```diff
@@ -79,7 +81,7 @@ public class C
         private class MyDTO
         {
             public DateTime timestamp { get; set; }
-            public decimal sqrt_calc { get; set; }
+            [BsonRepresentation(BsonType.Decimal128)] public decimal sqrt_calc { get; set; }
```
sanych-sun: As it was discussed: please revert the file if the attribute is not necessary.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4820Tests.cs
```diff
@@ -0,0 +1,114 @@
...
+    public void Update_Set_with_List_should_work()
+    {
+        var values = new List<string>() { "abc", "def" };
+        var update = Builders<C>.Update.Set(x => x.ReadOnlyCollection, values);
+        var serializerRegistry = BsonSerializer.SerializerRegistry;
+        var documentSerializer = serializerRegistry.GetSerializer<C>();
+
+        var rendered = (BsonDocument)update.Render(new (documentSerializer, serializerRegistry));
+
+        rendered.Should().Be("{ $set : { ReadOnlyCollection : ['[abc]', '[def]'] } }");
```
sanych-sun, line 57: Do we want to have this as an integration test? Should we execute the pipeline and assert the results?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5435Tests.cs
```diff
@@ -0,0 +1,225 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Serializers;
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+{
+    public class CSharp5435Tests : Linq3IntegrationTest
```
sanych-sun: Any particular reasons why this new test class is not inhereted from `LinqIntegrationTest<>`?

sanych-sun (changes requested):
Minor comments

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4593Tests.cs
rstam: Done

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp2472Tests.cs
rstam: Removed

Tests pass without it.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp4820Tests.cs
rstam: I converted all tests in this file to integration tests.

Note that the tests involve update definitions, not pipelines, but the suggestion to make them integration tests is equally valid here.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Jira/CSharp5435Tests.cs
rstam: Converted to use `LinqIntegrationTest<CSharp5435Tests.ClassFixture>`

Also had to use:

```
public sealed class ClassFixture : MongoCollectionFixture<MyDocument, BsonDocument>
```

since the inital data is in the form of `BsonDocuments`.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
rstam: Done a different way. See `IsTupleOrValueTuple` in `TypeExtensions.cs`

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitListInit.cs
rstam: See CSHARP-5853

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/MethodInfoExtensions.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/MethodInfoSet.cs
rstam: Done

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinder.cs
rstam: Removed

PR #1705, opened 2025-06-06
CSHARP-3550: CSOT: Server Selection

sanych-sun:


src/MongoDB.Driver/OperationCancellationContext.cs
```diff
@@ -0,0 +1,122 @@
...
+            _stopwatch = stopwatch;
+            Timeout = timeout;
+            CancellationToken = cancellationToken;
+        }
+
+        public CancellationToken CancellationToken { get; }
+
+        public TimeSpan Elapsed => _stopwatch.Elapsed;
+
+        public TimeSpan Timeout { get; }
```
rstam: I think we need TWO properties:

```
public TimeSpan OperationTimeout { get; }
public TimeSpan StepTimeout { get; }
```
to keep track of BOTH kinds of timeouts.

That probably also means we need two stopwatches. One for the operation as a whole, and the other for whatever sub step we are in (server selection, etc..).

The reason I say this is that when a timeout occurs we should be able to say in the message whether the whole operation timed out or whether a sub step timed out.

src/MongoDB.Driver/OperationCancellationContext.cs
```diff
@@ -0,0 +1,122 @@
...
+        public OperationCancellationContext(TimeSpan timeout, CancellationToken cancellationToken)
+            : this(Stopwatch.StartNew(), timeout, cancellationToken)
+        {
+        }
+
+        internal OperationCancellationContext(Stopwatch stopwatch, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            _stopwatch = stopwatch;
+            Timeout = timeout;
+            CancellationToken = cancellationToken;
```
rstam: I recommend not mixing backing fields and auto-properties in the same class.

src/MongoDB.Driver/OperationCancellationContext.cs
```diff
@@ -0,0 +1,122 @@
...
+            var remainingTimeout = RemainingTimeout;
+            if (remainingTimeout == System.Threading.Timeout.InfiniteTimeSpan)
+            {
+                return false;
+            }
+
+            return remainingTimeout < TimeSpan.Zero;
+        }
+
+        public OperationCancellationContext WithTimeout(TimeSpan timeout)
```
rstam: Maybe should be called `WithStepTimeout`?

src/MongoDB.Driver/OperationCancellationContext.cs
```diff
@@ -0,0 +1,122 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    internal sealed class OperationCancellationContext
```
rstam: `OperationContext`?

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: Interesting idea. What about having single timeout as is and `ParentContext` property instead? Somehow I would like to keep the Context kind of immutable.

rstam: Having a `ParentContext` might be a good idea in its own right.

You still need a way to detect when a timeout occurs whether it was the `OperationTimeout` or the `StepTimeout` that timed out. Does a `ParentContext` help you with that?

rstam (changes requested):

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: Pseudo-code to check which the timeout occurred: 

```
if (operationContext.IsTimedOut())
{
   var isLocalTimeout = operationContext?.ParentContext.IsTimedOut() == false;
}
```

So the logic is: timeout is "local" when there is no parent context or parent context is not timed out yet.

sanych-sun: This code works, because child context is always more strict then parent. It means there should not be a case when ParentContext is timed out, but child is not.

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: Done

src/MongoDB.Driver/OperationCancellationContext.cs
rstam: I don't think that's true. Depends on the configuration and how far into the operation we are.

rstam: For example, during server selection we could timeout either because:

1. the operation timeout timedout
2. the server selection timedout

sanych-sun: Accordingly to CSOT spec, we always should choose more strict timeout: either "local" timeout, or remaining portion of whole timeout.

sanych-sun: Ok. So in server selection scenario we will have 2 operation contexts:
1) operation context that represents whole operation
2) operation context that was produced from the first one and representing server-selection timeout. This operation context will have first one as a ParentContext.

So there could be the following combinations:
a) both are timed out - it means that on server selection context creation remaining timeout was smaller then server selection timeout - so it means whole operation has timed out.
b) both are not timed out - it means we OK to continue.
c) server selection context is timed out, but whole operation is not - it means server selection was timed out because of "local" timeout.
d) whole operation is timed out, but server selection is not - this is impossible state as per spec.

So we really interested in 2 cases, and code above looks like cover both states.

rstam: > Accordingly to CSOT spec, we always should choose more strict timeout: either "local" timeout, or remaining portion of whole timeout.

Agreed. The question is when we timeout, WHICH one is it?

> a) both are timed out 

This is an unlikely race condition. If they are truly equal, I would say attribute the timeout to the operation, not the step (or vice versa if you can think of a reason).

> c) server selection context is timed out, but whole operation is not 

Yes. But how do you KNOW which timeout timed out?

> d) whole operation is timed out, but server selection is not - this is impossible state as per spec.

Why? Doesn't it depend on the configured values? If I configure the operation timeout to be smaller than the server selection, then the whole operation *could* time out, even though server selection did not.

rstam: Sorry... I think when we say "BOTH" we don't mean the same thing. I mean both TIMEOUTS (the operation timeout and the step timeout). I just realized that I think you mean both OperationContexts?

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -47,7 +47,7 @@ private Exception CreateTimeoutException(Stopwatch stopwatch, string message)
                 var checkOutsForOtherCount = checkOutsCount - checkOutsForCursorCount - checkOutsForTransactionCount;
 
                 message =
-                    $"Timed out after {stopwatch.ElapsedMilliseconds}ms waiting for a connection from the connection pool. " +
+                    $"Timed out after {operationContext.Elapsed.TotalMilliseconds}ms waiting for a connection from the connection pool. " +
```
BorisDog: minor: Looks like `Elapsed.TotalMilliseconds` is frequently used, does this award a shortcut property:
`ElapsedMilleseconds`?

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -0,0 +1,122 @@
...
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    internal sealed class OperationContext
+    {
+        // TODO: this static field is temporary here and will be removed in a future PRs in scope of CSOT.
+        public static readonly OperationContext NoTimeout = new(System.Threading.Timeout.InfiniteTimeSpan, CancellationToken.None);
+
+        private readonly Stopwatch _stopwatch;
```
BorisDog: So we are saving the multiple stopwatches creation on operation execution path, it's minor but still nice.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -175,69 +174,62 @@ private sealed class AcquireConnectionHelper : IDisposable
...
+                            pooledConnection = connectionCreator.CreateOpenedOrReuse(operationContext);
                         }
 
-                        return EndCheckingOut(pooledConnection, stopwatch);
+                        return EndCheckingOut(pooledConnection, operationContext);
                     }
 
-                    stopwatch.Stop();
-                    throw CreateException(stopwatch);
+                    throw CreateException(operationContext);
```
BorisDog: Is the stopwatch stopped at some point?
Should it be stopped after first IsTimedOut is called (or StopAndValidateTimeout or similar)? Seems that it's more accurate to report the `elapsed` when first validated and not on creation of the actual report /log later.

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: Yep, I was talking about contexts.

rstam: In a way we are talkinig about the same thing, since a context represents (among other things) a timeout.

> d) whole operation is timed out, but server selection is not - this is impossible state as per spec.

I still think this is possible. I just have to configure the operation timeout to be smaller than the server selection timeout.

sanych-sun: Spec says we should use more strict timeout in such situation:

> If timeoutMS is set, drivers MUST use min(serverSelectionTimeoutMS, remaining timeoutMS), referred to as computedServerSelectionTimeout as the timeout for server selection and connection checkout.

https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#server-selection

rstam: I think we aren't talking about the same thing again.

I'm saying that the operation timeout can be less than the server selection timeout. Of course I can do that.

That's not the same thing as the "computedServerSelectionTimeout".

When the "computedServerSelectionTimeout" (NOT a configured value, but a value that is computed and is different evry time) times out, we probably need to know WHY it timed out. Did it timeout because the operation timed out or because server selection timed out? I think the error message should say why. Seems like the user would want to know which timeout timed out because they may want to know which timeout value to reconfigure if they actually didn't want to time out so quickly.

rstam: > d) whole operation is timed out, but server selection is not - this is impossible state as per spec.

I don't think the spec says this is impossible. The spec describes how to calculate the "computedServerSelectionTimeout", but if the configured operation timeout is less than the configured server selection timeout and we timeout during server selection it is NOT the server selection that timed out, but the WHOLE operation.

sanych-sun: As it was discussed offline:
We both were talking about the same thing, but there were small confusion regarding "configured" timeout and "computed" timeout. We confirmed that we DO need to distinguish what exactly timeout was occurred: was it "local" timeout or whole operation timeout. Also we decided to go with `ParentContext` approach as it seems to cover our needs and also let us have more the 2 level of nesting. To check whether it was local timeout occurred or whole operation - we have to examine the `ParentContext` property and see if it was timed out or not. If there is no `ParentContext` and the context is timed out - it means whole operation was timed out.

tests/MongoDB.Driver.Tests/Core/Clusters/ClusterTests.cs
```diff
@@ -256,11 +227,11 @@ public void SelectServer_should_return_second_server_if_first_cannot_be_found(
 
         [Theory]
         [ParameterAttributeData]
-        public void SelectServer_should_throw_if_no_servers_match(
+        public async Task SelectServer_should_throw_if_no_servers_match(
             [Values(false, true)]
             bool async)
         {
-            var subject = CreateSubject();
+            var subject = CreateSubject(serverSelectionTimeout: TimeSpan.FromMilliseconds(10));
```
sanych-sun, line 234: This adjusted `serverSelectionTimeout` is an optimization of the test: we probably do not have to wait whole 2 seconds here (the default server selection for this test class, see line 49).

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: It's not stopping, because the same context could be in use more. For example if `IsTimedOut` returned `false`, we could decide to wait more. May be we should replace `IsTimedOut` with `EnsureTimeout` method, that will throw in case of timeout and it could save the current elapsed time in the exception together with the timeout source (was it whole operation timeout or server selection timeout, or something else).

tests/MongoDB.Driver.Tests/OperationExecutorTests.cs
```diff
@@ -134,23 +134,13 @@ await Record.ExceptionAsync(() => subject.ExecuteWriteOperationAsync(null, opera
                 .Subject.ParamName.Should().Be("session");
         }
 
-        private OperationExecutor CreateSubject(out Mock<IClusterInternal> clusterMock, out Mock<ICoreSessionHandle> implicitSessionMock)
```
sanych-sun, line 137: Leftovers of the implicit session creation, that was factored out in one of the latest commits related to OperationExecutor refactoing.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Another idea might be to add `out` parameter to the `IsTimedOut` method. Let's discuss our options offline.

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: Fixed.

src/MongoDB.Driver/OperationCancellationContext.cs
sanych-sun: As it was discussed we will not introduce additional properties for now, so will leave the method name as is to match the property name.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: We decided to go with additional `out` parameter approach.

src/MongoDB.Driver/Core/Bindings/ChannelReadWriteBinding.cs
```diff
@@ -56,68 +55,68 @@ public void Dispose()
...
 
-        public Task<IChannelSourceHandle> GetReadChannelSourceAsync(CancellationToken cancellationToken)
+        public Task<IChannelSourceHandle> GetReadChannelSourceAsync(OperationContext operationContext)
         {
             ThrowIfDisposed();
             return Task.FromResult(GetChannelSourceHelper());
         }
 
-        public IChannelSourceHandle GetReadChannelSource(IReadOnlyCollection<ServerDescription> deprioritizedServers, CancellationToken cancellationToken)
+        public IChannelSourceHandle GetReadChannelSource(IReadOnlyCollection<ServerDescription> deprioritizedServers, OperationContext operationContext)
```
rstam: I think `operationContext` should always be the first parameter.

src/MongoDB.Driver/Core/Bindings/CoreSession.cs
```diff
@@ -145,6 +145,8 @@ public bool IsInTransaction
         {
             EnsureAbortTransactionCanBeCalled(nameof(AbortTransaction));
 
+            // TODO: CSOT implement proper way to obtain the operationCancellationContext
+            var operationCancellationContext = new OperationContext(Timeout.InfiniteTimeSpan, cancellationToken);
```
rstam: Variable name should be `operationContext`.

src/MongoDB.Driver/Core/Bindings/CoreSession.cs
```diff
@@ -537,21 +545,21 @@ private void EnsureTransactionsAreSupported()
             }
         }
 
-        private TResult ExecuteEndTransactionOnPrimary<TResult>(IReadOperation<TResult> operation, CancellationToken cancellationToken)
+        private TResult ExecuteEndTransactionOnPrimary<TResult>(IReadOperation<TResult> operation, OperationContext operationContext)
```
rstam: While it is true that `cancellationToken` is always the last parameter, I think that `operationContext` should always be the first parameter.

src/MongoDB.Driver/Core/Bindings/IBinding.cs
```diff
@@ -30,24 +30,24 @@ internal interface IReadBinding : IBinding
...
         ReadPreference ReadPreference { get; }
 
-        IChannelSourceHandle GetReadChannelSource(CancellationToken cancellationToken);
-        Task<IChannelSourceHandle> GetReadChannelSourceAsync(CancellationToken cancellationToken);
+        IChannelSourceHandle GetReadChannelSource(OperationContext operationContext);
+        Task<IChannelSourceHandle> GetReadChannelSourceAsync(OperationContext operationContext);
 
-        IChannelSourceHandle GetReadChannelSource(IReadOnlyCollection<ServerDescription> deprioritizedServers, CancellationToken cancellationToken);
-        Task<IChannelSourceHandle> GetReadChannelSourceAsync(IReadOnlyCollection<ServerDescription> deprioritizedServers, CancellationToken cancellationToken);
+        IChannelSourceHandle GetReadChannelSource(IReadOnlyCollection<ServerDescription> deprioritizedServers, OperationContext operationContext);
```
rstam: Seems like `operationContext` should always be the first parameter.

src/MongoDB.Driver/Core/Clusters/ICluster.cs
```diff
@@ -62,8 +61,8 @@ internal interface IClusterInternal : ICluster
 
         void Initialize();
 
-        IServer SelectServer(IServerSelector selector, CancellationToken cancellationToken);
-        Task<IServer> SelectServerAsync(IServerSelector selector, CancellationToken cancellationToken);
+        IServer SelectServer(IServerSelector selector, OperationContext operationContext);
```
rstam: Seems like `operationContext` should always be the first parameter.

src/MongoDB.Driver/Core/Clusters/Cluster.cs
```diff
@@ -246,27 +255,37 @@ public IServer SelectServer(IServerSelector selector, CancellationToken cancella
...
                         }
 
-                        helper.WaitingForDescriptionToChange();
-                        await WaitForDescriptionChangedAsync(helper.Selector, helper.Description, helper.DescriptionChangedTask, helper.TimeoutRemaining, cancellationToken).ConfigureAwait(false);
+                        await helper.WaitForDescriptionChangedAsync(serverSelectionOperationContext).ConfigureAwait(false);
                     }
                 }
+                catch (TimeoutException)
+                {
+                    var message = BuildTimeoutExceptionMessage(_settings.ServerSelectionTimeout, selector, helper.Description);
```
rstam, line 284: `BuildTimeoutExceptionMessage` doesn't inform the user whether the operation timed out or the server selection timed out.

src/MongoDB.Driver/Core/Clusters/IClusterExtensions.cs
```diff
@@ -29,7 +28,7 @@ public static IServer SelectServerAndPinIfNeeded(
             ICoreSessionHandle session,
             IServerSelector selector,
             IReadOnlyCollection<ServerDescription> deprioritizedServers,
-            CancellationToken cancellationToken)
+            OperationContext operationContext)
```
rstam: Seems like `operationContext` should always be the "first" parameter. In the case of an extension method that would be the "second" parameter, which "looks" like the first parameter when called.

src/MongoDB.Driver/Core/Operations/AggregateOperation.cs
```diff
@@ -270,27 +269,27 @@ public bool? UseCursor
 
         // methods
         /// <inheritdoc/>
-        public IAsyncCursor<TResult> Execute(IReadBinding binding, CancellationToken cancellationToken)
+        public IAsyncCursor<TResult> Execute(IReadBinding binding, OperationContext operationContext)
```
rstam: Seems like the `operationContext` should always be the first parameter.

src/MongoDB.Driver/Core/Bindings/ChannelReadWriteBinding.cs
rstam: Same everywhere. I stopped adding this comment after a while.

src/MongoDB.Driver/Core/Bindings/CoreSession.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Core/Bindings/ChannelReadWriteBinding.cs
sanych-sun: Done. Here and in all other places.

src/MongoDB.Driver/Core/Bindings/CoreSession.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Bindings/IBinding.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Clusters/ICluster.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Clusters/IClusterExtensions.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Operations/AggregateOperation.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/ConnectionPools/MaintenanceHelper.cs
```diff
@@ -113,21 +113,20 @@ private void RunMaintenance(MaintenanceExecutingContext maintenanceExecutingCont
...
                     if (!entered)
                     {
                         return;
                     }
 
-                    using (var connectionCreator = new ConnectionCreator(_connectionPool, minTimeout))
+                    using (var connectionCreator = new ConnectionCreator(_connectionPool))
                     {
-                        var connection = connectionCreator.CreateOpened(cancellationToken);
+                        var connection = connectionCreator.CreateOpened(operationContext);
```
BorisDog: The reason for tiny 20ms timeout is just try to take turn in the connecting queue without blocking checkouts.
This timeout is applied to every step individually: MaxConnections and MaxConnecting. 
Seems that this  code changes this and allocates 20ms to 
* MaxConnections
* MaxConnecting
* Whole Connection creation operation (excluding open)

If this thought is correct, it's really surprising that this was not caught by tests.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -175,69 +174,62 @@ private sealed class AcquireConnectionHelper : IDisposable
...
 
-            public IConnectionHandle AcquireConnection(CancellationToken cancellationToken)
+            public IConnectionHandle AcquireConnection(OperationContext operationContext)
             {
-                var stopwatch = new Stopwatch();
                 try
                 {
-                    StartCheckingOut(stopwatch);
-                    _poolQueueWaitResult = _pool._maxConnectionsQueue.WaitSignaled(_timeout, cancellationToken);
+                    StartCheckingOut();
```
BorisDog: Looks like our previous `ConnectionCheckedOutEvent` duration calculation was a bit different from [spec](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md?plain=1#L581):
It didn't include emitting (and processing) the `CheckoutStarted` event. 

Probably we should mention this nuance in release notes, as the duration will include client events processing code.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -281,50 +273,42 @@ private void AcquireWaitQueueSlot()
...
 
-            private Exception CreateException(Stopwatch stopwatch) =>
+            private Exception CreateException(OperationContext operationContext) =>
                 _poolQueueWaitResult switch
                 {
                     SemaphoreSlimSignalable.SemaphoreWaitResult.Signaled =>
                         MongoConnectionPoolPausedException.ForConnectionPool(_pool._endPoint),
                     SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut =>
-                        _pool.CreateTimeoutException(stopwatch, $"Timed out waiting for a connection after {stopwatch.ElapsedMilliseconds}ms."),
+                        _pool.CreateTimeoutException(operationContext.Elapsed, $"Timed out waiting for a connection after {operationContext.Elapsed.TotalMilliseconds}ms."),
```
BorisDog: Ideally we'd get the timeout closer to WaitSignaled result. But I don't see a very simple solution for this.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -846,41 +830,33 @@ public void UntrackInUseConnection(PooledConnection connection)
...
-                    _connectingWaitStatus = _pool._maxConnectingQueue.Wait(_connectingTimeout, cancellationToken);
-                    stopwatch.Stop();
-
+                    _connectingWaitStatus = _pool._maxConnectingQueue.Wait(operationContext.RemainingTimeout, operationContext.CancellationToken);
                     _pool._poolState.ThrowIfNotReady();
 
                     if (_connectingWaitStatus == SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut)
                     {
-                        _pool.CreateTimeoutException(stopwatch, $"Timed out waiting for in connecting queue after {stopwatch.ElapsedMilliseconds}ms.");
+                        _pool.CreateTimeoutException(operationContext.Elapsed, $"Timed out waiting for in connecting queue after {operationContext.Elapsed.TotalMilliseconds}ms.");
```
BorisDog: This one is not accurate, Elapsed includes more waits than just the maxConnecting queue.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -905,21 +879,19 @@ public PooledConnection CreateOpenedOrReuse(CancellationToken cancellationToken)
...
-                        _connectingWaitStatus = _pool._maxConnectingQueue.WaitSignaled(waitTimeout, cancellationToken);
+                        _connectingWaitStatus = _pool._maxConnectingQueue.WaitSignaled(operationContext.RemainingTimeout, operationContext.CancellationToken);
 
                         connection = _connectingWaitStatus switch
                         {
                             SemaphoreSlimSignalable.SemaphoreWaitResult.Signaled => _pool._connectionHolder.Acquire(),
-                            SemaphoreSlimSignalable.SemaphoreWaitResult.Entered => CreateOpenedInternal(cancellationToken),
-                            SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut => throw CreateTimeoutException(stopwatch),
+                            SemaphoreSlimSignalable.SemaphoreWaitResult.Entered => CreateOpenedInternal(operationContext),
+                            SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut => throw CreateTimeoutException(operationContext.Elapsed),
```
BorisDog: Same issue, the exception is just for `maxConnecting` queue accumulative waits.
Maybe we need to understand what is prescribed by spec here.

tests/MongoDB.Driver.Tests/OperationContextTests.cs
```diff
@@ -0,0 +1,245 @@
...
+            Thread.Sleep(10);
+            stopwatch.Stop();
+
+            var operationContext = new OperationContext(stopwatch, Timeout.InfiniteTimeSpan, CancellationToken.None);
+
+            operationContext.RemainingTimeout.Should().Be(Timeout.InfiniteTimeSpan);
+        }
+
+        [Fact]
+        public void RenainingTimeout_could_be_negative()
```
BorisDog: type: remaining

BorisDog (changes requested):

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            var remainingTimeout = RemainingTimeout;
+            if (timeout == System.Threading.Timeout.InfiniteTimeSpan)
+            {
+                timeout = remainingTimeout;
+            }
+            else if (remainingTimeout != System.Threading.Timeout.InfiniteTimeSpan && remainingTimeout < timeout)
+            {
+                timeout = remainingTimeout;
+            }
```
adelinowona: You could condense this into a single if statement with `or`

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -0,0 +1,155 @@
...
+            get
+            {
+                if (Timeout == System.Threading.Timeout.InfiniteTimeSpan)
+                {
+                    return System.Threading.Timeout.InfiniteTimeSpan;
+                }
+
+                return Timeout - Elapsed;
+            }
+        }
```
adelinowona: nit: alphabetize the properties

src/MongoDB.Driver/OperationOptionsBase.cs
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+using System;
+using System.Threading;
+
+namespace MongoDB.Driver
+{
+    internal abstract record OperationOptionsBase(TimeSpan Timeout)
+    {
+        public OperationContext CreateCancellationContext(CancellationToken cancellationToken)
```
adelinowona: This naming is a bit misleading, this doesn't really create a "cancellation context".  It just creates an OperationContext with the specified cancelationToken. Could you rename to `CreateOperationContext` or even `ToOperationContext`.

adelinowona (changes requested):

src/MongoDB.Driver/OperationOptionsBase.cs
sanych-sun: Good catch. This method was implemented before renaming `OperationCancellationContext` to `OperationContext`.

src/MongoDB.Driver/OperationContext.cs
sanych-sun: Honestly usually I'm trying to avoid mixing `and` and `or` in single statement. As it might be source of confusion (is it: `(a or b) and c` or `a or (b and c)`) and extra brackets might be subject of "Remove unnecessary parentheses" warning.

src/MongoDB.Driver/OperationContext.cs
sanych-sun: Done

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Reverted to use stopwatch here.

src/MongoDB.Driver/OperationContext.cs
sanych-sun: Yep, but as it was discussed had to revert the change =)

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Changes were reverted.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Reverted.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Reverted.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Reverted.

tests/MongoDB.Driver.Tests/OperationContextTests.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -178,58 +178,72 @@ public AcquireConnectionHelper(ExclusiveConnectionPool pool)
...
                 {
-                    StartCheckingOut();
+                    StartCheckingOut(stopwatch);
                     _poolQueueWaitResult = await _pool._maxConnectionsQueue.WaitSignaledAsync(operationContext.RemainingTimeout, operationContext.CancellationToken).ConfigureAwait(false);
 
                     if (_poolQueueWaitResult == SemaphoreSlimSignalable.SemaphoreWaitResult.Entered)
                     {
                         PooledConnection pooledConnection;
-                        EnsureTimeout(operationContext);
+                        if (operationContext.IsTimedOut())
```
BorisDog: I think `EnsureTimeout` can be repurposed to reuse this section of code.

rstam (approved):
LGTM

src/MongoDB.Driver/OperationExecutor.cs
```diff
@@ -50,9 +50,10 @@ public TResult ExecuteReadOperation<TResult>(
             Ensure.IsNotNull(session, nameof(session));
             ThrowIfDisposed();
 
+            var cancellationContext = options.ToOperationContext(cancellationToken);
```
adelinowona: rename this variable (and other instances) to operationContext?

tests/MongoDB.Driver.Tests/Core/Clusters/LoadBalancedClusterTests.cs
```diff
@@ -398,14 +387,10 @@ public void SelectServer_should_be_cancelled_by_cancellationToken(
...
                 {
-                    if (async)
-                    {
-                        exception = Record.Exception(() => subject.SelectServerAsync(Mock.Of<IServerSelector>(), cancellationTokenSource.Token).GetAwaiter().GetResult());
-                    }
-                    else
-                    {
-                        exception = Record.Exception(() => subject.SelectServer(Mock.Of<IServerSelector>(), cancellationTokenSource.Token));
-                    }
+                    var cancellationContext = new OperationContext(Timeout.InfiniteTimeSpan, cancellationTokenSource.Token);
```
adelinowona: rename variable to operationContext?

src/MongoDB.Driver/OperationContext.cs
adelinowona: Ok sounds good to me, I don't feel strongly about this.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -284,19 +290,17 @@ private void AcquireWaitQueueSlot()
...
-
                 stopwatch.Start();
-                _pool._poolState.ThrowIfNotReady();
 
+                _pool._poolState.ThrowIfNotReady();
                 AcquireWaitQueueSlot();
             }
 
-            private IConnectionHandle EndCheckingOut(PooledConnection pooledConnection, Stopwatch stopwatch)
+            private IConnectionHandle EndCheckingOut(Stopwatch stopwatch, PooledConnection pooledConnection)
```
BorisDog: Can revert this change.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -991,54 +969,59 @@ public void Dispose()
...
             {
                 _pool._eventLogger.LogAndPublish(new ConnectionPoolAddingConnectionEvent(_pool._serverId, EventContext.OperationId));
 
-                cancellationToken.ThrowIfCancellationRequested();
-
-                _stopwatch = Stopwatch.StartNew();
+                operationContext.CancellationToken.ThrowIfCancellationRequested();
+                if (operationContext.IsTimedOut())
+                {
+                    throw new TimeoutException();
```
BorisDog: Would a convenience method like
`operationContext.ThrowIfTimedOut()` similar to `ThrowIfCancellationRequested` be useful?
Maybe even `operationContext.ThrowIfTimedOutOrCanceled()` ?

src/MongoDB.Driver/Core/ConnectionPools/MaintenanceHelper.cs
BorisDog: @sanych-sun was this addressed?

tests/MongoDB.Driver.Tests/OperationContextTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                await Record.ExceptionAsync(() => operationContext.WaitTaskAsync(task)) :
+                Record.Exception(() => operationContext.WaitTask(task));
+
+            exception.Should().BeNull();
+        }
+    }
+
+    internal static class OperationContextReflector
+    {
+        private static readonly PropertyInfo __timeoutMemberInfo = typeof(OperationContext).GetProperty("Timeout", BindingFlags.Instance | BindingFlags.NonPublic);
```
BorisDog: We'd should avoid reflection when possible.
In this case I don't why `Timeout` can't be public, it's a property the context is initialized with. Also OperationContext is internal, so no harm in exposing Timeout.

BorisDog (changes requested):

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Then I should pass both `operationContext` and `stopwatch`. It seems like it's easier to leave the code inline.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/ConnectionPools/MaintenanceHelper.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Core/Clusters/IClusterExtensions.cs
```diff
@@ -41,9 +40,9 @@ public static IServer SelectServerAndPinIfNeeded(
                 ? new CompositeServerSelector(new[] { new PriorityServerSelector(deprioritizedServers), selector })
                 : selector;
 
-            // Server selection also updates the cluster type, allowing us to to determine if the server
+            // Server selection also updates the cluster type, allowing us to determine if the server
             // should be pinned.
-            var server = cluster.SelectServer(selector, cancellationToken);
+            var server = cluster.SelectServer(selector, operationContext);
```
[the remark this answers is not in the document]
sanych-sun: For some reasons we do not validate other parameters in the `IClusterExtensions` methods. Will keep it as is for now.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done.

tests/MongoDB.Driver.Tests/OperationContextTests.cs
sanych-sun: Done.

tests/MongoDB.Driver.Tests/Core/Clusters/LoadBalancedClusterTests.cs
sanych-sun: Done.

src/MongoDB.Driver/OperationExecutor.cs
sanych-sun: Done

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -287,6 +279,15 @@ private void AcquireWaitQueueSlot()
                 _enteredWaitQueue = true;
             }
 
+            private void EnsureTimeout(OperationContext operationContext, Stopwatch stopwatch)
```
rstam: Would `ThrowIfTimedOut` be a better name?

Normally we use `EnsureXyz` for argument validation.

src/MongoDB.Driver/Core/ConnectionPools/MaintenanceHelper.cs
BorisDog: What about `_maxConnectingQueue` inside ConnectionCreator ?
We don't want to block on that for whole timeout, just a quick try instead.

adelinowona (approved):
LGTM

tests/MongoDB.Driver.Tests/OperationContextTests.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        public static IEnumerable<object[]> IsTimedOut_test_cases =
+        [
+            [false, Timeout.InfiniteTimeSpan, TimeSpan.FromMilliseconds(5)],
+            [false, TimeSpan.FromMilliseconds(50), TimeSpan.FromMilliseconds(5)],
+            [true, TimeSpan.FromMilliseconds(5), TimeSpan.FromMilliseconds(10)],
+        ];
+
+        [Fact]
+        public void ThrowIfTimedOutOrCanceled_succeed_if_no_timeout_and_no_cancellation()
```
BorisDog: `ThrowIfTimedOutOrCanceled_should_not_throw_....`

tests/MongoDB.Driver.Tests/OperationContextTests.cs
```diff
@@ -0,0 +1,267 @@
...
+            var operationContext = new OperationContext(TimeSpan.FromMilliseconds(10), CancellationToken.None);
+            Thread.Sleep(20);
+
+            var exception = Record.Exception(() => operationContext.ThrowIfTimedOutOrCanceled());
+
+            exception.Should().BeOfType<TimeoutException>();
+        }
+
+        [Fact]
+        public void ThrowIfTimedOutOrCanceled_throws_on_cancellation()
```
BorisDog, line 126: Consider adding `ThrowIfTimedOutOrCanceled_throws_CancelledException_when_timedout_and_cancelled`.
To capture the order of checks.

tests/MongoDB.Driver.Tests/OperationContextTests.cs
```diff
@@ -0,0 +1,267 @@
...
+        [
+            [Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan],
+            [TimeSpan.FromMilliseconds(5), Timeout.InfiniteTimeSpan, TimeSpan.FromMilliseconds(5)],
+            [TimeSpan.FromMilliseconds(5), TimeSpan.FromMilliseconds(5), Timeout.InfiniteTimeSpan],
+            [TimeSpan.FromMilliseconds(5), TimeSpan.FromMilliseconds(5), TimeSpan.FromMilliseconds(10)],
+            [TimeSpan.FromMilliseconds(5), TimeSpan.FromMilliseconds(10), TimeSpan.FromMilliseconds(5)],
+        ];
+
+        [Fact]
+        public void WithTimeout_should_set_ParentContext()
```
BorisDog: Alphabetical order.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Renamed

tests/MongoDB.Driver.Tests/OperationContextTests.cs
sanych-sun: Done

tests/MongoDB.Driver.Tests/OperationContextTests.cs
sanych-sun: Done

tests/MongoDB.Driver.Tests/OperationContextTests.cs
sanych-sun: Done

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -846,41 +838,36 @@ public void UntrackInUseConnection(PooledConnection connection)
...
 
                     if (_connectingWaitStatus == SemaphoreSlimSignalable.SemaphoreWaitResult.TimedOut)
                     {
-                        _pool.CreateTimeoutException(stopwatch, $"Timed out waiting for in connecting queue after {stopwatch.ElapsedMilliseconds}ms.");
+                        _pool.CreateTimeoutException(stopwatch.Elapsed, $"Timed out waiting for in connecting queue after {stopwatch.ElapsedMilliseconds}ms.");
                     }
 
-                    return CreateOpenedInternal(cancellationToken);
+                    var operationContext = new OperationContext(Timeout.InfiniteTimeSpan, cancellationToken);
+                    return CreateOpenedInternal(operationContext);
```
BorisDog: minor: `return CreateOpenedInternal(new(Timeout.InfiniteTimeSpan, cancellationToken));`

BorisDog (approved):
LGTM

PR #1706, opened 2025-06-07
CSHARP-5611: Eliminate the temporary byte array allocation in ObjectId.ToString

ugurpelister:
ObjectId's `ToString` method is among the highest of all highly frequently used methods in this library. The temporary byte array allocation that takes place inside `ToString` is an unnecessary one that adds workload on the GC.

Given that the same functionality can be achieved by a leaner implementation for .NET6+ and .NETStandard2.1, we propose a Span-based implementation for these two platforms so that the resulting string can by created without leaving any garbage. The old implementation can continue to be used on .NET Framework 4.7.2.

### DIFF
--- a/src/MongoDB.Bson/ObjectModel/ObjectId.cs
+++ b/src/MongoDB.Bson/ObjectModel/ObjectId.cs
@@ -459,7 +459,7 @@ namespace MongoDB.Bson
         {
             if (destination.Length < 12)
             {
-                throw new ArgumentException("Not enough room in destination span.", "offset");
+                throw new ArgumentException("Not enough room in destination span.", "destination");
             }
 
             destination[0] = (byte)(_a >> 24);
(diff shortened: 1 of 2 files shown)


damieng:
I think like the previous two PRs for this there this is going to regress on big endian for very real-world gains.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -449,12 +449,86 @@ public void ToByteArray(byte[] destination, int offset)
...
+        /// <summary>
+        /// Converts the ObjectId to a byte buffer provided by the input span.
+        /// </summary>
+        /// <param name="destination">The destination byte span.</param>
+        /// <exception cref="ArgumentException">Not enough room in destination span.</exception>
+        public void ToByteSpan(Span<byte> destination)
+        {
+            if (destination.Length < 12)
+            {
+                throw new ArgumentException("Not enough room in destination span.", "destination");
```
BorisDog, line 462: I know that this is against the existing convention in the file, but for new methods we use should use `nameof(...)`.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -449,12 +449,86 @@ public void ToByteArray(byte[] destination, int offset)
             destination[offset + 11] = (byte)(_c);
         }
 
+#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+        /// <summary>
```
BorisDog, line 453: There is no need for `#if`, we include the `System.Memory` package.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -449,12 +449,86 @@ public void ToByteArray(byte[] destination, int offset)
...
+        /// <param name="destination">The span to fill the characters in.</param>
+        /// <exception cref="ArgumentException">Not enough room in destination span.</exception>
+        public void ToCharSpan(Span<char> destination)
+        {
+            if (destination.Length < 24)
+            {
+                throw new ArgumentException("Not enough room in destination span.", "destination");
+            }
+
+            Span<byte> span = stackalloc byte[12];
```
BorisDog, line 491: I think we should reuse the existing `ToString` implementation, to avoid the Span allocation (even thought it's on stack) and for easier review.

While `ToString()` should just fallback to `ToCharSpan`, as it does already, but for all TFMs.

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -449,12 +449,86 @@ public void ToByteArray(byte[] destination, int offset)
...
+#endif
+
         /// <summary>
         /// Returns a string representation of the value.
         /// </summary>
         /// <returns>A string representation of the value.</returns>
         public override string ToString()
         {
+#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+            return string.Create(24, this, static (span, input) =>
```
BorisDog, line 527: nit: expression-bodied method

src/MongoDB.Bson/ObjectModel/ObjectId.cs
```diff
@@ -449,12 +449,86 @@ public void ToByteArray(byte[] destination, int offset)
             destination[offset + 11] = (byte)(_c);
         }
 
+#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+        /// <summary>
+        /// Converts the ObjectId to a byte buffer provided by the input span.
+        /// </summary>
+        /// <param name="destination">The destination byte span.</param>
+        /// <exception cref="ArgumentException">Not enough room in destination span.</exception>
+        public void ToByteSpan(Span<byte> destination)
```
BorisDog, line 458: For completeness, please add `internal ObjectId(ReadOnlySpan<byte> span)` contstructor
And appropriate tests.

tests/MongoDB.Bson.Tests/ObjectModel/ObjectIdTests.cs
```diff
@@ -101,6 +106,11 @@ public void TestStringConstructor()
             Assert.Equal(BsonConstants.UnixEpoch.AddSeconds(0x01020304), objectId.CreationTime);
             Assert.Equal("0102030405060708090a0b0c", objectId.ToString());
             Assert.True(bytes.SequenceEqual(objectId.ToByteArray()));
+#if NET6_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+            Span<char> span = new char[24];
+            objectId.ToCharSpan(span);
+            Assert.True(span.SequenceEqual("0102030405060708090a0b0c"));
```
BorisDog, line 112: Please add separate tests for the new methods, testing different cases:
`ToByteSpan_should_return_expcted_results(byte[] data, byte[] expectedBytes)`
`ToCharSpan_should_return_expected_results(byte[] data, string expectedChars)` (this one can also test `Tostring`)

BorisDog (changes requested):

BorisDog:
Thanks @ugurpelister 
Addressed in https://github.com/mongodb/mongo-csharp-driver/pull/1960

PR #1707, opened 2025-06-09
CSHARP-5610: Cleanup unnecessary whitespaces in the codebase

adelinowona:
This PR accompanies: https://github.com/mongodb/mongo-csharp-driver/pull/1697 as the first change to be ignored in git blame.


benchmarks/MongoDB.Driver.Benchmarks/BenchmarkResult.cs
```diff
@@ -41,7 +41,7 @@ public BenchmarkResult(BenchmarkReport benchmarkReport)
                 Name = Categories.Contains(DriverBenchmarkCategory.BulkWriteBench)
                     ? benchmarkReport.BenchmarkCase.Descriptor.WorkloadMethod.Name
                     : benchmarkReport.BenchmarkCase.Descriptor.Type.Name;
-                
+
```
adelinowona, line 44: `git blame` has difficulties blaming this type of space changes. Git fundamentally sees this as creating a new empty line without spaces and deleting an old line with spaces rather than simply modifying the existing line. As such, `git blame` will fail to ignore this change.

BorisDog (approved):
LGTM

PR #1708, opened 2025-06-12
CSHARP-3458: Extend IAsyncCursor and IAsyncCursorSource to support IAsyncEnumerable

adelinowona:


BorisDog (approved):
LGTM!

sanych-sun:
As far as I know `IAsyncEnumerator` implements `IAsyncDisposable` interface to perform async clean up. I know that current implementation of `AsyncCursor` do not implements `IAsyncDisposable`, but there is an async method for clean up (see `CloseIfNotAlreadyClosedAsync`). Should we add support of `IAsyncDisposable` to `AsyncCursor` (probably in separate ticket/PR)?

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -72,6 +73,12 @@ public void Dispose()
             }
         }
 
+        public ValueTask DisposeAsync()
+        {
+            Dispose();
+            return new ValueTask(Task.CompletedTask);
```
sanych-sun: It could be simply `default`.

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -102,6 +109,28 @@ public bool MoveNext()
...
+            }
+
+            while (await _cursor.MoveNextAsync(_cancellationToken).ConfigureAwait(false))
+            {
+                _batchEnumerator?.Dispose();
+                _batchEnumerator = _cursor.Current.GetEnumerator();
+                return _batchEnumerator.MoveNext();
+            }
+
+            _batchEnumerator = null;
```
sanych-sun, line 132: Should we dispose `_batchEnumerator` here?

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerableAdapter.cs
```diff
@@ -13,27 +13,37 @@
...
 {
-    internal class AsyncCursorSourceEnumerableAdapter<TDocument> : IEnumerable<TDocument>
+    internal class AsyncCursorSourceEnumerableAdapter<TDocument> : IEnumerable<TDocument>, IAsyncEnumerable<TDocument>
     {
         // private fields
         private readonly CancellationToken _cancellationToken;
         private readonly IAsyncCursorSource<TDocument> _source;
 
         // constructors
+        public AsyncCursorSourceEnumerableAdapter(IAsyncCursorSource<TDocument> source)
```
sanych-sun: I would suggest to call another constructor here, like:
`: this(source, CancellationToken.None)`

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
adelinowona: Yes, great catch! I think the sync version needs to do this as well.

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
adelinowona: done

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerableAdapter.cs
adelinowona: done

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerableOneTimeAdapter.cs
```diff
@@ -21,18 +21,33 @@
...
 namespace MongoDB.Driver.Core.Operations
 {
-    internal sealed class AsyncCursorEnumerableOneTimeAdapter<TDocument> : IEnumerable<TDocument>
+    internal sealed class AsyncCursorEnumerableOneTimeAdapter<TDocument> : IEnumerable<TDocument>, IAsyncEnumerable<TDocument>
     {
         private readonly CancellationToken _cancellationToken;
         private readonly IAsyncCursor<TDocument> _cursor;
         private bool _hasBeenEnumerated;
 
+        public AsyncCursorEnumerableOneTimeAdapter(IAsyncCursor<TDocument> cursor)
```
rstam: I would not have added this constructor.

It makes it too easy for the caller to *forget* to pass in the cancellationToken.

Let the caller provide `CancellationToken.None` if it needs to.

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerableAdapter.cs
rstam: I would not have added this constructor.

It makes it too easy for the caller to *forget* to pass in the cancellationToken.

Let the caller provide `CancellationToken.None` if it needs to.

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -72,6 +73,12 @@ public void Dispose()
             }
         }
 
+        public ValueTask DisposeAsync()
+        {
+            Dispose();
```
rstam, line 79: I would add a comment like:

     Dispose(); // TODO: implement true async disposal

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -72,6 +73,12 @@ public void Dispose()
...
+            return default;
```
rstam: My first reaction was that `default` was a *bug*, that we should be returning` ValueTask.CompletedTask`.

It took me some research to figure out why you were returning `default`.

I suggest the following:

```
#if NET6_0_OR_GREATER
            return ValueTask.CompletedTask;
#else
            return default; // prior to NET6_0 you have to fake ValueTask.CompletedTask using default
#endif
```

It's a little more verbose, but it documents why we sometimes have to use `default` instead of `ValueTask.CompletedTask`, and when we remove support for netstandard2.1 this will get cleaned up.

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerableAdapter.cs
```diff
@@ -13,27 +13,37 @@
...
+
         public AsyncCursorSourceEnumerableAdapter(IAsyncCursorSource<TDocument> source, CancellationToken cancellationToken)
         {
             _source = Ensure.IsNotNull(source, nameof(source));
             _cancellationToken = cancellationToken;
         }
 
+        public IAsyncEnumerator<TDocument> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        {
+            var cursor = _source.ToCursor(cancellationToken);
```
rstam: Calling `_source.ToCursor` is a BLOCKING call that executes the query.

Presumable the caller did not want to execute the query synchronously.

Can we defer query execution until `asyncEnumerator.MoveNextAsync()` is called?

Thoughts?

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
sanych-sun: It's not a bug definitely. This is documented behavior:
> An instance created with the parameterless constructor or by the default(ValueTask) syntax (a zero-initialized structure) represents a synchronously, successfully completed operation.

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-9.0#remarks

rstam: I didn't say it was a bug. I meant it "looked" like a bug.

I'm saying that we should use `ValueTask.CompletedTask` when available.

sanych-sun (approved):
LGTM, but we probably need to create a ticket for proper `IAsyncDisposable` implementation.

rstam (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
sanych-sun: Yep, but it's technically the same.
image
`ValueTask.CompletedTask` is more readable though.

rstam: I didn't say it wasn't technically the same.

I'm saying that we should use ValueTask.CompletedTask when available. It's more readable than `default` (who wants to go hunt down the documentation to figure out what `default` means).

rstam (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
sanych-sun: I cannot agree that ifdef makes code more readable, but I do not insist.

rstam: It might be less readable in the sense that it is longer, but it is definitely more precise.

My request starts from the fact that it is not at all clear (without reading the documentation) what `default` means. I'd rather the meaning be clear WITHOUT having to consult the documentation.

I assumed that `default` was an uncompleted task. That's wrong of course, but it's a reasonable assumption. `default` could mean anything.

It will definitely be more readable when we remove support for older target frameworks and we can get rid of the ifdef altogether and simplly use `ValueTask.CompletedTask`.

rstam (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
adelinowona: I agree that `Value.CompletedTask` is what we want to use here and I see the conditional compilation as a way to document this architectural preference. AFAIK, we don't know when we will drop the older frameworks blocking right us now so we might end up trying to apply this pattern consistently across the codebase which will pollute the driver with #ifndef blocks for minor framework differences. That's my concern here.

As a compromise, I could add a comment specifying that `default` and `ValueTask.CompletedTask` are semantically identical to help clear any initial confusion.

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerableAdapter.cs
adelinowona: That's a fair, I'll revert the change.

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
rstam: I would be OK with a comment documenting why we use `default`. Something like:

```return default; // default is equivalent to ValueTask.CompletedTask which is not available on older target framewokrs```

The only difference is that we might never remember to change it to `ValueTask.CompletedTask`. Removing the `#if` would automatically accomplish that in the future.

My main point is that `default` all by itself is confusing. I assumed it was a bug, because defaults are usually "false" which to me implied "not completed". A comment or an `#if` are two ways of clarifying that.

File: src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs [hunk 8]
```diff
@@ -0,0 +1,94 @@
...
+            if (!_disposed)
+            {
+                _disposed = true;
+
+                if (_cursorEnumerator != null)
+                {
+                    await _cursorEnumerator.DisposeAsync().ConfigureAwait(false);
+                }
+
+                GC.SuppressFinalize(this);
```
sanych-sun: Should we have finalizer? Or why we call GC.SuppressFinalize?

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs
```diff
@@ -0,0 +1,94 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Operations
+{
+#pragma warning disable CA1001
+    internal class AsyncCursorSourceEnumerator<TDocument> : IAsyncEnumerator<TDocument>
```
sanych-sun: sealed?

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Operations
+{
+#pragma warning disable CA1001
```
sanych-sun, line 24: Can we investigate why we need to suppress this warning? Somehow I was thinking implementing `IAsyncDisposable` should be enough for analyzer not to raise the warning.

sanych-sun: It might be because we are using relatively old version of `Microsoft.CodeAnalysis.FxCopAnalyzers`. Which is deprecated on it's own - we might want to check if we can switch to `Microsoft.CodeAnalysis.NetAnalyzers` that is listed as alternative. Probably not in scope of this PR though.

tests/MongoDB.Driver.Tests/Core/IAsyncCursorExtensionsTests.cs
```diff
@@ -201,6 +203,55 @@ public void SingleOrDefault_should_throw_when_cursor_has_wrong_number_of_documen
...
         }
 
+        [Fact]
+        public void ToAsyncEnumerable_result_should_only_be_enumerable_one_time()
+        {
+            var cursor = CreateCursor(2);
+            var enumerable = cursor.ToAsyncEnumerable();
+            enumerable.GetAsyncEnumerator();
+
+            Action action = () => enumerable.GetAsyncEnumerator();
```
sanych-sun: I was thinking we prefer to use `Record.Exception(() => )` instead.

tests/MongoDB.Driver.Tests/Core/IAsyncCursorExtensionsTests.cs
```diff
@@ -201,6 +203,55 @@ public void SingleOrDefault_should_throw_when_cursor_has_wrong_number_of_documen
...
+        }
+
+        [Fact]
+        public async Task ToAsyncEnumerable_should_respect_cancellation_token()
+        {
+            var source = CreateCursor(5);
+            using var cts = new CancellationTokenSource();
+
+            var count = 0;
+            await Assert.ThrowsAsync<OperationCanceledException>(async () =>
```
sanych-sun: Should we use `Record.ExceptionAsync` instead?

tests/MongoDB.Driver.Tests/Core/Operations/AsyncCursorEnumerableOneTimeAdapterTests.cs
```diff
@@ -32,6 +33,37 @@ public void constructor_should_throw_when_cursor_is_null()
...
+        }
+
+        [Fact]
+        public void GetAsyncEnumerator_should_throw_when_called_more_than_once()
+        {
+            var mockCursor = new Mock<IAsyncCursor<BsonDocument>>();
+            var subject = new AsyncCursorEnumerableOneTimeAdapter<BsonDocument>(mockCursor.Object, CancellationToken.None);
+            subject.GetAsyncEnumerator();
+
+            Action action = () => subject.GetAsyncEnumerator();
```
sanych-sun: Use `Record.Exception()` instead.

tests/MongoDB.Driver.Tests/Core/Operations/AsyncCursorEnumeratorTests.cs
```diff
@@ -139,25 +163,61 @@ public void MoveNext_should_return_expected_result_when_there_are_two_batches()
...
+            [Values(false, true)] bool async)
         {
             var subject = CreateSubject(0);
-            subject.Dispose();
+            if (async)
+            {
+                await subject.DisposeAsync();
 
-            Action action = () => subject.MoveNext();
+                Func<Task> action = async () => await subject.MoveNextAsync();
```
sanych-sun: Use `Record.ExceptionAsync` instead.

tests/MongoDB.Driver.Tests/Core/Operations/AsyncCursorSourceEnumeratorTests.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        [Fact]
+        public async Task MoveNextAsync_should_throw_when_disposed()
+        {
+            var source = CreateCursorSource(1);
+            var enumerator = new AsyncCursorSourceEnumerator<BsonDocument>(source, CancellationToken.None);
+
+            await enumerator.DisposeAsync();
+
+            var exception = await Record.ExceptionAsync(async () => await enumerator.MoveNextAsync());
```
sanych-sun, line 132: Minor: no needs of async-await here and more in the file. Could be simplified to:
```
var exception = await Record.ExceptionAsync(() => enumerator.MoveNextAsync());
```

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs
adelinowona: Thanks for the catch, I think it was probably a residue when I first tried to follow the standard dispose pattern

tests/MongoDB.Driver.Tests/Core/Operations/AsyncCursorSourceEnumeratorTests.cs
adelinowona: That simplification requires `MoveNextAsync` to return a `Task` but we can't do that as `MoveNextAsync` has to return a `ValueTask<bool>` in order to implement the `IAsyncEnumerator` interface.

sanych-sun: Makes sense.

tests/MongoDB.Driver.Tests/Core/IAsyncCursorExtensionsTests.cs
adelinowona: I think I subconsciously tried to match the pattern in the current file. However, I'm happy to change to Record.Exception() as it is generally better.

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs
adelinowona: Thanks for the tip! I have added comments around this and will make a ticket to address updating our code analyzers packages.

src/MongoDB.Driver/Core/Operations/AsyncCursorSourceEnumerator.cs
adelinowona: I have made the other classes related here sealed as well.

tests/MongoDB.Driver.Tests/Core/Operations/AsyncCursorEnumeratorTests.cs
```diff
@@ -139,25 +163,60 @@ public void MoveNext_should_return_expected_result_when_there_are_two_batches()
...
+            [Values(false, true)] bool async)
         {
             var subject = CreateSubject(0);
-            subject.Dispose();
+            if (async)
+            {
+                await subject.DisposeAsync();
 
-            Action action = () => subject.MoveNext();
+                (await Record.ExceptionAsync(async () => await subject.MoveNextAsync()))
```
sanych-sun: MInor: I prefer to separate act and assert if possible:
```
var exception = await Record.ExceptionAsync(() => ...);

exception.Should().BeOfType<ObjectDisposedException>()
```

But I do not insist.

sanych-sun (approved):
LGTM + minor comment

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -82,24 +92,46 @@ public bool MoveNext()
                 return true;
             }
 
-            while (true)
+            while (_cursor.MoveNext(_cancellationToken))
```
BorisDog, line 95: Good refactoring

tests/MongoDB.Driver.Tests/Linq/Linq3ImplementationWithLinq2Tests/MongoQueryableTests.cs
```diff
@@ -78,6 +77,21 @@ public async Task AnyAsync_with_predicate()
...
         }
 
+        [Fact]
+        public async Task ToAsyncEnumerable()
+        {
+            var query = CreateQuery().Select(x => x.A);
+            var expectedResults = query.ToList();
+
+            var asyncResults = new List<string>();
+            await foreach (var item in query.ToAsyncEnumerable().ConfigureAwait(false))
```
BorisDog: Now need for `ConfigureAwait` in tests.

BorisDog (approved):
LGTM + minor comment

PR #1710, opened 2025-06-13
CSHARP-5512: Convert LINQ-related tests to use fixtures.

sanych-sun:
Another batch of LINQ-related tests converted to use Fixtures.

### DIFF
--- a/tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
+++ b/tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
@@ -15,6 +15,7 @@
     <NoWarn>
       1701;1702; <!--https://github.com/dotnet/roslyn/issues/19640-->
       CA1724; <!--The type name Model conflicts in whole or in part with the namespace name. Change either name to eliminate the conflict. (https://docs.microsoft.com/visualstudio/code-quality/ca1724-type-names-should-not-match-namespaces)-->
+      CA1714; <!-- Flags enums should have plural names -->
       xUnit1010; <!--The value is not convertible to the method parameter type (https://xunit.github.io/xunit.analyzers/rules/xUnit1010.html)-->
       xUnit1013; <!--Public method should be marked as test (https://xunit.github.io/xunit.analyzers/rules/xUnit1013.html)-->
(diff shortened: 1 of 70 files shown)


tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
```diff
@@ -15,6 +15,7 @@
     <NoWarn>
       1701;1702; <!--https://github.com/dotnet/roslyn/issues/19640-->
       CA1724; <!--The type name Model conflicts in whole or in part with the namespace name. Change either name to eliminate the conflict. (https://docs.microsoft.com/visualstudio/code-quality/ca1724-type-names-should-not-match-namespaces)-->
+      CA1714; <!-- Flags enums should have plural names -->
```
rstam: Are you sure you want to disable this warning project-wide?

rstam (changes requested):
One small question before I LGTM.

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
sanych-sun: We have to either suppress the warning or rename enums as recommended. I do remember we had similar problem with class named "Model" and we decided to suppress the warning instead of renaming (see line above the commented). But I'm OK with renaming if in this case we do not want to suppress the warning.

rstam: I'm fine with suppressing the warnings, but I think it should be done on a case by case basis instead of project wide.

Are there a lot of these? What motivated you to do it project wide instead of an a case by case basis?

rstam (changes requested):

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
sanych-sun: I'll check.

sanych-sun: Found 1 problem so far, suppressed with #pragma

rstam (approved):
LGTM

PR #1712, opened 2025-06-17
CSHARP-5618: CSOT: Race condition in ExclusiveConnectionPool could lead to ArgumentOutOfRangeException

sanych-sun:


src/MongoDB.Driver/OperationContext.cs
```diff
@@ -53,7 +53,13 @@ public TimeSpan RemainingTimeout
                     return System.Threading.Timeout.InfiniteTimeSpan;
                 }
 
-                return Timeout - Stopwatch.Elapsed;
+                var result = Timeout - Stopwatch.Elapsed;
```
sanych-sun, line 54: Many .net APIs that accepts timeouts validates the timeout and throws `ArgumentException` on negative values (for example [SemaphoreSlim](https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.wait?view=net-9.0#system-threading-semaphoreslim-wait(system-timespan-system-threading-cancellationtoken))). On other hand if it's `TimeSpan.Zero` it just considered as timeout out. So it's safer to use `TimeSpan.Zero` for timed out contexts.

tests/MongoDB.Bson.TestHelpers/Threading/ThreadingUtilities.cs
```diff
@@ -29,7 +29,7 @@ public static void ExecuteOnNewThreads(int threadsCount, Action<int> action, int
 
             if (exceptions.Any())
             {
-                throw exceptions.First();
+                throw new AggregateException(exceptions);
```
sanych-sun, line 32: throwing like it was before messes up the call stack, when `AggregateException` can preserve the original stack trace of InnerException.

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -151,6 +162,11 @@ public OperationContext WithTimeout(TimeSpan timeout)
                 timeout = remainingTimeout;
             }
 
+            if (timeout == TimeSpan.Zero)
+            {
+                throw new TimeoutException();
```
BorisDog: Are you sure about throwing here, can't we create timed out contexts?? Feels like throwing should be part of some explicit validation like `ThrowIfTimedOutOrCanceled`.

sanych-sun: Honestly not sure. Probably you are right, as throwing here might be not-expected, but throwing once we decide to Wait on something - will be more expected behavior.
Will revert.

sanych-sun: Reverted.

BorisDog (approved):
LGTM

PR #1713, opened 2025-06-18
CSHARP-5619: Replace IConnection.SendMessages with the method to send a single message

sanych-sun:


src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
```diff
@@ -1033,7 +1027,7 @@ public IByteBuffer EncodeMessages(CancellationToken cancellationToken, out List<
                 return buffer;
             }
 
-            public void EncodingMessages()
+            public void EncodingMessage()
             {
                 _connection._eventLogger.LogAndPublish(new ConnectionSendingMessagesEvent(_connection.ConnectionId, _requestIds.Value, EventContext.OperationId));
```
sanych-sun: _requestIds collection is a single element array. Is the published event format prescribed by any spec? Can we replace requestIds array with the requestId itself?

src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
```diff
@@ -582,22 +582,22 @@ private async Task SendBufferAsync(IByteBuffer buffer, CancellationToken cancell
...
             ThrowIfCancelledOrDisposedOrNotOpen(cancellationToken);
 
-            var helper = new SendMessagesHelper(this, messages, messageEncoderSettings);
+            var helper = new SendMessagesHelper(this, message, messageEncoderSettings);
             try
             {
-                helper.EncodingMessages();
-                using (var uncompressedBuffer = helper.EncodeMessages(cancellationToken, out var sentMessages))
+                helper.EncodingMessage();
+                using (var uncompressedBuffer = helper.EncodeMessage(cancellationToken, out var sentMessages))
```
BorisDog: minor: `sentMessage` (I hoped copilot would catch this). Also in `SendMessageAsync`.

src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
```diff
@@ -983,24 +981,24 @@ private class SendMessagesHelper
...
+            public SendMessagesHelper(BinaryConnection connection, RequestMessage message, MessageEncoderSettings messageEncoderSettings)
             {
                 _connection = connection;
-                _messages = messages.ToList();
+                _message = message;
                 _messageEncoderSettings = messageEncoderSettings;
 
                 _commandStopwatch = Stopwatch.StartNew();
-                _requestIds = new Lazy<List<int>>(() => _messages.Select(m => m.RequestId).ToList());
+                _requestIds = new Lazy<IReadOnlyList<int>>(() => [_message.RequestId]);
```
BorisDog: Maybe just `_requestIds = [_message.RequestId]` ? It's always used in happy path, so no need for Lazy.

src/MongoDB.Driver/Core/WireProtocol/Messages/RequestMessage.cs
```diff
@@ -39,14 +38,12 @@ public static int GetNextRequestId()
 
         // fields
         private readonly int _requestId;
-        private readonly Func<bool> _shouldBeSent;
         private bool _wasSent;
 
         // constructors
-        protected RequestMessage(int requestId, Func<bool> shouldBeSent = null)
```
BorisDog, line 46: Can you explain the removal of `shouldBeSent` please?

sanych-sun: It always either `null` or `() => true` (which is the same). So no way for us today to have `shouldBeSent` to return `false`. And also I do not see any usage for such functionality for now.

src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
sanych-sun: As we discussed offline we will update the events to use single requestId instead of array, this Lazy will be removed as well.

rstam (approved):
LGTM

src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
sanych-sun: Done. Events also work with single RequestId now.

src/MongoDB.Driver/Core/Events/ConnectionSendingMessagesEvent.cs
```diff
@@ -37,10 +38,25 @@ public struct ConnectionSendingMessagesEvent : IEvent
         /// <param name="connectionId">The connection identifier.</param>
         /// <param name="requestIds">The request ids.</param>
         /// <param name="operationId">The operation identifier.</param>
+        [Obsolete("Sending of multiple messages were deleted, use another constructor accepting single requestId instead.")]
```
BorisDog: Sending multiple messages is not supported any longer, use the constructor with single requestId instead.

src/MongoDB.Driver/Core/Events/Diagnostics/TraceSourceEventSubscriber.cs
```diff
@@ -250,17 +250,17 @@ private void Handle(ConnectionReceivingMessageFailedEvent @event)
 
         private void Handle(ConnectionSendingMessagesEvent @event)
         {
-            Debug(TraceSourceEventHelper.ConnectionIdBase + 9, "{0}: sending messages [{1}].", TraceSourceEventHelper.Label(@event.ConnectionId), string.Join(",", @event.RequestIds));
+            Debug(TraceSourceEventHelper.ConnectionIdBase + 9, "{0}: sending message [{1}].", TraceSourceEventHelper.Label(@event.ConnectionId), @event.RequestId);
```
BorisDog: I don't think we should be changing logs now. Might be breaking for any automation users might have.

BorisDog (changes requested):

src/MongoDB.Driver/Core/Events/ConnectionSendingMessagesEvent.cs
sanych-sun: Done

src/MongoDB.Driver/Core/Events/Diagnostics/TraceSourceEventSubscriber.cs
sanych-sun: Reverted

PR #1717, opened 2025-06-25
CSHARP-3537: CSOT: retryable reads and writes

sanych-sun:


src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
```diff
@@ -88,73 +80,82 @@ public RetryableWriteContext(IWriteBinding binding, bool retryRequested)
         public IChannelSourceHandle ChannelSource => _channelSource;
         public bool RetryRequested => _retryRequested;
 
-        public void DisableRetriesIfAnyWriteRequestIsNotRetryable(IEnumerable<WriteRequest> requests)
```
sanych-sun, line 91: As it turned out none of `request.IsRetryable` is using `ConnectionDescription`, so I've decided to remove this method and calculate `_retryRequested` upfront before the context creation.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
```diff
@@ -88,73 +80,82 @@ public RetryableWriteContext(IWriteBinding binding, bool retryRequested)
...
+                    else
+                    {
+                        throw;
+                    }
                 }
             }
         }
 
-        public void Dispose()
+        internal async Task InitializeAsync(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers)
```
sanych-sun: `Initialize` and `InitializeAsync` are internal now, because they have retry logic, that is also useful on operation retry, when we replacing the channel. Might need to be renamed though, because OperationExecutor could call this method several times.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -34,44 +34,48 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
-            }
-
-            try
-            {
-                return operation.ExecuteAttempt(operationContext, context, 2, transactionNumber);
-            }
-            catch (Exception ex) when (ShouldThrowOriginalException(ex))
-            {
-                throw originalException;
+            HashSet<ServerDescription> deprioritizedServers = null;
```
sanych-sun, line 37: Accordingly to spec's pseudo code we should provide only one previous server on retry in the list of deprioritized servers, but as it was discussed in the feature channel - we ought to maintain the list of previously attempted servers, because the original code was created when there was only 1 retry. At least Java and Python maintain the list as well.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -34,44 +34,48 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
+            var attempt = 1;
+            Exception originalException = null;
+
+            long? transactionNumber = AreRetriesAllowed(operation.WriteConcern, context, context.ChannelSource.ServerDescription) ? context.Binding.Session.AdvanceTransactionNumber() : null;
+
+            while (true)
+            {
+                operationContext.ThrowIfTimedOutOrCanceled();
+                var server = context.ChannelSource.ServerDescription;
```
sanych-sun, line 46: Need to preserve the server description BEFORE the operation, because in case of exception server's description might be reset, but we need the original one to determinate if the operation could be retried.

src/MongoDB.Driver/OperationContextExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+using System.Threading;
+
+namespace MongoDB.Driver
+{
+    internal static class OperationContextExtensions
+    {
+        public static bool IsOperationTimeoutConfigured(this OperationContext operationContext)
```
BorisDog: Should this be called `IsRootOperationTimeoutConfigured` ?
nit: Consider
```
IsOperationTimeoutConfigured(this OperationContext operationContext) =>
operationContext.GetRootOperationContext().Timeout != Timeout.InfiniteTimeSpan;
```

Also consider just having `IsRootTimeoutConfigured`/`RootContext` property for simplicity.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -96,50 +90,72 @@ public void Dispose()
...
+                    }
+                }
+            }
+        }
+
+        internal async Task InitializeAsync(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers)
+        {
+            var attempt = 1;
+            while (true)
+            {
```
BorisDog, line 116: Minor optimization, should we consider checking timeout on `operationContext` within the loop as well?

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -96,50 +90,72 @@ public void Dispose()
...
+                try
+                {
+                    ReplaceChannelSource(await Binding.GetReadChannelSourceAsync(operationContext, deprioritizedServers).ConfigureAwait(false));
+                    ReplaceChannel(await ChannelSource.GetChannelAsync(operationContext).ConfigureAwait(false));
+                    return;
+                }
+                catch (Exception ex)
+                {
+                    var innerException = ex is MongoAuthenticationException mongoAuthenticationException ? mongoAuthenticationException.InnerException : ex;
```
BorisDog: Why not to continue extracting the exception in `ShouldRetryOperation`? Just to reuse the code.
Then you also can keep using the `When` pattern, so more compact code.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -14,121 +14,112 @@
...
+                deprioritizedServers.Add(context.ChannelSource.ServerDescription);
 
-        public static async Task<TResult> ExecuteAsync<TResult>(OperationContext operationContext, IRetryableReadOperation<TResult> operation, IReadBinding binding, bool retryRequested)
-        {
-            using (var context = await RetryableReadContext.CreateAsync(operationContext, binding, retryRequested).ConfigureAwait(false))
-            {
-                return await ExecuteAsync(operationContext, operation, context).ConfigureAwait(false);
+                try
+                {
+                    context.Initialize(operationContext, deprioritizedServers);
```
BorisDog: Probably need a better naming. `Initialize` can run only once.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
BorisDog: Yes, it needs a better naming.

tests/MongoDB.Driver.Tests/Core/Operations/RetryableWriteOperationExecutorTests.cs
```diff
@@ -13,17 +13,14 @@
 * limitations under the License.
 */
 
-using System.Linq;
+using System;
+using System.Collections.Generic;
```
BorisDog, line 17: Copyright 2010

tests/MongoDB.Driver.Tests/Core/Operations/WriteCommandOperationTests.cs
```diff
@@ -13,6 +13,7 @@
 * limitations under the License.
 */
 
+using System.Collections.Generic;
```
BorisDog, line 16: Copyright 2010

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Core/Operations/WriteCommandOperationTests.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
sanych-sun: Done

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
sanych-sun: Done

src/MongoDB.Driver/OperationContextExtensions.cs
sanych-sun: Changed `OperationContext` to have `RootContext`. Decided not to change the method name though, as we have a single operation, where the timeout might be configured.

tests/MongoDB.Driver.Tests/Core/Operations/RetryableWriteOperationExecutorTests.cs
sanych-sun: Done

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -96,50 +90,58 @@ public void Dispose()
             }
         }
 
-        public void ReplaceChannel(IChannelHandle channel)
+        internal void AcquireOrReplaceChannel(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers)
```
BorisDog: No need for `internal`?

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -14,121 +14,118 @@
...
-            {
-                return operation.ExecuteAttempt(operationContext, context, attempt: 2, transactionNumber: null);
-            }
-            catch (Exception ex) when (ShouldThrowOriginalException(ex))
-            {
-                throw originalException;
-            }
-        }
+                deprioritizedServers ??= new HashSet<ServerDescription>();
+                deprioritizedServers.Add(context.ChannelSource.ServerDescription);
```
BorisDog: Do we want to be consistent with Write and save the original description
https://github.com/mongodb/mongo-csharp-driver/pull/1717/files#r2167227322
?

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -14,121 +14,118 @@
...
-            {
-                return operation.ExecuteAttempt(operationContext, context, attempt: 1, transactionNumber: null);
-            }
+            HashSet<ServerDescription> deprioritizedServers = null;
+            var attempt = 1;
+            Exception originalException = null;
 
-            Exception originalException;
-            try
+            while (true)
```
BorisDog: How do we prove to ourselves that infinite loop is not possible, because of `ShouldRetryOperation`? Worth adding a comment.

Or maybe something more evident like
```
var maxAtempts = hasTimeout ? int.Max : 2;
while (attempt++ < maxAtempts)
```

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
sanych-sun: It does not really matter here, because deprioritizedServers compared by EndPoint only, but will do that for consistency.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
sanych-sun: Yep, you are right. The class is `internal` already. Made the methods `public`.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
sanych-sun: As discussed: added comment + unit tests for `ShouldRetryOperation` for `RetryableReadOperationExecutor` (`RetryableWriteOperationExecutor` require a little refactoring probably to make `ShouldRetryOperation` easy testable).

src/MongoDB.Driver/OperationContextExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
```
BorisDog: `IsRootContextTimeoutConfigured`?

BorisDog (approved):
LGTM

PR #1718, opened 2025-07-01
CSHARP-5626 Add evergreen script to generate CycloneDX SBOM

jasonhills-mongodb:
## Overview

This PR automates `sbom.json` generation and upload using a two-stage pipeline:

1. **GitHub Actions** (`sbom.yml`): Triggered when NuGet dependency files change on `main`. Regenerates `sbom.json` and opens an automated PR if the content has changed. This stage runs in GitHub Actions because it needs GitHub API access for license resolution.

2. **Evergreen** (`upload-sbom` task/variant): Triggered when `sbom.json` changes. Publishes the SBOM via silkbomb directly to Dependency-Track and Kondukto. This stage runs in Evergreen because it needs access to internal ECR and IAM roles.

### New files
- **`MongoDB.Driver.sbom.slnf`** — Solution filter scoping SBOM generation to the 4 production packages (`MongoDB.Bson`, `MongoDB.Driver`, `MongoDB.Driver.Authentication.AWS`, `MongoDB.Driver.Encryption`). Replaces the per-package loop and `packages.sh`.
- **`.github/workflows/sbom.yml`** — GitHub Actions workflow that regenerates `sbom.json`, validates it with the CycloneDX CLI (SHA256-pinned binary), and opens an automated PR via `peter-evans/create-pull-request` when content changes.
- **`evergreen/upload-sbom.sh`** — Pulls silkbomb from ECR and runs `upload`, publishing `sbom.json` directly to Dependency-Track and Kondukto in one call.

### Modified files
- **`evergreen/generate-sbom.sh`** — Complete rewrite:
  - Upgraded from `cyclonedx-dotnet` 5.3.1 → **6.2.0**, which adds solution filter (`.slnf`) support, eliminating the need for Docker and the per-package loop + merge approach
  - Single `dotnet-CycloneDX` invocation on the `.slnf` instead of 4 separate invocations + Docker merge
  - `--spec-version 1.5` to prevent cyclonedx-dotnet 6.2.0 from defaulting to 1.7
  - `/p:TreatWarningsAsErrors=false` on restore to suppress NU1902/NU1903 vulnerability warnings from `MongoDB.ruleset`
  - Stable `serialNumber` (hardcoded UUID, same across all regenerations)
  - Monotonic `.version` field — only increments when SBOM content changes (compared without `metadata.timestamp` and `.version`)
  - `libmongocrypt` injected as a component after generation: version extracted from `LibMongoCryptVersion` in `MongoDB.Driver.Encryption.csproj` using `dotnet msbuild -getProperty:LibMongoCryptVersion`. Added to `MongoDB.Driver.Encryption`'s `dependsOn` in the `dependencies` graph.
  - Drops Docker dependency entirely for SBOM generation
  - Dev-tool package exclusion (`Microsoft.CodeAnalysis.FxCopAnalyzers`, `Microsoft.NETFramework.ReferenceAssemblies`, `Microsoft.SourceLink.GitHub`, and their orphaned children) is done via a `jq` post-processing step rather than `dotnet-CycloneDX`'s `--exclude-filter` — see "cyclonedx-dotnet bug" below.
- **`evergreen/evergreen.yml`**:
  - Adds `upload-sbom` function, task (`tags: ["ssdlc"]`), and `sbom` build variant with `paths: [sbom.json]` trigger
  - Removes the dead `generate-sbom` function (nothing invoked it — SBOM generation lives entirely in `.github/workflows/sbom.yml`)
  - Removes `generate-ssdlc-reports` task/matrix and its two Evergreen functions — see "RNDSEC-4060" below
  - Fixed permission bit on `evergreen/generate-sbom.sh`: was committed as `100644`, sibling scripts are `100755`

### Deleted files
- **`purls.txt`** — No longer referenced or needed.
- **`evergreen/download-augmented-sbom.sh`**, **`evergreen/fetch-kondukto-token.sh`**, **`evergreen/generate-ssdlc-report.sh`**, **`evergreen/template_ssdlc_compliance_report.md`** — see "RNDSEC-4060" below.

## cyclonedx-dotnet's `--exclude-filter` is broken for our input shape

While regenerating `sbom.json` from scratch, found that `dotnet-CycloneDX`'s `--exclude-filter` — originally used here to drop the 3 build-tool packages above — silently discards **every** component, not just the filtered ones, for any `.sln`/`.slnf`/`.slnx` input (or a single project without `--recursive`). Root cause: in cyclonedx-dotnet 6.2.0, the filter's orphan-removal pass runs before packages are flagged as direct references, so its reachability search starts from an empty set and intersects the whole graph down to nothing. No released version avoids this for a solution-filter input like ours. Filed upstream: https://github.com/CycloneDX/cyclonedx-dotnet/issues (search "exclude-filter" if the link needs updating once filed).

Fix applied here: dropped `--exclude-filter` from the `dotnet-CycloneDX` invocation, replaced with a `jq` step that removes the same packages (plus children orphaned by a related bug, [cyclonedx-dotnet#445](https://github.com/CycloneDX/cyclonedx-dotnet/issues/445)) directly from `sbom.cdx.json`, using a small reachability check so only things exclusively reachable through an excluded package are ever removed — genuine direct dependencies are untouched. Verified with `cyclonedx validate --fail-on-errors` (same check CI runs) and by diffing the dependency graph before/after for dangling references.

`sbom.json` has been regenerated with the fixed script, and also reflects dependency movement from the `main` merge: `AWSSDK.Core`/`AWSSDK.SecurityToken` 3.7.100.14 → 4.0.100.2, `libmongocrypt` 1.18.2 → 1.20.0, and new transitive packages (`System.Text.Encodings.Web`, `System.Text.Json`, `System.ValueTuple`).

## `augment` → `upload`, replicating mongo-go-driver#2154

`evergreen/upload-sbom.sh` originally ran silkbomb's `augment` subcommand, which only enriches an SBOM with Kondukto scan data locally — it doesn't publish anywhere by itself. That left two problems (both flagged in review):
- The `upload-sbom` task never fetched a Kondukto token or passed `--env-file`, unlike `download-augmented-sbom.sh`'s identical `augment` call, which does — real gap, not just a missing comment.
- A "significant changes" diff guard compared `augmented.sbom.json.new` against a baseline (`augmented.sbom.json`) that this task never created or downloaded, so it always took the "no baseline" branch and reported every run as significant — a no-op check.

mongo-go-driver's equivalent (`internal/cmd/upload-sbom/main.go`, from mongodb/mongo-go-driver#2154) uses silkbomb's `upload` subcommand instead, passing only AWS creds for the assumed Silkbomb IAM role — no Kondukto token needed, because `upload` publishes directly to Dependency-Track *and* Kondukto in one call. Replicated that pattern here: switched to `upload`, and removed the diff-guard entirely, since `upload` doesn't produce a local artifact to diff against in the first place (go-driver doesn't do any local diffing either).

## RNDSEC-4060: Remove deprecated SSDLC compliance report automation

Per [RNDSEC-4060](https://jira.mongodb.org/browse/RNDSEC-4060): the SSDLC compliance report (`ssdlc_compliance_report.md`) and the augmented-SBOM artifact it was published alongside (`augmented-sbom.json`, via Kondukto's `augment`) are a vestige of a retired SSDLC program — stale, unread attestations, tracked for removal across many repos. Per the ticket, nothing replaces this; SBOM generation and upload to Kondukto (this PR's `generate-sbom.sh`/`upload-sbom.sh`) continue unchanged, since that's the artifact with real consumers and current data.

Removed: the `generate-ssdlc-reports` task/matrix (`git_tag_only`, ran on every release tag) and the two Evergreen functions it alone used (`download-and-promote-augmented-sbom-to-s3-bucket`, `generate-ssdlc-report`), plus their four backing scripts/templates.

That task also ran `func: trace-artifacts` (Papertrail provenance tracing of the built `.nupkg`/`.snupkg` files) — an unrelated, still-valid supply-chain control that just happened to live in the same task. Moved it into `push-packages-nuget`, right before packages are pushed to NuGet.org, so it keeps running.

## Addressing previous review feedback

- **PURL patching bug** (`startswith("MongoDB.")` + `join(",")` stamping wrong PURLs): Entire approach replaced. The `.slnf` causes the 4 MongoDB packages to be treated as in-solution projects, so they never appear in `components[]` and no PURL patching is needed.
- **`/pwd/` path in `download-augmented-sbom.sh`**: moot — file removed (RNDSEC-4060, above).
- **Environment variable comment block**: Added at top of `generate-sbom.sh`.
- **Echo messages for GitHub license resolution**: Added (`enabled for user: X` / `disabled`).
- **Docker image version/digest concern**: Docker is no longer used for SBOM generation. `upload-sbom.sh` uses the `silkbomb:2.0` tag (consistent with other Evergreen-based silkbomb usage in the driver ecosystem).
- **`--exclude-dev` not covering `PrivateAssets="All"` packages**: `--exclude-dev` does correctly drop the top-level packages (confirmed 2 excluded as dev dependencies in local runs) — what it doesn't clean up is their transitive-only orphaned children ([cyclonedx-dotnet#445](https://github.com/CycloneDX/cyclonedx-dotnet/issues/445)), now handled by the `jq` step above.
- **Dead `generate-sbom` Evergreen function**: confirmed unused (nothing calls `func: generate-sbom`) and deleted, matching mongo-go-driver#2154's config, which has no Evergreen generation task either.
- **Missing Kondukto auth in `upload-sbom.sh`**: resolved by switching to `upload` (above), which needs no separate Kondukto token.
- **Broken "significant changes" diff guard**: resolved by removing it (above) rather than wiring up a real baseline — `upload` doesn't produce a local artifact to diff against.

## SBOM content comparison

The new output vs. the previous per-package + Docker merge approach:

| | Old (5.3.1 + docker merge) | New (6.2.0 + `.slnf`) |
|---|---|---|
| `specVersion` | 1.6 (cyclonedx-cli silently upgraded) | 1.5 ✓ |
| `serialNumber` | Random UUID each run | Stable across regenerations |
| Shape | Hierarchical: 4 MongoDB packages each with nested sub-components | Flat with `dependencies[]` graph |
| NuGet packages captured | 25 (same) | 34 (after `main` merge; excludes 6 build-tool-only packages) |
| `MongoDB.Bson` / `MongoDB.Driver` as components | Yes (via ProjectReference→PURL hack) | No (correctly omitted — source projects) |
| `libmongocrypt` | Not included | Included (1.20.0), linked to `MongoDB.Driver.Encryption` |
| Docker required | Yes | No |
| Version tracking | None | Monotonic `.version`, stable `serialNumber` |

## Testing

- `dotnet-CycloneDX` run locally against the `.slnf` confirmed: 34 NuGet components, `specVersion: 1.5`, all 6 excluded build-tool packages absent, all packages have NuGet PURLs, no dangling `dependencies[].dependsOn` references.
- `cyclonedx validate --input-file sbom.json --fail-on-errors` (same binary/version CI uses) passes.
- `dotnet msbuild -getProperty:LibMongoCryptVersion` extracts `1.20.0` directly — no GitHub API lookup needed.
- Reproduced the `--exclude-filter` bug in isolation against a single leaf project (no project references at all) and against the full `.slnf`, with and without `--recursive`, to confirm no CLI flag combination avoids it for this project's input type.
- `evergreen/evergreen.yml` and `evergreen/upload-sbom.sh` changes validated with `python3 -c "import yaml; yaml.safe_load(...)"` and `bash -n`; confirmed via `grep` that no task still references the removed functions/scripts/roles.


evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,87 @@
...
+  dotnet-CycloneDX src/${package}/${package}.csproj \
+    --disable-package-restore --set-type library --set-nuget-purl --exclude-dev --include-project-references \
+    --set-name ${package} --set-version ${PACKAGE_VERSION} --filename ${SBOM_FILE} \
+    --exclude-filter "${EXCLUDE_FILTER}" \
+    "${github_options[@]}"
+
+  # Patch JSON file with PURLs, as needed
+  for patch in $PURL_PATCHES; do
+    echo "Patching ${patch} with Nuget PURL"
+    contents=$(jq --arg package "$patch" --arg version "$PACKAGE_VERSION" '.components |= map(if [.name | startswith("MongoDB.")] and has("purl") == false then .purl = "pkg:nuget/\($package)@\($version)" else . end)' ${SBOM_FILE})
```
adelinowona: Question: If a package has multiple ProjectReferences (e.g., MongoDB.Driver references both MongoDB.Bson and MongoDB.Driver.Core), won't this logic:
  - Loop 1: Match ALL components starting with "MongoDB." that lack PURLs, assign all of them pkg:nuget/MongoDB.Bson@VERSION
  - Loop 2: All MongoDB components now have PURLs, so nothing gets patched with MongoDB.Driver.Core

  Should the condition be .name == $package instead of .name | startswith("MongoDB.")?

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+set -o errexit # Exit the script with error if any of the commands fail
+
```
adelinowona, line 8:
```suggestion
# Environment Variables:
  #   PACKAGE_VERSION (optional) - Version to use for SBOM generation
  #                                Defaults to output of get-version.sh
  #   GITHUB_USER (optional)     - GitHub username for license resolution
  #   GITHUB_APIKEY (optional)   - GitHub API token for license resolution
```
Add a comment block to describe variables used in the script.

evergreen/download-augmented-sbom.sh
```diff
@@ -21,4 +21,6 @@ echo "KONDUKTO_TOKEN=$kondukto_token" > ${PWD}/kondukto_credentials.env
 docker run --platform="linux/amd64" --rm -v ${PWD}:/pwd \
   --env-file ${PWD}/kondukto_credentials.env \
   artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:2.0 \
-  augment --repo mongodb/mongo-csharp-driver --branch main --sbom-in /pwd/sbom.json --sbom-out /pwd/${SSDLC_PATH}/augmented-sbom.json
\ No newline at end of file
+  augment --repo mongodb/mongo-csharp-driver --branch main --sbom-in /pwd/sbom.json --sbom-out /pwd/${SSDLC_PATH}/augmented-sbom.json
+
+cp /pwd/${SSDLC_PATH}/augmented-sbom.json ./vex.cdx.json
```
adelinowona: I don't think /pwd/  is available in this context since we are outside the docker command. I think it should be ${PWD}

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,87 @@
...
+  # we add the Nuget PURL to the JSON. This command lists the ProjectReferences for processing.
+  PURL_PATCHES=$(dotnet msbuild ./src/${package}/${package}.csproj -getItem:ProjectReference | jq -r '[.Items.ProjectReference[] | .Filename] | join(",")')
+  echo "Project References requiring added PURL: ${PURL_PATCHES}"
+
+  echo "+++++++++++++++++++++++++++++++++++++"
+
+  ## Run cyclonedx-dotnet
+  # Attempt GitHub license resolution only if GITHUB_USER and GITHUB_APIKEY are both set
+  if [[ -v GITHUB_USER && -v GITHUB_APIKEY ]]; then
+    github_options=(--enable-github-licenses --github-username ${GITHUB_USER} --github-token ${GITHUB_APIKEY})
```
adelinowona: Maybe add some echo messages to indicate whether github license resolution is enabled or not

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,87 @@
...
+
+done
+
+echo -e "\n================================="
+echo "Merging SBOMs using cyclonedx-cli"
+echo "================================="
+
+# Use cyclonedx-cli to merge the SBOMs into 1 hierarchical SBOM
+docker run --platform="linux/amd64" --rm -v ${PWD}:/pwd \
+  cyclonedx/cyclonedx-cli:0.28.2 \
```
adelinowona: Not sure if we should be using latest here or not. But if we use a specific version as now, should we include the digest as well for security?

adelinowona (changes requested):

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,87 @@
...
+
+  SBOM_FILE="sbom.${package}.cdx.json"
+  SBOM_FILES="${SBOM_FILES} /pwd/${SBOM_FILE}"
+
+  echo "SBOM file: ${SBOM_FILE}"
+
+  # There are nuances to how cyclonedx-dotnet handles <PackageReference> items in Directory.Build.props that lead to private packages being included in SBOM
+  # results even when PrivateAssets is set to "All". As a safeguard, this command lists the PackageReferences and adds the references with PrivateAssets="All"
+  # to an exclusion filter variable to be fed into cyclonedx-dotnet
+  EXCLUDE_FILTER=$(dotnet msbuild ./src/${package}/${package}.csproj -getItem:PackageReference | jq -r '[.Items.PackageReference[] | select(.PrivateAssets != null) | select(.PrivateAssets | test ("All"; "i")) | .Identity + "@" + .Version] | join(",")')
```
adelinowona: So the option `--exclude-dev` doesn't help with this?

evergreen/generate-sbom.sh
jasonhills-mongodb: You are right, though cyclonedx-cli has removed from the workflow.

evergreen/generate-sbom.sh
jasonhills-mongodb: The `--exclude-dev` only catches components with certain properties defined. The revamped approach improves on this.

evergreen/generate-sbom.sh
jasonhills-mongodb: Good idea. Done.

evergreen/download-augmented-sbom.sh
jasonhills-mongodb: Those options are getting fed into the Python script that lives inside the container.

evergreen/generate-sbom.sh
jasonhills-mongodb: Logic has been revamped

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,136 @@
...
+  "${github_options[@]+"${github_options[@]}"}"
+
+echo -e "\n================================="
+echo "Resolving libmongocrypt version"
+echo "================================="
+
+MONGOCRYPT_COMMIT=$(sed -n 's/.*<LibMongoCryptCommit>\([^<]*\)<\/LibMongoCryptCommit>.*/\1/p' \
+  src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj | head -1)
+
+echo "Looking up tag for libmongocrypt commit ${MONGOCRYPT_COMMIT}"
```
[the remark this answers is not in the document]
jasonhills-mongodb: Lines 67–70 add the guard that exits with a clear error message if `MONGOCRYPT_COMMIT` is empty

evergreen/evergreen.yml
```diff
@@ -222,6 +222,60 @@ functions:
...
+    - command: shell.exec
+      params:
+        working_dir: "mongo-csharp-driver"
+        env:
+          GITHUB_USER: ${github_user}
+          GITHUB_APIKEY: ${github_apikey}
+          PACKAGE_VERSION: ${PACKAGE_VERSION}
+        script: |
+          ${PREPARE_SHELL}
+          ./evergreen/generate-sbom.sh
```
adelinowona, line 235: The generate-sbom function is defined here but I can't find anything that invokes it — no task uses func: generate-sbom, and the sbom variant only runs upload-sbom. Since SBOM generation now lives entirely in the .github/workflows/sbom.yml GitHub Actions workflow, is this Evergreen function leftover from the earlier design? If so, let's delete it; if it's intended for manual/future use, a comment saying so would help.

evergreen/upload-sbom.sh
```diff
@@ -0,0 +1,41 @@
...
+new_json=$(mktemp)
+diff_txt=$(mktemp)
+trap 'rm -f "$old_json" "$new_json" "$diff_txt"' EXIT
+
+if [ -f ./augmented.sbom.json ]; then
+  jq -S 'del(.metadata.timestamp)' ./augmented.sbom.json > "$old_json"
+else
+  echo '{}' > "$old_json"
+fi
+jq -S 'del(.metadata.timestamp)' ./augmented.sbom.json.new > "$new_json"
```
adelinowona, line 35: The "significant changes" check diffs augmented.sbom.json.new against augmented.sbom.json, but nothing in the upload-sbom task ever creates or downloads augmented.sbom.json (the script only writes the .new file, and the task runs func: upload-sbom alone). So old_json always falls into the echo '{}' branch and the diff reports significant changes on every run, making the guard effectively constant-true. Was the intent to download the previously-augmented SBOM as the baseline first (analogous to download-augmented-sbom.sh), or to persist .new back as the baseline after upload?

adelinowona (changes requested):

evergreen/upload-sbom.sh
```diff
@@ -0,0 +1,41 @@
...
+  --branch "${branch_name}"
+  --sbom-in /pwd/sbom.json
+  --sbom-out /pwd/augmented.sbom.json.new
+  --no-update-sbom-version
+)
+
+docker run --rm -v "$(pwd):/pwd" \
+  --user "$(id -u):$(id -g)" \
+  --env 'AWS_ACCESS_KEY_ID' --env 'AWS_SECRET_ACCESS_KEY' --env 'AWS_SESSION_TOKEN' \
+  "${silkbomb}" augment "${silkbomb_augment_flags[@]}"
```
adelinowona, line 23: This runs the same silkbomb `augment` command as our existing `download-augmented-sbom.sh`, but the `upload-sbom` function assumes the silkbomb role and forwards only AWS creds + branch_name + DOCKER_CONFIG — it never runs `fetch-kondukto-token.sh` and never passes` --env-file kondukto_credentials.env / KONDUKTO_TOKEN`.

The download path goes through` kondukto_role_arn → fetch-kondukto-token.sh → --env-file` for the identical `augment` call, which suggests `augment` needs the `Kondukto` token to reach `Kondukto`. Does the silkbomb IAM role somehow cover `Kondukto` auth (e.g. silkbomb self-fetches the token from Secrets Manager), or is the f`etch-kondukto-token.sh + --env-file` step missing here?

evergreen/download-augmented-sbom.sh
adelinowona: Agreed the docker flags are fine — and the `cp` line outside the container is now fixed too, thanks

evergreen/evergreen.yml
jasonhills-mongodb: Removed the dead `generate-sbom` function (SBOM generation lives entirely in `.github/workflows/sbom.yml`)

evergreen/upload-sbom.sh
jasonhills-mongodb: The `augment` command no longer used/needed. Instead using `upload` without a diff.

evergreen/upload-sbom.sh
jasonhills-mongodb: Good callout. When running in Evergreen, `silkbomb` only needs AWS credentials. It retrieves the keys for Kondukto and Dependency Track automatically.

evergreen/evergreen.yml
```diff
@@ -222,72 +222,47 @@ functions:
...
-          ./evergreen/fetch-kondukto-token.sh
+        updates:
+          - key: DOCKER_CONFIG
+            value: "${workdir}/.docker"
     - command: ec2.assume_role
+      display_name: Assume ECR readonly IAM role
       params:
-        role_arn: ${ecr-evergreen-role}
-    - command: shell.exec
+        role_arn: arn:aws:iam::901841024863:role/ecr-role-evergreen-ro
```
adelinowona: why are we using hardcoded ARN role here? can we use `${ecr-evergreen-role}` instead?

evergreen/generate-sbom.sh
```diff
@@ -0,0 +1,154 @@
...
+NEW_CONTENT=$(jq -S 'del(.version, .metadata.timestamp)' sbom.cdx.json)
+OLD_CONTENT=$(jq -S 'del(.version, .metadata.timestamp)' sbom.json 2>/dev/null || echo '{}')
+
+if [ "$NEW_CONTENT" = "$OLD_CONTENT" ]; then
+  NEW_VERSION=$CURRENT_VERSION
+  echo "SBOM content unchanged, keeping version ${NEW_VERSION}"
+else
+  NEW_VERSION=$((CURRENT_VERSION + 1))
+  echo "SBOM content changed, incrementing version to ${NEW_VERSION}"
+fi
```
adelinowona: This always takes the "changed" branch — it's comparing `sbom.cdx.json` (cyclonedx's random serial) to `sbom.json` (our stable one), and `serialNumber` isn't in the `del()`. So `.version` bumps every run, not just on real changes. Add `.serialNumber` to both `del()` calls and it correctly says "unchanged, keeping version 5" when only the serial numbers changed.

Not a real problem in CI since the workflow does its own diff, but would rather have it fixed while we are we still here.

adelinowona (changes requested):

evergreen/generate-sbom.sh
jasonhills-mongodb: Excellent catch. Fixed in dfd126e1ac by adding `.serialNumber` to both `del()` calls. Verified by running the script twice back-to-back: first run correctly bumped the version, second run correctly reported "SBOM content unchanged, keeping version 6" instead of bumping again.

PR #1720, opened 2025-07-07
CSHARP-3538: CSOT: Auth

sanych-sun:
This PR contains changes related to multiple CSOT-related tickets:
- [CSHARP-3538](https://jira.mongodb.org/browse/CSHARP-3538): CSOT: Auth
- [CSHARP-3545](https://jira.mongodb.org/browse/CSHARP-3545): CSOT: MongoDB Handshake
- [CSHARP-3548](https://jira.mongodb.org/browse/CSHARP-3548): CSOT: Server Discovery and Monitoring

### DIFF
--- a/src/MongoDB.Driver/Authentication/SaslAuthenticator.cs
+++ b/src/MongoDB.Driver/Authentication/SaslAuthenticator.cs
@@ -16,7 +16,6 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
-using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization.Serializers;
@@ -72,7 +71,7 @@ namespace MongoDB.Driver.Authentication
 
         public string Name => Mechanism.Name;
 
-        public void Authenticate(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken)
+        public void Authenticate(OperationContext operationContext, IConnection connection, ConnectionDescription description)
         {
             Ensure.IsNotNull(connection, nameof(connection));
             Ensure.IsNotNull(description, nameof(description));
@@ -91,7 +90,10 @@ namespace MongoDB.Driver.Authentication
 
                 while (currentStep != null)
                 {
-                    var executionResult = currentStep.Execute(conversation, result?["payload"]?.AsByteArray, cancellationToken);
+                    operationContext.ThrowIfTimedOutOrCanceled();
+#pragma warning disable CS0618 // Type or member is obsolete
+                    var executionResult = currentStep.Execute(conversation, result?["payload"]?.AsByteArray, operationContext.CombinedCancellationToken);
+#pragma warning restore CS0618 // Type or member is obsolete
                     if (executionResult.BytesToSendToServer == null)
                     {
                         currentStep = executionResult.NextStep;
@@ -109,8 +111,7 @@ namespace MongoDB.Driver.Authentication
                     try
                     {
                         var protocol = CreateCommandProtocol(command);
-                        // TODO: CSOT: implement operationContext support for Auth.
-                        result = protocol.Execute(new OperationContext(Timeout.InfiniteTimeSpan, cancellationToken), connection);
+                        result = protocol.Execute(operationContext, connection);
                         conversationId ??= result?.GetValue("conversationId").AsInt32;
                     }
                     catch (MongoException ex)
@@ -136,7 +137,7 @@ namespace MongoDB.Driver.Authentication
             }
         }
 
-        public async Task AuthenticateAsync(IConnection connection, ConnectionDescription description, CancellationToken cancellationToken)
+        public async Task AuthenticateAsync(OperationContext operationContext, IConnection connection, ConnectionDescription description)
         {
             Ensure.IsNotNull(connection, nameof(connection));
             Ensure.IsNotNull(description, nameof(description));
@@ -155,7 +156,10 @@ namespace MongoDB.Driver.Authentication
 
                 while (currentStep != null)
                 {
-                    var executionResult = await currentStep.ExecuteAsync(conversation, result?["payload"]?.AsByteArray, cancellationToken).ConfigureAwait(false);
+                    operationContext.ThrowIfTimedOutOrCanceled();
+#pragma warning disable CS0618 // Type or member is obsolete
+                    var executionResult = await currentStep.ExecuteAsync(conversation, result?["payload"]?.AsByteArray, operationContext.CombinedCancellationToken).ConfigureAwait(false);
+#pragma warning restore CS0618 // Type or member is obsolete
                     if (executionResult.BytesToSendToServer == null)
                     {
                         currentStep = executionResult.NextStep;
@@ -173,8 +177,7 @@ namespace MongoDB.Driver.Authentication
                     try
                     {
                         var protocol = CreateCommandProtocol(command);
-                        // TODO: CSOT: implement operationContext support for Auth.
-                        result = await protocol.ExecuteAsync(new OperationContext(Timeout.InfiniteTimeSpan, cancellationToken), connection).ConfigureAwait(false);
+                        result = await protocol.ExecuteAsync(operationContext, connection).ConfigureAwait(false);
                         conversationId ??= result?.GetValue("conversationId").AsInt32;
                     }
                     catch (MongoException ex)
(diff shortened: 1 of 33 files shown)


src/MongoDB.Driver/Core/Configuration/ClusterBuilder.cs
```diff
@@ -252,22 +249,23 @@ private IServerMonitorFactory CreateServerMonitorFactory()
...
-                heartbeatSocketTimeout = heartbeatConnectTimeout;
+                heartbeatTimeout = heartbeatConnectTimeout;
             }
             var serverMonitorTcpStreamSettings = new TcpStreamSettings(_tcpStreamSettings)
                 .With(
                     connectTimeout: heartbeatConnectTimeout,
-                    readTimeout: heartbeatSocketTimeout,
-                    writeTimeout: heartbeatSocketTimeout
+                    readTimeout: null,
+                    writeTimeout: null
```
sanych-sun, line 261: Do not have to set the timeouts on the stream level, now we are using OperationContext to enforce the timeouts.

src/MongoDB.Driver/Core/Connections/IConnection.cs
```diff
@@ -32,12 +31,10 @@ internal interface IConnection : IDisposable
         bool IsExpired { get; }
         ConnectionSettings Settings { get; }
 
-        // TODO: CSOT: remove this in scope of MongoDB Handshake
-        void SetReadTimeout(TimeSpan timeout);
```
sanych-sun, line 36: This method was used in a single place, where we were setting the timeout for streaming heartbeat. Now we are using OperationContext instead.

src/MongoDB.Driver/Core/Connections/ConnectionInitializer.cs
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2013-present MongoDB Inc.
+/* Copyright 2010-present MongoDB Inc.
```
adelinowona, line 1: @BorisDog is every file supposed to be dated to 2010? I've seen some other files dated from different years.

src/MongoDB.Driver/Core/Servers/ServerMonitor.cs
```diff
@@ -282,14 +282,21 @@ private HelloResult GetHelloResult(
             CommandWireProtocol<BsonDocument> helloProtocol,
             CancellationToken cancellationToken)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            var timeout = _serverMonitorSettings.HeartbeatTimeout;
+            if (IsUsingStreamingProtocol(connection.Description.HelloResult))
+            {
+                timeout += _serverMonitorSettings.HeartbeatInterval;
+            }
```
adelinowona: This timeout calculation (`HeartbeatTimeout + HeartbeatInterval`) seems ok but the previous timeout set by `SetReadTimeout` in InitializeHelloProtocol was (`ConnectTimeout + HeartbeatInterval`). I doubt `ConnectTimeout` and `HeartbeatTimeout` are the same thing so this means we might be using a different timeout calculation in this current approach. Also, I can't find something similar to `HeartbeatTimeout` in the server monitoring spec so maybe our driver deviated from the spec a bit. We need to decide if it's ok for `HeartbeatTimeout` to replace `ConnectTimeout` in this timeout calculation. 
cc @BorisDog

adelinowona: After looking further, it seems unless explicitly set HeartbeatTimeout is essentially equal to ConnectTimeout so I'm ok with the current approach.

sanych-sun: As it was discussed with @BorisDog , we decided to rollback the logic for now (to use `connectTimeout + HeartbeatInterval`), but new ticket was created to deprecate the `HeartbeatTimeout` and follow the spec more precisely.
https://jira.mongodb.org/browse/CSHARP-5638

src/MongoDB.Driver/Authentication/SaslAuthenticator.cs
```diff
@@ -155,7 +155,9 @@ public async Task AuthenticateAsync(IConnection connection, ConnectionDescriptio
 
                 while (currentStep != null)
                 {
-                    var executionResult = await currentStep.ExecuteAsync(conversation, result?["payload"]?.AsByteArray, cancellationToken).ConfigureAwait(false);
+#pragma warning disable CS0618 // Type or member is obsolete
```
BorisDog, line 160: Do we want to an optimization of manually checking for cancellation at every iteration (in addition to relying on the step cancellation)?

src/MongoDB.Driver/Core/Connections/ConnectionInitializer.cs
BorisDog: Yes, we change that when we touch a file.

src/MongoDB.Driver/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -128,15 +131,17 @@ private void MonitorServer()
...
+                        InitializeConnection(operationContext); // sets _roundTripTimeConnection
                     }
                     else
                     {
                         var helloCommand = HelloHelper.CreateCommand(_serverApi, helloOk, loadBalanced: _roundTripTimeConnection.Settings.LoadBalanced);
                         var helloProtocol = HelloHelper.CreateProtocol(helloCommand, _serverApi);
 
                         var stopwatch = Stopwatch.StartNew();
-                        var helloResult = HelloHelper.GetResult(_roundTripTimeConnection, helloProtocol, _cancellationToken);
+                        using var operationContext = new OperationContext(Timeout.InfiniteTimeSpan, _cancellationToken);
```
BorisDog: The params here don't change, should we create on instance instead and change the loop to
`while (!operationContext.IsTimedOut())` ?

src/MongoDB.Driver/Core/Servers/ServerMonitorSettings.cs
```diff
@@ -14,28 +14,24 @@
...
+            HeartbeatTimeout = heartbeatTimeout;
+            MinHeartbeatInterval = minHeartbeatInterval.WithDefault(TimeSpan.FromMilliseconds(500));
+            ServerMonitoringMode = serverMonitoringMode.WithDefault(ServerMonitoringMode.Auto);
         }
 
-        public TimeSpan ConnectTimeout => _connectTimeout;
-        public TimeSpan HeartbeatInterval => _heartbeatInterval;
-        public TimeSpan MinHeartbeatInterval => _minHeartbeatInterval;
-        public ServerMonitoringMode ServerMonitoringMode => _serverMonitoringMode;
+        public TimeSpan ConnectTimeout { get; }
```
BorisDog: +1
Record with a primary constructor instead of properties would be even more concise.

sanych-sun: Changing to record would probably require to decide what to do with `Optional<T>` parameters of constructor.

src/MongoDB.Driver/Authentication/SaslAuthenticator.cs
sanych-sun: Done

src/MongoDB.Driver/Core/Servers/RoundTripTimeMonitor.cs
sanych-sun: Done.

src/MongoDB.Driver/Core/Servers/ServerMonitorSettings.cs
BorisDog: Nullables?  Up to you.

src/MongoDB.Driver/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -122,21 +125,23 @@ private void MonitorServer()
             _logger?.LogDebug(_serverId, "Monitoring started");
 
             var helloOk = false;
-            while (!_cancellationToken.IsCancellationRequested)
+            using var operationContext = new OperationContext(Timeout.InfiniteTimeSpan, _cancellationToken);
```
BorisDog, line 128: Do we want `OperationContext.Infinite(cancellationToken=default) `helper :) ?

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -632,7 +630,9 @@ private bool ShouldAddTransientTransactionError(MongoException exception)
 
         private void ThrowIfRemainingTimeoutLessThenRoundTripTime(OperationContext operationContext)
         {
-            if (operationContext.RemainingTimeout == Timeout.InfiniteTimeSpan || operationContext.RemainingTimeout > _roundTripTime)
+            if (operationContext.RemainingTimeout == Timeout.InfiniteTimeSpan ||
```
BorisDog, line 633: Do we want `operationContext.IsInfinite()`?

BorisDog (changes requested):
minor comments

src/MongoDB.Driver/Core/Servers/RoundTripTimeMonitor.cs
sanych-sun: Not sure. May be not yet. We definitely need to discuss my operations refactoring prototype, as it might affect our vision on `OperationContext`.

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
sanych-sun: Not sure how it helps. May be we need `IsTimeoutGreaterThen` method, which could encapsulate `TimeSpan` comparison. But so far this method will be used in a single place.

src/MongoDB.Driver/Core/Servers/ServerMonitorSettings.cs
sanych-sun: Done. Let's discuss what we've got here.

PR #1721, opened 2025-07-10
CSHARP-3549: CSOT: Add timeoutMS to settings

sanych-sun:


src/MongoDB.Driver/Core/Misc/Ensure.cs
```diff
@@ -459,12 +460,12 @@ public static string IsNullOrNotEmpty(string value, string paramName)
         /// <returns>The value of the parameter.</returns>
         public static TimeSpan IsValidTimeout(TimeSpan value, string paramName)
         {
-            if (value < TimeSpan.Zero && value != Timeout.InfiniteTimeSpan)
+            if (value > TimeSpan.Zero || value == Timeout.InfiniteTimeSpan)
```
rstam: Why did you flip the logic in this method?

All the other methods end with `return value`.

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -376,13 +376,14 @@ private Type0CommandMessageSection<BsonDocument> CreateType0Section(OperationCon
 
             if (operationContext.IsRootContextTimeoutConfigured() && _roundTripTime > TimeSpan.Zero)
             {
-                var serverTimeout = operationContext.RemainingTimeout - _roundTripTime;
-                if (serverTimeout < TimeSpan.Zero)
+                var serverTimeout = operationContext.RemainingTimeout;
+                if (serverTimeout != Timeout.InfiniteTimeSpan)
                 {
-                    throw new TimeoutException();
```
rstam, line 382: Is it OK that we no longer throw an exception?

src/MongoDB.Driver/ClientSessionExtensions.cs
```diff
@@ -13,15 +13,22 @@
  * limitations under the License.
  */
 
-using System;
-using System.Threading;
+namespace MongoDB.Driver;
 
-namespace MongoDB.Driver
+internal static class ClientSessionExtensions
```
rstam: Should this file/class be called `IClientSessionExtensions`?

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -87,7 +87,7 @@ public CancellationToken CombinedCancellationToken
         }
         private Stopwatch Stopwatch { get; }
 
-        public TimeSpan Timeout { get; }
+        public TimeSpan? Timeout { get; }
```
rstam, line 90: I find it really unhelpful when a class has BOTH fields and auto-properties.

I'm not a fan of auto-properties in general, but I'm specially not a fan of mixing fields and auto-properties in the same class. Mixing them makes it much harder to understand what state a class maintains.

src/MongoDB.Driver/MongoCollectionImpl.cs
```diff
@@ -114,13 +109,13 @@ public override IAsyncCursor<TResult> Aggregate<TResult>(IClientSessionHandle se
             if (isAggregateToCollection)
             {
                 var aggregateOperation = CreateAggregateToCollectionOperation(renderedPipeline, options);
-                ExecuteWriteOperation(session, aggregateOperation, cancellationToken);
+                ExecuteWriteOperation(session, aggregateOperation, options.Timeout, cancellationToken);
```
rstam, line 112: Sometime I see `options.Timeout` and other times I see `options?.Timeout`.

Should they ALL have `?`?

src/MongoDB.Driver/MongoUrlBuilder.cs
```diff
@@ -296,7 +297,7 @@ public TimeSpan HeartbeatTimeout
             get { return _heartbeatTimeout; }
             set
             {
-                if (value < TimeSpan.Zero && value != Timeout.InfiniteTimeSpan)
+                if (value < TimeSpan.Zero && value != System.Threading.Timeout.InfiniteTimeSpan)
```
rstam, line 300: Why this change?

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
sanych-sun: Yes, you are right. We should throw for negative timeouts. Will return the exception.

src/MongoDB.Driver/Core/Misc/Ensure.cs
sanych-sun: Somehow this way it's easier to understand for me. But not critical, will revert to follow other methods pattern.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Let me double-check. I think in some cases there is no-way to have null options.

src/MongoDB.Driver/MongoUrlBuilder.cs
sanych-sun: Now as `MongoUrlBuilder` has property named `Timeout`, we should use full name here.

src/MongoDB.Driver/MongoCollectionImpl.cs
sanych-sun: Yep. There are number of cases when options is guaranteed to be not-null. In all other cases there is Null-conditional operators.

src/MongoDB.Driver/Core/Misc/Ensure.cs
```diff
@@ -444,11 +444,12 @@ public static string IsNullOrNotEmpty(string value, string paramName)
         /// <returns>The value of the parameter.</returns>
         public static TimeSpan? IsNullOrValidTimeout(TimeSpan? value, string paramName)
         {
-            if (value != null)
+            if (value == null)
```
rstam: There was probably no need to touch this file at all.

src/MongoDB.Driver/MongoCollectionImpl.cs
rstam: That's a pain. I leave it to you then to check that there are no missing `?`.

src/MongoDB.Driver/MongoUrlBuilder.cs
rstam: That's unfortunate.

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
```diff
@@ -376,13 +376,18 @@ private Type0CommandMessageSection<BsonDocument> CreateType0Section(OperationCon
...
-                }
+                    serverTimeout -= _roundTripTime;
+                    if (serverTimeout < TimeSpan.Zero)
```
rstam: Should this be changed to `<=`?

rstam (approved):
LGTM with minor comments

src/MongoDB.Driver/Core/Operations/FindOperation.cs
```diff
@@ -254,7 +254,6 @@ public BsonDocument CreateCommand(ConnectionDescription connectionDescription, I
             var effectiveComment = _comment;
             var effectiveHint = _hint;
             var effectiveMax = _max;
-            var effectiveMaxTime = _maxTime;
             var effectiveMin = _min;
             var effectiveReturnKey = _returnKey;
             var effectiveShowRecordId = _showRecordId;
```
adelinowona: What's the point of all this effective variables? can't we just remove them while we are here?

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
adelinowona: +1

src/MongoDB.Driver/Core/Misc/Ensure.cs
adelinowona: I don't mind leaving these changes in. I think they improve this code a bit.

adelinowona (approved):
LGTM! left a comment you can tackle if you think it's ok to do it now.

src/MongoDB.Driver/Core/Misc/Ensure.cs
rstam: Maybe not...

In general I don't like making random changes to files that have nothing to do with the PR.

If we think these methods could be written in a better way we should make a ticket for it.

The reason that all methods end with `return value` is that in general an `Ensure` method might throw more than one exception, so the general pattern is:

* check for invalid condition 1 and throw exception with suitable message
* check for invalid condition 2 and throw exception with suitable message
* ...
* return value

rstam (changes requested):

src/MongoDB.Driver/Core/Misc/Ensure.cs
sanych-sun: Same story here as with previous change in the file: some how "quick returns" works better for me to understand the code. But I do not mind to revert the code.
And technically this method is kind of related to the PR: we used to have unused method `Ensure.IsValidTimeout`. Instead of implementing new method I've decided to adopt it and was making sure the logic is appropriate for the usage. That's the reason why I touched the file.

rstam: OK

sanych-sun: Anyhow: reverted :slightly_smiling_face:

src/MongoDB.Driver/Core/WireProtocol/CommandUsingCommandMessageWireProtocol.cs
sanych-sun: Done, thank you!

src/MongoDB.Driver/Core/Operations/FindOperation.cs
sanych-sun: Done.

adelinowona (approved):

PR #1722, opened 2025-07-10
CSHARP-5633: BsonSerializer.IsTypeDiscriminated is not thread-safe

papafe:
[CSHARP-5633](https://jira.mongodb.org/browse/CSHARP-5633)

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonSerializer.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializer.cs
@@ -39,7 +39,7 @@ namespace MongoDB.Bson.Serialization
         private static Dictionary<Type, IIdGenerator> __idGenerators = new Dictionary<Type, IIdGenerator>();
         private static Dictionary<Type, IDiscriminatorConvention> __discriminatorConventions = new Dictionary<Type, IDiscriminatorConvention>();
         private static Dictionary<BsonValue, HashSet<Type>> __discriminators = new Dictionary<BsonValue, HashSet<Type>>();
-        private static HashSet<Type> __discriminatedTypes = new HashSet<Type>();
+        private static ConcurrentDictionary<Type, bool> __discriminatedTypes = new ();
         private static BsonSerializerRegistry __serializerRegistry;
         private static TypeMappingSerializationProvider __typeMappingSerializationProvider;
         // ConcurrentDictionary<Type, object> is being used as a concurrent set of Type. The values will always be null.
@@ -321,20 +321,7 @@ namespace MongoDB.Bson.Serialization
         /// <returns>True if the type is discriminated.</returns>
         public static bool IsTypeDiscriminated(Type type)
         {
-            if (type.GetTypeInfo().IsInterface)
-            {
-                return true;
-            }
-
-            __configLock.EnterReadLock();
-            try
-            {
-                return __discriminatedTypes.Contains(type);
-            }
-            finally
-            {
-                __configLock.ExitReadLock();
-            }
+            return type.GetTypeInfo().IsInterface || __discriminatedTypes.ContainsKey(type);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -321,8 +321,20 @@ internal static bool IsDiscriminatorConventionRegisteredAtThisLevel(Type type)
...
         public static bool IsTypeDiscriminated(Type type)
         {
-            var typeInfo = type.GetTypeInfo();
-            return typeInfo.IsInterface || __discriminatedTypes.Contains(type);
+            if (type.GetTypeInfo().IsInterface)
+            {
+                return true;
+            }
+
+            __configLock.EnterReadLock();
```
sanych-sun: Could you please double-check if introducing this synchronization affects performance. If so, we could replace `HashSet<Type>` with `ConcurrentDictionary<Type, bool>`.

papafe: I changed it to use a `ConcurrentDictionary`, thanks for taking time to do the benchmarks :)

src/MongoDB.Bson/Serialization/BsonSerializer.cs
```diff
@@ -321,8 +321,7 @@ internal static bool IsDiscriminatorConventionRegisteredAtThisLevel(Type type)
         /// <returns>True if the type is discriminated.</returns>
         public static bool IsTypeDiscriminated(Type type)
         {
-            var typeInfo = type.GetTypeInfo();
-            return typeInfo.IsInterface || __discriminatedTypes.Contains(type);
+            return type.GetTypeInfo().IsInterface || __discriminatedTypes.ContainsKey(type);
```
rstam: `GetTypeInfo` is left over from when we used to target a very old .NET Core framework.

I think this line can be just:

```return type.IsInterface || __discriminatedTypes.ContainsKey(type);```

File: src/MongoDB.Bson/Serialization/BsonSerializer.cs [hunk 3]
```diff
@@ -587,7 +586,7 @@ public static void RegisterDiscriminator(Type type, BsonValue discriminator)
                     // mark all base types as discriminated (so we know that it's worth reading a discriminator)
                     for (var baseType = typeInfo.BaseType; baseType != null; baseType = baseType.GetTypeInfo().BaseType)
                     {
-                        __discriminatedTypes.Add(baseType);
+                        __discriminatedTypes.TryAdd(baseType, true);
```
rstam, line 590: Is it OK that we are ignoring the return value of `TryAdd`?

I understand that we expect it to *always* return true given the check on line 582.

Should we document that we are ignoring the return value on purpose by coding like this:

```_ = __discriminatedTypes.TryAdd(baseType, true);```

rstam (changes requested):
Looks good. Two small comments.

sanych-sun (approved):
LGTM

src/MongoDB.Bson/Serialization/BsonSerializer.cs
papafe: I think it's safe to ignore the return value of `TryAdd` in the same way we ignored `Add` before (the return value has the same semantic). 
Maybe adding a comment would make it clearer, I'm not a super fan of adding the discard operator.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
papafe: Corrected.

src/MongoDB.Bson/Serialization/BsonSerializer.cs
rstam: Interesting... I *am* a fan of the discard variable. I always prefer code to comments to document something.

But it's fine either way.

rstam (approved):
LGTM

PR #1724, opened 2025-07-14
CSHARP-5588: Unable to select KeyValuePair Key when using camelCase convention

sanych-sun:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/NewKeyValuePairExpressionToAggregationExpressionTranslatorTests.cs
```diff
@@ -16,17 +16,27 @@
...
         : base(fixture)
     {
+        // TODO: uncomment the code to demonstrate the fix, revert changes in the test class before merging into the main
+        // _useCamelCase = true;
+
+        if (_useCamelCase)
+        {
+            var defaultPack = new ConventionPack { new CamelCaseElementNameConvention() };
+            ConventionRegistry.Register("camelCase", defaultPack, c => true);
+        }
```
adelinowona: I think we should keep a permanent tests around this to verify that registered conventions are respected instead of just having this for demonstration. You can just remove the convention once the test is done.

Not sure if we want to keep the test here or do what @rstam usually does which is create a test file with the ticket number and test the particular code the user had as repro code.

sanych-sun: I would love to have such tests permanently, but not sure if we can. As far as I know there is no way to have multiple serialization registries for now. If I leave the convention pack registered, it will affect all tests that are run after this particular class.

PR #1730, opened 2025-07-16
CSHARP-5379: Add mixed case tests for read preference tags

papafe:


sanych-sun (approved):
LGTM, but please fix the ticket number. It should be [CSHARP-5379](https://jira.mongodb.org/browse/CSHARP-5379) (both PR's title and commit message)

PR #1731, opened 2025-07-16
CSHARP-734: SOCKS5 Proxy Support

papafe:


src/MongoDB.Driver/Core/Connections/Socks5ProxySettingsBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+        _authentication = Socks5AuthenticationSettings.UsernamePassword(username, password);
+        return this;
+    }
+
+    /// <summary>
+    /// Builds the <see cref="Socks5ProxySettings"/> instance with the specified settings.
+    /// </summary>
+    public Socks5ProxySettings Build()
+    {
+        return new Socks5ProxySettings(_host, _port, _authentication);
```
BorisDog: minor: `public Socks5ProxySettings Build() => new(_host, _port, _authentication);`

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
```diff
@@ -0,0 +1,103 @@
...
+    public Socks5AuthenticationSettings Authentication { get; }
+
+    internal Socks5ProxySettings(string host, int? port, Socks5AuthenticationSettings authentication)
+    {
+        Host = Ensure.IsNotNullOrEmpty(host, nameof(host));
+        Port = port is null ? DefaultPort : Ensure.IsBetween(port.Value, 1, 65535, nameof(port));
+        Authentication = authentication ?? Socks5AuthenticationSettings.None;
+    }
+
+    // Convenience method used internally.
```
BorisDog: No need for comment.

evergreen/evergreen.yml
```diff
@@ -1041,6 +1041,40 @@ functions:
...
+            evergreen/add-ca-certs.sh
+          SSL=${SSL} \
+          MONGODB_URI="${MONGODB_URI}" \
+          TOPOLOGY=${TOPOLOGY} \
+          OS=${OS} \
+          FRAMEWORK=${FRAMEWORK} \
+          TARGET="TestSocks5Proxy" \
+            evergreen/run-tests.sh
+          OS=${OS} \
+            evergreen/cleanup-test-resources.sh
```
sanych-sun: Do not have to call it here. It is being executed as a part of `cleanup` function, called from `post` steps.

sanych-sun: BTW, as far as I understood we are starting some proxy in `setup-socks5-proxy`. Should we stop it on clean up?

evergreen/evergreen.yml
```diff
@@ -1041,6 +1041,40 @@ functions:
...
+    - command: shell.exec
+      params:
+        background: true
+        script: |
+          ${PREPARE_SHELL}
+          MONGODB_URI="${MONGODB_URI}"
+          # Read the MongoDB URI connection string and extract an arbitrary member's host:port
+          HOST_PORT=$(echo "$MONGODB_URI" | sed 's|mongodb://||' | cut -d',' -f1)
+
+          MAP_ARG="localhost:12345 to $HOST_PORT"
```
sanych-sun, line 1082: Hardcoded port number looks suspicious here: `localhost:12345`

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -995,6 +1042,55 @@ private void ParseOption(string name, string value)
...
                     _minPoolSize = ParseInt32(name, value);
                     break;
+                case "proxyhost":
+                    if (!string.IsNullOrEmpty(_proxyHost))
+                    {
+                        throw new MongoConfigurationException("Multiple proxyHost options are not allowed.");
+                    }
+                    
+                    _proxyHost = value;
+                    if (_proxyHost.Length == 0)
```
sanych-sun, line 1060: `string.IsNullOrEmpty(_proxyHost)`?

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -995,6 +1042,55 @@ private void ParseOption(string name, string value)
...
+                    }
+                    break;
+                case "proxyport":
+                    if (_proxyPort != null)
+                    {
+                        throw new MongoConfigurationException("Multiple proxyPort options are not allowed.");
+                    }
+
+                    var proxyPortValue = ParseInt32(name, value);
+                    if (proxyPortValue is < 0 or > 65535)
```
sanych-sun: Is 0 legit value here?

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -995,6 +1042,55 @@ private void ParseOption(string name, string value)
...
+                    _proxyPort = proxyPortValue;
+                    break;
+                case "proxyusername":
+                    if (!string.IsNullOrEmpty(_proxyUsername))
+                    {
+                        throw new MongoConfigurationException("Multiple proxyUsername options are not allowed.");
+                    }
+
+                    _proxyUsername = value;
+                    if (_proxyUsername.Length == 0)
```
sanych-sun, line 1085: `string.IsNullOrEmpty`?

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -995,6 +1042,55 @@ private void ParseOption(string name, string value)
...
+                    }
+                    break;
+                case "proxypassword":
+                    if (!string.IsNullOrEmpty(_proxyPassword))
+                    {
+                        throw new MongoConfigurationException("Multiple proxyPassword options are not allowed.");
+                    }
+
+                    _proxyPassword = value;
+                    if (_proxyPassword.Length == 0)
```
sanych-sun, line 1097: `string.IsNullOrEmpty`?

src/MongoDB.Driver/Core/Configuration/TcpStreamSettings.cs
```diff
@@ -52,7 +55,8 @@ public TcpStreamSettings(
             Optional<int> receiveBufferSize = default(Optional<int>),
             Optional<int> sendBufferSize = default(Optional<int>),
             Optional<Action<Socket>> socketConfigurator = default(Optional<Action<Socket>>),
-            Optional<TimeSpan?> writeTimeout = default(Optional<TimeSpan?>))
+            Optional<TimeSpan?> writeTimeout = default(Optional<TimeSpan?>),
+            Optional<Socks5ProxySettings> socks5ProxySettings = default(Optional<Socks5ProxySettings>))
```
sanych-sun: How come this does not lead to ambiguous method call?

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
```diff
@@ -0,0 +1,140 @@
...
+        Host = Ensure.IsNotNullOrEmpty(host, nameof(host));
+        Port = Ensure.IsBetween(port, 1, 65535, nameof(port));
+        Authentication = Ensure.IsNotNull(authentication, nameof(authentication));
+    }
+
+    // This is a convenience method to create Socks5ProxySettings from the connection string parameters.
+    internal static Socks5ProxySettings Create(string host, int? port, string username, string password)
+    {
+        var authentication =
+            !string.IsNullOrEmpty(username) && !string.IsNullOrEmpty(password)
```
sanych-sun: I suppose we should throw in case one of the parameters is provided, but other i not, instead of silently ignoring wrong parameters configuration.

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
```diff
@@ -0,0 +1,140 @@
...
+        sb.Append("{ Host : ");
+        sb.Append(Host);
+        sb.Append(", Port : ");
+        sb.Append(Port);
+        sb.Append(", Authentication : ");
+
+        sb.Append(Authentication switch
+        {
+            Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings up =>
+                $"UsernamePassword (Username: {up.Username}, Password: {up.Password})",
```
sanych-sun: Are you sure we should let username/password here?

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Socks5ProxySettings"/> class with the specified host, port, and authentication settings.
+    /// </summary>
+    /// <param name="host">The proxy host.</param>
+    /// <param name="port">The proxy port.</param>
+    /// <param name="authentication">The proxy authentication settings.</param>
+    public Socks5ProxySettings(string host, int port, Socks5AuthenticationSettings authentication)
+    {
+        Host = Ensure.IsNotNullOrEmpty(host, nameof(host));
+        Port = Ensure.IsBetween(port, 1, 65535, nameof(port));
```
sanych-sun, line 84: Does 1 allowed port number?

src/MongoDB.Driver/Core/Connections/TcpStreamFactory.cs
```diff
@@ -48,19 +48,36 @@ public TcpStreamFactory(TcpStreamSettings settings)
...
                     Connect(socket, resolved[i], cancellationToken);
-                    return CreateNetworkStream(socket);
+                    var stream = CreateNetworkStream(socket);
+
+                    if (useProxy)
+                    {
+                        Socks5Helper.PerformSocks5Handshake(stream, endPoint, socks5ProxySettings.Authentication, cancellationToken);
+                    }
+
+                    return stream;
```
sanych-sun, line 82: I know we did not clean up in case of exception before, but should we dispose socket and stream if something went wrong during connection establishment?

src/MongoDB.Driver/MongoUrlBuilder.cs
```diff
@@ -438,6 +446,59 @@ public string Password
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException"></exception>
+        public int? ProxyPort
+        {
+            get => _proxyPort;
+            set
+            {
+                if (value is < 0 or > 65535)
```
sanych-sun: Is 0 allowed port number?

sanych-sun: Also why not use Ensure helpers here?

src/MongoDB.Driver/MongoUrlBuilder.cs
```diff
@@ -980,6 +1041,22 @@ public override string ToString()
...
+            {
+                query.AppendFormat("proxyPort={0}&", _proxyPort);
+            }
+            if (!string.IsNullOrEmpty(_proxyUsername))
+            {
+                query.AppendFormat("proxyUsername={0}&", _proxyUsername);
+            }
+            if (!string.IsNullOrEmpty(_proxyPassword))
+            {
+                query.AppendFormat("proxyPassword={0}&", _proxyPassword);
```
sanych-sun, line 1072: Now sure if we ought to print credentials here.

tests/MongoDB.Driver.Tests/Core/Configuration/ConnectionStringTests.cs
```diff
@@ -374,6 +374,10 @@ public void When_nothing_is_specified(string connectionString)
             subject.MaxPoolSize.Should().Be(null);
             subject.MinPoolSize.Should().Be(null);
             subject.Password.Should().BeNull();
+            subject.ProxyHost.Should().Be(null);
```
sanych-sun: `BeNull()`?

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <param name="username">The username</param>
+    /// <param name="password">The password</param>
+    /// <returns></returns>
+    public static Socks5AuthenticationSettings UsernamePassword(string username, string password)
+        => new UsernamePasswordAuthenticationSettings(username, password);
+
+    /// <summary>
+    /// Represents settings for no authentication in SOCKS5.
+    /// </summary>
+    public sealed class NoAuthenticationSettings : Socks5AuthenticationSettings
```
sanych-sun: Can we move this classes outside of abstract `Socks5AuthenticationSettings` class or make them internal?

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
```diff
@@ -0,0 +1,99 @@
...
+            return false;
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return new Hasher()
+                .Hash(Username)
+                .Hash(Password)
+                .GetHashCode();
```
sanych-sun, line 92: This class looks immutable, we can calculate the hashcode once.

src/MongoDB.Driver/Core/Connections/TcpStreamFactory.cs
```diff
@@ -48,19 +48,36 @@ public TcpStreamFactory(TcpStreamSettings settings)
         // methods
         public Stream CreateStream(EndPoint endPoint, CancellationToken cancellationToken)
         {
+            var socks5ProxySettings = _settings.Socks5ProxySettings;
+            var useProxy = socks5ProxySettings != null;
+            var targetEndpoint = useProxy ? new DnsEndPoint(socks5ProxySettings.Host, socks5ProxySettings.Port) : endPoint;
```
sanych-sun: Am I right to think that proxy vs non-proxy scenario is different in 2 points:
1) What endpoint to connect
2) Do some additional logic after connecting the socket

If so - can we instead of tweaking the `TcpStreamFactory` class, create a new wrapper that will replace the endpoint, call the TcpStreamFactory.CreateStream as usual, and then do proxy negotiation after.
We are doing something similar in `SslStreamFactory`.

sanych-sun (changes requested):

evergreen/evergreen.yml
papafe: What this mapping does is saying to the proxy server "if someone says they want to connect to localhost:12345, instead connect them to $HOST_PORT". 
This is coming directly from the prose tests description, and it's used to verify in our tests that we actually use the test proxy server. In fact, some of the tests (the ones with `mappedHost`) use `localhost:12345` as the URI of the mongo server.
Essentially if we didn't have this, the tests could all be passing if we completely disregarded the proxy server and tried to connect directly with mongo.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: It's a valid port number, but usually the port numbers less than 1024 are reserved. I tried to be more relaxed here with the validation, just excluding values that are definitely not valid.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: If we reached here, we already validated the options with a regex, so the strings cannot be null. 
They could be empty though if the connection string looks something like (with empty spaces):
`....proxyHost=  &proxyPort=2020`.
So ideally we should only check that is not empty. We could also still use `isNullOrEmpty` for readability, but it's not necessary here. 

This made me also realise that we should just check for null a couple of lines earlier, so I removed `IsNullOrEmpty` and added a test that verifies it.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: Same as before

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: Same as before

src/MongoDB.Driver/Core/Configuration/TcpStreamSettings.cs
papafe: Because there are two versions of the constructor now:
- one with 7 parameters, no optional
- one with 8 parameters (including the socks5 settings), all optional


When choosing which constructor to use:
- if the method call has less than 7 parameters => uses the second constructor 
- if the method has exactly 7 parameters => uses the first constructors because it has the exact number of parameters
- if the method has 8 parameters => uses the second constructor

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
papafe: I added validation. Hopefully the exception should never be reached though, as this method is only used internally for convenience.

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
papafe: That is a good question... I'm not sure either. I think we have a similar issue in `MongoUrlBuilder`, and in that case we're even printing username and password of the mongo connection.

tests/MongoDB.Driver.Tests/Core/Configuration/ConnectionStringTests.cs
papafe: So... `BeNull` can't be used with `proxyPort` because of the type constraints on the fluent assertions, but I'll change the others.

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
papafe: Same as before. I've decided to uniform all the check to start from 1.

src/MongoDB.Driver/MongoUrlBuilder.cs
papafe: I've added Ensure, but I've also needed to add a check for `null`. Not sure if using a double Ensure for this would make it more readable or not.

src/MongoDB.Driver/MongoUrlBuilder.cs
papafe: Yeah... We have a comment about this also some lines before, as we also print username and password for mongo. I think we can decide to remove it at some point, but it should be done with both.

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
papafe: I've made them internal.

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
papafe: I agree we could, but I suppose this method is method is going be called almost never. Do you think it's worth to cache it vs keeping the code shorter?

src/MongoDB.Driver/Core/Connections/TcpStreamFactory.cs
papafe: That is a good point... I've added it.

evergreen/evergreen.yml
papafe: I've added a cleanup stage taking inspiration from `cleanup-test-resources.sh`. The unix/macos version works fine, but the windows one just kills the python process, it's a little too aggressive. 
I need to modify it to save the PID and then kill directly with those.

evergreen/evergreen.yml
```diff
@@ -1041,6 +1041,40 @@ functions:
...
+            evergreen/add-ca-certs.sh
+          SSL=${SSL} \
+          MONGODB_URI="${MONGODB_URI}" \
+          TOPOLOGY=${TOPOLOGY} \
+          OS=${OS} \
+          FRAMEWORK=${FRAMEWORK} \
+          TARGET="TestSocks5Proxy" \
+            evergreen/run-tests.sh
+          OS=${OS} \
+            evergreen/cleanup-proxy.sh
```
sanych-sun: You probably should add this script to `cleanup-test-resources.sh` or call the `cleanup-proxy.sh` from post steps. Otherwise we could skip on cleaning proxy if some previous step failed with an error.

evergreen/cleanup-proxy.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -o xtrace   # Write all commands first to stderr
+
+# Environment variables used as input:
+#   OS                                               The current operating system
+
+echo "Attempt to kill proxy server process if present on ${OS}"
+if [[ "$OS" =~ Windows|windows ]]; then
+  tasklist -FI "IMAGENAME eq python.exe"
```
sanych-sun: Killing all Python processes might be too much. Let's investigate if there is anything else we can do.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -995,6 +1042,55 @@ private void ParseOption(string name, string value)
...
+                case "proxyport":
+                    if (_proxyPort != null)
+                    {
+                        throw new MongoConfigurationException("Multiple proxyPort options are not allowed.");
+                    }
+
+                    var proxyPortValue = ParseInt32(name, value);
+                    if (proxyPortValue is < 1 or > 65535)
+                    {
+                        throw new MongoConfigurationException("proxyPort must be between 0 and 65535.");
```
sanych-sun: I would probably suggest to change the error message to:
`Invalid proxy port: {proxyPortValue}: must be between 1 and 65535, inclusive`

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
sanych-sun: Makes sense, I would probably prefer comparing to `String.Empty` then, but I'm not insisting on this.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
sanych-sun: 0 is probably wrong value for the port, as it means "random available port" as far as I know.

src/MongoDB.Driver/MongoUrlBuilder.cs
sanych-sun: Ok, in this case it's probably the expected behavior, as ToString documentation says:
`Returns the canonical URL based on the settings in this MongoUrlBuilder`

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
sanych-sun: `MongoUrlBuilder.ToString` is being used to build the Url from provided parameters, so having passwords there is probably the expected behavior (I suppose we ought to have another method for that, something like `BuildUrl`). However this class is user-facing settings class, there is a bigger chances it could be converted to string and logged or even worse outputted as a part of exception. I've checked `MongoClientSettings` - it does not look like it can leak passwords in the similar way.

src/MongoDB.Driver/MongoUrlBuilder.cs
sanych-sun: We have Ensure.IsNullOrBetween if this is what you are looking for.

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
sanych-sun: If this class is not used as a key (or part of the key) in some dictionaries, then probably it's OK to keep the code as is.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
rstam: And I would prefer `if (_proxy == "")` LOL

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+    private const byte AddressTypeIPv6 = 0x04;
+    private const byte AddressTypeDomain = 0x03;
+
+    // Largest possible message size when using username and password auth.
+    private const int BufferSize = 513;
+
+    public static void PerformSocks5Handshake(Stream stream, EndPoint endPoint, Socks5AuthenticationSettings authenticationSettings, CancellationToken cancellationToken)
+    {
+        var (targetHost, targetPort) = endPoint.GetHostAndPort();
+        var buffer = ArrayPool<byte>.Shared.Rent(BufferSize);
```
BorisDog, line 99: We already have similar `ThreadStaticBuffer` that is used for smaller buffers. Do you know how it compares to `ArrayPool.Shared` ?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+            var skip = ProcessConnectResponse(buffer);
+            await stream.ReadBytesAsync(buffer, 0, skip, cancellationToken).ConfigureAwait(true);
+        }
+        finally
+        {
+            ArrayPool<byte>.Shared.Return(buffer);
+        }
+    }
+
+    private static int CreateGreetingRequest(byte[] buffer, bool useAuth)
```
BorisDog, line 170: Minor:
Looks like a static data, would it be simpler to create static buffers once?
```
byte[]  __greetingNoAuth = [ProtocolVersion5, ..],
byte[]  __greetingAuth = [ProtocolVersion5, ..],
```
?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+            AddressTypeDomain => buffer[4] + 2,
+            _ => throw new IOException("Unknown address type in SOCKS5 reply.")
+        };
+    }
+
+    private static void VerifyProtocolVersion(byte version)
+    {
+        if (version != ProtocolVersion5)
+        {
+            throw new IOException("Invalid SOCKS version in response.");
```
BorisDog: Add actual and expected versions?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.GridFS;
+
+namespace MongoDB.Driver.Core.Connections;
+
+internal static class Socks5Helper
```
BorisDog, line 29: Nice work here.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+            {
+                throw new IOException("SOCKS5 proxy requires authentication, but no credentials were provided.");
+            }
+
+            return true;
+        }
+
+        if (acceptedMethod != MethodNoAuth)
+        {
+            throw new IOException("SOCKS5 proxy requires unsupported authentication method.");
```
BorisDog: Add the `acceptedMethod` to the exception?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            var useAuth = authenticationSettings is Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings;
+
+            var greetingRequestLength = CreateGreetingRequest(buffer, useAuth);
+            await stream.WriteAsync(buffer, 0, greetingRequestLength, cancellationToken).ConfigureAwait(false);
+
+            await stream.ReadBytesAsync(buffer, 0, 2, cancellationToken).ConfigureAwait(false);
+            var acceptsUsernamePasswordAuth = ProcessGreetingResponse(buffer, useAuth);
+
+            // If we have username and password, but the proxy doesn't need them, we skip.
```
BorisDog: minor: Should comment say "we skip the authentication step" ?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    // Reads the SOCKS5 connect response and returns the number of bytes to skip in the buffer.
+    private static int ProcessConnectResponse(byte[] buffer)
+    {
+        VerifyProtocolVersion(buffer[0]);
+
+        if (buffer[1] != Socks5Success)
+        {
+            throw new IOException($"SOCKS5 connect failed");
```
BorisDog: Add `buffer[1]` value?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+        buffer[2 + usernameLength] = passwordLength;
+
+        return 3 + usernameLength + passwordLength;
+    }
+
+    private static void ProcessAuthenticationResponse(byte[] buffer)
+    {
+        if (buffer[0] != SubnegotiationVersion || buffer[1] != Socks5Success)
+        {
+            throw new IOException("SOCKS5 authentication failed.");
```
BorisDog: Add `buffer[1]` value?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+            var useAuth = authenticationSettings is Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings;
+
+            var greetingRequestLength = CreateGreetingRequest(buffer, useAuth);
+            await stream.WriteAsync(buffer, 0, greetingRequestLength, cancellationToken).ConfigureAwait(false);
+
+            await stream.ReadBytesAsync(buffer, 0, 2, cancellationToken).ConfigureAwait(false);
+            var acceptsUsernamePasswordAuth = ProcessGreetingResponse(buffer, useAuth);
+
+            // If we have username and password, but the proxy doesn't need them, we skip.
+            if (useAuth && acceptsUsernamePasswordAuth)
```
BorisDog: Looks like in the case of `acceptsUsernamePasswordAuth`, `useAuth` is validated in `ProcessGreetingResponse`.
So technically `useAuth` doesn't need to be checked here.
Should `acceptsUsernamePasswordAuth` be named something like `useAuthenticationStep` ?

tests/MongoDB.Driver.Tests/Specifications/socks5-support/Socks5SupportProseTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            foreach (var useTls in new[] { true, false })
+            {
+                foreach (var isAsync in new[] { true, false })
+                {
+                    var id = $"{index++}_{(useTls ? "Tls" : "NoTls")}_{(isAsync ? "Async" : "Sync")}";
+                    yield return [id, connectionString, expectedResult, useTls, isAsync];
+                }
+            }
+        }
```
BorisDog: minor: another way to write the same, if you'd prefer:
```
return  from (connectionString, expectedResult) in testCases
      from useTls in new[] { true, false }
      from isAsync in new[] { true, false }
      select new object[] {..., useTls, isAsync};
```

BorisDog (changes requested):
Looks great overall!

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: I think you're right :)

src/MongoDB.Driver/MongoUrlBuilder.cs
papafe: Ah yes, I missed that! Thank you

src/MongoDB.Driver/Core/Connections/Socks5ProxySettings.cs
papafe: Mmh, makes sense, I've removed it.

src/MongoDB.Driver/Core/Connections/Socks5AuthenticationSettings.cs
papafe: Yes, it should not be used in that way.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: Reasonable.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: So, I didn't see `ThreadStaticBuffer` before, so it's good that you pointed it 😁 
Regarding the use of it, I suppose the main disadvantage is here is that `ThreadStaticBuffer` is not thread safe, so we can't use it in the async version. Another advantage of `ArrayPool.Shared` is that you can "rent" as many buffer you need per thread.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: I suppose we could create the static buffer once, but then we're exchanging a very little time in needing to fill the rented array with the memory necessary to keep the static buffer. 
I think it would make sense to keep it as it is.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Sure.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: sure.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Sure.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Sure.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Sure.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Makes sense!

tests/MongoDB.Driver.Tests/Specifications/socks5-support/Socks5SupportProseTests.cs
papafe: I've used something similar now, hopefully it's readable enough.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Thanks!

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
BorisDog: `ThreadStaticBuffer` is designed to be threadsafe, leveraging the `[ThreadStatic]` technique, which I believe what `ArrayPool<byte>.Shared` does.
You are correct about the limitation for a single rent per thread, but looks like that's what is needed here.

I am thinking that if we don't have a good reason to use one over another, than we should fallback to consistency principle. But going forward, if `ArrayPool<byte>.Shared` is as performant, no reason not switch to it in all places in the future.
I do like the option of  `using` in `RentedBuffer`, which eliminated the need for `finally`, so we still might want to create wrapper around `ArrayPool<byte>.Shared`.

papafe: If we want to use `ConfigureAwait(false)` in the async version, then we can't use `ThreadStaticBuffer` as it's not thread safe.

evergreen/cleanup-proxy.sh
papafe: Yup, I've changed it to save the PID of the server processes, and kill them afterwards. The killing almost certainly has been already done by the evergreen process cleaning, but I left the cleanup script just in case, as they have suggested.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
BorisDog: I haven't though about the async version, thanks!

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+        // We skip the last bytes of the response as we do not need them.
+        // We skip length(dst.address) + length(dst.port) - 1 --- length(dst.port) is always 2
+        // -1 because we already ready the first byte of the address type
+        // (necessary for the variable length domain-type addresses)
+        return buffer[3] switch
+        {
+            AddressTypeIPv4 => 5,
+            AddressTypeIPv6 => 17,
+            AddressTypeDomain => buffer[4] + 2,
+            _ => throw new IOException("Unknown address type in SOCKS5 reply.")
```
BorisDog: It's worth adding the actual value, here and in other exceptions where applicable.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
BorisDog: Need to update the comment to match the sync version.

tests/MongoDB.Driver.Tests/Specifications/socks5-support/Socks5SupportProseTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                x.ConnectionString,
+                x.ExpectedResult,
+                x.useTls,
+                x.isAsync
+            });
+    }
+
+    [Theory]
+    [MemberData(nameof(GetTestCombinations))]
+    public async Task TestConnectionStrings(string id, string connectionString, bool expectedResult, bool useTls, bool async)
```
BorisDog, line 81: Let's add  GitHub permalink to the prose test.

src/MongoDB.Driver/Core/Configuration/Socks5ProxyStreamSettings.cs
```diff
@@ -0,0 +1,50 @@
...
+    public Socks5ProxySettings Socks5ProxySettings => _socks5ProxySettings;
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Socks5ProxyStreamSettings"/> with the specified SOCKS5 proxy settings.
+    /// </summary>
+    /// <param name="socks5ProxySettings"></param>
+    /// <returns></returns>
+    public Socks5ProxyStreamSettings With(Socks5ProxySettings socks5ProxySettings)
+    {
+        return new Socks5ProxyStreamSettings(socks5ProxySettings);
```
BorisDog: nit: expression body?

src/MongoDB.Driver/Core/Configuration/Socks5ProxyStreamSettings.cs
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+using MongoDB.Driver.Core.Connections;
+
+namespace MongoDB.Driver.Core.Configuration;
+
+/// <summary>
+/// Represents settings for a SOCKS5 proxy stream.
+/// </summary>
+public class Socks5ProxyStreamSettings
```
BorisDog: sealed?

src/MongoDB.Driver/MongoUrlBuilder.cs
```diff
@@ -439,6 +447,55 @@ public string Password
             set { _password = value; }
         }
 
+        /// <summary>
+        ///
```
BorisDog: Comments

src/MongoDB.Driver/Core/Configuration/ClusterBuilder.cs
```diff
@@ -157,6 +158,19 @@ public ClusterBuilder ConfigureTcp(Func<TcpStreamSettings, TcpStreamSettings> co
             return this;
         }
 
+        /// <summary>
+        /// Configures the SOCKS5 proxy settings for the cluster.
+        /// </summary>
+        /// <param name="configurator"> The SOCKS5 proxy settings configurator delegate.</param>
+        /// <returns> A reconfigured cluster builder.</returns>
```
BorisDog: nit: remove the space before 'A'

BorisDog (changes requested):

tests/MongoDB.Driver.Tests/Specifications/socks5-support/Socks5SupportProseTests.cs
papafe: Sure.

src/MongoDB.Driver/Core/Configuration/Socks5ProxyStreamSettings.cs
papafe: Probably a good idea.

src/MongoDB.Driver/Core/Configuration/ClusterBuilder.cs
papafe: Done.

src/MongoDB.Driver/MongoUrlBuilder.cs
papafe: Yes 🤦

src/MongoDB.Driver/Core/Configuration/Socks5ProxyStreamSettings.cs
papafe: Done.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Done.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5ProxySettingsTest.cs
```diff
@@ -0,0 +1,172 @@
...
+
+    [Fact]
+    public void Socks5AuthenticationSettings_None_should_return_NoAuthenticationSettings_instance()
+    {
+        var none = Socks5AuthenticationSettings.None;
+        none.Should().BeOfType<Socks5AuthenticationSettings.NoAuthenticationSettings>();
+    }
+
+    [Fact]
+    public void Socks5AuthenticationSettings_UsernamePassword_should_return_UsernamePasswordAuthenticationSettings_instance_with_correct_values()
```
BorisDog: Should this be in `Socks5AuthenticationSettingsTests`?

BorisDog (changes requested):
Few minor comment + unaddressed comment

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Corrected.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5ProxySettingsTest.cs
papafe: I've added to its own file. In the beginning I didn't as I thought that the tests were very few and related to the ones already here, so I put them in the same one.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+        var acceptedMethod = buffer[1];
+        if (acceptedMethod == MethodUsernamePassword)
+        {
+            if (!useAuth)
+            {
+                throw new IOException("SOCKS5 proxy requires authentication, but no credentials were provided.");
+            }
+
+            return true;
+        }
```
vbabanin: As i understand, If the server cannot find any acceptable authentication methods offered by the client, it replies with `0xFF` ("no acceptable methods"). The only way the server can select username/password `(0x02)` is if the client offered it. A case where the server replies with `0x02` but the client did not offer it should not occur in a compliant implementation - that would indicate a protocol violation by the server.

To handle this defensively, what do you think about throwing an error such as:
"Unexpected server response: server replied with <method> but expected one of <offered methods>"?

This would highlight the unexpected behavior to users and, in edge cases where a proxy forces `0x02`, might help discourage clients from inadvertently sending credentials.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+
+        if (acceptedMethod != MethodNoAuth)
+        {
+            throw new IOException($"SOCKS5 proxy requires unsupported authentication method. Unsupported method: {acceptedMethod}");
+        }
+
+        return false;
+    }
+
+    private static int CreateAuthenticationRequest(byte[] buffer, Socks5AuthenticationSettings authenticationSettings)
```
vbabanin, line 212: [nitpick] The method signature looks generic, but internally it casts from `Socks5AuthenticationSettings` to `UsernamePasswordAuthenticationSettings`, so it effectively only works with the `0x02` method. Would it make sense to make the signature explicit (e.g. only accept `UsernamePasswordAuthenticationSettings`)?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+        return 3 + usernameLength + passwordLength;
+    }
+
+    private static void ProcessAuthenticationResponse(byte[] buffer)
+    {
+        if (buffer[0] != SubnegotiationVersion || buffer[1] != Socks5Success)
+        {
+            throw new IOException($"SOCKS5 authentication failed. Version: {buffer[0]}, Status: {buffer[1]}.");
+        }
+    }
```
vbabanin: According to the [RFC-1929](https://datatracker.ietf.org/doc/html/rfc1929), authentication only fails when `buffer[1] != Socks5Success`. If `buffer[0] != SubnegotiationVersion` (despite the client specifying 1), that suggests a protocol-noncompliant response from the server/proxy.

The RFC states:
> “The VER field contains the current version of the subnegotiation, which is X'01'.”

I'd propose making the failure reason explicit here. When `buffer[0] != SubnegotiationVersion`, we should indicate that this is an unexpected server response rather than a credentials error. For example: `"SOCKS5 auth subnegotiation: unexpected VER from server"`. This would also make troubleshooting easier. What do you think?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+            throw new IOException($"Invalid SOCKS version in response. Expected version {ProtocolVersion5}, but received {version}.");
+        }
+    }
+
+    private static byte EncodeString(string input, byte[] buffer, int offset, string parameterName)
+    {
+        try
+        {
+            var written = Encoding.UTF8.GetBytes(input, 0, input.Length, buffer, offset);
+            return checked((byte)written);
```
vbabanin, line 305: Casting username or password length to byte works correctly if the `int` values are ≤ `0xFF`. Otherwise, we end up reporting incorrect lengths and writing more bytes than the proxy expects, which could silently corrupt data and result in unexpected responses from the server.

I’d propose enforcing a maximum length (in UTF-8 bytes, not chars/code points) for username and password in `UsernamePasswordAuthenticationSettings`. That way, it is validate at the API edge and fail fast with a clear error message.

The same applies to `var hostLength = EncodeString(targetHost, buffer, 5, nameof(targetHost));`. Per [RFC 1035 §2.3.4](https://www.rfc-editor.org/rfc/rfc1035.html#section-2.3.4), the maximum domain name length is 255 bytes (0xFF), which i'd propose enforcing in `Socks5ProxySettings`.

What do you think?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    // Reads the SOCKS5 connect response and returns the number of bytes to skip in the buffer.
+    private static int ProcessConnectResponse(byte[] buffer)
+    {
+        VerifyProtocolVersion(buffer[0]);
+
+        if (buffer[1] != Socks5Success)
+        {
+            throw new IOException($"SOCKS5 connect failed. Error code: {buffer[1]}");
```
vbabanin: Should we provide a more user-friendly message here indicating what type of error the server returned?

The RFC defines the following failure reply types:
> 0x01 General SOCKS server failure
> 0x02 Connection not allowed by ruleset
> 0x03 Network unreachable
> 0x04 Host unreachable
> 0x05 Connection refused
> 0x06 TTL expired
> 0x07 Command not supported
> 0x08 Address type not supported

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Yes, this should not happen unless the server is behaving badly, I'm trying to be defensive here. 

Regarding the message, I agree, it's a little vague, I'll try to rephrase it so it highlights the issue better.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: I think we can keep the method generic, but throw an exception if the authentication is not for the correct type (that should not happen), so we keep the method open to extension in the future if needed.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: I agree!

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: This is what the `checked` statement is doing, it will throw an `OverflowException` in case the value cannot be safely converted to `byte`. 
I agree that probably we can be a little more user friendly here, instead of throwing a generic exception. I'll try to add some more validation to our settings class.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Makes sense! Added.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Added.

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
rstam: See:

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1820

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+    [InlineData(TooLong, "pass")]
+    [InlineData("user", TooLong)]
+    public void UsernamePassword_should_throw_when_username_or_password_are_too_long(string username, string password)
+    {
+        var ex = Record.Exception(() => Socks5AuthenticationSettings.UsernamePassword(username, password));
+        ex.Should().BeAssignableTo<ArgumentException>();
+    }
+
+    [Fact]
+    public void NoAuthenticationSettings_Equals_should_return_true_for_any_Socks5AuthenticationSettings()
```
BorisDog: The name of this test is confusing, do we really want the `NoAuthenticationSettings` to be equal to any of the settings?

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        none.Should().BeOfType<Socks5AuthenticationSettings.NoAuthenticationSettings>();
+    }
+
+    [Fact]
+    public void UsernamePassword_should_return_UsernamePasswordAuthenticationSettings_instance_with_correct_values()
+    {
+        var up = Socks5AuthenticationSettings.UsernamePassword("user", "pass");
+        up.Should().BeOfType<Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings>();
+        var upcast = (Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings)up;
+        upcast.Username.Should().Be("user");
```
BorisDog, line 40: minor: 
```
var upcast = up.Should()
   .BeOfType<Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings>()
   .Subject;
```

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    [Fact]
+    public void UsernamePasswordAuthenticationSettings_Equals_and_GetHashCode_should_work()
+    {
+        var a = Socks5AuthenticationSettings.UsernamePassword("u", "p");
+        var b = Socks5AuthenticationSettings.UsernamePassword("u", "p");
+        a.Equals(b).Should().BeTrue();
+        a.GetHashCode().Should().Be(b.GetHashCode());
+        var c = Socks5AuthenticationSettings.UsernamePassword("u", "x");
+        a.Equals(c).Should().BeFalse();
```
BorisDog: Is it worth splitting the test cases with `InlineData`?

BorisDog (changes requested):
Few minor comments

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Corrected.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: It should be `NoAuthenticationSettings_Equals_and_GetHashCode_should_work_correctly` like the following test, I'll fix it.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: I've refactored the test a little bit, so I don't think it's necessary to use `InlineData`

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
papafe: So it seems that using ` _proxy == String.Empty` is the non recommended way of doing it, while using `IsNullOrEmpty` or checking the length (and nullability if necessary) are better. To be honest this is definitely not a hot path, so performance is not something we should be worried about 😁

src/MongoDB.Driver/MongoClientSettings.cs
```diff
@@ -1165,6 +1188,10 @@ public override string ToString()
             sb.AppendFormat("MaxConnectionLifeTime={0};", _maxConnectionLifeTime);
             sb.AppendFormat("MaxConnectionPoolSize={0};", _maxConnectionPoolSize);
             sb.AppendFormat("MinConnectionPoolSize={0};", _minConnectionPoolSize);
+            if (_socks5ProxySettings!= null)
```
BorisDog: nit: space before `!=`

src/MongoDB.Driver/Core/Connections/Socks5ProxyStreamFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.IO;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Connections;
+
+internal class Socks5ProxyStreamFactory: IStreamFactory
```
BorisDog: minor: sealed, space before `:`

BorisDog (dismissed):
LGTM + few minor comments

src/MongoDB.Driver/Core/Configuration/ClusterBuilder.cs
```diff
@@ -286,6 +300,12 @@ private IServerMonitorFactory CreateServerMonitorFactory()
         private IStreamFactory CreateTcpStreamFactory(TcpStreamSettings tcpStreamSettings)
         {
             var streamFactory = (IStreamFactory)new TcpStreamFactory(tcpStreamSettings);
+
+            if (_socks5ProxyStreamSettings != null)
```
sanych-sun, line 308: Is it possible to use ssl connection together with socks proxy? Have we tested such setup?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,344 @@
...
+    // Connect response
+    // +----+-----+-------+------+----------+----------+
+    // |VER | REP |  RSV  | ATYP | DST.ADDR | DST.PORT |
+    // +----+-----+-------+------+----------+----------+
+    // | 1  |  1  | X'00' |  1   | Variable |    2     |
+    // +----+-----+-------+------+----------+----------+
+
+    //General use constants
+    private const byte ProtocolVersion5 = 0x05;
+    private const byte Socks5Success = 0x00;
```
sanych-sun, line 79: This constant is not in use. Should we remove it or use in `EnsureSocksSuccess`?

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
```diff
@@ -0,0 +1,344 @@
...
+using System.Buffers;
+using System.Buffers.Binary;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.GridFS;
```
sanych-sun: Having reference to GridFS here is weird. I've figured out you are using extensions method to stream from that namespace. Should we move that extension methods to some more generic namespaces?

src/MongoDB.Driver/Core/Connections/Socks5ProxyStreamFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+        Socks5Helper.PerformSocks5Handshake(stream, endPoint, _settings.Socks5ProxySettings.Authentication, cancellationToken);
+        return stream;
+    }
+
+    public async Task<Stream> CreateStreamAsync(EndPoint endPoint, CancellationToken cancellationToken)
+    {
+        var proxyEndpoint = new DnsEndPoint(_settings.Socks5ProxySettings.Host, _settings.Socks5ProxySettings.Port);
+        var stream = await _wrapped.CreateStreamAsync(proxyEndpoint, cancellationToken).ConfigureAwait(false);
+        await Socks5Helper.PerformSocks5HandshakeAsync(stream, endPoint, _settings.Socks5ProxySettings.Authentication, cancellationToken).ConfigureAwait(false);
+        return stream;
```
sanych-sun: Should we dispose the stream in case of errors?

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        var none = Socks5AuthenticationSettings.None;
+        none.Should().BeOfType<Socks5AuthenticationSettings.NoAuthenticationSettings>();
+    }
+
+    [Fact]
+    public void UsernamePassword_should_return_UsernamePasswordAuthenticationSettings_instance_with_correct_values()
+    {
+        var up = Socks5AuthenticationSettings.UsernamePassword("user", "pass");
+        var upcast = up.Should() .BeOfType<Socks5AuthenticationSettings.UsernamePasswordAuthenticationSettings>().Subject;
```
sanych-sun: _Minor_: extra space before `.BeOfType<>`

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        upcast.Username.Should().Be("user");
+        upcast.Password.Should().Be("pass");
+    }
+
+    [Theory]
+    [InlineData(null, "pass")]
+    [InlineData("user", null)]
+    [InlineData("", "pass")]
+    [InlineData("user", "")]
+    public void UsernamePassword_should_throw_when_username_or_password_is_null_or_empty(string username, string password)
```
sanych-sun: _Minor_: I would split this into 2 theories (separate for username and separate for password), and assert on the ParamName as well.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        var ex = Record.Exception(() => Socks5AuthenticationSettings.UsernamePassword(username, password));
+        ex.Should().BeAssignableTo<ArgumentException>();
+    }
+
+    [Theory]
+    [InlineData(TooLong, "pass")]
+    [InlineData("user", TooLong)]
```
sanych-sun: _Minor_: same as above.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        var up = Socks5AuthenticationSettings.UsernamePassword("a", "b");
+        none.Equals(up).Should().BeFalse();
+        none.GetHashCode().Should().NotBe(up.GetHashCode());
+    }
+
+    [Fact]
+    public void UsernamePasswordAuthenticationSettings_Equals_and_GetHashCode_should_work_correctly()
+    {
+        var up1 = Socks5AuthenticationSettings.UsernamePassword("u", "p");
```
sanych-sun: Should we turn this into a theory? With following cases:
```
"u", "p", "u", "p", true
"u", "p", "u", "x", false
"u", "p", "x", "p", false
```

sanych-sun (changes requested):

src/MongoDB.Driver/MongoClientSettings.cs
papafe: Corrected

src/MongoDB.Driver/Core/Connections/Socks5ProxyStreamFactory.cs
papafe: Fixed

src/MongoDB.Driver/Core/Configuration/ClusterBuilder.cs
papafe: Yes, it's possible. For all the connection strings we are using in the prose tests we test both SSL/NoSSL Sync/Async.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: We should use it in `EnsureSocksSuccess`, got removed by mistake during a refactoring.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Corrected

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Reasonable! Done it.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Done it.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Reasonable, I've added this.

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: Makes sense... For now I've moved it to the `Support` namespace, as there are other useful extension classes there, but I'm open to other possibilities.

sanych-sun: Should we put there extensions into `StreamExtensionMethods.cs`?

BorisDog (dismissed):
LGTM!

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    public void UsernamePassword_should_throw_when_username_is_null_or_empty(string username)
+    {
+        var ex = Record.Exception(() => Socks5AuthenticationSettings.UsernamePassword(username, "pass"));
+        ex.Should().BeAssignableTo<ArgumentException>();
+        ex.Message.Should().Contain("username");
```
sanych-sun: We should assert that `ex.ParamName.Should().Be("username")` instead.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    [InlineData(null)]
+    [InlineData("")]
+    public void UsernamePassword_should_throw_when_password_is_null_or_empty(string password)
+    {
+        var ex = Record.Exception(() => Socks5AuthenticationSettings.UsernamePassword("username", password));
+        ex.Should().BeAssignableTo<ArgumentException>();
+        ex.Message.Should().Contain("password");
+    }
+
+    [Fact]
```
sanych-sun: _Minor_: I would prefer to group test that using a member together (for example all tests that checking `username` parameter behavior). So it would be easier to see the whole picture of how the member should work.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    [Theory]
+    [InlineData("u", "p", "u", "p", true)]
+    [InlineData("u", "p", "u", "x", false)]
+    [InlineData("u", "p", "x", "p", false)]
+    public void UsernamePasswordAuthenticationSettings_Equals_and_GetHashCode_should_work_correctly(string u1, string p1, string u2, string p2, bool areEqual)
+    {
+        var up1 = Socks5AuthenticationSettings.UsernamePassword(u1, p1);
+
+        var none = Socks5AuthenticationSettings.None;
```
sanych-sun: _Minor_: testing equality with `Socks5AuthenticationSettings.None` probably could be done as separate test.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Connections/Socks5Helper.cs
papafe: That seems the more appropriate place 😁 
Moved it there.

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Done

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Done

tests/MongoDB.Driver.Tests/Core/Connections/Socks5AuthenticationSettingsTests.cs
papafe: Makes sense. This made me discover I forgot to pass the `paramName` for one of the errors.

sanych-sun (approved):
LGTM

PR #1732, opened 2025-07-16
CSHARP-5536: Test sharded clusters with requireApiVersion=1

adelinowona:
Previously we were only testing with standalone servers. We are now unblocked from testing on sharded clusters, but not yet on replica sets.


sanych-sun (approved):
LGTM, there are some failures in new variants, but it seems to be related to ongoing problem with the latest build on Windows.

PR #1738, opened 2025-07-17
CSHARP-5373: Add test that PoolClearedEvent is emitted before ConnectionCheckedInEvent/ConnectionCheckOutFailedEvent

adelinowona:


rstam (approved):
LGTM

PR #1739, opened 2025-07-17
CSHARP-5391: Add spec test for $$type operator with "number"

adelinowona:


tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
```diff
@@ -193,22 +195,30 @@ private void AssertValuesMatch(BsonValue actual, BsonValue expected, bool isRoot
...
+                        t.AsString == actualTypeName ||
+                        (t.AsString == "number" && NumericTypes.Contains(actualTypeName)))
+                    .Should().BeTrue();
             }
             else
             {
                 throw new FormatException($"Unexpected $$type value BsonType: '{expectedTypes.BsonType}'.");
             }
-
-            actualTypeName.Should().BeOneOf(expectedTypeNames);
```
rstam: I don't like seeing this line removed. This is a nice one liner that summarizes WHAT we are asserting.

Instead, I'd rather see how `expectedTypeNmes` is computed refactored.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
```diff
@@ -27,6 +28,7 @@ namespace MongoDB.Driver.Tests.UnifiedTestOperations.Matchers
     public class UnifiedValueMatcher
     {
         private UnifiedEntityMap _entityMap;
+        private static readonly HashSet<string> NumericTypes = ["int", "long", "double", "decimal"];
```
rstam: I would put this new static field in front of any instance fields and leave a blank between static fields and instance fields.

Also, follow our naming conventions for static fields:

    private static readonly List<string> __numericTypes = ["int", "long", "double", "decimal"];
                                                                                               
    private UnifiedEntityMap _entityMap;                                                       

Finally... it is known that for very small sets `List<string>` is more efficient than `HashSet<string>`.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/Matchers/UnifiedValueMatcher.cs
```diff
@@ -193,22 +195,30 @@ private void AssertValuesMatch(BsonValue actual, BsonValue expected, bool isRoot
         private void AssertExpectedType(BsonValue actual, BsonValue expectedTypes)
         {
             var actualTypeName = GetBsonTypeNameAsString(actual.BsonType);
-            List<string> expectedTypeNames;
 
             if (expectedTypes.IsString)
             {
-                expectedTypeNames = new List<string> { expectedTypes.AsString };
+                var expectedType = expectedTypes.AsString;
```
rstam, line 203: Instead of asserting differently in more than one place keep the existing assert at the end of this method and change how `expectedTypeNames` is computed:

            if (expectedTypes.IsString)
            {
                var expectedType = expectedTypes.AsString;
                expectedTypeNames = expectedType == "number" ? __numericTypes : [expectedType];
            }
            else if (expectedTypes.IsBsonArray)
            {
                expectedTypeNames = expectedTypes.AsBsonArray
                    .Select(t => t.AsString)
                    .SelectMany(t => t == "number" ? __numericTypes : [t])
                    .ToList();
            }
            else
            {
                throw new FormatException($"Unexpected $$type value BsonType: '{expectedTypes.BsonType}'.");
            }

rstam: I leave it to you to verify that all tests still pass with this proposed refactoring.

rstam: Also, does `"number"` actually ever appear as ONE element of an array?

In the json file you are adding it always appears by itself.

So the changes to the `else if` part might not actually be needed...

adelinowona: I decided to also modify the BsonArray case as the "number" value could be used anywhere. Currently there are no instances in our spec tests where it is used as an element of an array but in the future we could have a spec test that uses the "number" value in an array. However, it is untested functionality currently and updating the array use case in the future will be easy so I'll defer it.

PR #1748, opened 2025-08-07
CSHARP-5672: Support sorting by value in PushEach operation

adelinowona:


src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -288,4 +311,26 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
             return new BsonDocument(renderedField.FieldName, value);
         }
     }
+
+    internal sealed class NoFieldDirectionalSortDefinition<TDocument> : SortDefinition<TDocument>
```
rstam: I suggest naming this class `ValueDirectionalSortDefinition`.

I'd rather emphasize that we are sorting by "value". The fact that there is no field name is a side effect, not the main feature.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -288,4 +311,26 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+    {
+        private readonly SortDirection _direction;
+
+        public NoFieldDirectionalSortDefinition(SortDirection direction)
+        {
+            _direction = direction;
+        }
+
+        public override BsonDocument Render(RenderArgs<TDocument> args)
```
rstam, line 317: Do we want a `Direction` property like `DirectionalSortDefinition` has?

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -288,4 +311,26 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+        {
+            BsonValue value = _direction switch
+            {
+                SortDirection.Ascending => 1,
+                SortDirection.Descending => -1,
+                _ => throw new InvalidOperationException("Unknown value for " + typeof(SortDirection) + ".")
```
rstam: I would have done this:

```_ => throw new InvalidOperationException($"Invalid sort direction: {_direction}.")```

but I suppose then we might want to change `DirectionalSortDefinition` also...

File: src/MongoDB.Driver/SortDefinitionBuilder.cs [hunk 4]
```diff
@@ -288,4 +311,26 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+            };
+
+            return new BsonDocument("direction",  value);
```
rstam: When we use "magic" values like this we usually enclose them in angle brackets to make it visually apparent that something is special about this value:

```return new BsonDocument("<direction>",  value);```

We only have a field name here because `Render` is already declared to return `BsonDocument` and changing it to return `BsonValue` would be a breaking change.

File: src/MongoDB.Driver/UpdateDefinitionBuilder.cs [hunk 5]
```diff
@@ -1685,7 +1685,13 @@ public override BsonValue Render(RenderArgs<TDocument> args)
 
             if (_sort != null)
             {
-                document["$push"][renderedField.FieldName]["$sort"] = _sort.Render(args.WithNewDocumentType((IBsonSerializer<TItem>)itemSerializer));
+                var newArgs = args.WithNewDocumentType((IBsonSerializer<TItem>)itemSerializer);
+
+                var renderedSort = _sort is NoFieldDirectionalSortDefinition<TItem>
+                    ? _sort.Render(newArgs)["direction"]
```
rstam: ```? _sort.Render(newArgs)["<direction>"]```

Note the added angle brackets.

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
```diff
@@ -31,6 +32,14 @@ public void Ascending()
             Assert(subject.Ascending("a"), "{a: 1}");
         }
 
+        [Fact]
+        public void Ascending_no_field()
```
rstam: I would name this test `Ascending_value`

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
```diff
@@ -84,6 +103,14 @@ public void Descending()
             Assert(subject.Descending("a"), "{a: -1}");
         }
 
+        [Fact]
+        public void Descending_no_field()
```
rstam: I would name this test `Descending_value`.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -288,4 +311,26 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+        private readonly SortDirection _direction;
+
+        public NoFieldDirectionalSortDefinition(SortDirection direction)
+        {
+            _direction = direction;
+        }
+
+        public override BsonDocument Render(RenderArgs<TDocument> args)
+        {
+            BsonValue value = _direction switch
```
BorisDog: We do the `SortDirection` to `BsonValue` conversion in 2 different places already.
Consider extorting this to internal `SortDirectionExtensions` or similar?

src/MongoDB.Driver/SortDefinitionBuilder.cs
BorisDog: So what happens in other cases (which are not `UpdateDefinitionBuilder`), will it be rendered as 
`sort: { direction: ... }` and do we want that?

We could add
`internal virtual BsonValue RenderAsBsonValue(RenderArgs<TDocument> args) => Render(args);`
to `SortDefinition`, and refactor `Render` to return `BsonValue` in 4.0.

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: We currently have no use for that property, so I decided to omit it for now to avoid introducing unused code.

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: Nice, I like this! I thought of something similar but thought adding something to SortDefinition would be a breaking change but I suppose adding an internal virtual method is not? Learning some library maintainer tricks :)

adelinowona: With this updated approach, I am also thinking of throwing an exception when a call to `Render` is made on a `ValueDirectionalSortDefinition`. A call to render for `ValueDirectionalSortDefinition` is suggesting someone is trying to use this where a field-based sort is expected which is invalid.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -272,20 +295,25 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+        private readonly SortDirection _direction;
+
+        public ValueDirectionalSortDefinition(SortDirection direction)
+        {
+            _direction = direction;
+        }
+
+        internal override BsonValue RenderAsBsonValue(RenderArgs<TDocument> args) => _direction.ToBsonValue();
+
+        public override BsonDocument Render(RenderArgs<TDocument> args)
```
adelinowona, line 317: For now, I'm using the exception thrown during rendering as the mechanism to inform users that value-based sorts aren't supported. Since I've only updated call sites where value-based sorts are explicitly supported, any `Render` call that returns a `BsonDocument` will naturally fail in unsupported contexts, keeping this PR's changes minimal. However, Alex raised a point that when we update `Render` to return `BsonValue` in 4.0, we'll need to add validation at all call sites to check for value-based sort support before rendering. The question is whether to implement this validation now or defer it to 4.0. Implementing it now would distribute some of the 4.0 migration work and provide users with cleaner exceptions higher in the stack trace (and potentially a better exception type). I suppose we'll probably want to do that but what's the preference? @BorisDog @rstam

src/MongoDB.Driver/SortDefinition.cs
```diff
@@ -47,6 +47,9 @@ public abstract class SortDefinition<TDocument>
         /// <returns>A <see cref="BsonDocument"/>.</returns>
         public abstract BsonDocument Render(RenderArgs<TDocument> args);
 
+        // TODO: remove this and refactor Render to return a BsonValue in 4.0
+        internal virtual BsonValue RenderAsBsonValue(RenderArgs<TDocument> args) => Render(args);
```
rstam, line 51: The `RenderAsBsonValue` method should probably be placed *after* the `Render` method.

src/MongoDB.Driver/SortDirectionExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+using System;
+using MongoDB.Bson;
+
+namespace MongoDB.Driver
+{
+    internal static class SortDirectionExtensions
+    {
+        internal static BsonValue ToBsonValue(this SortDirection direction)
```
rstam: I think this method should probably be named `Render`.

src/MongoDB.Driver/SortDefinitionBuilder.cs
rstam: I think the way you have it now is fine.

We can address whatever changes (if any) are needed later.

Most places that currently call `Render` *should* work with a `BsonValue` also. If not there is some work to be done a the CALL site, *not* here.

BorisDog: I agree with @rstam, we can defer the work to 4.0.

src/MongoDB.Driver/SortDirectionExtensions.cs
BorisDog: Nice, would be good to use it in other places as well, like `DirectionalIndexKeyDefinition`.

src/MongoDB.Driver/SortDefinition.cs
adelinowona: It already is? Maybe you got a defective diff

src/MongoDB.Driver/SortDirectionExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using MongoDB.Bson;
+
+namespace MongoDB.Driver
+{
+    internal static class SortDirectionExtensions
+    {
+        internal static BsonValue Render(this SortDirection direction)
+        {
+            return direction switch
```
BorisDog: minor: you can use expression body definition.

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
```diff
@@ -31,6 +32,14 @@ public void Ascending()
             Assert(subject.Ascending("a"), "{a: 1}");
         }
 
+        [Fact]
+        public void Ascending_value()
+        {
+            var subject = CreateSubject<BsonDocument>();
+
+            Assert(subject.Ascending(), "{direction: 1}");
```
BorisDog: Fix the tests, and add new one to test exception?

src/MongoDB.Driver/SortDefinition.cs
rstam: I tagged the wrong one. It's the one on line 311 of SortDefinitionBuilder.cs that's out of order.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -272,20 +295,25 @@ public override BsonDocument Render(RenderArgs<TDocument> args)
...
+    internal sealed class ValueDirectionalSortDefinition<TDocument> : SortDefinition<TDocument>
+    {
+        private readonly SortDirection _direction;
+
+        public ValueDirectionalSortDefinition(SortDirection direction)
+        {
+            _direction = direction;
+        }
+
+        internal override BsonValue RenderAsBsonValue(RenderArgs<TDocument> args) => _direction.Render();
```
rstam: Move `RenderAsBsonValue` to after `Render`.

rstam (changes requested):
Looks good with one minor change requested.

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: Done!

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
```diff
@@ -117,10 +144,20 @@ public void MetaTextScore()
             Assert(subject.MetaTextScore("awesome"), "{awesome: {$meta: 'textScore'}}");
         }
 
+        [Fact]
+        public void CallingRenderOnValueBasedSortShouldThrow()
```
BorisDog: Use the underscores naming styling, like you did in
`Combine_with_value_based_sort_and_additional_sort_should_throw `

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -130,6 +130,15 @@ public static SortDefinition<TDocument> MetaTextScore<TDocument>(this SortDefini
     /// <typeparam name="TDocument">The type of the document.</typeparam>
     public sealed class SortDefinitionBuilder<TDocument>
     {
+        /// <summary>
+        /// Creates a value ascending sort.
+        /// </summary>
```
BorisDog, line 136: I am wondering if should add a short example of what "value sort" is. As it might be not that evident to our users.
Something like "used for expressions that are not requiring a field, for example "`$sort: -1`"
cc @rstam

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
adelinowona: done

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: done

tests/MongoDB.Driver.Tests/SortDefinitionBuilderTests.cs
```diff
@@ -117,10 +144,20 @@ public void MetaTextScore()
             Assert(subject.MetaTextScore("awesome"), "{awesome: {$meta: 'textScore'}}");
         }
 
+        [Fact]
+        public void Calling_render_on_value_based_sort_should_throw()
```
BorisDog: Alphabetical order

src/MongoDB.Driver/SortDefinitionBuilder.cs
rstam: This comment needs to be reworked a bit.

The statement "when documents are simple values like strings" doesn't make sense to me. A document can never be a simple value, it's a document.

rstam: Something like this:

```
/// Create an ascending sort on a value rather than on a field of a document. For example, "$sort : 1".
/// This is used when sorting primitive values like strings or numbers, but can also be used to sort whole documents.
```

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -131,7 +131,19 @@ public static SortDefinition<TDocument> MetaTextScore<TDocument>(this SortDefini
...
+        /// <returns>A value ascending sort.</returns>
+        public SortDefinition<TDocument> Ascending()
+        {
+            return new ValueDirectionalSortDefinition<TDocument>(SortDirection.Ascending);
+        }
+
+        /// <summary>
+        /// Creates an ascending sort based on a specific field within the document. For example, "$sort: {field: 1}"
+        /// This is used when documents are complex objects, and you want to sort by a particular field's value.
```
rstam: I would reword slightly:

```
/// Creates an ascending sort based on a specific field within the document. For example, "$sort : { field : 1 }"
/// This is used when values are documents, and you want to sort by a particular field's value.                  
```

rstam: Note the extra white space in `$sort : { field : 1 }` which is our custom for readability.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -141,7 +153,8 @@ public SortDefinition<TDocument> Ascending(FieldDefinition<TDocument> field)
         }
 
         /// <summary>
-        /// Creates an ascending sort.
+        /// Creates an ascending sort based on a specific field within the document. For example, "$sort: {field: 1}"
+        /// This is used when documents are complex objects, and you want to sort by a particular field's value.
```
rstam: Same as previous comment.

src/MongoDB.Driver/SortDefinitionBuilder.cs
```diff
@@ -171,7 +184,19 @@ public SortDefinition<TDocument> Combine(IEnumerable<SortDefinition<TDocument>>
         }
 
         /// <summary>
-        /// Creates a descending sort.
+        /// Creates a value descending sort that sorts documents by their entire document value. For example, "$sort: -1"
+        /// This is typically used when documents are simple values (like strings, numbers, or dates)
+        /// rather than complex objects with multiple fields.
```
rstam: Same as ascending comment for the three `Descending` methods.

rstam (changes requested):
Minor changes to comments suggested.

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: Done

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: done

src/MongoDB.Driver/SortDefinitionBuilder.cs
adelinowona: done

PR #1752, opened 2025-08-11
CSHARP-3547: CSOT: read-write concern

sanych-sun:


tests/MongoDB.Driver.Tests/Core/Operations/CreateViewOperationTests.cs
```diff
@@ -295,24 +290,39 @@ public void CreateCommand_should_return_expected_result_when_Collation_is_set(
...
+            {
+                writeConcern ??= WriteConcern.Acknowledged;
+                writeConcern = writeConcern.With(wTimeout: TimeSpan.FromMilliseconds(wtimeout.Value));
+            }
+
             var subject = new CreateViewOperation(_databaseNamespace, _viewName, _collectionNamespace.CollectionName, _pipeline, _messageEncoderSettings)
             {
                 WriteConcern = writeConcern
             };
+            var operationContext = hasOperationTimeout ? new OperationContext(TimeSpan.FromSeconds(42), CancellationToken.None) : OperationContext.NoTimeout;
```
adelinowona, line 308: I noticed you used this 42 magic number for setting the operation timeout when we need. Should we extract this into a constant that can be used across all the tests?

tests/MongoDB.Driver.Tests/Core/Operations/CreateCollectionOperationTests.cs
```diff
@@ -356,22 +357,36 @@ public void CreateCommand_should_return_expected_result_when_Validator_is_set(
...
+            {
+                writeConcern ??= WriteConcern.Acknowledged;
+                writeConcern = writeConcern.With(wTimeout: TimeSpan.FromMilliseconds(wtimeout.Value));
+            }
+
             var subject = new CreateCollectionOperation(_collectionNamespace, _messageEncoderSettings)
             {
                 WriteConcern = writeConcern
             };
+            var operationContext = hasOperationTimeout ? new OperationContext(TimeSpan.FromSeconds(42), CancellationToken.None) : OperationContext.NoTimeout;
```
adelinowona, line 375: I noticed you used this 42 magic number for setting the operation timeout when we need. Should we extract this into a constant that can be used across all the tests?

sanych-sun: Do not think so 🤣, 42 - does not have any special meaning =) Other then it is "answer to the ultimate question of life, the universe, and everything". It could be any number, simply not-zero timeout.

adelinowona: I was suggesting extracting this to a constant so that if we need to change this value in the future, we'll only need to update it in one place rather than hunting down and replacing all instances of the magic number 42 throughout the codebase. But I don't insist, up to you.

PR #1753, opened 2025-08-11
CSHARP-3540: CSOT: Client-side encryption

sanych-sun:


rstam (approved):
LGTM

src/MongoDB.Driver/Core/WireProtocol/CommandMessageFieldEncryptor.cs
```diff
@@ -41,17 +40,23 @@ public CommandMessageFieldEncryptor(IBinaryCommandFieldEncryptor commandFieldEnc
         }
 
         // public static methods
```
adelinowona, line 42: the comment suggests the methods below are static but they aren't. So maybe update that comment? same for the private methods too.

adelinowona (approved):
LGTM just left a single comment you can address if you agree.

PR #1754, opened 2025-08-13
CSHARP-3984: Remove BinaryConnection.DropBox

sanych-sun:


src/MongoDB.Driver/Core/Connections/BinaryConnection.cs
```diff
@@ -482,11 +385,19 @@ public ResponseMessage ReceiveMessage(
             try
             {
                 helper.ReceivingMessage();
-                using (var buffer = ReceiveBuffer(operationContext, responseTo))
+                while (true)
```
sanych-sun, line 406: We need this because driver could stop waiting on response from the server at any moment, and if it happened we should ignore server's response that is not the current request. We used to have exactly the same behavior with DropBox, with only difference that DropBox would store such response forever.

sanych-sun: And just to have full picture: we will change this in https://jira.mongodb.org/browse/DRIVERS-2884

Accordingly to the DRIVERS-2884 Driver should perform reading of pending response BEFORE sending request.

rstam (dismissed):
LGTM

PR #1755, opened 2025-08-13
CSHARP-5675: Where possible, return null for average over the empty set

ajcvickers:
Fixes [CSHARP-5675: Average over empty collection of nullable types throws](https://jira.mongodb.org/browse/CSHARP-5675)

For the terminating operator, the fix is to use the SingleOrDefault finalizer when the type is nullable.

### DIFF
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToExecutableQueryTranslators/AverageMethodToExecutableQueryTranslator.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToExecutableQueryTranslators/AverageMethodToExecutableQueryTranslator.cs
@@ -36,7 +36,8 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToExecut
         // private static fields
         private static readonly MethodInfo[] __averageMethods;
         private static readonly MethodInfo[] __averageWithSelectorMethods;
-        private static readonly IExecutableQueryFinalizer<TOutput, TOutput> __finalizer = new SingleFinalizer<TOutput>();
+        private static readonly IExecutableQueryFinalizer<TOutput, TOutput> __singleFinalizer = new SingleFinalizer<TOutput>();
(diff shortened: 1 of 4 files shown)


src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToExecutableQueryTranslators/AverageMethodToExecutableQueryTranslator.cs
```diff
@@ -155,10 +156,14 @@ public static ExecutableQuery<TDocument, TOutput> Translate<TDocument>(MongoQuer
...
 
+                var returnType = expression.Type;
+
                 return ExecutableQuery.Create(
                     provider,
                     pipeline,
-                    __finalizer);
+                    !returnType.IsValueType || returnType.IsNullable()
+                        ? __singleOrDefaultFinalizer
+                        : __singleFinalizer);
```
rstam, line 166: I don't think it is necessary to test for `!returnType.IsValueType`.

I would simplify this to:

```
var finalizer = expression.Type.IsNullable() ? __singleOrDefaultFinalizer : __singleFinalizer;
return ExecutableQuery.Create(provider, pipeline, finalizer);                                 
```
    
I think it helps to keep parameters to method calls as simple as possible, and move argument computation outside the call.

ajcvickers: @rstam So you are asserting that numeric types will always be value types? Or that if there was a numeric reference type, then it should behave differently?

rstam: What I'm really asserting is that `T` in `Nullable<T>` must be a value type.

And yes, all numeric types are value types currently. Not sure we need to be considering the remote possibility that a future numeric type might not be a value type.

PR #1762, opened 2025-08-23
CSHARP-5500: Support auto encryption in unified tests

adelinowona:
This is the first of three PRs implementing support for text indexes with auto encryption. Most relevant spec tests have been synced and necessary changes in the unified test runner implemented. 

Note: The `maxWireVersion.json` spec test has been intentionally excluded because it requires MongoDB Server 4.0, which is no longer supported in our test matrix.

### DIFF
--- a/tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
+++ b/tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
@@ -193,24 +193,31 @@ namespace MongoDB.Driver.Tests.UnifiedTestOperations
 
                 var database = client.GetDatabase(databaseName).WithWriteConcern(WriteConcern.WMajority);
 
-                _logger.LogDebug("Dropping {0}", collectionName);
-                using var session = client.StartSession();
-                database.DropCollection(session, collectionName);
-
-                // https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#why-are-enxcol_-collections-dropped
-                _logger.LogDebug("Dropping QE state collections for {0}", collectionName);
-                database.DropCollection(session, $"enxcol_.{collectionName}.esc");
-                database.DropCollection(session, $"enxcol_.{collectionName}.ecoc");
-
+                var createCollectionOptions = new CreateCollectionOptions();
                 if (dataItem.AsBsonDocument.Contains("createOptions"))
                 {
-                    var encryptedFields = dataItem["createOptions"]["encryptedFields"].AsBsonDocument;
-                    database.CreateCollection(session, collectionName, new CreateCollectionOptions { EncryptedFields = encryptedFields });
+                    var options = dataItem.AsBsonDocument["createOptions"].AsBsonDocument;
+                    foreach (var option in options)
+                    {
+                        switch (option.Name)
+                        {
+                            case "encryptedFields":
+                                createCollectionOptions.EncryptedFields = option.Value.AsBsonDocument;
+                                break;
+                            default:
+                                throw new FormatException($"Invalid createOptions argument name: '{option.Name}'.");
+                        }
+                    }
                 }
-                else
-                {
-                    database.CreateCollection(session, collectionName);
-                }
+
+                _logger.LogDebug("Dropping {0}", collectionName);
+                using var session = client.StartSession();
+
+                // For some QE spec tests we need to drop QE state collections (enxcol_.*.esc, enxcol_.*.ecoc).
+                // DropCollection with EncryptedFields automatically handles cleanup of those QE state collections
+                database.DropCollection(session, collectionName, new DropCollectionOptions { EncryptedFields = createCollectionOptions.EncryptedFields });
+
+                database.CreateCollection(session, collectionName, createCollectionOptions);
 
                 if (documents.Any())
                 {


tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/ClientSideEncryptionTestRunner.cs
```diff
@@ -39,7 +39,7 @@ public class ClientSideEncryptionTestRunner : MongoClientJsonDrivenTestRunnerBas
         public ClientSideEncryptionTestRunner(ITestOutputHelper testOutputHelper)
             : base(testOutputHelper)
         {
-            CoreTestConfiguration.SkipMongocryptdTests_SERVER_106469();
+            CoreTestConfiguration.SkipMongocryptdTests_SERVER_106469(true);
```
adelinowona, line 42: Resolves the issue where all legacy CSFLE tests were being skipped on the latest MongoDB version on windows, regardless of whether they were using mongocryptd.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -84,7 +84,14 @@ public void Run(JsonDrivenTestCase testCase)
             var outcome = testCase.Test.GetValue("outcome", null)?.AsBsonArray;
             var async = testCase.Test["async"].AsBoolean; // cannot be null
 
-            Run(schemaVersion, testSetRunOnRequirements, entities, initialData, runOnRequirements, skipReason, operations, expectEvents, expectedLogs, outcome, async);
+            // WORKAROUND: The localSchema.json spec test contains empty UUID fields that are
+            // intended to be used as document IDs. However, our driver interprets empty UUIDs
+            // as missing IDs and automatically generates new ones, causing the test to fail.
+            // We disable missing ID validation here to maintain compatibility with the spec test
+            // while preserving the expected test behavior.
+            var shouldDisableAssignId = testCase.Name.Contains("localSchema.json");
```
adelinowona: Our driver treats empty GUID IDs as "not set" and generates new IDs, which follows .NET conventions where `Guid` is a struct that defaults to an empty GUID when uninitialized. However, other drivers allow empty GUID IDs to be inserted and retrieved as-is given that mongo allows for empty UUIDs. This behavioral difference creates compatibility issues for users working with multiple drivers against the same database, as documents may have different ID handling depending on which driver created them. The driver's behavior is technically correct per .NET conventions, but consistency across drivers may be more important for user experience. 

We first should document this behavior in our driver documentation.

specifications/client-side-encryption/tests/unified/localSchema.yml
```diff
@@ -0,0 +1,103 @@
+description: localSchema
```
BorisDog, line 1: We need to be synced with the spec repo, so `maxVersion.json` needs be excluded in code.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
```diff
@@ -337,12 +337,19 @@ public static IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> P
...
+                                                               kmsProviderDocument.Contains("sessionToken");
+
                                 foreach (var providedKmsInfo in kmsProviderDocument)
                                 {
                                     kmsOptions.Add(
                                         providedKmsInfo.Name,
                                         IsPlaceholder(providedKmsInfo.Value)
-                                            ? GetFromEnvVariables(kmsProvider.Name, providedKmsInfo.Name) // use initial kms name
+                                            ? GetFromEnvVariables(
+                                                shouldUseTempCredentials ? "awsTemporary" : kmsProvider.Name,
```
BorisDog: You could compute `kmsProviderName` once outside the loop.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedEntityMap.cs
```diff
@@ -202,7 +201,7 @@ public Dictionary<string, Dictionary<string, LogLevel>> LoggingComponents
             }
         }
 
-        public Dictionary<string, BsonValue> Resutls
+        public Dictionary<string, BsonValue> Results
```
BorisDog, line 204: +1

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedEntityMap.cs
```diff
@@ -764,7 +827,7 @@ private ClientEncryption CreateClientEncryption(Dictionary<string, IMongoClient>
                         options.SetKeyExpiration(keyExpiration);
                         break;
                     default:
-                        throw new FormatException($"Invalid {nameof(ClientEncryptionOptions)} argument name: '{element.Name}'.");
+                        throw new FormatException($"Invalid clientEncryption argument name: '{element.Name}'.");
```
BorisDog, line 830: +1

BorisDog (changes requested):
Few comments

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs
adelinowona: Done

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
adelinowona: I have created a [PR](https://github.com/mongodb/specifications/pull/1831) to update the localSchema.json test data in the spec repo so we don't need to do this workaround anymore.

BorisDog (approved):
LGTM!

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -196,7 +196,22 @@ private BsonDocument AddInitialData(IMongoClient client, BsonArray initialData)
                 _logger.LogDebug("Dropping {0}", collectionName);
                 using var session = client.StartSession();
                 database.DropCollection(session, collectionName);
-                database.CreateCollection(session, collectionName);
+
+                // https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#why-are-enxcol_-collections-dropped
```
sanych-sun: MongoDatabase.DropCollection already has some logic of dropping additional collections related to encryption. See `DropCollectionOperation.CreateEncryptedDropCollectionOperationIfConfigured` helper. There might be chance that we do not need to do that explicitly here.

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
```diff
@@ -196,7 +196,22 @@ private BsonDocument AddInitialData(IMongoClient client, BsonArray initialData)
...
+                _logger.LogDebug("Dropping QE state collections for {0}", collectionName);
+                database.DropCollection(session, $"enxcol_.{collectionName}.esc");
+                database.DropCollection(session, $"enxcol_.{collectionName}.ecoc");
+
+                if (dataItem.AsBsonDocument.Contains("createOptions"))
```
sanych-sun: _nit_ : I would always call `database.CreateCollection` overload with options, but left it `null` if no `createOptions` provided.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestRunner.cs
adelinowona: Looked into it and confirmed that the driver includes a mechanism for dropping QE state collections, but it only triggers when `encryptedFields` options are detected on the client. Updated the code to pass the `encryptedFields` option from the `initialData` config to `MongoDatabase.DropCollection` when available. This ensures QE state collections are properly dropped along with the main collection.

PR #1764, opened 2025-08-27
Upgraded Snappier to 1.2.0 to let license compliance checks find its license

ffissore:
At the moment, the driver depends on [Snappier 1.0.0](https://www.nuget.org/packages/Snappier/1.0.0) which is probably badly annotated and doesn't provide a license identifier (though it provides a link to the license text)

This upgrade is to support automated license checks, which are required to ensure legal compliance.


ffissore:
Superseded by #1989

PR #1768, opened 2025-09-02
CSHARP-5664: Remove allocation in BsonClassMap.GetDiscriminatorConvention

BorisDog:


src/MongoDB.Bson/Serialization/BsonClassMap.cs
```diff
@@ -1329,21 +1329,26 @@ internal IDiscriminatorConvention GetDiscriminatorConvention()
...
-                            $"The discriminator element name cannot be {discriminatorConvention.ElementName} " +
-                            $"because it is already being used by the {fieldOrProperty} {conflictingMemberMap.MemberName} of type {_classType.FullName}");
-                    }
+                    EnsureNoMemberMapConflicts(discriminatorConvention.ElementName);
                 }
             }
 
             return discriminatorConvention;
 
+            void EnsureNoMemberMapConflicts(string elementName)
```
sanych-sun, line 1338: Should we add a comment why this was extracted into a local method?

BorisDog: I thought about it. I am not sure, if taking that path, we might end up adding a comment in every "what if it was written in a other specific way" scenario, which is endless.
Another alternative would be to rewrite the code, so it wouldn't be possible to capture `discriminatorConvention`, but it's an overkill imo.

sanych-sun (approved):
LGTM

PR #1770, opened 2025-09-03
CSHARP-5625: Support text indexes for explicit encryption

adelinowona:
Added TextPreview support for Prefix/Suffix/Substring Indexes

### DIFF
--- a/src/MongoDB.Driver.Encryption/EncryptOptions.cs
+++ b/src/MongoDB.Driver.Encryption/EncryptOptions.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Linq;
 using MongoDB.Bson;
 using MongoDB.Driver.Core.Misc;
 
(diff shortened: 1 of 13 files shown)


src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -91,6 +92,220 @@ public RangeOptions(
...
+    /// <summary>
+    /// Prefix options.
+    /// </summary>
+    /// <remarks>
+    /// PrefixOptions is used with TextOptions and provides further options to support "prefixPreview" queries.
+    /// PrefixOptions is currently unstable API and subject to backwards breaking changes.
+    /// </remarks>
+    public sealed class PrefixOptions
+    {
+        private readonly int _strMaxQueryLength;
```
BorisDog, line 104: maybe auto props instead?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -91,6 +92,220 @@ public RangeOptions(
...
+        /// Querying with a shorter string will error.
+        /// </remarks>
+        public int StrMinQueryLength => _strMinQueryLength;
+    }
+
+    /// <summary>
+    /// Text options.
+    /// </summary>
+    /// <remarks>
+    /// The TextPreview algorithm is experimental only. It is not intended for public use.
```
BorisDog: "It is not intended for public use." is this accurate? We do expose this for experimental use, so maybe just the usual disclaimer:
"... is currently unstable API and subject to backwards breaking changes..."

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -91,6 +92,220 @@ public RangeOptions(
...
+    /// TextOptions specifies options for a Queryable Encryption field supporting "textPreview" queries.
+    /// TextOptions only applies when algorithm is "textPreview".
```
BorisDog: minor: will "encryption algorithm" be more clear phrasing?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -257,14 +547,63 @@ public EncryptOptions With(
...
+        /// <param name="queryType">The query type.</param>
+        /// <returns>A new EncryptOptions instance.</returns>
+        public EncryptOptions With(
+            TextOptions textOptions,
+            Optional<string> algorithm = default,
+            Optional<string> alternateKeyName = default,
+            Optional<Guid?> keyId = default,
+            Optional<long?> contentionFactor = default,
+            Optional<string> queryType = default
+            )
```
BorisDog: minor: move the closing bracket to the end of the previous line.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -257,14 +547,63 @@ public EncryptOptions With(
...
+            Ensure.That(!(_queryType != null && (_algorithm != EncryptionAlgorithm.Indexed.ToString() && _algorithm != EncryptionAlgorithm.Range.ToString() && _algorithm != EncryptionAlgorithm.TextPreview.ToString())), "QueryType only applies for Indexed, Range, or TextPreview algorithm.");
             Ensure.That(!(_rangeOptions != null && _algorithm != EncryptionAlgorithm.Range.ToString()), "RangeOptions only applies for Range algorithm.");
+            Ensure.That(!(_textOptions != null && _algorithm != EncryptionAlgorithm.TextPreview.ToString()), "TextOptions only applies for TextPreview algorithm.");
+
+            if (_algorithm == EncryptionAlgorithm.TextPreview.ToString() && _queryType != null)
+            {
+                Ensure.That(
+                    ValidTextQueryTypes.Contains(_queryType),
+                    $"QueryType '{_queryType}' is not valid for TextPreview algorithm. Use: prefixPreview, suffixPreview, or substringPreview.");
```
BorisDog: nit for your consideration:
`$"QueryType '{_queryType}' is not valid for TextPreview algorithm. Use: {string.join(',', ValidTextQueryTypes)."`

src/MongoDB.Driver.Encryption/EncryptionOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+using MongoDB.Bson;
+
+namespace MongoDB.Driver.Encryption;
+
+internal static class EncryptionOptionsExtensions
+{
+    public static BsonDocument CreateDocument(this RangeOptions rangeOptions)
```
BorisDog: minor: Here and below
```
public static BsonDocument CreateDocument(...) =>
    new ()
    {
         { "min", rangeOptions.Min, rangeOptions.Min != null },
         ...
    }
```

src/MongoDB.Driver.Encryption/EncryptionOptionsExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using MongoDB.Bson;
+
+namespace MongoDB.Driver.Encryption;
+
+internal static class EncryptionOptionsExtensions
```
BorisDog, line 20: +1

tests/MongoDB.Driver.Tests/Encryption/EncryptOptionsTests.cs
```diff
@@ -69,6 +69,83 @@ public void Constructor_should_fail_when_rangeOptions_and_algorithm_is_not_range
             e.Message.Should().Be("RangeOptions only applies for Range algorithm.");
         }
 
+        [Fact]
+        public void Constructor_should_fail_when_textOptions_is_null()
+        {
+            var exception = Record.Exception(() => new EncryptOptions(algorithm: "test", textOptions: null));
+            exception.Should().BeOfType<ArgumentNullException>();
```
BorisDog: Check the param name as well:
```
exception.Should().BeOfType<ArgumentNullException>().Which.ParamName...
```

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2730,6 +2730,225 @@ PipelineDefinition<BsonDocument, BsonDocument> CreatePipeline(string pipelineJso
...
+                        true,
+                        substringOptions: new SubstringOptions(10, 10, 2))),
+                    valueToEncrypt,
+                    async);
+
+                Insert(substringCollection, async, new BsonDocument { { "_id", 0 }, { "encryptedText", encryptedText } });
+
+                RunTestCase(clientEncryption, prefixSuffixCollection, substringCollection);
+            }
+            return;
```
BorisDog: minor: no need for `return`?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2730,6 +2730,225 @@ PipelineDefinition<BsonDocument, BsonDocument> CreatePipeline(string pipelineJso
...
+
+            BsonDocument CreateFindFilter(string operation, string fieldName, BsonValue encryptedValue)
```
BorisDog: minor: expression body, with `new() {...}`

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2730,6 +2730,225 @@ PipelineDefinition<BsonDocument, BsonDocument> CreatePipeline(string pipelineJso
...
+                                    { "substring", encryptedValue, operation == "$encStrContains" },
+                                    { "suffix", encryptedValue, operation == "$encStrEndsWith" }
+                                }
+                            }
+                        }
+                    }
+                };
+            }
+
+            void DropAndCreateCollection(IMongoClient client, CollectionNamespace collectionNamespace, BsonDocument encryptedFields)
```
BorisDog: Let's place these at the end, despite the non alphabetical order, so the `RunTestCase` is more visible.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -2730,6 +2730,225 @@ PipelineDefinition<BsonDocument, BsonDocument> CreatePipeline(string pipelineJso
...
+
+            BsonDocument CreateFindFilter(string operation, string fieldName, BsonValue encryptedValue)
+            {
+                return new BsonDocument
+                {
+                    {
+                        "$expr", new BsonDocument
+                        {
+                            {
+                                $"{operation}", new BsonDocument
```
BorisDog: @rishitb-mongodb we probably will need builders for $strEndsWith, $strStartsWith, $strContains.
But only when it will be out of preview. WDYT?

BorisDog (changes requested):
Looks good overall, few changes requested

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: I tried to follow the pattern already present in this file. I didn't want to mix auto props pattern with the existing backing field pattern. As a matter of fact, I don't think we have a general guideline in the driver, do we generally prefer auto props pattern vs backing field pattern?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
adelinowona: I was following the rider suggestion to include an explicit return to enhance the clarity of the method control flow for readers which is fair given that the method has a multiple local functions. Do we generally not care about this?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -91,6 +92,220 @@ public RangeOptions(
         public long? Sparsity => _sparsity;
     }
 
+    /// <summary>
+    /// Prefix options.
+    /// </summary>
+    /// <remarks>
+    /// PrefixOptions is used with TextOptions and provides further options to support "prefixPreview" queries.
+    /// PrefixOptions is currently unstable API and subject to backwards breaking changes.
```
sanych-sun, line 100: Should we mark the API with `ExperimentalAttribute` if it is known to be unstable?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -150,6 +368,42 @@ public EncryptOptions(
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EncryptOptions"/> class.
+        /// </summary>
+        /// <param name="algorithm">The encryption algorithm.</param>
+        /// <param name="textOptions">The text options.</param>
+        /// <param name="alternateKeyName">The alternate key name.</param>
+        /// <param name="keyId">The key Id.</param>
+        /// <param name="contentionFactor">The contention factor.</param>
+        /// <param name="queryType">The query type.</param>
+        public EncryptOptions(
```
sanych-sun, line 380: Should we invoke this new constructor from the "old" one on line 345 to reduce code duplication?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -150,6 +368,42 @@ public EncryptOptions(
...
+            string algorithm,
+            TextOptions textOptions,
+            Optional<string> alternateKeyName = default,
+            Optional<long?> contentionFactor = default,
+            Optional<Guid?> keyId = default,
+            Optional<string> queryType = default)
+        {
+            Ensure.IsNotNull(algorithm, nameof(algorithm));
+            Ensure.IsNotNull(textOptions, nameof(textOptions));
```
sanych-sun, line 389: Why do we require not null here? What the harm to have explicit value of `null` here?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
BorisDog: I am not sure the attribute exists in net472. ifdef?

sanych-sun: Yep, ifdef. But at least users of the newer targets could see the warning.

BorisDog: I am not opposed to experimenting with `ExperimentalAttribute`. Let's have a quick discussion in the channel.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
BorisDog: We don't have a general guideline, and we do have mixed styles. I do prefer auto props and recommend for others as well.
I don't see an issue with mixing styles within the same file. Also `EncryptOptions` can be refactored to use autoprops.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
sanych-sun: Oh... it was introduced in net8. We cannot use it yet =(

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
BorisDog: I don't think we've notices that in other places. From a first glance having empty returns feels awkward.
Maybe not in this case, but in general if the method in not very readable, we probably should just refactor it.

@sanych-sun wdyt about empty returns?

BorisDog (approved):
LGTM!

src/MongoDB.Driver.Encryption/EncryptOptions.cs
rstam: I strongly dislike auto-properties for all but the very simplest data only classes, and maybe not even then.

My issue is that when reading a class it is helpful to know what state the class maintains. When you have explicit fields at the beginning of the file you can easily tell what state the class maintains.

Auto-properties are harder to find, and so the state the class maintains is spread out a bit more.

The worst-case scenario is a class that has some explicit fields and some auto-properties, because then the state the class maintains is really hard to discern. For that reason, we should never mix the two.

To me auto properties are just lazy syntactic sugar that usually just make code harder to understand.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: I added this new constructor because modifying the old one would be a breaking change for users. So if a user decides to use this new constructor it means they wanted to use textOptions so it probably shouldn't be `null`. That's why check for it here.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: What does that look like? what are you envisioning here?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: Yeah we can't use it yet

src/MongoDB.Driver.Encryption/EncryptOptions.cs
sanych-sun: The new constructor duplicates code from the old one + 1 additional parameter (that meant to be optional, but we cannot do it really optional).
So instead of copying logic, we could have all code in the "new constructor", and simply call it from the "old one", by providing `null` as the value for the new parameter.

adelinowona: ah I see, will do

sanych-sun (approved):
LGTM

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: I'll leave the changes as is just because I don't like mixing patterns.

PR #1772, opened 2025-09-09
CSHARP-5716: Update cake version to support newer Ubuntu versions

BorisDog:
Older cake version fail to load on newer Ubuntu version, due to https://github.com/cake-build/cake/issues/3931


sanych-sun (approved):
LGTM

PR #1776, opened 2025-09-18
CSHARP-3985: Serialization config objects belong to a serialization domain

rstam:


src/MongoDB.Bson/Serialization/Serializers/BsonDocumentSerializer.cs
```diff
@@ -89,7 +89,7 @@ public bool GetDocumentId(
             if (bsonDocument.TryGetValue("_id", out idBsonValue))
             {
                 id = idBsonValue;
-                idGenerator = BsonSerializer.LookupIdGenerator(id.GetType());
+                idGenerator = BsonSerializationDomain.Default.LookupIdGenerator(id.GetType());
```
papafe, line 92: Shouldn't this be dependent on the actual domain?

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -50,15 +50,44 @@ public static TValue Deserialize<TValue>(this IBsonSerializer<TValue> serializer
             return serializer.Deserialize(context, args);
         }
 
+        internal static IBsonSerializer GetSerializerForBaseType(this IBsonSerializer serializer, Type baseType)
```
papafe, line 53: Is this used?

src/MongoDB.Bson/Serialization/BsonSerializationDomain.cs
```diff
@@ -0,0 +1,907 @@
...
+            defaultSerializationDomain.ConfigureDefaults();
+            __default =  defaultSerializationDomain;
+        }
+
+        /// <summary>
+        /// Gets the default serialization domain.
+        /// </summary>
+        public static IBsonSerializationDomain Default => __default;
+
+        internal static IBsonSerializationDomain Create(string name)
```
papafe, line 51: This is not used, but also I suppose that a completely "empty" domain without defaults is not useful?

papafe (commented):
Overall it looks very good and I also really like that it's a much slimmer PR. 

I like the direction, and I would like to try to make some example domain-construction/building API to explore how a developer would build it. 

I think that it would be a nice API, if we manage to create the domain starting from serializers/conventions/etc instead of viceversa, but I'll explore that.

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
rstam: Not yet...

This is somewhat orthogonal to serialization domains and could be postponed and done later.

The idea is that we should always ask the serializer FIRST for the related base or derived type serializer rather than just looking it up in the registry. That would allow a family of polymorphic serializers to work together properly without using the registry.

src/MongoDB.Bson/Serialization/Serializers/BsonDocumentSerializer.cs
rstam: I think the `IdGenerator` for `BsonValue` types is hard coded and can't be changed. If not, it should be. So the result should be the same no matter which serialization domain we look it up in.

For that matter, we probably don't even need to be looking this up in any serialization domain. This could just be a mapping implemented with a switch statement.

src/MongoDB.Bson/Serialization/BsonSerializationDomain.cs
rstam: Hard to say. If I wanted to create a fully customized serialization domain from scratch I wouldn't want it to be polluted with all the defaults.

papafe: True, I still think it's a very daunting task to create a domain completely from the ground up.

src/MongoDB.Bson/Serialization/Serializers/BsonDocumentSerializer.cs
papafe: I agree.

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
papafe: I see!  Yes, probably we should do it separately, so we don't put more irons on the fire.

src/MongoDB.Bson/Serialization/BsonSerializationContext.cs
```diff
@@ -121,9 +121,10 @@ internal Builder(BsonSerializationContext other, IBsonWriter writer)
                 }
                 else
                 {
+                    var bsonDefaults = BsonSerializationDomain.Default.BsonDefaults;
```
papafe, line 124: How can we inject the domain here so that we don't refer to the static defaults? 

This is a problem in general with the properties of `BsonDefaults`.

src/MongoDB.Bson/IO/BsonBinaryReaderSettings.cs
```diff
@@ -30,7 +31,7 @@ public class BsonBinaryReaderSettings : BsonReaderSettings
         private UTF8Encoding _encoding = Utf8Encodings.Strict;
         private bool _fixOldBinarySubTypeOnInput = true;
         private bool _fixOldDateTimeMaxValueOnInput = true;
-        private int _maxDocumentSize = BsonDefaults.MaxDocumentSize;
+        private int _maxDocumentSize = BsonSerializationDomain.Default.BsonDefaults.MaxDocumentSize;
```
papafe, line 34: How can we inject the domain here?

rstam: I think we should not bother to put `BsonDefaults` into the domain.

The `MaxDocumentSize` shouldn't be coming from `BsonDefaults` anyway, in actual use it is determined by the max document size the server tells us to use.

papafe: We got 4 properties in `BsonDefaults` at the moment: `DynamicArraySerializer`, `DynamicDocumentSerializer`, `MaxDocumentSize` and `MaxSerializationDepth`. 
Do you think they should all be removed from the defaults? 
Giving a brief look it seems that `DynamicArraySerializer` and `DynamicDocumentSerializer` are quite circumstantial and used only by `ObjectSerializer` and `ExpandoObjectSerializer`. Maybe they should be configuration options on those serializers instead?
`MaxDocumentSize` you already said it should be removed from the defaults, what about `MaxSerializationDepth`?

rstam: I think `MaxSerializationDepth` can be left global. It doesn't seem like it would need to vary by serialization domain.

I wouldn't remove it entirely though. We need a limit to prevent infinite recursion if someone attempts to serialize a circular structure.

papafe: I agree we can't simply remove `MaxSerializationDepth`, not sure we should leave it as global, but we can decide later about it. 
Regarding the two `DynamicArraySerializer/ DynamicDocumentSerializer`, do you also agree we could move them to the respective serializers where they are needed?

rstam: I'm actually unclear how these properties are used... but it does look like they could be configured in either the `ObjectSerializer` or the `ExpandoObjectSerializer`.

src/MongoDB.Bson/Serialization/BsonSerializationDomain.cs
rstam: A domain that is fully pre-configured is probably the most common use case, making it easy to add just a few tweaks on top of the defaults.

But if you want to completely replace the defaults you would want to start with an empty domain.

PR #1781, opened 2025-09-22
CSHARP-5742: Do not run netstandard tests on ARM platform

sanych-sun:
build.cake makes us to switch to "net6.0" target for ARM platforms. What means test run with $FRAMEWORK set to "net6.0" and "netstandard2.1" are actually duplicates and both run using "net6.0" target framework.
This PR conditionally removes `netstandard` task from ARM platform.
(it removes like a 30 duplicated test variants)


adelinowona (approved):
LGTM

PR #1782, opened 2025-09-30
CSHARP-5705: Use standard RID paths in MongoDB.Driver.Encryption packaging

adelinowona:
- Local testing: Verified that when an application consuming the encryption package is published for a specific platform, the correct native library is automatically copied to the publish output directory
- Cross-platform validation: Used Docker containers to test applications on different target platforms, confirming they can successfully locate and load the appropriate native libraries
- Ensured no NETSDK1152 errors occur during builds targeting specific Runtime Identifiers (RIDs)

### DIFF
--- a/src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
+++ b/src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
@@ -103,42 +103,42 @@
 
   <ItemGroup>
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/osx/native/libmongocrypt.dylib">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Link>runtimes\osx\native\libmongocrypt.dylib</Link>
       <Pack>true</Pack>
       <PackagePath>runtimes\osx\native</PackagePath>
     </Content>
 
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/linux-x64/native/libmongocrypt.so">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Link>runtimes\linux-x64\native\libmongocrypt.so</Link>
       <Pack>true</Pack>
       <PackagePath>runtimes\linux-x64\native</PackagePath>
     </Content>
 
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/linux-arm64/native/libmongocrypt.so">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Link>runtimes\linux-arm64\native\libmongocrypt.so</Link>
       <Pack>true</Pack>
       <PackagePath>runtimes\linux-arm64\native</PackagePath>
     </Content>
 
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/linux-musl-arm64/native/libmongocrypt.so">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Link>runtimes\linux-musl-arm64\native\libmongocrypt.so</Link>
       <Pack>true</Pack>
       <PackagePath>runtimes\linux-musl-arm64\native</PackagePath>
     </Content>
 
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/linux-musl-x64/native/libmongocrypt.so">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Link>runtimes\linux-musl-x64\native\libmongocrypt.so</Link>
       <Pack>true</Pack>
       <PackagePath>runtimes\linux-musl-x64\native</PackagePath>
     </Content>
 
     <Content Include="$(MSBuildProjectDirectory)/$(LibMongoCryptVersionPath)/runtimes/win-x64/native/mongocrypt.dll">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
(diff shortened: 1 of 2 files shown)


src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
```diff
@@ -31,47 +31,60 @@
...
     </PropertyGroup>
 
     <MSBuild Projects ="$(MSBuildProjectFullPath)"
              Properties="TargetFramework=once;LibMongoCryptSourceUrl=$(LibMongoCryptSourceUrl);LibMongoCryptSourcePath=$(LibMongoCryptSourcePath);LibMongoCryptPackagePath=$(LibMongoCryptPackagePath)"
              Targets="DownloadNativeBinary" />
   </Target>
 
-  <Target Name="DownloadNativeBinaries_Alpine" BeforeTargets="BeforeBuild" Condition="!Exists('$(MSBuildProjectDirectory)/runtimes/linux/native/alpine/libmongocrypt.so')">
+
+  <Target Name="DownloadNativeBinaries_AlpineAMD64" BeforeTargets="BeforeBuild" Condition="!Exists('$(MSBuildProjectDirectory)/runtimes/linux-musl-x64/native/libmongocrypt.so')">
```
adelinowona: Added explicit `linux-musl-x64` RID mapping to ensure correct native library selection

During local testing, I discovered that MSBuild's RID fallback mechanism was causing incorrect library resolution. According to the [RID compatibility graph](https://github.com/dotnet/runtime/blob/a2f46fc059ce2ce06a3f36c4d63d965163be2200/src/libraries/Microsoft.NETCore.Platforms/src/PortableRuntimeIdentifierGraph.json#L264-L267), `linux-musl-x64` is compatible with `linux-x64`, which caused MSBuild to use the `linux-x64` native library as a fallback for `linux-musl-x64` targets.

Since we provide a dedicated native library built specifically for `linux-musl-x64`, this change ensures that the platform-specific library is selected instead of relying on the fallback mechanism.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
```diff
@@ -91,33 +104,39 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Pack>true</Pack>
-      <PackagePath>runtimes\linux\native\x64</PackagePath>
+      <PackagePath>runtimes\linux-arm64\native</PackagePath>
     </Content>
 
-    <Content Include="$(MSBuildProjectDirectory)/runtimes/linux/native/arm64/libmongocrypt.so">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    <Content Include="$(MSBuildProjectDirectory)/runtimes/linux-musl-arm64/native/libmongocrypt.so">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
adelinowona: Switched to `PreserveNewest` as it follows MSBuild best practices by only copying files when they've been modified, rather than copying on every build. While this should theoretically improve build performance, the impact will be negligible given our small number of native libraries. Made the change for consistency with recommended practices.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
```diff
@@ -91,33 +104,39 @@
 
   <ItemGroup>
     <Content Include="$(MSBuildProjectDirectory)/runtimes/osx/native/libmongocrypt.dylib">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Pack>true</Pack>
       <PackagePath>runtimes\osx\native</PackagePath>
```
adelinowona, line 109: The macOS paths remain unchanged because our packaged library is a universal binary that supports both `arm64` and `x64` architectures. According to MSBuild's RID compatibility fallback graph, both `osx-x64` and `osx-arm64` fall back to `osx`, so the current structure already handles both platforms correctly without requiring separate RID-specific paths.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.targets
```diff
@@ -1,8 +1,8 @@
...
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <ItemGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
-        <Content Include="$(MSBuildThisFileDirectory)../runtimes/win/native/mongocrypt.dll">
-            <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true' AND $([MSBuild]::IsOsPlatform('Windows'))">
+        <Content Include="$(MSBuildThisFileDirectory)../runtimes/win-x64/native/mongocrypt.dll">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
             <Link>mongocrypt.dll</Link>
         </Content>
     </ItemGroup>
```
adelinowona, line 8: From my understanding this target file provides backward compatibility for older, non-SDK-style projects (e.g., .NET Framework with packages.config) that do not automatically handle native assets from the 'runtimes' folder. 

The condition ensures this logic ONLY runs for non-SDK-style projects. Modern .NET SDK projects (which set '`UsingMicrosoftNETSdk`' to '`true`') will handle native asset copying automatically based on the RID.

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
BorisDog: This might be problematic is some weird corner cases, and given that it's not a huge toll on our build I'd suggest leaving 'Always' just to be on the safe side. 

cc @sanych-sun what do you think?

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.targets
adelinowona: I have reverted part of this change. I removed the condition on `UsingMicrosoftNETSdk` which allows the target file to also run for SDK style projects on windows targeting .Net framework. This was motivated by smoketests failing on windows variants and a Microsoft recommending using the targets file for supporting SDK style AnyCPU projects targeting .Net Framework -> https://learn.microsoft.com/en-us/nuget/create-packages/native-files-in-net-packages#projects-targeting-net-framework

src/MongoDB.Driver.Encryption/MongoDB.Driver.Encryption.csproj
sanych-sun: I'm OK with any setting here. As on EG, while the packages creation, we will always start from scratch and there are no difference.

sanych-sun (dismissed):
LGTM

BorisDog (approved):
LGTM

PR #1783, opened 2025-09-30
CSHARP-5737: Add legacy representation for TimeOnly

papafe:


src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -145,6 +172,20 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
                     bsonWriter.WriteString(value.ToString("o"));
                     break;
 
+                case BsonType.Document:
+                    bsonWriter.WriteStartDocument();
+                    bsonWriter.WriteInt32("Hour", value.Hour);
+                    bsonWriter.WriteInt32("Minute", value.Minute);
+                    bsonWriter.WriteInt32("Second", value.Second);
+                    bsonWriter.WriteInt32("Millisecond", value.Millisecond);
+#if NET7_0_OR_GREATER
```
papafe: `Microsecond` and `Nanosecond` were introduced in .NET 7.0

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -75,6 +90,17 @@ public TimeOnlySerializer(BsonType representation, TimeOnlyUnits units)
 
             _representation = representation;
             _units = units;
+
+            _helper = new SerializerHelper
+            (
+                new SerializerHelper.Member("Hour", Flags.Hour, isOptional: true),
```
rstam: It's a little weird that these are optional.

In understand that we ignore them during deserialization (we only use `Ticks`). But that doesn't necessarily mean they are optional.

We also might want to validate during deserialization that the `Hour`, `Minutes`, `Second` etc.. fields are consistent with the `Ticks` value. If WE wrote the values they should be consistent, but if the data came from outside the C# driver they might not be.

src/MongoDB.Bson/Serialization/Attributes/BsonTimeOnlyOptionsAttribute.cs
```diff
@@ -44,7 +44,7 @@ public BsonTimeOnlyOptionsAttribute(BsonType representation)
         /// Initializes a new instance of the BsonTimeOnlyOptionsAttribute class.
         /// </summary>
         /// <param name="representation">The external representation.</param>
-        /// <param name="units">The TimeOnlyUnits.</param>
+        /// <param name="units">The TimeOnlyUnits. Ignored if representation is BsonsType.Document.</param>
```
rstam: Replace `BsonsType.Document` with `BsonType.Document` (fix typo)

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -67,6 +81,7 @@ public TimeOnlySerializer(BsonType representation, TimeOnlyUnits units)
                 case BsonType.Int32:
                 case BsonType.Int64:
                 case BsonType.String:
+                case BsonType.Document:
```
rstam: Put `case BsonType.Document` first (keep in alphabetical order).

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
rstam: Consider the following change:

```
_helper = new SerializerHelper
(
    new SerializerHelper.Member("Hour", Flags.Hour, isOptional: false),
    new SerializerHelper.Member("Minute", Flags.Minute, isOptional: false),
    new SerializerHelper.Member("Second", Flags.Second, isOptional: false),
    new SerializerHelper.Member("Millisecond", Flags.Millisecond, isOptional: false),
    new SerializerHelper.Member("Microsecond", Flags.Microsecond, isOptional: true), // not serialized by older versions of the driver
    new SerializerHelper.Member("Nanosecond", Flags.Nanosecond, isOptional: true), // not serialized by older versions of the driver  
    new SerializerHelper.Member("Ticks", Flags.Ticks, isOptional: false)                                                              
);                                                                                                                                    
```

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -102,6 +128,7 @@ public override TimeOnly Deserialize(BsonDeserializationContext context, BsonDes
                 BsonType.Int64 =>  FromInt64(bsonReader.ReadInt64(), _units),
                 BsonType.Int32 =>  FromInt32(bsonReader.ReadInt32(), _units),
                 BsonType.Double =>  FromDouble(bsonReader.ReadDouble(), _units),
+                BsonType.Document => FromDocument(context),
```
rstam, line 130: Put `BsonType.Document` first (alphabetical order).

While you are at it, fix the order of the other ones:

```
return bsonType switch
{
    BsonType.Document => FromDocument(context),
    BsonType.Double =>  FromDouble(bsonReader.ReadDouble(), _units),
    BsonType.Int32 =>  FromInt32(bsonReader.ReadInt32(), _units),
    BsonType.Int64 =>  FromInt64(bsonReader.ReadInt64(), _units),
    BsonType.String => TimeOnly.ParseExact(bsonReader.ReadString(), "o"),
    _ => throw CreateCannotDeserializeFromBsonTypeException(bsonType)    
};                                                                       
```

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -145,6 +172,20 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
                     bsonWriter.WriteString(value.ToString("o"));
                     break;
 
+                case BsonType.Document:
```
rstam: Put `case BsonType.Document` first in the switch statement (alphabetical order).

rstam: Consider the following refactoring so that we consistently write the `Microsecond` and `Nanosecond` properties even on older versions of .NET:

```
                case BsonType.Document:
#if NET7_0_OR_GREATER
                    var microsecond = value.Microsecond;
                    var nanosecond = value.Nanosecond;
#else
                    var microsecond = ComputeMicrosecondFromTicks(value.Ticks);
                    var nanosecond = ComputeNanosecondFromTicks(value.Ticks);
#endif
                    bsonWriter.WriteStartDocument();
                    bsonWriter.WriteInt32("Hour", value.Hour);
                    bsonWriter.WriteInt32("Minute", value.Minute);
                    bsonWriter.WriteInt32("Second", value.Second);
                    bsonWriter.WriteInt32("Millisecond", value.Millisecond);
                    bsonWriter.WriteInt32("Microsecond", microsecond);
                    bsonWriter.WriteInt32("Nanosecond", nanosecond);
                    bsonWriter.WriteInt64("Ticks", value.Ticks);
                    bsonWriter.WriteEndDocument();
                    break;
```

rstam: The `ComputeMicrosecondFromTicks` and `ComputeNanosecondFromTicks` helper methods can be something like:

```
private int ComputeMicrosecondFromTicks(long ticks)
{
    return (int)(ticks / (TimeSpan.TicksPerMillisecond * 1_000)) % 1000;
}
 
private int ComputeNanosecondFromTicks(long ticks)
{
    return (int)(ticks / (TimeSpan.TicksPerMillisecond * 1_000_000)) % 1000;
}                                                                           
```

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +237,29 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
...
+            {
+                switch (flag)
+                {
+                    case Flags.Hour:
+                    case Flags.Minute:
+                    case Flags.Second:
+                    case Flags.Millisecond:
+                    case Flags.Microsecond:
+                    case Flags.Nanosecond:
+                        bsonReader.SkipValue();  break; // ignore value (use Ticks instead)
```
rstam: Do we really want to ignore these values?

Should we check that they are in agreement with the `Ticks` value?

rstam (changes requested):
Overall looks good. Let me know what you think of the comments.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Regarding the use of the preprocessor directives, that was my first implementation but I thought that given that the implementation is the same, we would not need to necessarily make a differentiation.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: I agree, that probably those should not be optional, I'll answer the question about validation in the other comment.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: This is similar to what we do with `DateTimeSerializer` and `DateOnlySerializer`, where we ignore the `DateTime` value when the representation is `BsonType.Document` and use only the ticks. 
The difference is that here we have many more fields that we ignoring. I was trying to be the most permissive here, but I understand if want to be more strict.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Added validation and fixed optionality.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: I've added validation here.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -75,6 +90,20 @@ public TimeOnlySerializer(BsonType representation, TimeOnlyUnits units)
 
             _representation = representation;
             _units = units;
+
+            _helper = new SerializerHelper
+            (
+                //TimeOnlySerializer was introduced in version 3.0.0 of the driver. Prior to that, TimeOnly was serialized
+                //as a POCO. Due to that, Microsecond and Nanosecond could be missing, as they were introduced in .NET 7.
+                //To avoid deserialization issues, we treat Microsecond and Nanosecond as optional members.
```
rstam: Nits: we usually put a space after `//` in comments.

Also, instead of "was serialized as a POCO" say "was serialized as a class mapped POCO".

    // TimeOnlySerializer was introduced in version 3.0.0 of the driver. Prior to that, TimeOnly was serialized
    // as a class mapped POCO. Due to that, Microsecond and Nanosecond could be missing, as they were introduced in .NET 7.
    // To avoid deserialization issues, we treat Microsecond and Nanosecond as optional members.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -145,6 +188,7 @@ public override void Serialize(BsonSerializationContext context, BsonSerializati
                     bsonWriter.WriteString(value.ToString("o"));
                     break;
 
+
```
rstam: Remove extra blank line.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +240,59 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
                 : new TimeOnly(value * TicksPerUnit(units));
         }
 
+        private TimeOnly FromDocument(BsonDeserializationContext context)
```
rstam, line 221: Put in alphabetical order.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -231,6 +328,18 @@ private long ToInt64(TimeOnly timeOnly, TimeOnlyUnits units)
                 : timeOnly.Ticks / TicksPerUnit(units);
         }
 
+        private int GetNanosecondsComponent(long ticks)
```
rstam: Put in alphabetical order.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -231,6 +328,18 @@ private long ToInt64(TimeOnly timeOnly, TimeOnlyUnits units)
...
+        {
+            // ticks % 10 * 100
+            return (int)(ticks % TicksPerUnit(TimeOnlyUnits.Microseconds) * 100);
```
rstam: I'm not sure that computation is correct. I suggest:

    var nanosecondsPerTick = 100;
    return (int)(ticks * nanosecondsPerTick % 1000);

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -231,6 +328,18 @@ private long ToInt64(TimeOnly timeOnly, TimeOnlyUnits units)
...
+        }
+
+        private int GetMicrosecondsComponent(long ticks)
+        {
+            // ticks / 10 % 1000
+            return (int)(ticks / TicksPerUnit(TimeOnlyUnits.Microseconds) % 1000);
```
rstam: Let's introduce an intermediate variable to make the computation easier to understand:

    var ticksPerMicrosecond = TimeSpan.TicksPerMillisecond / 1000;
    return (int)(ticks / ticksPerMicrosecond % 1000);

rstam (changes requested):
Some minor changes requested.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Deal.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: I'll add an intermediate variable, but I'll keep the call to `TicksPerUnit` so we don't have code duplication.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: The computation is correct, it takes the last digit of the ticks and multiplies it by 100, but I can change it to make it clearer.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -175,6 +218,61 @@ public TimeOnlySerializer WithRepresentation(BsonType representation, TimeOnlyUn
         }
 
         // private methods
+
```
rstam: Did you mean to add a blank line here?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +294,20 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
                 : new TimeOnly(value * TicksPerUnit(units));
         }
 
+        private int GetNanosecondsComponent(long ticks)
```
rstam: Put in alphabetical order?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +294,20 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
...
+        {
+            // ticks * 100 % 1000
```
rstam: This comment isn't very helpful. That's exactly what the code says.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +294,20 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
...
+            var nanosecondsPerTick = 100;
+            return (int)(ticks * nanosecondsPerTick % 1000);
+        }
+
+        private int GetMicrosecondsComponent(long ticks)
+        {
+            // ticks / 10 % 1000
```
rstam: This comment isn't very helpful. That's exactly what the code says.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -196,6 +294,20 @@ private TimeOnly FromInt64(long value, TimeOnlyUnits units)
...
+            var ticksPerMicrosecond = TicksPerUnit(TimeOnlyUnits.Microseconds);
```
rstam, line 298: It's fine to the `TicksPerUnit` method if you want. My thought was to eliminate a method call for what was essentially a constant.

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
```diff
@@ -231,6 +343,7 @@ private long ToInt64(TimeOnly timeOnly, TimeOnlyUnits units)
                 : timeOnly.Ticks / TicksPerUnit(units);
         }
 
+
```
rstam: Did you mean to add a blank line here?

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: 🤦

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Nope, it was hidden by the comments on my IDE 😄

src/MongoDB.Bson/Serialization/Serializers/TimeOnlySerializer.cs
papafe: Nope

PR #1859, opened 2026-01-21
CSHARP-5723: Add exponential backoff to operation retry loop for server overloaded errors

papafe:


src/MongoDB.Driver/Core/Operations/ClientBulkWriteOperation.cs
```diff
@@ -286,7 +286,9 @@ private IAsyncCursor<BsonDocument> GetIndividualResultsCursor(RetryableWriteCont
                 0,
                 0,
                 BsonDocumentSerializer.Instance,
-                MessageEncoderSettings);
+                MessageEncoderSettings,
+                maxTime: null,
+                canBeRetried: RetryRequested); //TODO This is a little strange, but AsyncCursor do read operation, but this a write operation. Do we keep it like this...? It should be CanBeRetried anyways, once we add it to write operations.
```
sanych-sun: Accordingly to the spec (before Backpressure) whenever we run into an error while reading the result cursor - we have to restart the entire bulk write if it's retriable. With Backpressure we should retry read only if SystemOverloaded label is present.
https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#retrying-getmores

sanych-sun: Please note there is additional check in `GetEffectiveRetryRequested` that validates if the bulk operation could be really retriable.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -534,7 +536,8 @@ private ListDatabasesOperation CreateListDatabasesOperation(ListDatabasesOptions
                 Comment = options.Comment,
                 Filter = options.Filter?.Render(new(BsonDocumentSerializer.Instance, BsonSerializer.SerializerRegistry, translationOptions: translationOptions)),
                 NameOnly = options.NameOnly,
-                RetryRequested = _settings.RetryReads
+                RetryRequested = _settings.RetryReads,
+                CanBeRetried = _settings.RetryReads
```
sanych-sun: What the difference between `RetryRequested` and `CanBeRetried`?

src/MongoDB.Driver/Core/Clusters/Cluster.cs
```diff
@@ -70,31 +72,25 @@ protected Cluster(ClusterSettings settings, IClusterableServerFactory serverFact
             _serverSessionPool = new CoreServerSessionPool(this);
             _clusterEventLogger = loggerFactory.CreateEventLogger<LogCategories.SDAM>(eventSubscriber);
             _serverSelectionEventLogger = loggerFactory.CreateEventLogger<LogCategories.ServerSelection>(eventSubscriber);
+            _maxAdaptiveRetries = maxAdaptiveRetries;
+            _enableOverloadRetargeting = enableOverloadRetargeting;
```
sanych-sun: I'm confused: why do we need this settings in both in cluster and in Operations. It looks like cluster do not really use this settings.

src/MongoDB.Driver/ClusterKey.cs
```diff
@@ -155,13 +161,15 @@ public ClusterKey(
...
         public bool DirectConnection { get { return _directConnection; } }
+        public bool EnableOverloadRetargeting => _enableOverloadRetargeting;
         public TimeSpan HeartbeatInterval { get { return _heartbeatInterval; } }
         public TimeSpan HeartbeatTimeout { get { return _heartbeatTimeout; } }
         public bool IPv6 { get { return _ipv6; } }
         public LibraryInfo LibraryInfo { get { return _libraryInfo; } }
         public bool LoadBalanced => _loadBalanced;
         public TimeSpan LocalThreshold { get { return _localThreshold; } }
         public LoggingSettings LoggingSettings { get { return _loggingSettings; } }
+        public int MaxAdaptiveRetries => _maxAdaptiveRetries;
```
sanych-sun: I'm not really sure if `_enableOverloadRetargeting` and `_maxAdaptiveRetries` should be a part of ClusterKey, because both settings done not affect cluster or connection pool behavior. They affects operations retriability only.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -40,17 +43,35 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
                     {
-                        throw originalException ?? ex;
+                        throw originalException;
                     }
 
-                    originalException ??= ex;
-                    deprioritizedServers = UpdateServerList(server, deprioritizedServers, ex);
+                    // We bail early if the backoff would exceed the CSOT deadline.
+                    var remaining = operationContext.RemainingTimeout;
+                    if (remaining != Timeout.InfiniteTimeSpan && operationContext.IsRootContextTimeoutConfigured() && remaining < backoff)
```
sanych-sun: No needs to check for root context here, as long as RemainingTimeout == Infinite or more then `backoff` - we are OK.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -40,17 +43,35 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
+                        context.Binding.Session.CurrentTransaction.HasCompletedCommand = true;
+                    }
+
+                    return operationResult;
                 }
                 catch (Exception ex)
                 {
-                    if (!ShouldRetryOperation(operationContext, context, ex, totalAttempts))
+                    originalException ??= ex;
+                    overloadErrorSeen |= RetryabilityHelper.IsSystemOverloadedException(ex);
```
sanych-sun: I think we need ||= here instead.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -72,43 +94,96 @@ public static async Task<TResult> ExecuteAsync<TResult>(OperationContext operati
...
```
sanych-sun: `||=`

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -50,27 +54,55 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
                 }
                 catch (Exception ex)
                 {
-                    if (!ShouldRetryOperation(operationContext, context.Channel is null, operation.WriteConcern, context, server, ex, totalAttempts))
+                    if (originalException == null || WritesWereAttempted(ex, context.Channel is not null))
+                    {
+                        originalException = ex;
+                    }
+
+                    overloadErrorSeen |= RetryabilityHelper.IsSystemOverloadedException(ex);
```
sanych-sun: `||=`

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -50,27 +54,55 @@ public static TResult Execute<TResult>(OperationContext operationContext, IRetry
...
+                    }
+
+                    if (operation is EndTransactionOperation endTransactionOperation)
+                    {
+                        endTransactionOperation.OnRetry(context, ex);
+                    }
+
+                    // We bail early if the backoff would exceed the CSOT deadline.
+                    var remaining = operationContext.RemainingTimeout;
+                    if (remaining != Timeout.InfiniteTimeSpan && operationContext.IsRootContextTimeoutConfigured() && remaining < backoff)
```
sanych-sun: No needs to check for root context.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -79,80 +111,144 @@ public static async Task<TResult> ExecuteAsync<TResult>(OperationContext operati
...
-                    if (!ShouldRetryOperation(operationContext, !channelAcquisitionSuccessful, operation.WriteConcern, context, server, ex, totalAttempts))
+                    if (originalException == null || WritesWereAttempted(ex, context.Channel is not null))
                     {
-                        throw originalException ?? ex;
+                        originalException = ex;
                     }
 
-                    originalException ??= ex;
-                    deprioritizedServers = UpdateServerList(server, deprioritizedServers, ex);
+                    overloadErrorSeen |= RetryabilityHelper.IsSystemOverloadedException(ex);
```
sanych-sun: `||=`

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -79,80 +111,144 @@ public static async Task<TResult> ExecuteAsync<TResult>(OperationContext operati
...
+
+                    await Task.Delay(backoff, operationContext.CancellationToken).ConfigureAwait(false);
+                    deprioritizedServers = UpdateServerList(server, deprioritizedServers, ex, context.EnableOverloadRetargeting);
                 }
             }
         }
 
         // private static methods
-        private static bool ShouldRetryOperation(OperationContext operationContext, bool errorDuringChannelAcquisition, WriteConcern writeConcern, RetryableWriteContext context, ServerDescription server, Exception exception, int totalAttempts)
+        private static bool ShouldRetry(OperationContext operationContext,
```
sanych-sun, line 169: _Minor_: Each parameter on new line.

src/MongoDB.Driver/Core/Operations/RetryableUpdateCommandOperation.cs
```diff
@@ -105,7 +105,7 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             var maxBatchCount = Math.Min(MaxBatchCount ?? int.MaxValue, channel.ConnectionDescription.MaxBatchCount);
             var maxDocumentSize = channel.ConnectionDescription.MaxWireDocumentSize;
-            var payload = new Type1CommandMessageSection<UpdateRequest>("updates", _updates, UpdateRequestSerializer.Instance, NoOpElementNameValidator.Instance, maxBatchCount, maxDocumentSize);
+            var payload = new Type1CommandMessageSection<UpdateRequest>("updates", updates, UpdateRequestSerializer.Instance, NoOpElementNameValidator.Instance, maxBatchCount, maxDocumentSize);
```
sanych-sun: I thought we decided to rollback all fixes to bulkWrite and create a separate ticket to investigate and fix them.

src/MongoDB.Driver/Core/Operations/BulkMixedWriteOperation.cs
```diff
@@ -136,6 +150,8 @@ public bool RetryRequested
             set { _retryRequested = value; }
         }
 
+        public bool IsOperationRetryable => _requests.All(r => r.IsRetryable());
```
sanych-sun: We probably better to calculate the value once, otherwise we will iterate over the Requests on each usage of the property.

src/MongoDB.Driver/Core/Operations/BulkUnmixedWriteOperationBase.cs
```diff
@@ -114,6 +128,8 @@ public bool RetryRequested
             set { _retryRequested = value; }
         }
 
+        public bool IsOperationRetryable => _requests.All(r => r.IsRetryable());
```
sanych-sun: We probably better to calculate the value once, otherwise we will iterate over the Requests on each usage of the property.

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
```diff
@@ -30,7 +30,8 @@ namespace MongoDB.Driver.Core.Operations
     /// </summary>
     /// <typeparam name="TDocument">The type of the output documents.</typeparam>
     /// <seealso cref="MongoDB.Driver.IAsyncCursor{TOutput}" />
-    internal sealed class ChangeStreamCursor<TDocument> : IChangeStreamCursor<TDocument>
+    internal sealed class
```
sanych-sun: _Minor_: looks like unnecessary change.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
```diff
@@ -57,7 +73,10 @@ public ServerDescription SelectServer(OperationContext operationContext, IReadOn
             try
             {
                 operationContext.ThrowIfTimedOutOrCanceled();
-                var writeChannelSource = Binding.GetWriteChannelSource(operationContext, deprioritizedServers);
+                //TODO The implementation of the two overloads is different, need to understand if it's an issue and we can just call the second overload from the first one.
```
sanych-sun: Do we still need to investigate this? Is there any problem with this? SHould we create a ticket?

tests/MongoDB.Driver.Tests/Specifications/client-backpressure/prose-tests/ClientBackpressureProseTests.cs
```diff
@@ -0,0 +1,284 @@
...
+
+namespace MongoDB.Driver.Tests.Specifications.client_backpressure.prose_tests;
+
+// Test 1 requires controlling the RNG for precise timing assertions, so it uses unit tests with mocks.
+public class ClientBackpressureProseTestsUnit
+{
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    // https://github.com/mongodb/specifications/blob/master/source/client-backpressure/tests/README.md#test-1-operation-retry-uses-exponential-backoff
```
sanych-sun: It's better to use perma-link instead.

tests/MongoDB.Driver.Tests/Specifications/mongodb-handshake/prose-tests/MongoDbHandshakeProseTests.cs
```diff
@@ -99,6 +104,32 @@ public async Task DriverAcceptsArbitraryAuthMechanism([Values(false, true)] bool
 
             subject._state().Should().Be(3); // 3 - open.
         }
+
+        [Fact]
+        // https://github.com/baileympearson/specifications/blob/530e727dd5cc0d0eb2606ea6db1cf144968597e7/source/mongodb-handshake/tests/README.md#test-9-handshake-documents-include-backpressure-true
+        public async Task HandshakeDocumentsIncludeBackpressureTrue()
+        {
+            //TODO A test
```
sanych-sun: ?

tests/MongoDB.Driver.Tests/Specifications/mongodb-handshake/prose-tests/MongoDbHandshakeProseTests.cs
```diff
@@ -99,6 +104,32 @@ public async Task DriverAcceptsArbitraryAuthMechanism([Values(false, true)] bool
...
+            using var client = DriverTestConfiguration.CreateMongoClient(eventCapturer);
+
+            var database = client.GetDatabase("admin");
+            await database.RunCommandAsync<BsonDocument>(new BsonDocument("ping", 1));
+
+            var commandStartedEvents = eventCapturer.Events.OfType<CommandStartedEvent>().ToList();
+            commandStartedEvents.Should().NotBeEmpty();
+            foreach (var doc in commandStartedEvents.Select(ev => ev.Command))
+            {
+                //TODO The messages have been added for tesing purposes only. Remove them later.
```
sanych-sun: It's probably already "later" :-)

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
```diff
@@ -317,6 +319,153 @@ await collection.FindAsync(Builders<BsonDocument>.Filter.Empty) :
...
+            var eventCapturer = new EventCapturer().CaptureCommandEvents("find");
+            using var client = DriverTestConfiguration.CreateMongoClient(s =>
+            {
+                s.RetryReads = true;
+                s.ClusterConfigurator = b =>
+                {
+                    b.Subscribe(eventCapturer);
+                    b.Subscribe<CommandFailedEvent>(e =>
+                    {
+                        if (e is { CommandName: "find", Failure: MongoCommandException { Code: 91 } } && !secondFailPointConfigured)
```
sanych-sun, line 368: How come this test is not flaky? I'm not sure I understand why do we have such requirement to setup a failpoint inside of event handler. It's definitely outside of this PR scope, but we might want to try use skip parameter of FailPoint instead.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
papafe: It does not exist as an operator in c#. I'll put the long hand form so we get the short circuit and maybe it's a little easier to read?

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
papafe: Fixed

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
papafe: As before.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
papafe: As before.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
papafe: As before.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/RetryableUpdateCommandOperation.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/BulkMixedWriteOperation.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/BulkUnmixedWriteOperationBase.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
papafe: Fixed.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
papafe: I don't think it's a problem, it just feels a little strange to me that their implementation is different. It was more a TODO about understanding if there was something fishy here.

tests/MongoDB.Driver.Tests/Specifications/client-backpressure/prose-tests/ClientBackpressureProseTests.cs
papafe: Added.

tests/MongoDB.Driver.Tests/Specifications/mongodb-handshake/prose-tests/MongoDbHandshakeProseTests.cs
papafe: Removed those todos 😬

tests/MongoDB.Driver.Tests/Specifications/mongodb-handshake/prose-tests/MongoDbHandshakeProseTests.cs
papafe: Done.

src/MongoDB.Driver/Core/Clusters/Cluster.cs
papafe: So, it didn't need to be on cluster key, that was unnecessary. But before it was passed to the cluster because we need those values in `CoreSession`, where we start the `abort/commit` operations. I've removed it from those classes, but I added it to `CoreSessionOption`, as the straightest path. Still, not the cleanest.

src/MongoDB.Driver/ClusterKey.cs
papafe: As before.

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
papafe: The spec test requires to "configure a failpoint via command monitoring CommandFailedEvent".
In this case we want to have the error labels change after the first error. What is the skip parameter? I could not find info about it.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
sanych-sun: Hm... I'm hallucinated this operator :-)

tests/MongoDB.Driver.Tests/Specifications/retryable-reads/RetryableReadsProseTests.cs
sanych-sun: For failPoint we can specify `mode: { skip: 2 }`, which means "first and second attempt are lucky, but after that - do the failpoint", this might help us to configure more then one failPoint. But it is definitely out of this PR scope. The test is not flaky probably because we have a synchronous events, but anyhow it looks suspicious.

sanych-sun (approved):
LGTM, but please run the entire test suite on the custom patch before merging. Just in case.

sanych-sun (approved):
LGTM

PR #1882, opened 2026-02-17
CSHARP-5887: Simplify retryable read and writes

papafe:


src/MongoDB.Driver/Core/Operations/ClientBulkWriteOperation.cs
```diff
@@ -110,7 +110,7 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
                     bulkWriteResults.TopLevelException = commandException;
                     serverResponse = commandException.Result;
                 }
-                catch (Exception exception)
+                catch (Exception exception) when (!context.ErrorDuringLastChannelAcquisition)
```
papafe: This avoids an exception in `EnsureCanProceedNextBatch` and `ToFinalResultsOrThrow`, as the `context` does not have the channel anymore, as there was an exception. 
This error was not visible before, as the channel acquisition was done outside of the try catch and it will just raise the exception on the whole method. 
The question here is we want the non-retryable channel acquisition to make the whole method throw (like it was before), or we need to find another way to make  `EnsureCanProceedNextBatch` and `ToFinalResultsOrThrow` work.

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
```diff
@@ -41,11 +41,21 @@ protected CommandOperationBase(
             MessageEncoderSettings messageEncoderSettings)
         {
             _databaseNamespace = Ensure.IsNotNull(databaseNamespace, nameof(databaseNamespace));
-            _command = Ensure.IsNotNull(command, nameof(command));
+            _command = command; //can be null
```
papafe: This is used so that we can modify the command once the `ReadCommandOperation` has been created. 
This is done because some operations need to have the `operationContext` to properly create the command, but now the `operationContext` is not available until later. 
I'm not a fan of this implementation, to be honest. 
We got two other possibilities in my opinion:
1. We make `ReadCommandOperation` equivalent to `WriteCommandOperation`, so it does not concern with retryability. Then we make a new class, `RetryableReadCommandOperation` that has retryability and uses `ReadCommandOperation` inside
2. Instead of having the possibility of setting the command, we can add a method to modify it when we got the operation context.

I think nr1 is more desirable, but requires further much more refactoring.

src/MongoDB.Driver/Core/Operations/RetryableInsertCommandOperation.cs
```diff
@@ -89,7 +89,8 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             else
             {
-                documents = new BatchableSource<TDocument>(_documents.Items, _documents.Offset, _documents.ProcessedCount, canBeSplit: false);
+                documents = new BatchableSource<TDocument>(_documents.Items, _documents.Offset, _documents.Count, _documents.ProcessedCount, canBeSplit: false);
+                //TODO This looked wrong, is it tested somewhere? In the update command it was not even used.
```
papafe: It seems this was not working at all before, do we have tests?

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -92,39 +66,43 @@ public void Dispose()
...
-                }
-                catch (Exception ex) when (RetryableReadOperationExecutor.ShouldConnectionAcquireBeRetried(operationContext, this, ex, attempt))
-                {
-                    attempt++;
-                }
+                _lastAcquiredServer = ChannelSource.ServerDescription;
+                ReplaceChannel(ChannelSource.GetChannel(operationContext));
+
+                ChannelPinningHelper.PinChannellIfRequired(ChannelSource, Channel,
+                    Binding.Session); //TODO We should do it only the first time, as an improvement we could pass the attempt number.
```
papafe: I don't think it's a problem to call this method multiple times, is it?

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -26,92 +26,69 @@ internal static class RetryableReadOperationExecutor
         public static TResult Execute<TResult>(OperationContext operationContext, IRetryableReadOperation<TResult> operation, RetryableReadContext context)
         {
             HashSet<ServerDescription> deprioritizedServers = null;
-            var attempt = 1;
+            var totalAttempts = 0;
```
papafe, line 29: Renamed to be aligned with the write version, and starting from 0 to be more compatible with the future client backpressure integration

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -26,37 +26,9 @@ internal sealed class RetryableReadContext : IDisposable
     {
         #region static
 
-        public static RetryableReadContext Create(OperationContext operationContext, IReadBinding binding, bool retryRequested)
+        public static RetryableReadContext Create(IReadBinding binding, bool retryRequested)
```
sanych-sun: Do we still need this factory method? As far as you've removed the connection resolving logic, I would say there is no reasons to keep it.

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
```diff
@@ -41,11 +41,21 @@ protected CommandOperationBase(
...
             _resultSerializer = Ensure.IsNotNull(resultSerializer, nameof(resultSerializer));
             _messageEncoderSettings = messageEncoderSettings;
         }
 
+        //TODO Move to the right place
+        /// <summary>
+        /// Sets the command to be executed. This is used by derived classes that build commands dynamically.
+        /// </summary>
+        /// <param name="command">The command.</param>
+        protected void SetCommand(BsonDocument command)
```
sanych-sun: I would suggest to remove the `_command` variable and introduce abstract `CreateCommand` method instead, so derived classes could create the command on demand.

src/MongoDB.Driver/Core/Operations/RetryableInsertCommandOperation.cs
sanych-sun: Agree, it looks like our retry was not working properly if an error occurs on non-first batch.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
papafe: I'll remove it.

src/MongoDB.Driver/Core/Operations/DistinctOperation.cs
```diff
@@ -151,14 +151,17 @@ public BsonDocument CreateCommand(OperationContext operationContext, ICoreSessio
             };
         }
 
-        private EventContext.OperationNameDisposer BeginOperation() => EventContext.BeginOperation("distinct");
+        private EventContext.OperationIdDisposer BeginOperation() => EventContext.BeginOperation(null, "distinct");
```
papafe: It is unclear to me the difference between the overloads

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -26,145 +26,124 @@ internal static class RetryableWriteOperationExecutor
...
-                }
-
-                if (!AreRetryableWritesSupported(context.ChannelSource.ServerDescription))
-                {
-                    throw originalException;
-                }
-
-                attempt++;
+                deprioritizedServers ??= [];
+                deprioritizedServers.Add(context.LastAcquiredServer);
```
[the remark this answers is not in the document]
papafe: If channel acquisition fails before `LastAcquiredServer` is set, then we'll get a non-retryable exception, so the method will throw and we'll never reach here, but maybe it's a good idea to add this.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -92,39 +57,39 @@ public void Dispose()
...
-                try
-                {
-                    ReplaceChannel(ChannelSource.GetChannel(operationContext));
-                    return;
-                }
-                catch (Exception ex) when (RetryableReadOperationExecutor.ShouldConnectionAcquireBeRetried(operationContext, this, ex, attempt))
-                {
-                    attempt++;
-                }
+                _lastAcquiredServer = ChannelSource.ServerDescription;
```
sanych-sun: Do we really need to store the last server? As far as I understood we need it only for servers deprioritization, but having it stored could lead to the situation when `_lastAcquiredServer` and `Channel` is out of sync. Even this version of code: If we run into an exception while the connection check out (line 66) - we will have no Channel assigned, but _lastAcquiredServer will have the just selected server, what is probably wrong - because we haven't try any operation against the server yet.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
```diff
@@ -92,41 +59,48 @@ public void Dispose()
...
+                var writeChannelSource = Binding.GetWriteChannelSource(operationContext, deprioritizedServers);
+                ReplaceChannelSource(writeChannelSource);
+                _lastAcquiredServer = ChannelSource.ServerDescription;
+                ReplaceChannel(ChannelSource.GetChannel(operationContext));
+
+                ChannelPinningHelper.PinChannellIfRequired(ChannelSource, Channel, Binding.Session);
+            }
+            catch
+            {
+                _errorDuringLastChannelAcquisition = true;
```
sanych-sun: Why do we need this variable/property?

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
papafe: If we have an error during channel acquisition then we need to check the retryability according to the retryable reads spec, and this variable is used for that check. 
With this variable we can keep one try-catch in `RetryableWriteOperationExecutor.Execute`, without the need to have multiple ones. 
This is also used by `ClientBulkWriteOperation`, and there is a comment about I doubt I have regarding this.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
papafe: I think that maybe the more correct name would be `lastSelectedServer`, I agree. 
We use it for server deprioritization and also for checking if the server allows retries, so we should have it even if we get an exception during connection establishment / channel acquisition on line 66.
Another possibility would be to return a specific POCO from this method, that contains both the exception and the selected server. I tried with an out parameter when I was working with this, but that won't work with the async version.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -84,50 +47,95 @@ public void Dispose()
...
             {
-                _channelSource?.Dispose();
-                _channel?.Dispose();
+                DisposeChannelAndSource();
                 _disposed = true;
             }
         }
 
-        public void AcquireOrReplaceChannel(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers)
+        public ServerDescription DoServerSelection(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers)
```
papafe: The naming is temporary, I'm not sure of how we want to call these two methods. If we want to go with our current internal naming the two new methods should be:
`AcquireChannelSource` and `AcquireChannel`.

src/MongoDB.Driver/Core/Operations/DistinctOperation.cs
```diff
@@ -153,14 +153,18 @@ public BsonDocument CreateCommand(OperationContext operationContext, ICoreSessio
             };
         }
 
-        private EventContext.OperationNameDisposer BeginOperation() => EventContext.BeginOperation(OperationName);
+        private EventContext.OperationIdDisposer BeginOperation() => EventContext.BeginOperation(null, OperationName);
```
papafe, line 156: This is something strange. If we use this overloads, then the name of the command gets set properly. 
This happens because before this PR, we were connecting to the server when creating the `RetryableContext`, and at that time the command name was set properly. Now this is done inside the retryable operation executor, and by that time (for example) `ReadCommandOperation.Execute` is called and that contains `BeginOperation(null, null)`, that puts the command name to null. This means that the command name in `ClusterSelectingServerEvent` is null. 
We need to understand if this is the way we want to go, and so do it for other read operations as well.

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
```diff
@@ -146,6 +146,15 @@ protected async Task<TCommandResult> ExecuteProtocolAsync(
             }
         }
 
+        /// <summary>
+        /// Sets the command to be executed. This is used by derived classes that build commands dynamically.
+        /// </summary>
+        /// <param name="command">The command.</param>
+        protected void SetCommand(BsonDocument command)
```
sanych-sun: As far as I understood we need this method so inhereted classes could set the command just before executing it. Also as far as I understood we do not expose the _command (other then for testing purposes). Should we instead of `SetCommand` have a `CreateCommand` abstarct method and use it from `ExecuteProtocol`, so each inherited class will be able to produce a command.

src/MongoDB.Driver/Core/Operations/RetryableDeleteCommandOperation.cs
```diff
@@ -94,7 +94,7 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             else
             {
-                deletes = new BatchableSource<DeleteRequest>(_deletes.Items, _deletes.Offset, _deletes.ProcessedCount, canBeSplit: false);
+                deletes = new BatchableSource<DeleteRequest>(_deletes.Items, _deletes.Offset, _deletes.Count, _deletes.ProcessedCount, canBeSplit: false);
```
sanych-sun: I do agree it looks like a bug in the previous implementation. But also it means our collection bulk write could lost some operations in case of error/retries. Right? Can we create a test to reproduce the error on the current codebase and proof it's gone after the fix? Let's discuss this tomorrow, because it looks like a important fix. We might even separate it into own ticket.

src/MongoDB.Driver/Core/Operations/RetryableInsertCommandOperation.cs
```diff
@@ -91,7 +91,7 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             else
             {
-                documents = new BatchableSource<TDocument>(_documents.Items, _documents.Offset, _documents.ProcessedCount, canBeSplit: false);
+                documents = new BatchableSource<TDocument>(_documents.Items, _documents.Offset, _documents.Count, _documents.ProcessedCount, canBeSplit: false);
```
sanych-sun: Same as with RetriableDelete.

src/MongoDB.Driver/Core/Operations/RetryableUpdateCommandOperation.cs
```diff
@@ -101,11 +101,11 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             else
             {
-                updates = new BatchableSource<UpdateRequest>(_updates.Items, _updates.Offset, _updates.ProcessedCount, canBeSplit: false);
+                updates = new BatchableSource<UpdateRequest>(_updates.Items, _updates.Offset, _updates.Count, _updates.ProcessedCount, canBeSplit: false);
```
sanych-sun: Same as with RetriableDelete.

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
```diff
@@ -143,5 +138,11 @@ private void ReplaceChannelSource(IChannelSourceHandle channelSource)
             _channelSource = channelSource;
             _channel = null;
         }
+
+        private void DisposeChannelAndSource()
+        {
```
sanych-sun, line 143: Should we null them up after the disposal? Otherwise in case it's a retry and we've got an error in SelectServer method, we will have non-null Channel and Source (disposed one from the previous attempt)

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
```diff
@@ -26,98 +26,72 @@ internal static class RetryableReadOperationExecutor
...
-                var server = context.ChannelSource.ServerDescription;
+                ServerDescription server = null;
+
                 try
                 {
-                    return operation.ExecuteAttempt(operationContext, context, attempt, transactionNumber: null);
+                    server = context.SelectServer(operationContext, deprioritizedServers);
+                    context.AcquireChannel(operationContext);
+
+                    return operation.ExecuteAttempt(operationContext, context, totalAttempts, transactionNumber: null);
```
sanych-sun, line 43: Should we track separatelly executionAttempts? Because in no-CSOT scenario we have only 2 attempts, in case there will be an error on server selection or channel aquision - we might exit after just 1 attempt.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
```diff
@@ -26,151 +26,131 @@ internal static class RetryableWriteOperationExecutor
...
-                var server = context.ChannelSource.ServerDescription;
+                ServerDescription server = null;
+                bool channelAcquisitionSuccessful = false;
+
                 try
                 {
-                    return operation.ExecuteAttempt(operationContext, context, attempt, transactionNumber);
+                    server = context.SelectServer(operationContext, deprioritizedServers);
+                    context.AcquireChannel(operationContext);
+                    channelAcquisitionSuccessful = true;
```
sanych-sun: _Minor_: We can replace this variable with check if `context.Channel != null`.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
```diff
@@ -145,5 +138,11 @@ private void ReplaceChannelSource(IChannelSourceHandle channelSource)
             _channelSource = channelSource;
             _channel = null;
         }
+
+        private void DisposeChannelAndSource()
+        {
+            _channelSource?.Dispose();
```
sanych-sun, line 144: Should we null up them?

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/RetryableReadContext.cs
papafe: Agreed.

src/MongoDB.Driver/Core/Operations/RetryableWriteContext.cs
papafe: Agreed.

src/MongoDB.Driver/Core/Operations/RetryableReadOperationExecutor.cs
papafe: Are we talking about the attempt variable that goes into `operation.ExecuteAttempt`? If so then this variable it is not actually used for read operation, but instead we have a difference for write operation in `RetryableWriteOperationExecutor`, where we have both `totalAttempts` and `operationExecutionAttempts`. We could add it here too for the sake of symmetry, but it won't be used as well. 

If it's about counting how many attempts we have done, and so to count how many attempts we have left, those errors should be counted as well as far as I have understood from the retryable reads and writes specs.

src/MongoDB.Driver/Core/Operations/RetryableWriteOperationExecutor.cs
papafe: Agreed.

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
papafe: Your comment made me realise we could actually just pass the `command` to `ExecuteProtocol`, so I removed the `SetCommand` method.

src/MongoDB.Driver/Core/Operations/RetryableDeleteCommandOperation.cs
papafe: It's unclear if this was a bug or not in the first place, so I've restored this and created a ticket to investigate further: https://jira.mongodb.org/browse/CSHARP-5936

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
```diff
@@ -41,7 +41,7 @@ protected CommandOperationBase(
             MessageEncoderSettings messageEncoderSettings)
         {
             _databaseNamespace = Ensure.IsNotNull(databaseNamespace, nameof(databaseNamespace));
-            _command = Ensure.IsNotNull(command, nameof(command));
+            _command = command; //can be null
```
sanych-sun: I would prefer to get rid of _command at all, especially after adding that as a parameter to ExecuteProtocol.

src/MongoDB.Driver/Core/Operations/RetryableUpdateCommandOperation.cs
```diff
@@ -105,7 +105,7 @@ protected override IEnumerable<BatchableCommandMessageSection> CreateCommandPayl
             }
             var maxBatchCount = Math.Min(MaxBatchCount ?? int.MaxValue, channel.ConnectionDescription.MaxBatchCount);
             var maxDocumentSize = channel.ConnectionDescription.MaxWireDocumentSize;
-            var payload = new Type1CommandMessageSection<UpdateRequest>("updates", _updates, UpdateRequestSerializer.Instance, NoOpElementNameValidator.Instance, maxBatchCount, maxDocumentSize);
+            var payload = new Type1CommandMessageSection<UpdateRequest>("updates", updates, UpdateRequestSerializer.Instance, NoOpElementNameValidator.Instance, maxBatchCount, maxDocumentSize);
```
sanych-sun: I think we should revert this change too. Again it clearly looks like an issue, but it could be the reason why we did not face the missing operations on retry. Let's do this in separate PR/ticket after deeper investigation.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/CommandOperationBase.cs
papafe: Agreed, done.

src/MongoDB.Driver/Core/Operations/RetryableUpdateCommandOperation.cs
papafe: Done.

sanych-sun (approved):
LGTM, but please run the entire test suite to see if there are some unexpected failures.

PR #1913, opened 2026-03-13
CSHARP-5905: Fix bug when using EnumRepresentationConvention or ObjectSerializerAllowedTypesConvention

papafe:


tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
```diff
@@ -31,7 +32,12 @@ private class TestClass
             public object ObjectProp { get; set; }
             public object[] ArrayOfObjectProp { get; set; }
             public object[][] ArrayOfArrayOfObjectProp { get; set; }
+            public Dictionary<string, object> DictionaryStringObjectProp { get; set; }
```
papafe, line 38: This was not supported before, but given we're already working on this part of the code, it's a good occasion to add it.

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
```diff
@@ -162,26 +162,27 @@ public void Apply(BsonMemberMap memberMap)
...
+
             if (reconfiguredSerializer is not null)
             {
                 memberMap.SetSerializer(reconfiguredSerializer);
             }
 
             bool CouldApply(Type type)
-                => type == typeof(object) || type.IsNullable() || type.IsArray || typeof(IEnumerable).IsAssignableFrom(type);
+                => type == typeof(object) ||
+                   type.IsNullable() ||
```
adelinowona: Also I think this check is unnecessary? From my understanding, For `ObjectSerializerAllowedTypesConvention`, we're looking for `ObjectSerializer` instances to reconfigure. The question is: can a `Nullable<T>` ever contain an object? That's a no since `Nullable<T> `only works with value types. By having this check, it seems `ObjectSerializerAllowedTypesConvention` always does wasteful work for `Nullable<T> `types when it doesn't need to. I suggest we remove it.

adelinowona (changes requested):

src/MongoDB.Bson/Serialization/Conventions/ObjectSerializerAllowedTypesConvention.cs
papafe: Correct, I've removed that check.

adelinowona (approved):

adelinowona (approved):

PR #1918, opened 2026-03-18
CSHARP-5935: Command activities may be skipped when using pooled connection

ajcvickers:
I was looking at flaky tests, and found MongoClient_should_create_activities_when_tracing_enabled was failing occasionally. Turns out it was an actual bug.

New fix by Claude to avoid tracking state.

## Summary

I've successfully refactored the code to fix the bug where activity listeners were being checked when the connection was created (at which point none may be set), but not checked again when the connection was pulled from the pool.

### Key Changes:

1. **Constructor (`CommandEventHelper.cs:60`)**: Removed the listener check from `_shouldTrackState` initialization. Now `_shouldTrackState` only depends on whether event tracking is needed (`_shouldTrackSucceeded || _shouldTrackFailed`), not on tracing with listeners.

2. **Property Getters (`CommandEventHelper.cs:71-94`)**: Updated all property getters (`ShouldCallBeforeSending`, `ShouldCallAfterSending`, etc.) to dynamically check `MongoTelemetry.ActivitySource.HasListeners()` at execution time, ensuring that activity tracing is detected when connections are obtained from the pool.

3. **Lazy State Initialization (`CommandEventHelper.cs:37, 741-745`)**: Changed `_state` from `readonly` to allow lazy initialization. The state dictionary is now only created when needed - either for event tracking or when tracing with listeners is detected at runtime.

4. **Null Safety (`CommandEventHelper.cs:132, 165, 192, 216, 233`)**: Added null checks before accessing `_state` in all methods to handle cases where state tracking hasn't been initialized.

5. **TrackCommandState (`CommandEventHelper.cs:734-770`)**: Updated to check for listeners at execution time and lazily initialize the state dictionary only when tracing is actually needed.

6. **Test Update (`CommandEventHelperTests.cs:122-124`)**: Updated the test to reflect the new behavior where `_shouldTrackState` only reflects event tracking, not tracing with listeners (which is now checked dynamically).

### Benefits:

- **Bug Fixed**: Activity listeners are now checked when connections are obtained from the pool, not just when they're created
- **No Unnecessary State Tracking**: State tracking is only enabled when actually needed (either for events or when listeners are registered), avoiding the performance overhead of always tracking state
- **Backward Compatible**: The behavior is the same for all scenarios, just more efficient

The changes ensure that the driver properly detects activity listeners whenever a connection is used, whether it's newly created or retrieved from the pool, while avoiding unnecessary state tracking when no listeners are registered.

----

Junie (Opus 4.6) says:

### Root Cause

`_shouldTrace` in `CommandEventHelper` was set **once** at connection construction time via `MongoTelemetry.ActivitySource.HasListeners()`. Since connections are pooled and reused, if a connection was created before an `ActivityListener` was registered (or by a test with tracing disabled), `_shouldTrace` remained `false` permanently for that connection — command activities were never created even when a listener was later active.

### DIFF
--- a/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
+++ b/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
@@ -78,6 +78,8 @@ namespace MongoDB.Driver.Core.Connections
 
         public bool ShouldCallErrorReceiving => _eventsNeedState || ShouldTraceWithActivityListener();
 
+        public bool ShouldCallConnectionFailed => (_shouldTrackFailed || ShouldTraceWithActivityListener()) && _state != null;
+
         private bool ShouldTraceWithActivityListener()
             => !_tracingDisabled && MongoTelemetry.ActivitySource.HasListeners();
 
@@ -227,16 +229,6 @@ namespace MongoDB.Driver.Core.Connections
 
         public void ConnectionFailed(ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            if (!_shouldTrackFailed && !ShouldTraceWithActivityListener())
-            {
-                return;
-            }
-
-            if (_state == null)
-            {
-                return;
-            }
-
             CompleteFailedCommandActivity(exception);
 
             var requestIds = _state.Keys;
--- a/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
+++ b/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
@@ -36,9 +36,9 @@ namespace MongoDB.Driver.Core.Connections
         private readonly EventLogger<LogCategories.Command> _eventLogger;
         private ConcurrentDictionary<int, CommandState> _state;
 
-        private readonly bool _shouldProcessRequestMessages;
+        private readonly bool _eventsNeedBeforeSending;
         private readonly bool _shouldTrackStarted;
-        private readonly bool _shouldTrackState;
+        private readonly bool _eventsNeedState;
         private readonly bool _shouldTrackFailed;
         private readonly bool _shouldTrackSucceeded;
         private readonly bool _tracingDisabled;
@@ -57,10 +57,10 @@ namespace MongoDB.Driver.Core.Connections
             _tracingDisabled = tracingOptions?.Disabled == true;
             _queryTextMaxLength = tracingOptions?.QueryTextMaxLength ?? 0;
 
-            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed;
-            _shouldProcessRequestMessages = _shouldTrackStarted || _shouldTrackState;
+            _eventsNeedState = _shouldTrackSucceeded || _shouldTrackFailed;
+            _eventsNeedBeforeSending = _shouldTrackStarted || _eventsNeedState;
 
-            if (_shouldTrackState)
+            if (_eventsNeedState)
             {
                 // we only need to track state if we have to raise
                 // a succeeded or failed event or for tracing
@@ -68,15 +68,15 @@ namespace MongoDB.Driver.Core.Connections
             }
         }
 
-        public bool ShouldCallBeforeSending => _shouldProcessRequestMessages || ShouldTraceWithActivityListener();
+        public bool ShouldCallBeforeSending => _eventsNeedBeforeSending || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallAfterSending => _shouldTrackState || ShouldTraceWithActivityListener();
+        public bool ShouldCallAfterSending => _eventsNeedState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallErrorSending => _shouldTrackState || ShouldTraceWithActivityListener();
+        public bool ShouldCallErrorSending => _eventsNeedState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallAfterReceiving => _shouldTrackState || ShouldTraceWithActivityListener();
+        public bool ShouldCallAfterReceiving => _eventsNeedState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallErrorReceiving => _shouldTrackState || ShouldTraceWithActivityListener();
+        public bool ShouldCallErrorReceiving => _eventsNeedState || ShouldTraceWithActivityListener();
 
         private bool ShouldTraceWithActivityListener()
             => !_tracingDisabled && MongoTelemetry.ActivitySource.HasListeners();
@@ -742,16 +742,12 @@ namespace MongoDB.Driver.Core.Connections
         {
             var shouldTraceCommand = ShouldTraceWithActivityListener() && !shouldRedactCommand && !skipLogging;
 
-            if (!_shouldTrackState && !shouldTraceCommand)
+            if (!_eventsNeedState && !shouldTraceCommand)
             {
                 return;
             }
 
-            if (!_shouldTrackState && shouldTraceCommand)
-            {
-                // Lazy initialize state dictionary when tracing is needed but event tracking is not
-                _state ??= new ConcurrentDictionary<int, CommandState>();
-            }
+            _state ??= new ConcurrentDictionary<int, CommandState>();
 
             var commandState = new CommandState
             {
--- a/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
+++ b/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
@@ -41,7 +41,7 @@ namespace MongoDB.Driver.Core.Connections
         private readonly bool _shouldTrackState;
         private readonly bool _shouldTrackFailed;
         private readonly bool _shouldTrackSucceeded;
-        private readonly bool _shouldTrace;
+        private readonly bool _tracingDisabled;
         private readonly int _queryTextMaxLength;
 
         private Activity _currentCommandActivity;
@@ -54,10 +54,10 @@ namespace MongoDB.Driver.Core.Connections
             _shouldTrackFailed = _eventLogger.IsEventTracked<CommandFailedEvent>();
             _shouldTrackStarted = _eventLogger.IsEventTracked<CommandStartedEvent>();
 
-            _shouldTrace = tracingOptions?.Disabled != true && MongoTelemetry.ActivitySource.HasListeners();
+            _tracingDisabled = tracingOptions?.Disabled == true;
             _queryTextMaxLength = tracingOptions?.QueryTextMaxLength ?? 0;
 
-            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || _shouldTrace;
+            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || !_tracingDisabled;
             _shouldProcessRequestMessages = _shouldTrackStarted || _shouldTrackState;
 
             if (_shouldTrackState)
@@ -223,7 +223,7 @@ namespace MongoDB.Driver.Core.Connections
 
         public void ConnectionFailed(ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            if (!_shouldTrackFailed && !_shouldTrace)
+            if (!_shouldTrackFailed && _tracingDisabled)
             {
                 return;
             }
@@ -746,7 +746,7 @@ namespace MongoDB.Driver.Core.Connections
                 ShouldRedactReply = shouldRedactCommand
             };
 
-            if (_shouldTrace && !shouldRedactCommand && !skipLogging)
+            if (!_tracingDisabled && MongoTelemetry.ActivitySource.HasListeners() && !shouldRedactCommand && !skipLogging)
             {
                 _currentCommandActivity = MongoTelemetry.StartCommandActivity(
                     commandName,
--- a/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
+++ b/src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
@@ -34,7 +34,7 @@ namespace MongoDB.Driver.Core.Connections
     internal class CommandEventHelper
     {
         private readonly EventLogger<LogCategories.Command> _eventLogger;
-        private readonly ConcurrentDictionary<int, CommandState> _state;
+        private ConcurrentDictionary<int, CommandState> _state;
 
         private readonly bool _shouldProcessRequestMessages;
         private readonly bool _shouldTrackStarted;
@@ -57,7 +57,7 @@ namespace MongoDB.Driver.Core.Connections
             _tracingDisabled = tracingOptions?.Disabled == true;
             _queryTextMaxLength = tracingOptions?.QueryTextMaxLength ?? 0;
 
-            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || !_tracingDisabled;
+            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed;
             _shouldProcessRequestMessages = _shouldTrackStarted || _shouldTrackState;
 
             if (_shouldTrackState)
@@ -68,30 +68,18 @@ namespace MongoDB.Driver.Core.Connections
             }
         }
 
-        public bool ShouldCallBeforeSending
-        {
-            get { return _shouldProcessRequestMessages; }
-        }
+        public bool ShouldCallBeforeSending => _shouldProcessRequestMessages || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallAfterSending
-        {
-            get { return _shouldTrackState; }
-        }
+        public bool ShouldCallAfterSending => _shouldTrackState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallErrorSending
-        {
-            get { return _shouldTrackState; }
-        }
+        public bool ShouldCallErrorSending => _shouldTrackState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallAfterReceiving
-        {
-            get { return _shouldTrackState; }
-        }
+        public bool ShouldCallAfterReceiving => _shouldTrackState || ShouldTraceWithActivityListener();
 
-        public bool ShouldCallErrorReceiving
-        {
-            get { return _shouldTrackState; }
-        }
+        public bool ShouldCallErrorReceiving => _shouldTrackState || ShouldTraceWithActivityListener();
+
+        private bool ShouldTraceWithActivityListener()
+            => !_tracingDisabled && MongoTelemetry.ActivitySource.HasListeners();
 
         public void CompleteFailedCommandActivity(Exception exception)
         {
@@ -129,8 +117,12 @@ namespace MongoDB.Driver.Core.Connections
 
         public void AfterSending(RequestMessage message, ConnectionId connectionId, ObjectId? serviceId, bool skipLogging)
         {
-            CommandState state;
-            if (_state.TryGetValue(message.RequestId, out state) &&
+            if (_state == null)
+            {
+                return;
+            }
+
+            if (_state.TryGetValue(message.RequestId, out var state) &&
                 state.ExpectedResponseType == ExpectedResponseType.None)
             {
                 state.Stopwatch.Stop();
@@ -157,12 +149,16 @@ namespace MongoDB.Driver.Core.Connections
 
         public void ErrorSending(RequestMessage message, ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            CommandState state;
-            if (_state.TryRemove(message.RequestId, out state))
+            CompleteCommandActivityWithException(exception);
+
+            if (_state == null)
+            {
+                return;
+            }
+
+            if (_state.TryRemove(message.RequestId, out var state))
             {
                 state.Stopwatch.Stop();
-
-                CompleteCommandActivityWithException(exception);
 
                 _eventLogger.LogAndPublish(new CommandFailedEvent(
                         state.CommandName,
@@ -179,8 +175,12 @@ namespace MongoDB.Driver.Core.Connections
 
         public void AfterReceiving(ResponseMessage message, IByteBuffer buffer, ConnectionId connectionId, ObjectId? serviceId, MessageEncoderSettings encoderSettings, bool skipLogging)
         {
-            CommandState state;
-            if (!_state.TryRemove(message.ResponseTo, out state))
+            if (_state == null)
+            {
+                return;
+            }
+
+            if (!_state.TryRemove(message.ResponseTo, out var state))
             {
                 // this indicates a bug in the sending portion...
                 return;
@@ -198,16 +198,20 @@ namespace MongoDB.Driver.Core.Connections
 
         public void ErrorReceiving(int responseTo, ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            CommandState state;
-            if (!_state.TryRemove(responseTo, out state))
+            CompleteCommandActivityWithException(exception);
+
+            if (_state == null)
+            {
+                return;
+            }
+
+            if (!_state.TryRemove(responseTo, out var state))
             {
                 // this indicates a bug in the sending portion...
                 return;
             }
 
             state.Stopwatch.Stop();
-
-            CompleteCommandActivityWithException(exception);
 
             _eventLogger.LogAndPublish(new CommandFailedEvent(
                 state.CommandName,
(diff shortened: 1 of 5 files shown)


tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/LibmongocryptTests.cs
```diff
@@ -59,7 +59,8 @@ public void Explicit_encryption_with_libmongocrypt_package_works()
                 kmsProviders.Add("local", localKey);
 
                 var keyVaultNamespace = CollectionNamespace.FromFullName("encryption.__keyVault");
-                var keyVaultMongoClient = new MongoClient();
```
ajcvickers, line 62: This is a simple test fix to use the configured local connection string. Not related to the bug fix--I can create another PR if necessary.

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
```diff
@@ -43,6 +45,8 @@ public void MongoClient_should_create_activities_when_tracing_enabled()
                 collection.InsertOne(new BsonDocument("name", "test"));
                 collection.Find(Builders<BsonDocument>.Filter.Empty).FirstOrDefault();
                 collection.DeleteOne(Builders<BsonDocument>.Filter.Eq("name", "test"));
+
+                SpinWait.SpinUntil(() => capturedActivities.Count >= 6, millisecondsTimeout: 10000);
```
ajcvickers: I'm not sure this is needed--I added it when I thought the test was flaky. But I don't think it does harm.

tests/MongoDB.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingXunitTestAssemblyRunner.cs
```diff
@@ -41,7 +41,7 @@ public TimeoutEnforcingXunitTestAssemblyRunner(
                 executionMessageSink,
                 executionOptions)
         {
-            _unobservedExceptionTrackingTestCase = testCases.SingleOrDefault(IsUnobservedExceptionTrackingTestCase);
+            _unobservedExceptionTrackingTestCase = testCases.FirstOrDefault(IsUnobservedExceptionTrackingTestCase);
         }
```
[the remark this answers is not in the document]
ajcvickers: This is being reviewed in a different PR.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -223,7 +223,7 @@ public void ErrorReceiving(int responseTo, ConnectionId connectionId, ObjectId?
 
         public void ConnectionFailed(ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            if (!_shouldTrackFailed && !_shouldTrace)
+            if (!_shouldTrackFailed && _tracingDisabled)
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
```diff
@@ -13,9 +13,11 @@
 * limitations under the License.
 */
 
+using System.Collections.Concurrent;
 using System.Collections.Generic;
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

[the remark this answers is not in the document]
ajcvickers: Still needed for ReadOnlyConnection

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/LibmongocryptTests.cs
```diff
@@ -59,7 +59,8 @@ public void Explicit_encryption_with_libmongocrypt_package_works()
...
+                var connectionString = Environment.GetEnvironmentVariable("MONGODB_URI");
+                var keyVaultMongoClient = string.IsNullOrWhiteSpace(connectionString) ? new MongoClient() : new MongoClient(connectionString);
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -54,10 +54,10 @@ public CommandEventHelper(EventLogger<LogCategories.Command> eventLogger, Tracin
             _shouldTrackFailed = _eventLogger.IsEventTracked<CommandFailedEvent>();
             _shouldTrackStarted = _eventLogger.IsEventTracked<CommandStartedEvent>();
 
-            _shouldTrace = tracingOptions?.Disabled != true && MongoTelemetry.ActivitySource.HasListeners();
+            _tracingDisabled = tracingOptions?.Disabled == true;
             _queryTextMaxLength = tracingOptions?.QueryTextMaxLength ?? 0;
 
-            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || _shouldTrace;
+            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || !_tracingDisabled;
             _shouldProcessRequestMessages = _shouldTrackStarted || _shouldTrackState;
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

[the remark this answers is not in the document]
ajcvickers: While copilot is not wrong here, making this change re-introduces the bug, since if we're not tracking state, then we never get to the point of creating the activity.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -223,7 +223,7 @@ public void ErrorReceiving(int responseTo, ConnectionId connectionId, ObjectId?
 
         public void ConnectionFailed(ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            if (!_shouldTrackFailed && !_shouldTrace)
+            if (!_shouldTrackFailed && (_tracingDisabled || !MongoTelemetry.ActivitySource.HasListeners()))
```
BorisDog: minor: probably it' cleaner to extract to `IsTracingDisabled` method.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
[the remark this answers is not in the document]
BorisDog: It's important not to track stack when there are no listeners.
If it comes to that, we might need a more creative solution, or adjust our testing.

BorisDog (changes requested):

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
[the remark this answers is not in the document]
ajcvickers: Yep, I think you are right.

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/LibmongocryptTests.cs
ajcvickers: Moved to a different PR, so this one is just the otel bug.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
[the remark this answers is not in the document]
ajcvickers: New fix up.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -157,12 +149,16 @@ public void AfterSending(RequestMessage message, ConnectionId connectionId, Obje
 
         public void ErrorSending(RequestMessage message, ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            CommandState state;
-            if (_state.TryRemove(message.RequestId, out state))
+            CompleteCommandActivityWithException(exception);
```
adelinowona: no need to move this activity completion outside of finding the state entry? I would rather keep it gated on finding the state entry as that is more consistent with my original intent (complete activity only if we tracked the command) and easier to reason about.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -198,17 +198,21 @@ public void AfterReceiving(ResponseMessage message, IByteBuffer buffer, Connecti
 
         public void ErrorReceiving(int responseTo, ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            CommandState state;
-            if (!_state.TryRemove(responseTo, out state))
+            CompleteCommandActivityWithException(exception);
```
adelinowona: Same as the comment above about keeping this gated on finding the state entry.

tests/MongoDB.Driver.Tests/Core/Connections/CommandEventHelperTests.cs
adelinowona, line 1: I think the test `ShouldTrackState_should_be_correct` should drop the traceCommands parameter entirely — it's testing `_shouldTrackState` which is now purely about event tracking, so tracing config is irrelevant.

As for the test `ShouldTrackState_should_be_correct_with_activity_listener`, I think we should repurpose it. The interesting behavior to test now is: "when no events are subscribed but a listener is registered, the dynamic properties return true." Something like asserting `ShouldCallBeforeSending` is true even when `_shouldTrackState` is false, as long as tracing isn't disabled and a listener exists.

tests/MongoDB.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingXunitTestAssemblyRunner.cs
adelinowona, line 1: You mentioned to copilot that this change is being reviewed in a different PR. Should it be dropped from this PR then?

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
adelinowona: So is this test failing because of the pooled connections bug or because of timing for the capturedActivities? Do you have a link to an evergreen failure for this test?

adelinowona (changes requested):

tests/MongoDB.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingXunitTestAssemblyRunner.cs
ajcvickers: Yes. I just merged the other PR, so I will rebase. It was here too because it was blocking me from running tests locally.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: Will give it a try.

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
ajcvickers: The failure was caused the pooled connection bug together with the other tests run at the same time and their order. I have no evidence that things ever failed because this code ran before the activities were recorded. That being said, it's not clear to me that everything about the activities flow is guaranteed to be complete before getting here. If we are sure that is the case, then this code can be removed.

tests/MongoDB.Driver.Tests/Core/Connections/CommandEventHelperTests.cs
ajcvickers: Updated tests.

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
ajcvickers: I ran the failing set in a loop for about 20 minutes without the spin check and it didn't fail, so I have removed this. As long as we await every async operation before getting here there should not be a race.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: Updated.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: Updated.

tests/MongoDB.Driver.Tests/Core/Connections/CommandEventHelperTests.cs
```diff
@@ -119,7 +111,27 @@ public void ShouldTrackState_should_be_correct_with_activity_listener(
...
+                commandHelper.ShouldCallErrorReceiving.Should().Be(captureCommandSucceeded || captureCommandFailed || logCommands);
+
+                listener = new ActivityListener
+                {
+                    ShouldListenTo = source => source.Name == "MongoDB.Driver",
+                    Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllData
+                };
+                ActivitySource.AddActivityListener(listener);
+
+                // With listeners registered, these always return true when unless everything is disabled.
```
adelinowona: nit: typo in the comment, remove "when"

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
```diff
@@ -84,9 +85,9 @@ public void MongoClient_should_not_create_activities_when_tracing_disabled()
             capturedActivities.Should().BeEmpty();
         }
 
-        private static ActivityListener CreateActivityListener(out List<Activity> capturedActivities)
+        private static ActivityListener CreateActivityListener(out IReadOnlyCollection<Activity> capturedActivities)
         {
-            var activities = new List<Activity>();
+            var activities = new ConcurrentBag<Activity>();
```
adelinowona: I don't think we need this ConcurrentBag change given you removed the spin check code you had earlier.

adelinowona (changes requested):

tests/SmokeTests/MongoDB.Driver.SmokeTests.Sdk/OpenTelemetryTests.cs
ajcvickers: Fixed.

tests/MongoDB.Driver.Tests/Core/Connections/CommandEventHelperTests.cs
ajcvickers: Fixed.

adelinowona (approved):

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -54,10 +54,10 @@ public CommandEventHelper(EventLogger<LogCategories.Command> eventLogger, Tracin
             _shouldTrackFailed = _eventLogger.IsEventTracked<CommandFailedEvent>();
             _shouldTrackStarted = _eventLogger.IsEventTracked<CommandStartedEvent>();
 
-            _shouldTrace = tracingOptions?.Disabled != true && MongoTelemetry.ActivitySource.HasListeners();
+            _tracingDisabled = tracingOptions?.Disabled == true;
             _queryTextMaxLength = tracingOptions?.QueryTextMaxLength ?? 0;
 
-            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed || _shouldTrace;
+            _shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed;
```
BorisDog: While reviewing I got a bit confused between all the namings. 
So `_shouldTrackFailed` and `_shouldTrackStarted` correspond to Event tracking or Logging enabled/
Should `_shouldTrackState` naming reflect that as well? Maybe it should be `_shouldTrackEventsOrLogs`? Otherwise we can be get into situation where `_shouldTrackState == false` but state is still tracked and `_state != null`?

Or maybe there are better ways to simplify this?

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -731,11 +740,19 @@ private void TrackCommandState(
...
         {
-            if (!_shouldTrackState)
+            var shouldTraceCommand = ShouldTraceWithActivityListener() && !shouldRedactCommand && !skipLogging;
+
+            if (!_shouldTrackState && !shouldTraceCommand)
             {
                 return;
             }
 
+            if (!_shouldTrackState && shouldTraceCommand)
```
BorisDog: No need for this condition?
Would just simple
`_state ??= new ConcurrentDictionary<int, CommandState>();`
be more readable?

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -129,8 +117,12 @@ public void BeforeSending(
 
         public void AfterSending(RequestMessage message, ConnectionId connectionId, ObjectId? serviceId, bool skipLogging)
         {
-            CommandState state;
-            if (_state.TryGetValue(message.RequestId, out state) &&
+            if (_state == null)
+            {
+                return;
```
BorisDog, line 124: Is this code path expected, under assumption that it's guarded by `ShouldCallAfterSending()` ?
If it's not, should we throw instead of return?

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
```diff
@@ -223,12 +227,17 @@ public void ErrorReceiving(int responseTo, ConnectionId connectionId, ObjectId?
 
         public void ConnectionFailed(ConnectionId connectionId, ObjectId? serviceId, Exception exception, bool skipLogging)
         {
-            if (!_shouldTrackFailed && !_shouldTrace)
+            if (!_shouldTrackFailed && !ShouldTraceWithActivityListener())
```
BorisDog: Maybe not in scope of this PR, but we are already here: Should we be consistent with same pattern here as in methd `ShouldCallAfterReceiving` => AfterReceiving()?
Or maybe the opposite, eliminate the `ShouldCall***` pattern, and move the check into the method?

BorisDog (changes requested):

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: Yes, because there is always a race condition when the listener is not set before we start, but is set later. This is fine, it just means there needs to be a null check.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: Updated names and reasoning from Claude:

Current flags and what they actually mean
    
    ┌───────────────────────────────┬──────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                      
    │             Flag              │ Current Name │                                                                                                              What it actually does                                                                                                               │                                                                      
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
    │ _shouldTrackSucceeded         │ OK           │ True when CommandSucceededEvent has subscribers                                                                                                                                                                                                  │
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
    │ _shouldTrackFailed            │ OK           │ True when CommandFailedEvent has subscribers                                                                                                                                                                                                     │                                                                      
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
    │ _shouldTrackStarted           │ OK           │ True when CommandStartedEvent has subscribers                                                                                                                                                                                                    │                                                                      
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
    │ _shouldTrackState             │ Misleading   │ _shouldTrackSucceeded || _shouldTrackFailed — "do events need state tracking?" — but the name implies it's the sole control over whether state is tracked, which is no longer true since tracing also drives state tracking (lazily, at runtime) │
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
    │ _shouldProcessRequestMessages │ Vague        │ _shouldTrackStarted || _shouldTrackState — "do events require the BeforeSending phase?" — the name describes implementation mechanics rather than intent                                                                                         │
    ├───────────────────────────────┼──────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                      
    │ _tracingDisabled              │ OK           │ True when tracing is disabled by configuration                                                                                                                                                                                                   │
    └───────────────────────────────┴──────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

Proposed renames

_shouldTrackState → _eventsNeedState

The old name reads as if it's the single gatekeeper for all state tracking. But after the CSHARP-5935 fix, tracing also causes state to be tracked — lazily, checked dynamically via ShouldTraceWithActivityListener(). The flag only captures the event half. The new name makes the scope explicit:

// Before (misleading — doesn't cover tracing-driven state)                                                                                                                                                                                                                                                                                                              
_shouldTrackState = _shouldTrackSucceeded || _shouldTrackFailed;

// After (accurate — this flag is only about events)                                                                                                                                                                                                                                                                                                                     
_eventsNeedState = _shouldTrackSucceeded || _shouldTrackFailed;

_shouldProcessRequestMessages → _eventsNeedBeforeSending

"Process request messages" describes a mechanism, not a purpose. The flag answers: "do any events require the BeforeSending callback?" — which is true when we need to fire a CommandStartedEvent or need to capture state for later succeeded/failed events:

// The property reads naturally with the new name:                                                                                                                                                                                                                                                                                                                       
public bool ShouldCallBeforeSending => _eventsNeedBeforeSending || ShouldTraceWithActivityListener();
                                                                                                                                                                                                                                                                                                                                                                     
The three _shouldTrack{Started,Succeeded,Failed} flags are consistent and accurate as-is. The two renamed flags are the ones whose names no longer matched their semantics after the lazy-state-initialization refactor.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: I'm not really sure what you are suggesting here. Could you explain more?

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
ajcvickers: It's not really needed, since ??= will still work if _state is already not-null. Will remove.

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
BorisDog: Oh I see, in a situation where `ShouldCallBeforeSending == false` and `ShouldCallAfterSending == true`.
Thanks!

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
BorisDog: Currently `ConnectionFailed` is always called, and the check `!_shouldTrackFailed && !ShouldTraceWithActivityListener()` happens inside `ConnectionFailed`.
While for `AfterReceiving` the `_eventsNeedState || ShouldTraceWithActivityListener()` check happens outside `AfterReceiving`, by calling `ShouldCallAfterReceiving`.

Should we have the same pattern for all methods?

```
If (condition)
   CallMethod
```

or

```
CallMethod() { 
 if (not condition)
   return
}
```

src/MongoDB.Driver/Core/Connections/CommandEventHelper.cs
BorisDog: Looks good!

BorisDog (approved):
There is a minor question pending. But it's not a blocker.
Everything else LGTM!

PR #1920, opened 2026-03-19
CSHARP-5930: Enable running tests locally using Docker

ajcvickers:
I have one env var set: MONGODB_URI

I'm using Atlas Local 8.2, which is a standalone cluster by default, but tests that pass were being excluded from running. If people are using non-Atlas and this breaks, then we'll have to figure something out.


tests/MongoDB.Driver.Examples/TransactionExamplesForDocs/WithTransactionExample1.cs
```diff
@@ -29,7 +29,7 @@ public class WithTransactionExample1
         [Fact]
         public void Example1()
         {
-            RequireServer.Check().ClusterTypes(ClusterType.ReplicaSet, ClusterType.Sharded).Supports(Feature.Transactions);
+            RequireServer.Check().ClusterTypes(ClusterType.ReplicaSet, ClusterType.Sharded, ClusterType.Standalone).Supports(Feature.Transactions);
```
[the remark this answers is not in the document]
ajcvickers: Atlas Local reports Standalone, but supports transactions.

src/MongoDB.Driver/Encryption/AutoEncryptionProviderRegistry.cs
```diff
@@ -28,10 +28,16 @@ internal sealed class AutoEncryptionProviderRegistry : IAutoEncryptionProviderRe
...
         public void Register(Func<IMongoClient, AutoEncryptionOptions, IAutoEncryptionLibMongoCryptController> factory)
         {
+            if(ReferenceEquals(factory, _autoCryptClientControllerFactory))
+            {
+                return;
+            }
+
             if (_autoCryptClientControllerFactory != null)
             {
                 throw new MongoConfigurationException("AutoEncryption Provider already registered.");
```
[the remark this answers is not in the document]
ajcvickers: Except this is only called from one place, and it always uses the method group, so this works.

tests/MongoDB.Driver.Examples/ChangeStreamExamples.cs
```diff
@@ -135,22 +147,24 @@ public void ChangestreamExample4()
             var database = client.GetDatabase("ChangeStreamExamples");
             database.DropCollection("inventory");
 
-            using var cancelationTokenSource = new CancellationTokenSource();
-            try
+            var workerEnabled = true;
+            var document = new BsonDocument("username", "alice");
+
+            var thread = new Thread(() =>
             {
```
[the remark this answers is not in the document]
ajcvickers: You are wrong that this causes the build to fail, but removed the unneeded using anyway.

tests/MongoDB.TestHelpers/XunitExtensions/TimeoutEnforcing/TimeoutEnforcingXunitTestAssemblyRunner.cs
```diff
@@ -41,7 +41,7 @@ public TimeoutEnforcingXunitTestAssemblyRunner(
                 executionMessageSink,
                 executionOptions)
         {
-            _unobservedExceptionTrackingTestCase = testCases.SingleOrDefault(IsUnobservedExceptionTrackingTestCase);
+            _unobservedExceptionTrackingTestCase = testCases.FirstOrDefault(IsUnobservedExceptionTrackingTestCase);
         }
```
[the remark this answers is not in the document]
ajcvickers: Being reviewed in a different PR.

tests/MongoDB.Driver.Examples/ChangeStreamExamples.cs
```diff
@@ -135,22 +147,24 @@ public void ChangestreamExample4()
...
-                var document = new BsonDocument("username", "alice");
+                var inventoryCollection = database.GetCollection<BsonDocument>("inventory");
 
-                Task.Run(() =>
+                while (workerEnabled)
                 {
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

tests/MongoDB.Driver.Examples/StableApiExamples.cs
```diff
@@ -73,7 +73,7 @@ public void ConfigureServerApiDeprecationErrors()
         [Fact]
         public void StableAPI_Strict_Migration_Example()
         {
-            var connectionString = "mongodb://localhost";
+            var connectionString = InfrastructureUtilities.MongoUri;
             var serverApi = new ServerApi(ServerApiVersion.V1, strict: true);
             var mongoClientSettings = MongoClientSettings.FromConnectionString(connectionString);
             mongoClientSettings.ServerApi = serverApi;
```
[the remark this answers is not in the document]
ajcvickers: Will fix.

tests/MongoDB.Driver.Examples/TransactionExamplesForDocs/WithTransactionExample1.cs
sanych-sun: I have a provocative question: why do we even have Examples project? We do not run this as a part of our CI, and also as far as I know Docs Team has their own repo with runnable examples. Should we remove this or at least convert to non-tests like project?

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
```diff
@@ -172,7 +172,7 @@ public RequireServer Supports(params Feature[] features)
 
         public RequireServer SupportsCausalConsistency()
         {
-            return ClusterTypes(Clusters.ClusterType.Sharded, Clusters.ClusterType.ReplicaSet).SupportsSessions();
+            return ClusterTypes(Clusters.ClusterType.Sharded, Clusters.ClusterType.ReplicaSet, Clusters.ClusterType.Standalone).SupportsSessions();
```
sanych-sun: We probably should not allow this for any standalone server, but probably do some explicit check for Atlas. Is there any way to figure out the server is Atlas one?

tests/MongoDB.Driver.Examples/TransactionExamplesForDocs/WithTransactionExample1.cs
ajcvickers: @sanych-sun Very good questions! I don't know if we need this code, and if we don't, then it should be removed. @BorisDog?

If we do need this code, then it is better that it can be run. Dead code (code that is never run) rots. EF Core has runnable examples in the docs repo. This code did not run as part of the CI, so each year before the release I would load and run each example, and fix the ones where the rot had set in. This was a pain, but better than having completely un-run code. We could do the same thing here, but I'm not sure it is better than just having the code in tests that we can run easily. Like the unit tests, it's not like these are slow.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: Agreed, and that's what I am going to investigate.

tests/MongoDB.Driver.Examples/TransactionExamplesForDocs/WithTransactionExample1.cs
ajcvickers: Removed the cluster checks here, since I think Supports(Feature.Transactions) is sufficient.

tests/MongoDB.Driver.Examples/ClientSideEncryption2Examples.cs
```diff
@@ -37,7 +37,7 @@ public class ClientSideEncryption2Examples
         [Fact]
         public void FLE2AutomaticEncryption()
         {
-            RequireServer.Check().Supports(Feature.Csfle2).ClusterTypes(ClusterType.ReplicaSet, ClusterType.Sharded, ClusterType.LoadBalanced);
+            RequireServer.Check().Supports(Feature.Csfle2).ClusterTypes(ClusterType.ReplicaSet, ClusterType.Sharded, ClusterType.LoadBalanced, ClusterType.Standalone);
```
ajcvickers: Removed the cluster checks here, since I think Supports(Feature.Csfle2) is sufficient.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: So I could not find a way to detect Atlas, and Claude spent lots of tokens then also said that it was not possible to do it reliably and was only able to suggest some quite dodgy heuristics.

Instead, I've put some code in to check if the connection string for running Atlas tests is set to the same as the regular connection string, and if it is, then we still run these tests, since we know the connection string is for Atlas. These means a local setup using Atlas local will always run these, but usually they won't run elsewhere unless the cluster is not Standalone.

tests/MongoDB.Driver.Examples/TransactionExamplesForDocs/WithTransactionExample1.cs
sanych-sun: Generally speaking we have to check for topology here. Because `Supports(feature)` checks wire version only, but some features are not available on every topology (in this case Transactions require ReplicaSet and Sharded cluster). Other examples could be: change streams and casual consistency.

ajcvickers:
I'm closing this for now. The latest test don't pass on any container image I have access to. Until I can figure out how we can reliably get container images to use, I don't think there is much use in attempting to make the inner-loop better. I'm very sad about this, because I want to be able to work efficiently on the driver, but I think it's just the reality of where we are.

PR #1923, opened 2026-03-24
CSHARP-5942: Fix serializer resolution for dictionary paths with numeric string keys

damieng:
## Summary

- Fix `DictionarySerializerBase<TDictionary, TKey, TValue>.TryGetItemSerializationInfo()` to return `false` for Document-represented dictionaries with non-numeric key types
- Add tests covering `AddToSet` with dictionary-based nested collection paths, including numeric string keys

## Background

A regression was introduced in 3.7.0 via CSHARP-5572 (#1700) that changed `DictionarySerializerBase<TDictionary, TKey, TValue>.TryGetItemSerializationInfo()` to always return `true`, even for `DictionaryRepresentation.Document`. Previously it returned `false` for Document representation.

This broke string-based field path resolution for dictionaries with all-digit keys (e.g., `"Buckets.10"`), because `StringFieldDefinitionHelper.Resolve` treats all-digit path segments as array indices and calls `IBsonArraySerializer.TryGetItemSerializationInfo()`. When the dictionary serializer now returns `true`, the resolver picks up a `KeyValuePairSerializer` instead of the dictionary's value serializer. Operations like `AddToSet`, `Push`, and `Pull` then fail because `KeyValuePairSerializer` does not implement `IBsonArraySerializer`.

## Root cause

In CSHARP-5572, `DictionarySerializerBase.TryGetItemSerializationInfo()` was simplified to unconditionally return `true`. The original code returned `false` for `DictionaryRepresentation.Document`, which caused `StringFieldDefinitionHelper.Resolve` to correctly fall through the array-index branch and set `resolvedFieldSerializer = null`, letting `AddToSet` resolve the item serializer from the registry.

## Fix

Restore the guard in `TryGetItemSerializationInfo()` for Document-represented dictionaries, but with a refinement: allow `true` when `TKey` is a numeric type (int, long, short, byte, etc.), since dictionaries with integer keys serialized as Document look like BSON arrays with numeric indices. For string-keyed dictionaries — the common case — the method returns `false` for Document representation, matching the pre-3.7.0 behavior.

```csharp
if (_dictionaryRepresentation == DictionaryRepresentation.Document && !IsNumericType(typeof(TKey)))
{
    serializationInfo = null;
    return false;
}
```


src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
```diff
@@ -510,6 +516,14 @@ public bool TryGetItemSerializationInfo(out BsonSerializationInfo serializationI
             return true;
         }
 
+        private static bool IsNumericType(Type type)
```
sanych-sun, line 519: We have helper method to detect if the Type is integer in `TypeExtensions.IsIntegral`. I think we can use it instead.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/UpdateDefinitionBuilderTests.cs
```diff
@@ -66,6 +67,113 @@ public void AddToSetEach_Typed()
...
+
+            Assert(subject.AddToSet("Buckets.myKey", "newValue"), "{$addToSet: {'buckets.myKey': 'newValue'}}");
+        }
+
+        [Fact]
+        public void AddToSet_with_numeric_string_key_through_dictionary()
+        {
+            // When a dictionary key is all-digits (e.g., "10"), the string path resolver
+            // treats it as an array index rather than a dictionary member name.
+            // This is the exact scenario reported in the user bug: "Buckets.10" where
```
sanych-sun: We should either remove this comment or specify what bug/ticket we are talking about.

src/MongoDB.Bson/Serialization/Serializers/DictionarySerializerBase.cs
damieng: That's in the Driver asssembly, this is in the Bson assembly.

tests/MongoDB.Driver.Tests/UpdateDefinitionBuilderTests.cs
```diff
@@ -66,6 +67,113 @@ public void AddToSetEach_Typed()
...
+        {
+            var subject = CreateSubject<DocumentWithDictionaryOfHashSets>();
+
+            Assert(subject.AddToSet(x => x.Buckets["myKey"], "newValue"), "{$addToSet: {'buckets.myKey': 'newValue'}}");
+        }
+
+        [Fact]
+        public void AddToSet_with_string_path_through_array_rep_dictionary()
+        {
+            // When dictionary uses ArrayOfDocuments representation, string path resolution
```
sanych-sun: I think the comment is wrong, the issue was about the case when dictionary represented as Document. Checks in `DictionarySerializerBase.TryGetItemSerializationInfo` suggests there are no problem with ArrayOfDocuments representation too.

sanych-sun (approved):
LGTM, but please double-check the comment in tests, it seems to be misleading.

PR #1928, opened 2026-03-27
CSHARP-5947: Increase SingleServerReadBinding timeout

papafe:


src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
```diff
@@ -79,15 +79,22 @@ public void Dispose()
...
+        private IChannelSourceHandle GetChannelSourceHelper(OperationContext operationContext)
         {
             // server might be in unknown state due to previous failed operation, allow description to be updated
             // this is done instead of server selection
             // TODO parameterize timeout and avoid busy wait, or offload waiting to server level
             // Should be addressed by CSHARP-3556
             if (_server.Description.State == ServerState.Disconnected)
             {
-                SpinWait.SpinUntil(() => _server.Description.State == ServerState.Connected, SingleServerSelectionTimeoutMS);
+                using var serverSelectionContext = operationContext.WithTimeout(_serverSelectionTimeout);
```
papafe, line 90: What do you think about this approach? The risk of just putting the usual server timeout is that this will go on for 30s even if the operation was cancelled or timed out.

sanych-sun: It's good idea, it helps to calculate the actual timeout.

src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
```diff
@@ -24,18 +24,18 @@ namespace MongoDB.Driver.Core.Bindings
...
         private readonly IServer _server;
+        private readonly TimeSpan _serverSelectionTimeout;
         private readonly ICoreSessionHandle _session;
 
-        public SingleServerReadBinding(IServer server, ReadPreference readPreference, ICoreSessionHandle session)
+        public SingleServerReadBinding(IServer server, ReadPreference readPreference, ICoreSessionHandle session, TimeSpan serverSelectionTimeout)
         {
             _server = Ensure.IsNotNull(server, nameof(server));
             _readPreference = Ensure.IsNotNull(readPreference, nameof(readPreference));
             _session = Ensure.IsNotNull(session, nameof(session));
```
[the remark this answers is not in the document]
sanych-sun: Agree with CoPilot.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
```diff
@@ -79,15 +79,22 @@ public void Dispose()
...
             // server might be in unknown state due to previous failed operation, allow description to be updated
             // this is done instead of server selection
             // TODO parameterize timeout and avoid busy wait, or offload waiting to server level
             // Should be addressed by CSHARP-3556
             if (_server.Description.State == ServerState.Disconnected)
             {
-                SpinWait.SpinUntil(() => _server.Description.State == ServerState.Connected, SingleServerSelectionTimeoutMS);
+                using var serverSelectionContext = operationContext.WithTimeout(_serverSelectionTimeout);
+
+                SpinWait.SpinUntil(() =>
```
sanych-sun: Should we throw in case the condition was not meet until the timeout? It will throw anyway. Current exception reads as "Connection pool is in paused state", but we can throw something more readable here and more aligned with what we throw if server selection is timed out.

papafe: Makes sense, fixed it.

src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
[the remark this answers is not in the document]
papafe: Done.

src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
sanych-sun: `SpinWait.SpinUntil` returns bool value as far as I remember: `true` - is the condition was met, `false` - if the timeout reached. We can collapse the code a little:

```
if(SpinWait.SpinUntil() => ...)
{
    serverSelectionContext.CancellationToken.ThrowIfCancellationRequested();
}
else
{
    throw new TimeoutException(...);
}
```

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Bindings/SingleServerReadBinding.cs
papafe: Ah, makes sense, changed it.

sanych-sun (approved):
LGTM

PR #1929, opened 2026-03-27
CSHARP-2150 Add check that the serializer's ValueType matches the type when registering the serializer

kyra-rk:
`RegisterSerializer` and `TryRegisterSerializer` now throw an `ArgumentException` if the serializer's `ValueType` is not compatible with the registered type (e.g. registering an `Int32Serializer` for `typeof(long)`). Serializers registered for a base type remain valid for derived types.

### DIFF
--- a/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
@@ -141,6 +141,12 @@ namespace MongoDB.Bson.Serialization
             }
             EnsureRegisteringASerializerForThisTypeIsAllowed(type);
 
+            if (type != serializer.ValueType)
+            {
+                var message = string.Format("A serializer of type {0} cannot be registered for type {1}.", BsonUtils.GetFriendlyTypeName(serializer.ValueType), BsonUtils.GetFriendlyTypeName(type));
+                throw new BsonSerializationException(message);
+            }
+
             if (_cache.TryAdd(type, serializer))
             {
                 return true;
--- a/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
@@ -191,7 +191,7 @@ namespace MongoDB.Bson.Serialization
                 var message = string.Format("Generic type {0} has unassigned type parameters.", BsonUtils.GetFriendlyTypeName(type));
                 throw new ArgumentException(message, "type");
             }
-            if (type != serializer.ValueType)
+            if (!serializer.ValueType.IsAssignableFrom(type))
             {
                 throw new BsonSerializationException($"A serializer for {BsonUtils.GetFriendlyTypeName(serializer.ValueType)} cannot be registered for type {BsonUtils.GetFriendlyTypeName(type)}.");
             }
--- a/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
+++ b/src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
@@ -93,13 +93,7 @@ namespace MongoDB.Bson.Serialization
             {
                 throw new ArgumentNullException("serializer");
             }
-            EnsureRegisteringASerializerForThisTypeIsAllowed(type);
-
-            if (type != serializer.ValueType)
-            {
-                var message = string.Format("A serializer of type {0} cannot be registered for type {1}.", BsonUtils.GetFriendlyTypeName(serializer.ValueType), BsonUtils.GetFriendlyTypeName(type));
-                throw new BsonSerializationException(message);
-            }
+            EnsureRegisteringASerializerForThisTypeIsAllowed(serializer, type);
 
             if (!_cache.TryAdd(type, serializer))
             {
@@ -139,13 +133,7 @@ namespace MongoDB.Bson.Serialization
             {
                 throw new ArgumentNullException(nameof(serializer));
             }
-            EnsureRegisteringASerializerForThisTypeIsAllowed(type);
-
-            if (type != serializer.ValueType)
-            {
-                var message = string.Format("A serializer of type {0} cannot be registered for type {1}.", BsonUtils.GetFriendlyTypeName(serializer.ValueType), BsonUtils.GetFriendlyTypeName(type));
-                throw new BsonSerializationException(message);
-            }
+            EnsureRegisteringASerializerForThisTypeIsAllowed(serializer, type);
 
             if (_cache.TryAdd(type, serializer))
             {
@@ -190,7 +178,7 @@ namespace MongoDB.Bson.Serialization
             throw new BsonSerializationException(message);
         }
 
-        private void EnsureRegisteringASerializerForThisTypeIsAllowed(Type type)
+        private void EnsureRegisteringASerializerForThisTypeIsAllowed(IBsonSerializer serializer, Type type)
         {
             var typeInfo = type.GetTypeInfo();
             if (typeof(BsonValue).GetTypeInfo().IsAssignableFrom(type))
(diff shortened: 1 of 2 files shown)


src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -95,6 +95,12 @@ public void RegisterSerializer(Type type, IBsonSerializer serializer)
             }
             EnsureRegisteringASerializerForThisTypeIsAllowed(type);
 
+            if (type != serializer.ValueType)
```
papafe: Very small thing, but nowadays we mostly use string interpolation for formatting messages.

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -95,6 +95,12 @@ public void RegisterSerializer(Type type, IBsonSerializer serializer)
...
+            {
+                var message = string.Format("A serializer of type {0} cannot be registered for type {1}.", BsonUtils.GetFriendlyTypeName(serializer.ValueType), BsonUtils.GetFriendlyTypeName(type));
```
papafe: The message is a little confusing, it should be something like "A serializer for {--} cannot be registered for type {---}".

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
```diff
@@ -97,6 +97,19 @@ public void RegisterSerializer_should_throw_when_serializer_is_already_registere
             exception.Message.Should().Contain("There is already a serializer registered for type Object");
         }
 
+        [Fact]
```
papafe, line 100: I would say that normally one test should be for one method, so you probably need two tests.

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -135,6 +141,12 @@ public bool TryRegisterSerializer(Type type, IBsonSerializer serializer)
             }
             EnsureRegisteringASerializerForThisTypeIsAllowed(type);
```
papafe, line 137: I would put the input validation you're doing before "Ensure...", also in the other method, together with the rest of the input validation.

papafe (changes requested):

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
kyra-rk: Soungds good

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
kyra-rk: I originally did it this way to match the exception messages in the same function, but I updated it to use string interpolation—is this considered an OK design variation? (I know usually we tend to match the style to the existing code in the file.)

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -191,6 +191,10 @@ private void EnsureRegisteringASerializerForThisTypeIsAllowed(Type type)
                 var message = string.Format("Generic type {0} has unassigned type parameters.", BsonUtils.GetFriendlyTypeName(type));
                 throw new ArgumentException(message, "type");
             }
+            if (type != serializer.ValueType)
```
papafe: I think that it would make sense to do this validation before this method, as it does a different thing, as it was done before. 
it just needs to happen after the other validations and before this method is called.

kyra-rk: Ah, I see what you meant now. I tried moving it, but it causes a few tests to fail since `EnsureRegisteringASerializerForThisTypeIsAllowed` is expected to do some validation too. For example `TryRegisterSerializer_should_throw_when_type_is_not_closed_generic_type` fails because the mock serializer returns `null` for `ValueType` by default, so the type != `serializer.ValueType` check fires early and calls `BsonUtils.GetFriendlyTypeName(null)` to throw an `ArgumentNullException` before we ever reach the open-generic check. 

I could update these tests to avoid mock serializers, but I think it makes more logical sense to validate the type independently before before checking that it matches the serializer. I'd prefer keeping it where it is (or moving it out of and after the function), but lmk if you have a different opinion.

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
```diff
@@ -97,6 +97,78 @@ public void RegisterSerializer_should_throw_when_serializer_is_already_registere
...
+        private class Person
+        {
+            public string Name { get; set; }
+        }
+
+        private class PeopleSerializer : SerializerBase<IEnumerable<Person>>
+        {
+            public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, IEnumerable<Person> value)
+            {
+                context.Writer.WriteStartArray();
```
papafe: We're not actually using this implementation, just throw a `NotImplementedException`.

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
```diff
@@ -97,6 +97,78 @@ public void RegisterSerializer_should_throw_when_serializer_is_already_registere
...
+        [Fact]
+        public void TryRegisterSerializer_should_throw_when_type_and_serializer_do_not_match()
+        {
+            var subject = new BsonSerializerRegistry();
+            var intSerializer = new Int32Serializer();
+
+            var tryException = Record.Exception(() => subject.TryRegisterSerializer(typeof(long), intSerializer));
+
+            tryException.Should().BeOfType<BsonSerializationException>();
+            tryException.Message.Should().Contain("A serializer for Int32 cannot be registered for type Int64.");
```
papafe: If you use the full string comparison it should use `Be` instead of `Contain`

src/MongoDB.Bson/Serialization/BsonSerializerRegistry.cs
```diff
@@ -191,6 +191,10 @@ private void EnsureRegisteringASerializerForThisTypeIsAllowed(Type type)
                 var message = string.Format("Generic type {0} has unassigned type parameters.", BsonUtils.GetFriendlyTypeName(type));
                 throw new ArgumentException(message, "type");
             }
+            if (!serializer.ValueType.IsAssignableFrom(type))
```
papafe, line 200: I would move this check in his own method, as it's not doing what the rest of the method is doing. 
Also this should be an ArgumentException, as it's more related to the arguments being passed.

papafe (changes requested):
Overall it looks good, some minor comments.

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
kyra-rk: Nice catch! Will fix

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
```diff
@@ -97,6 +97,68 @@ public void RegisterSerializer_should_throw_when_serializer_is_already_registere
...
+            var subject = new BsonSerializerRegistry();
+            var serializer = new PeopleSerializer();
+
+            var result = subject.TryRegisterSerializer(typeof(List<Person>), serializer);
+            result.Should().BeTrue();
+
+            subject.GetSerializer(typeof(List<Person>)).Should().BeSameAs(serializer);
+        }
+
+        private class Person
```
papafe: Sorry for noticing before, usually we keep utility class and other utility methods after all the test methods, so at the end of the whole test class.

papafe (approved):
LGTM! Just one minor thing

tests/MongoDB.Bson.Tests/Serialization/BsonSerializerRegistryTests.cs
kyra-rk: No worries, will change it!

PR #1932, opened 2026-04-01
CSHARP-5952: SerializerFinder resolve wrong serializer for BsonDocument members

sanych-sun:
LINQ Provider does not support translation of BsonDocument members other then indexer and properties inhireted from BsonValue. This PR fixes SerializerFinder bug that resolved BsonValueSerializer for BsonDocument properties (like ElementCount), it should return UnknowableSerializer instead.


src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
```diff
@@ -46,6 +46,7 @@ protected override Expression VisitMember(MemberExpression node)
 
                 var resultSerializer = node.Member switch
                 {
+                    _ when typeof(BsonDocument).IsAssignableFrom(declaringType) && !node.Member.IsIndexer() => UnknowableSerializer.Create(node.Type), // we do not support translation of BsonDocument properties other than indexer
```
adelinowona: Is the `IsIndexer()` check here necessary? I think in C# expression trees, indexer access (`d["name"]`) is represented as a `MethodCallExpression` (calling `get_Item`), not a `MemberExpression`. So `VisitMember` should never see an indexer. The check is harmless but this maybe a code path that never executes. Did you verify this?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/BsonDocumentTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        var exception = Record.Exception(() => serializerMap.GetSerializer(expression.Body));
+        exception.Should().BeOfType<ExpressionNotSupportedException>();
+    }
+
+    public static readonly object[][] SupportedTestCases =
+    [
+        [TestHelpers.MakeLambda((BsonDocument d) => d["name"]), BsonValueSerializer.Instance]
+    ];
+
+    public static readonly object[][] NotSupportedTestCases =
```
adelinowona, line 61: Should we also add test cases using `BsonDocument` subclasses to validate the `IsAssignableFrom` behavior?

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/BsonDocumentTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        serializerMap.IsKnown(expression.Body, out var serializer).Should().BeTrue();
+        serializer.Should().BeOfType(typeof(UnknowableSerializer<>).MakeGenericType(expression.Body.Type));
+        var exception = Record.Exception(() => serializerMap.GetSerializer(expression.Body));
+        exception.Should().BeOfType<ExpressionNotSupportedException>();
+    }
+
+    public static readonly object[][] SupportedTestCases =
+    [
+        [TestHelpers.MakeLambda((BsonDocument d) => d["name"]), BsonValueSerializer.Instance]
```
adelinowona, line 58: Just noting that this doesn't exercise the code path you changed since Indexer will go through VisitMethodCall instead of VisitMember. Still good to have this here though.

adelinowona (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMember.cs
sanych-sun: Nice catch! Thank you. Removed.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/BsonDocumentTests.cs
sanych-sun: Done.

adelinowona (approved):

PR #1933, opened 2026-04-02
CSHARP-5769: Implement hasAncestor, hasRoot, and returnScope for Atlas Search

ajcvickers:
Add hasAncestor and hasRoot search operators for use within embeddedDocument queries, allowing searches to reference fields at ancestor or root document levels. Add returnScope support to $search and $searchMeta pipeline stages, enabling results to be returned from a nested embedded document scope rather than the root document. Add Clone() to SearchOptions and its nested options classes to avoid mutating caller-supplied options when returnScope is set.


src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1437,16 +1437,47 @@ public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
...
+                    {
+                        outputSerializer = (IBsonSerializer<TOutput>)args.DocumentSerializer;
+                    }
+                    else
+                    {
+                        outputSerializer = args.SerializerRegistry.GetSerializer<TOutput>();
+                        renderedSearchDefinition.Add("returnScope", new BsonDocument { { "path", returnScope!.Render(args).FieldName } });
+                        searchOptions = searchOptions.Clone();
+                        searchOptions.ReturnStoredSource = true;
+                    }
```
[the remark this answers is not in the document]
ajcvickers: You are wrong. The clone method ensures that the options passed in are not mutated.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1437,16 +1437,47 @@ public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
...
+        /// <typeparam name="TOutput">The type of the output documents.</typeparam>
+        /// <param name="searchDefinition">The search definition.</param>
+        /// <param name="returnScope">The level of nested documents to return.</param>
+        /// <param name="searchOptions">The search options.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TOutput> Search<TInput, TOutput>(
+            SearchDefinition<TInput> searchDefinition,
+            FieldDefinition<TInput, IEnumerable<TOutput>> returnScope,
+            SearchOptions<TInput> searchOptions)
         {
```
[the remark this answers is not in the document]
ajcvickers: This is not appropriate. We need to be testing against a functional server, in which has this is fully validated.

src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -440,15 +441,71 @@ IAggregateFluent<TResult> Search(
...
+        /// <param name="searchDefinition">The search definition.</param>
+        /// <param name="returnScope">The level of nested documents to return.</param>
+        /// <param name="searchOptions">The search options.</param>
+        /// <returns>
+        /// The fluent aggregate interface.
+        /// </returns>
+        IAggregateFluent<TNewResult> Search<TNewResult>(
+            SearchDefinition<TResult> searchDefinition,
+            Expression<Func<TResult, IEnumerable<TNewResult>>> returnScope,
+            SearchOptions<TResult> searchOptions = null);
```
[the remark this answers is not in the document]
ajcvickers: We should be fine breaking this going forward, or we need to discuss different designs.

Agents.md
```diff
@@ -0,0 +1,47 @@
...
+- Run all tests: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0`
+- Run a single test class: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ClassName"`
+
+A MongoDB connection is always available locally, so all tests can be run without filtering. Some test suites also require additional environment variables — if you need to run those tests and the variables are not set, stop and tell the user which variables are needed rather than working around it.
+
+| Feature area | Required environment variables |
+|---|---|
+| Atlas Search | `ATLAS_SEARCH_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| Atlas Search index helpers | `ATLAS_SEARCH_INDEX_HELPERS_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| CSFLE / auto-encryption | `CRYPT_SHARED_LIB_PATH` |
```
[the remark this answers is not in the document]
ajcvickers: Fixed the text. You are wrong about the markdown.

Agents.md
```diff
@@ -0,0 +1,47 @@
+# Agents.md - CSharpDriver
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multitargetted from .NET Framework 4.7.2 through .NET 10
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
```diff
@@ -280,10 +281,185 @@ public class AtlasSearchTests : LoggableTestClass
...
+                    }
+                    """));
+
+            var collection = GetReturnScopeCollection<BsonDocument>();
+            if (!collection.AsQueryable().Any())
+            {
+                collection.InsertMany(BsonSerializer.Deserialize<BsonArray>(directors).Select(e => e.AsBsonDocument));
+            }
+
+            var indexModels = new[] { returnScopeIndex1, returnScopeIndex2, returnScopeIndex3 };
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1437,16 +1437,47 @@ public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
...
                 args =>
                 {
                     ClientSideProjectionHelper.ThrowIfClientSideProjection(args.DocumentSerializer, operatorName);
                     var renderedSearchDefinition = searchDefinition.Render(args);
+
+                    IBsonSerializer<TOutput> outputSerializer;
+                    if (returnScope == null)
+                    {
+                        outputSerializer = (IBsonSerializer<TOutput>)args.DocumentSerializer;
+                    }
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -440,15 +441,71 @@ IAggregateFluent<TResult> Search(
...
+        /// <summary>
+        /// Appends a $search stage to the pipeline, returning documents from a nested scope.
+        /// </summary>
+        /// <param name="searchDefinition">The search definition.</param>
+        /// <param name="returnScope">The level of nested documents to return.</param>
+        /// <param name="searchOptions">The search options.</param>
+        /// <returns>
+        /// The fluent aggregate interface.
+        /// </returns>
+        IAggregateFluent<TNewResult> Search<TNewResult>(
```
sanych-sun: I think we easily can avoid this breaking change by implementing it as an extensions method that uses `AppendStage`. We can consider moving this back to the interface as a part of v4 release.

ajcvickers: Done.

Agents.md
```diff
@@ -0,0 +1,47 @@
+# Agents.md - CSharpDriver
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multi-targeted from .NET Framework 4.7.2 through .NET 10
```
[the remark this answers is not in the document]
ajcvickers: Removed from this PR.

sanych-sun (changes requested):

src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq.Expressions;
```
sanych-sun: It seems like not necessary change.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
```diff
@@ -280,10 +281,186 @@ public class AtlasSearchTests : LoggableTestClass
         #endregion
 
         private readonly IMongoClient _mongoClient;
+        private readonly EventCapturer _eventCapturer;
+        private static bool __indexesCreated;
+        private static readonly string __databaseUniquifier = Guid.NewGuid().ToString();
```
sanych-sun: Should we move this 2 variables to the other static variables inside the static region or rename/delete the region?

sanych-sun: Why do we need that unique suffix in the database name? Is this cluster shared and permanent? If so - should we drop the generated collection when tests are done?

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1437,16 +1437,53 @@ public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
...
+                        {
+                            throw new InvalidOperationException(
+                                $"The search output type '{typeof(TOutput).Name}' must be the same as the input type '{typeof(TInput).Name}' when 'returnScope' is not specified. Use the overload that specifies 'returnScope' to return documents of a nested collection type.");
+                        }
+
+                        outputSerializer = (IBsonSerializer<TOutput>)args.DocumentSerializer;
+                    }
+                    else
+                    {
+                        outputSerializer = args.SerializerRegistry.GetSerializer<TOutput>();
```
sanych-sun: As far as I understood the returnScope is a path inside of the parent document. In such case we have a little more info to resolve the proper serializer. Instead of Lookup, we can try to get member info if DocumentSerializer is IBsonDocumentSerializer. See TryGetMemberSerializationInfo method.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1437,16 +1437,53 @@ public static PipelineStageDefinition<TInput, TInput> Search<TInput>(
...
+                        renderedSearchDefinition.Add("returnScope", new BsonDocument { { "path", returnScope.Render(args).FieldName } });
+                        searchOptions = searchOptions.Clone();
+                        searchOptions.ReturnStoredSource = true;
```
sanych-sun: I'm not sure if I like the idea to alter user's provided options silently. I know it's a clone. But may be we should simply document it "if your search request uses returnScope, make sure to set ReturnStoredSource to true." BTW, why do we need to set it to true? Does it described somewhere?

sanych-sun: Also am I right to think that we need all this clone-and-update stuff, only to use in like a 10 lines of code under? May be we can have a local variable instead?

src/MongoDB.Driver/Search/SearchTrackingOptions.cs
```diff
@@ -34,6 +34,12 @@ public string SearchTerms
             set => _searchTerms = Ensure.IsNotNullOrEmpty(value, nameof(value));
         }
 
+        /// <summary>
+        /// Creates a clone of the options.
+        /// </summary>
+        /// <returns>A clone of the options.</returns>
+        public SearchTrackingOptions Clone() => new() { SearchTerms = SearchTerms };
```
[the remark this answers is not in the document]
ajcvickers: Updated.

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -313,6 +313,50 @@ public SearchDefinition<TDocument> GeoWithin<TCoordinates>(
...
                 new GeoWithinSearchDefinition<TDocument, TCoordinates>(path, area, score);
 
+        /// <summary>
+        /// Creates a search definition that moves the context back up the tree of embedded documents
+        /// to one of the ancestors of the current embedded document such that ancestor fields can be used in searches
+        /// of embedded documents.
+        /// </summary>
+        /// <typeparam name="TField">The type of the ancestor documents.</typeparam>
+        /// <param name="path">The path from the root to the ancestor.</param>
+        /// <param name="operator">The operator to execute at the root.</param>
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
ajcvickers: I'm trying to move in the direction of proper shared resource management, without any needed pre-configuration. The current code isn't very clean in this respect because it's not currently the way these tests are set up. For now, I can make sure that the database is deleted after all the tests that use it have run. (I generally don't like this approach for a few reasons, but it would be okay here for now.) The database name needs to be unique because if the same tests are immediately run again (e.g. Run until Fail in Rider), then the server can get confused between the old and the new database--we saw this a lot on EF tests.

ajcvickers: Will move.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
ajcvickers: Will do.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
ajcvickers: The reason it gets automatically set is that, conceptually, returnScope is partially an enhancement to ReturnStoredSource. So if you use returnScope, then you are implicitly using ReturnStoredSource. So if we let the server throw, we're basically saying, "We know you're using stored source, because you're using returnScope, but you didn't set the other flag to tell us this, so we're now going to fail." This is generally bad API design.

On the other hand, if it were reasonable, conceptually, to choose different values here, then it would be appropriate to pass whatever value is set through to the server and let it decide if the combination is valid or not. We can still do this, it would just be slightly annoying to users specifying essentially the same thing twice.

Since it's not clear which approach we should take here, I'll switch to the second approach, which is less code for us, and isn't a breaking change if, somehow, both are allowed in the future.

ajcvickers: The clone is needed to avoid mutating the settings that the application has a reference to, which would be very bad form. (Hence settings are very often immutable.) I don't follow what you mean about using a local variable.

src/MongoDB.Driver/IAggregateFluent.cs
ajcvickers: Fixed.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
ajcvickers: Added a temporary cleanup mechanism. Claude says:

 2. Countdown cleanup pattern: The __returnScopeTestCount = 12 counter, decremented in try/finally blocks, and checked in DisposeInternal, is a workable but fragile approach:
    - The hardcoded 12 must be kept in sync with the test count manually.                                                                                                                                                                                                                                                                                                  
    - The decrement/disposal comments (e.g., "3 & 4") correctly account for [Theory] running twice.                                                                                                                                                                                                                                                                        
    - No thread-safety (non-volatile, no Interlocked), but fine for sequential test execution.                                                                                                                                                                                                                                                                             
    - The DisposeInternal check on the counter is sound: the finally blocks run before DisposeInternal, so the counter hits 0 before the disposing call that triggers the DB drop.                                                                                                                                                                                         
    - Comments mark this as temporary, which is appropriate.

sanych-sun: If we need to handle shared resources for test class in xUnit, I believe we have to use ClassFixture (initialize the database/collection in ctor and remove them in Dispose). We sure can refactor this outside of this PR, but probably better to have a Jira ticket for that.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
sanych-sun: About local variable. What if instead of clone and update options approach we will go something like:
```
var shouldReturnStoredSource = options.ReturnStoredSource;
if (returnScope != null)
{
   shouldReturnStoredSource = true;
}

... skipped lines until 1490 where we actually render the value.
renderedSearchDefinition.Add("returnStoredSource", shouldReturnStoredSource, shouldReturnStoredSource);

```
This way we do not need to clone, but we will include the `returnStoredSource` automatically.

tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs
sanych-sun: Just read your comment in code - yep, agree. In this commit it's OK to go as is, but let's create a Jira ticket to implement the Fixture.

src/MongoDB.Driver/Search/SearchOptions.cs
```diff
@@ -69,5 +69,24 @@ public sealed class SearchOptions<TDocument>
         /// When set, the search retrieves documents starting immediately before the specified reference point.
         /// </summary>
         public string SearchBefore { get; set; }
+
+        /// <summary>
+        /// Creates a clone of the options.
+        /// </summary>
+        /// <returns>A clone of the options.</returns>
+        public SearchOptions<TDocument> Clone()
```
sanych-sun: Do we still need this and other added Clone methods as I cannot find if it's still in use?

sanych-sun (changes requested):

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
sanych-sun: Great! It took me some time to understand your solution, but you are right. Use of renderedField.ValueSerializer is even better.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
ajcvickers: Yeah, we could do that in this case. But since I changed the code here, we don't need to clone anymore. Nevertheless, in general settings should always be immutable or have a way of cloning, because modifying base settings without clobbering the base is a commonly needed pattern. Ideally, settings would be immutable, and hence we wouldn't be having this discussion.

src/MongoDB.Driver/Search/SearchOptions.cs
ajcvickers: No, we don't "need" it, but having settings that are mutable and for which there is no way to clone is an anti-pattern.

ajcvickers: @sanych-sun Clone method removed.

sanych-sun (approved):
LGTM

PR #1934, opened 2026-04-02
CSHARP-5943: Initial small AGENTS.md file for the C# driver

ajcvickers:
See notes in DRIVERS-3428.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,47 @@
+# AGENTS.md - CSharpDriver
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multi-targeted to various .NET versions from .NET Framework 4.7.2 up
+- xUnit + FluentAssertions for testing
+
+## Project Structure
+- `src/MongoDB.Bson/` - BSON for MongoDB
+- `src/MongoDB.Driver/` - C# driver for MongoDB
+- `src/MongoDB.Driver.Encryption/` - Client encryption (CSFLE with KMS).
+- `src/MongoDB.Driver.Authentication.AWS/` - AWS IAM authentication
+- `tests/MongoDB.Driver.Tests/` - Main C# driver tests
+- `tests/MongoDB.Bson.Tests/` - BSON handling tests
+- `tests/*/TestHelpers` - Common test utilities
+- `tests/*` - Specialized tests; less common
+- `tests/MongoDB.Driver.Tests/Specifications/` are JSON-driven tests using a common runner.
+
+## Commands
+- Build: `dotnet build CSharpDriver.sln`
+- Run all tests: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0`
+- Run a single test class: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ClassName"`
+
+A MongoDB connection is always available locally, so "integration" tests can be run as well as unit tests. Some test suites also require additional environment variables — if you need to run those tests and the variables are not set, stop and tell the user which variables are needed rather than working around it.
+
+| Feature area | Required environment variables |
+|---|---|
+| Atlas Search | `ATLAS_SEARCH_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| Atlas Search index helpers | `ATLAS_SEARCH_INDEX_HELPERS_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| CSFLE / auto-encryption | `CRYPT_SHARED_LIB_PATH` |
+| CSFLE with KMS mock servers | `KMS_MOCK_SERVERS_ENABLED` |
+| CSFLE with AWS KMS | `CSFLE_AWS_TEMPORARY_CREDS_ENABLED` |
+| CSFLE with Azure KMS | `CSFLE_AZURE_KMS_TESTS_ENABLED` |
+| CSFLE with GCP KMS | `CSFLE_GCP_KMS_TESTS_ENABLED` |
+| AWS authentication | `AWS_TESTS_ENABLED` |
+| GSSAPI / Kerberos | `GSSAPI_TESTS_ENABLED`, `AUTH_HOST`, `AUTH_GSSAPI` |
+| OIDC authentication | `OIDC_ENV` |
+| X.509 authentication | `MONGO_X509_CLIENT_CERTIFICATE_PATH`, `MONGO_X509_CLIENT_CERTIFICATE_PASSWORD` |
+| PLAIN authentication | `PLAIN_AUTH_TESTS_ENABLED` |
+| SOCKS5 proxy | `SOCKS5_PROXY_SERVERS_ENABLED` |
+
+## Commit and PR Conventions
+
+- Commit and PR messages start with a JIRA number: `CSHARP-1234: Description`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,12 +19,17 @@ The C# driver for MongoDB.
 - `tests/*` - Specialized tests; less common
 - `tests/MongoDB.Driver.Tests/Specifications/` are JSON-driven tests using a common runner.
 
+## Editing
+- Be careful to preserve file BOMs.
+
 ## Commands
 - Build: `dotnet build CSharpDriver.sln`
 - Run all tests: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0`
 - Run a single test class: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ClassName"`
 
-A MongoDB connection is always available locally, so "integration" tests can be run as well as unit tests. Some test suites also require additional environment variables — if you need to run those tests and the variables are not set, stop and tell the user which variables are needed rather than working around it.
+## Testing
+- Tests cannot be run in parallel.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+- `tests/*/TestHelpers` - Common test utilities
+- `tests/*` - Specialized tests; less common
+- `tests/MongoDB.Driver.Tests/Specifications/` are JSON-driven tests using a common runner.
+
+## Commands
+- Build: `dotnet build CSharpDriver.sln`
+- Run all tests: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0`
+- Run a single test class: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ClassName"`
+
+A MongoDB connection is always available locally, so "integration" tests can be run as well as unit tests. Some test suites also require additional environment variables — if you need to run those tests and the variables are not set, stop and tell the user which variables are needed rather than working around it.
```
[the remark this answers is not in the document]
ajcvickers: Local developers must ensure the Mongo connection is available. It's unlikely this will just be the default localhost going forward (which is likely to be removed) so this isn't really a good update.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+| Feature area | Required environment variables |
+|---|---|
+| Atlas Search | `ATLAS_SEARCH_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| Atlas Search index helpers | `ATLAS_SEARCH_INDEX_HELPERS_TESTS_ENABLED`, `ATLAS_SEARCH_URI` |
+| CSFLE / auto-encryption | `CRYPT_SHARED_LIB_PATH` |
```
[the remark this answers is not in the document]
ajcvickers: You are wrong.

Agents.md
```diff
@@ -0,0 +1,47 @@
+# Agents.md - CSharpDriver
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

Agents.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multi-targeted from .NET Framework 4.7.2 through .NET 10
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

Agents.md
```diff
@@ -0,0 +1,47 @@
...
+- xUnit + FluentAssertions for testing
+
+## Project Structure
+- `src/MongoDB.Bson/` - BSON for MongoDB
+- `src/MongoDB.Driver/` - C# driver for MongoDB
+- `MongoDB.Driver.Encryption` - Client encryption (CSFLE with KMS).
+- `MongoDB.Driver.Authentication.AWS` - AWS IAM authentication
```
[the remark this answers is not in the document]
ajcvickers: Fixed.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
+# AGENTS.md - CSharpDriver
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multi-targeted to various .NET versions from .NET Framework 4.7.2 up
+- xUnit + FluentAssertions for testing
```
papafe, line 9: This part about FluentAssertions.. I think we need to discuss about it in the team. Sometime ago we had a discussion about using the basic assertions against the fluent ones.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+| AWS authentication | `AWS_TESTS_ENABLED` |
+| GSSAPI / Kerberos | `GSSAPI_TESTS_ENABLED`, `AUTH_HOST`, `AUTH_GSSAPI` |
+| OIDC authentication | `OIDC_ENV` |
+| X.509 authentication | `MONGO_X509_CLIENT_CERTIFICATE_PATH`, `MONGO_X509_CLIENT_CERTIFICATE_PASSWORD` |
+| PLAIN authentication | `PLAIN_AUTH_TESTS_ENABLED` |
+| SOCKS5 proxy | `SOCKS5_PROXY_SERVERS_ENABLED` |
+
+## Commit and PR Conventions
+
+- Commit and PR messages start with a JIRA number: `CSHARP-1234: Description`
```
papafe: This is true only for the first commit in a branch, doesn't it? 
In my file I also have that branch names contain the JIRA ticket number. And also that it can use the ticket number from a branch to look up context from JIRA when needed.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+- `tests/MongoDB.Driver.Tests/` - Main C# driver tests
+- `tests/MongoDB.Bson.Tests/` - BSON handling tests
+- `tests/*/TestHelpers` - Common test utilities
+- `tests/*` - Specialized tests; less common
+- `tests/MongoDB.Driver.Tests/Specifications/` are JSON-driven tests using a common runner.
+
+## Commands
+- Build: `dotnet build CSharpDriver.sln`
+- Run all tests: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0`
+- Run a single test class: `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net10.0 --filter "FullyQualifiedName~ClassName"`
```
papafe, line 28: In my file I also have a note about the fact that tests cannot be run in parallel.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Overview
+The C# driver for MongoDB.
+
+## Tech Stack
+- .NET library projects producing NuGet packages
+- Multi-targeted to various .NET versions from .NET Framework 4.7.2 up
+- xUnit + FluentAssertions for testing
+
+## Project Structure
```
papafe, line 11: Something that I found useful was also to give the path to the specification repo that I have locally. It's super useful for certain kind of feature implementation.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+| CSFLE with AWS KMS | `CSFLE_AWS_TEMPORARY_CREDS_ENABLED` |
+| CSFLE with Azure KMS | `CSFLE_AZURE_KMS_TESTS_ENABLED` |
+| CSFLE with GCP KMS | `CSFLE_GCP_KMS_TESTS_ENABLED` |
+| AWS authentication | `AWS_TESTS_ENABLED` |
+| GSSAPI / Kerberos | `GSSAPI_TESTS_ENABLED`, `AUTH_HOST`, `AUTH_GSSAPI` |
+| OIDC authentication | `OIDC_ENV` |
+| X.509 authentication | `MONGO_X509_CLIENT_CERTIFICATE_PATH`, `MONGO_X509_CLIENT_CERTIFICATE_PASSWORD` |
+| PLAIN authentication | `PLAIN_AUTH_TESTS_ENABLED` |
+| SOCKS5 proxy | `SOCKS5_PROXY_SERVERS_ENABLED` |
+
```
papafe, line 49: Another line I have:

"**Do not strip BOMs**: Source files may have a UTF-8 BOM (`U+FEFF`). Prefer using the `Edit` tool over `Write` to avoid accidentally removing it."

(not sure if it's still an issue)

AGENTS.md
ajcvickers: I have added FluentAssertions to my list of things to bring up in retrospective. For now, I think we should leave it--Claude explicitly called out that it was useful.

AGENTS.md
ajcvickers: I will clarify that it is the first commit.

Sometimes branch names don't match--even Robert had names with "b" appended or similar occasionally. That's why I didn't add it, but I can, and we can see if any exceptions cause real issues.

AGENTS.md
ajcvickers: Have you found this to be an issue? I would expect that Claude would not try to do that given that tests are not configured that way by default.

AGENTS.md
ajcvickers: Interesting. I think we would have to standardize on the location to put that here, but it could be a good candidate for each of our user-specific agents file.

AGENTS.md
ajcvickers: What issues have we had with BOMs?

papafe: On some files both Claude and OpenCode were removing BOMs, so I was getting a change on the first line of some files.

AGENTS.md
papafe: I agree, for sure we should not standardize the location, but it's a good add.

AGENTS.md
papafe: It happened to me when it was trying to run tests on multiple classes, I think it had multiple sub-agents trying to run tests in parallel so I had to add it.

AGENTS.md
sanych-sun: Hm... just to leave the record: I do remember discussion about FluentAssertion turned to be commercial, and we should either stop using it, or pin the version and never bump it until it possible. But I do not remember if we really made any decisions. We also discussed possibility to switch to [Shouldly](https://docs.shouldly.org/).

papafe: Exactly what you said. I also think that at the time we didn't decide which direction to take though.

papafe: I think this is the only more "contentious point". Should we leave it out and let the AI check the follow the conventions that are already in the file? Or check similar tests/tests for similar classes.

papafe (changes requested):

ajcvickers:
I don't think the FluentAssertions line should be contentious. It's accurate for the current code base and Claude indicated it was useful. If we start transitioning away from using FluentAssertions, then we should remove it and tell it to favor whatever we are moving to.

papafe (approved):
Looks good!

A suggestion: we probably can add `CLAUDE.local.md` to our gitignore, so we can keep our own file still

ajcvickers:
> A suggestion: we probably can add CLAUDE.local.md to our gitignore, so we can keep our own file still

I'm going to merge this so that we have something in, then I'll include this in a new PR with some other updates.

PR #1936, opened 2026-04-02
CSHARP-5828: Add Rerank stage builder

adelinowona:


src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1604,6 +1604,94 @@ public static PipelineStageDefinition<TInput, TOutput> RankFusion<TInput, TOutpu
...
+        /// <param name="query">The rerank query.</param>
+        /// <param name="path">The field to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank (1–1000).</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Rerank<TInput, TField>(
+            RerankQuery query,
+            Expression<Func<TInput, TField>> path,
+            int numDocsToRerank,
+            string model)
```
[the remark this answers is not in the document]
adelinowona: Fixed.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1604,6 +1604,94 @@ public static PipelineStageDefinition<TInput, TOutput> RankFusion<TInput, TOutpu
...
+        public static PipelineStageDefinition<TInput, TInput> Rerank<TInput>(
+            RerankQuery query,
+            FieldDefinition<TInput> path,
+            int numDocsToRerank,
+            string model)
+            => Rerank(
+                query,
+                [path],
+                numDocsToRerank,
+                model);
```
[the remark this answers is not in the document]
adelinowona: Fixed.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1604,6 +1604,94 @@ public static PipelineStageDefinition<TInput, TOutput> RankFusion<TInput, TOutpu
...
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Rerank<TInput>(
+            RerankQuery query,
+            IEnumerable<FieldDefinition<TInput>> paths,
+            int numDocsToRerank,
+            string model)
+        {
+            Ensure.IsNotNull(query, nameof(query));
+            Ensure.IsNotNull(paths, nameof(paths));
+            Ensure.IsGreaterThanZero(numDocsToRerank, nameof(numDocsToRerank));
```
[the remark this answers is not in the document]
adelinowona: Fixed. removed the limit numbers in the comments

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -1081,6 +1081,104 @@ public static PipelineDefinition<TInput, TOutput> RankFusion<TInput, TIntermedia
...
+        /// <param name="pipeline">The pipeline.</param>
+        /// <param name="query">The rerank query.</param>
+        /// <param name="path">The field to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>A new pipeline with an additional stage.</returns>
+        public static PipelineDefinition<TInput, TOutput> Rerank<TInput, TField, TOutput>(
+            this PipelineDefinition<TInput, TOutput> pipeline,
+            RerankQuery query,
+            Expression<Func<TOutput, TField>> path,
```
sanych-sun, line 1099: Should we put `path` parameter as the last one? So the overload that takes more then 1 path would be more aligned.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -1081,6 +1081,104 @@ public static PipelineDefinition<TInput, TOutput> RankFusion<TInput, TIntermedia
...
+        /// <param name="pipeline">The pipeline.</param>
+        /// <param name="query">The rerank query.</param>
+        /// <param name="paths">The fields to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>A new pipeline with an additional stage.</returns>
+        public static PipelineDefinition<TInput, TOutput> Rerank<TInput, TOutput>(
+            this PipelineDefinition<TInput, TOutput> pipeline,
+            RerankQuery query,
+            IEnumerable<FieldDefinition<TOutput>> paths,
```
sanych-sun, line 1147: Let's either use IEnumerable parameter or params in both cases: here and on the line 1126.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -1081,6 +1081,104 @@ public static PipelineDefinition<TInput, TOutput> RankFusion<TInput, TIntermedia
...
+        /// </summary>
+        /// <typeparam name="TInput">The type of the input documents.</typeparam>
+        /// <typeparam name="TOutput">The type of the output documents.</typeparam>
+        /// <param name="pipeline">The pipeline.</param>
+        /// <param name="query">The rerank query.</param>
+        /// <param name="paths">The fields to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>A new pipeline with an additional stage.</returns>
+        public static PipelineDefinition<TInput, TOutput> Rerank<TInput, TOutput>(
```
sanych-sun, line 1144: Should we have corresponding overload in `IAggregateFluentExtensions.cs`?

src/MongoDB.Driver/IAggregateFluentExtensions.cs
```diff
@@ -1041,6 +1041,71 @@ public static IAggregateFluent<TNewResult> Unwind<TResult, TNewResult>(this IAgg
...
+        /// <param name="path">The field to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>The fluent aggregate interface.</returns>
+        public static IAggregateFluent<TResult> Rerank<TResult>(
+            this IAggregateFluent<TResult> aggregate,
+            RerankQuery query,
+            FieldDefinition<TResult> path,
+            int numDocsToRerank,
+            string model)
```
sanych-sun, line 1153: Any ideas if list of models are fixed? Should we have `enum` here instead? Or at least static class with string constants for names of the available for now models?

src/MongoDB.Driver/RerankQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+            _rendered = rendered;
+        }
+
+        /// <summary>
+        /// Creates a text-based rerank query.
+        /// </summary>
+        /// <param name="text">The text to rerank against.</param>
+        /// <returns>A text rerank query.</returns>
+        public static RerankQuery Text(string text) =>
+            new RerankQuery(new BsonDocument("text", Ensure.IsNotNull(text, nameof(text))));
```
sanych-sun: `Ensure.IsNotNullOrEmpty`?

src/MongoDB.Driver/RerankQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        internal BsonDocument Render() => _rendered;
```
sanych-sun: I'm not sure if it's safe to have the pre-rendered document here. Because we return it as a `BsonDocument` which is mutable, it means it might be populated/altered (I do not know, for example we could add _id field), reusing the same `BsonDocument` might lead to some weird behavior. I would suggest to render to a new document on each call.

tests/MongoDB.Driver.Tests/PipelineStageDefinitionBuilderTests.cs
```diff
@@ -653,6 +653,45 @@ public void RankFusion_with_incorrect_params_should_throw_expected_exception()
             });
         }
 
+        [Fact]
+        public void Rerank_with_incorrect_params_should_throw_expected_exception()
```
sanych-sun: We should split this to several Facts validating each parameter separately. Or a theory with test case for each parameter.

sanych-sun (changes requested):

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
adelinowona: I'd prefer keeping `(query, path, numDocsToRerank, model)` for the non-params overloads — it reads more naturally: `"rerank this query against this field, take N docs, use this model."` The params overload has path last because C# requires it, and that should be a well-understood language constraint for users. Reordering the common overloads to match the params one would make the typical usage less intuitive. Thoughts?

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
adelinowona: The two multi-path overloads serve different use cases: the `params Expression[]` one is for inline type-safe expressions `(x => x.Title, x => x.Plot)`, and the `IEnumerable<FieldDefinition>` one is for pre-built or dynamic collections of string-based paths. Using `params` for `FieldDefinition` wouldn't add much since there's no type inference benefit, and using `IEnumerable<Expression>` would lose the `params` convenience. I think the inconsistency is justified by the different use cases here. 
Thoughts?

src/MongoDB.Driver/IAggregateFluentExtensions.cs
adelinowona: I considered an enum but went with plain string to stay forward-compatible — model names are server-defined and the accepted list can change without a driver release. An enum or constants class would need a driver update each time a model is added or removed. The server already validates the model name and returns a clear error listing accepted values if invalid.

src/MongoDB.Driver/RerankQuery.cs
adelinowona: I was going to defer to the server validation but I suppose the early catch is worth it here.

src/MongoDB.Driver/RerankQuery.cs
adelinowona: Nice catch, will do.

tests/MongoDB.Driver.Tests/PipelineStageDefinitionBuilderTests.cs
adelinowona: will separate to different Facts

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
adelinowona: Yep, missed that

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
sanych-sun: I see this as a different way to provide the same parameters. Having one as an array and another as a params looks confusing a little. About different use cases: both expressions and FieldDefinitions could be used as fixed list of values or dynamic collections I believe. I think I would prefer consistency between overloads if it's possible. @BorisDog @papafe @ajcvickers WDYT?

ajcvickers: Agree with @adelinowona.

sanych-sun (approved):
LGTM

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1684,6 +1684,118 @@ public static PipelineStageDefinition<TInput, TOutput> ReplaceWith<TInput, TOutp
...
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Rerank<TInput, TField>(
+            RerankQuery query,
+            Expression<Func<TInput, TField>> path,
+            int numDocsToRerank,
+            string model)
+            => Rerank(
+                query,
+                new ExpressionFieldDefinition<TInput>(path),
```
BorisDog: minor: probably need a null check for `path` here, for consistent exception with other overload.

src/MongoDB.Driver/RerankQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
```
BorisDog: minor: filescoped namespaces.

src/MongoDB.Driver/RerankQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Represents a query for the $rerank aggregation stage.
+    /// </summary>
```
BorisDog: minor: Could also add a reference to relevant builder in the doc.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
```diff
@@ -1684,6 +1684,118 @@ public static PipelineStageDefinition<TInput, TOutput> ReplaceWith<TInput, TOutp
...
+        /// <param name="query">The rerank query.</param>
+        /// <param name="path">The field to send to the reranker.</param>
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <returns>The stage.</returns>
+        public static PipelineStageDefinition<TInput, TInput> Rerank<TInput, TField>(
+            RerankQuery query,
+            Expression<Func<TInput, TField>> path,
+            int numDocsToRerank,
+            string model)
```
BorisDog, line 1754: Double checking as  docs are a bit confusing, `model` is not optional?

src/MongoDB.Driver/RerankQuery.cs
```diff
@@ -0,0 +1,43 @@
...
+            _text = text;
+        }
+
+        /// <summary>
+        /// Creates a text-based rerank query.
+        /// </summary>
+        /// <param name="text">The text to rerank against.</param>
+        /// <returns>A text rerank query.</returns>
+        public static RerankQuery Text(string text) =>
+            new RerankQuery(Ensure.IsNotNullOrEmpty(text, nameof(text)));
```
BorisDog: minor: tests.

BorisDog (commented):
Minor comments

tests/MongoDB.Driver.Tests/PipelineStageDefinitionBuilderTests.cs
```diff
@@ -653,6 +653,69 @@ public void RankFusion_with_incorrect_params_should_throw_expected_exception()
...
+        }
+
+        [Fact]
+        public void Rerank_should_throw_when_query_is_null_for_params_overload()
+        {
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                PipelineStageDefinitionBuilder.Rerank<BsonDocument, string>(null, 10, "rerank-2.5");
+            }).ParamName.Should().Be("query");
+        }
```
BorisDog: Do we have a ticket for Atlas tests?

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
```diff
@@ -1081,6 +1081,104 @@ public static PipelineDefinition<TInput, TOutput> RankFusion<TInput, TIntermedia
...
+        /// <param name="numDocsToRerank">The maximum number of documents to rerank.</param>
+        /// <param name="model">The reranking model name.</param>
+        /// <param name="paths">The fields to send to the reranker.</param>
+        /// <returns>A new pipeline with an additional stage.</returns>
+        public static PipelineDefinition<TInput, TOutput> Rerank<TInput, TField, TOutput>(
+            this PipelineDefinition<TInput, TOutput> pipeline,
+            RerankQuery query,
+            int numDocsToRerank,
+            string model,
+            params Expression<Func<TOutput, TField>>[] paths)
```
BorisDog: Yeah as you mentioned the inconsistency in path and paths arguments position is not ideal. But no strong opinion.
Let's see what @ajcvickers thinks.

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
adelinowona: Correct, model is required. The original spec proposed it as optional with a "latest" default, but the server implementation requires an explicit model and rejects "latest".

tests/MongoDB.Driver.Tests/PipelineStageDefinitionBuilderTests.cs
adelinowona: Created CSHARP-5982 for this. Depending on the timing of cloud-qa readiness, I might just add the test in the current PR though and close that ticket.

src/MongoDB.Driver/RerankQuery.cs
adelinowona: done

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
adelinowona: done

src/MongoDB.Driver/RerankQuery.cs
adelinowona: done

src/MongoDB.Driver/RerankQuery.cs
adelinowona: done

src/MongoDB.Driver/IAggregateFluentExtensions.cs
ajcvickers: I think just the string is fine here, but another option is to have an overload with well-known model names which can be updated when we release, but still has the string overload for any newer model. However, I don't really think that is needed here, and we could add it later if necessary, so I'd just have the string for now.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
ajcvickers: I didn't notice the change in argument ordering before. I prefer the ordering where the paths are immediately after the query--this feels like what we commonly do with paths, and feels natural here like Adelin said. So, given that the params doesn't add a lot in modern C#, I think I would drop it and keep the order the same for overloads. So, yeah, I'm changing my mind. :-) But I don't feel strongly.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
ajcvickers: Changing my mind, as I already commented. I don't agree with, "Having one as an array and another as a params looks confusing a little. " but I do like keeping the parameter order the same between the overloads.

ajcvickers (approved):
LGTM, except I marginally prefer to drop the params array and keep the overloads having the same parameter order.

src/MongoDB.Driver/PipelineDefinitionBuilder.cs
BorisDog: I agree with Arthur here (and also don't feel strongly).

src/MongoDB.Driver/PipelineStageDefinitionBuilder.cs
BorisDog: nit: You could keep the expression body and have 
```
  return Rerank(
                query,
                new ExpressionFieldDefinition<TInput>( Ensure.IsNotNull(path, nameof(path)))
```

BorisDog (approved):
LGTM + 2 minor suggestions.

PR #1937, opened 2026-04-03
CSHARP-2868 Allow maxIdleTimeMS=0 and treat it as no idle time limit

kyra-rk:


kyra-rk:
Rebased to grab [Adelin's version bump](https://github.com/mongodb/mongo-csharp-driver/commit/98c4a1d37a2d32247bd5e2023a66ec1ba6ef1f36)

tests/MongoDB.Driver.Tests/Core/Connections/BinaryConnectionTests.cs
```diff
@@ -713,6 +713,44 @@ public async Task SendMessage_should_put_the_message_on_the_stream_and_raise_the
...
+                tracingOptions: null,
+                socketReadTimeout: TimeSpan.FromMilliseconds(1000),
+                socketWriteTimeout: TimeSpan.FromMilliseconds(1000));
+
+            subject.Open(OperationContext.NoTimeout);
+            subject.IsExpired.Should().BeFalse();
+        }
+
+        [Fact]
+        public void IsExpired_should_return_true_when_maxIdleTime_is_exceeded()
```
papafe: This test is a little misleading. We would expect some kind of "sleep" somewhere to wait for the max idle time to pass but we don't have it.
For the way that `IsExpired` work, this test will pass with any kind of positive `maxIdleTime`, and so its actual value is a red herring. 
I would change the title to `IsExpired_should_return_true_when_maxIdleTime_is_positive_and_last_used_is_stale`

papafe (changes requested):
Just a small comment, but looks good for the rest!

tests/MongoDB.Driver.Tests/Core/Connections/BinaryConnectionTests.cs
kyra-rk: Ah I think you're right. Looks like `_lastUsedAtUtc`, which is used in the `IsExpired` condition is only set when we send and receive. I'm not sure I fully understand what you meant by "stale". Would something like this fit - `IsExpired_should_return_true_when_maxIdleTime_is_positive_and_connection_was_unused`?

papafe: Yes, the name is fine.

papafe (approved):
LGTM! 
Please open a new ticket to investigate if we should do the same with `maxLifeTime`

kyra-rk:
> LGTM! Please open a new ticket to investigate if we should do the same with `maxLifeTime`

I think as you pointed out in [this slack discussion](https://mongodb.slack.com/archives/GQ27C9N7J/p1776172438750299), there's already a [ticket](https://jira.mongodb.org/browse/CSHARP-3960) to change the default for `maxLifeTime` to infinite. I can un-backlog this if we want to re-triage this?

PR #1938, opened 2026-04-06
CSHARP-5629 Respect BsonElement attribute in NamedIdMemberConvention

kyra-rk:


tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
```diff
@@ -14,19 +14,83 @@
...
+using MongoDB.Bson.Serialization.Attributes;
 using MongoDB.Bson.Serialization.Conventions;
 using Xunit;
 
 namespace MongoDB.Bson.Tests.Serialization.Conventions
 {
     public class ElementNameConventionsTests
     {
-        private class TestClass
+
```
papafe: Extra line

tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
```diff
@@ -14,19 +14,83 @@
...
+        public void TestBsonElementOnIdFallsThroughToLowercaseId()
+        {
+            var classMap = new BsonClassMap<BookWithBsonElementAndFallthrough>();
+            classMap.AutoMap();
+            classMap.Freeze();
+
+            Assert.Equal("id", classMap.IdMemberMap.MemberName);
+            Assert.Equal("_id", classMap.GetMemberMap(x => x.id).ElementName);
+            Assert.Equal("notId", classMap.GetMemberMap(x => x.Id).ElementName);
         }
```
papafe, line 104: I would add another test where using just `BsonElement` on a property called `Id` assigns the Id property correctly. It should work, so mostly a safety net.

papafe: I would also add another test, still mostly as a safety net. We can verify the projection that was shown in the JIRA ticket. That test could live in `ProjectionDefinitionBuilderTests.cs`

papafe (changes requested):
Looks good, some small comments about possible additional tests.

tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
kyra-rk: I added both, but I'm not sure if adding a test to `ProjectionDefinitionBuilderTests.cs` mirroring the JIRA description is the right thing to do. It tests a super specific edge case for attributes/convention that feels awkward in the projection builders test. If there aren't any "integration" tests for attributes or conventions where this could go, then I'm leaning towards just dropping it. (I know the CSHARP-XXXX tests in the Jira folder used to be the place to add these, but we're moving away from this.) Lmk what you think!

tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
```diff
@@ -14,19 +14,93 @@
 */
 
 using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Attributes;
 using MongoDB.Bson.Serialization.Conventions;
 using Xunit;
```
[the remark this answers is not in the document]
kyra-rk: The file's namespace is`MongoDB.Bson.Tests.Serialization.Conventions,` so the compiler is able to resolve `ObjectId` automatically from the namespace chain — it builds fine locally.

tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
```diff
@@ -58,5 +132,61 @@ public void TestCamelCaseElementNameConvention()
...
+            public ObjectId Id { get; set; }
+
+            [BsonId]
+            public string Title { get; set; } // should be set to _id
+        }
+
+        private class BookWithExplicitBsonElementId
+        {
+            [BsonElement("_id")]
+            public string Id { get; set; } // should not be set to _id
```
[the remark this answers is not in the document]
kyra-rk: Will change it to `// should be set to _id` to match my style so far

tests/MongoDB.Bson.Tests/Serialization/Conventions/ElementNameConventionsTests.cs
papafe: I agree with you, let's delete the projection test.

papafe (approved):
Let's delete the projection test, and then it would be ready to merge. I think it would be worth discussing with the team if we should put any particular attention to this in the release notes before merging.

src/MongoDB.Bson/Serialization/Conventions/NamedIdMemberConvention.cs
```diff
@@ -119,6 +120,12 @@ private bool IsValidIdMember(BsonClassMap classMap, MemberInfo member)
                     return false;
                 }
             }
+
+            var bsonElement = member.GetCustomAttribute<BsonElementAttribute>();
```
sanych-sun, line 124: What if I create a `ClassMap` manually and assign name for  "Id" property via `MapProperty` prior to calling AutoMap()? Should we handle that case too?

sanych-sun (changes requested):

src/MongoDB.Bson/Serialization/Conventions/NamedIdMemberConvention.cs
kyra-rk: Ah, great catch! I added logic to the same place to handle if there is already a declared member map and to skip the convention if the user mapped it to a non-Id/_id/id element name. I added 3 tests to cover these cases.

sanych-sun (approved):
LGTM

papafe (approved):

papafe:
Added the breaking change label to make this more visible in the release.

PR #1940, opened 2026-04-07
CSHARP-5949: $convert should allow any type to be converted to string

papafe:
We do not need to do anything specific, as we did not have any driver/c# constraint on what could have been converted.


sanych-sun (approved):
LGTM

PR #1941, opened 2026-04-07
CSHARP-5950: Support base conversion in $convert

papafe:


src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -202,6 +203,11 @@ public class Feature
         /// </summary>
         public static Feature ClusteredIndexes => __clusteredIndexes;
 
+        /// <summary>
+        /// Gets the base conversion in $convert feature.
+        /// </summary>
+        public static Feature ConvertOperatorBaseConversion => __convertOperatorBaseConversion;
```
BorisDog, line 216: Doesn't have to be resolved in this PR:
We'll have multiple 8.3 features like this. Should we continue with tracking each feature independently in Feature.cs, or just check for wire version in tests instead? 

cc @sanych-sun

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -500,6 +500,52 @@ public void Convert_should_work(
             AssertOutcome(collection, queryable, expectedStages, expectedValue);
         }
 
+        [Theory]
+        [InlineData(22, 2, "100001")]
+        [InlineData(22, 8, "41")]
+        [InlineData(22, 16, "21")]
+        public void Convert_to_string_with_base_should_work(int id, int @base, string expectedResult)
```
BorisDog: @sanych-sun is the new LINQ testing approach relevant here?

BorisDog (changes requested):
The changes look good.
Added @sanych-sun for LINQ review + testing question.

src/MongoDB.Driver/ConvertOptions.cs
```diff
@@ -22,10 +22,20 @@ namespace MongoDB.Driver
     /// </summary>
     public abstract class ConvertOptions
     {
+        private int? _base;
         private ByteOrder? _byteOrder;
         private string _format;
         private BsonBinarySubType? _subType;
 
+        /// <summary>
+        /// The base parameter. Must be 2, 8, 10, or 16.
```
sanych-sun: If it's a strict list of predefined items, should we have `enum` instead? Also we might have `enum` with `Base10` default value. This way we can even make it non-nullable.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
sanych-sun: I would say "yes", but also it looks like a minor extension for existing functionality. I think we can let @papafe decide if he wants to convert existing tests into the updated approach.

src/MongoDB.Driver/Core/Misc/Feature.cs
sanych-sun: I think it's better to have this as a list of separate features, same way as we have all other features. Unless we want to remove the Feature class from the public API.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
```diff
@@ -96,7 +101,7 @@ public AstConvertExpression Update(
                 return this;
             }
 
-            return new AstConvertExpression(input, to, _subType, _byteOrder, _format, onError, onNull);
+            return new AstConvertExpression(input, to, _base, _subType, _byteOrder, _format, onError, onNull);
```
sanych-sun, line 104: We must add comparison of _base to quick return condition on line 99.

sanych-sun (changes requested):

src/MongoDB.Driver/ConvertOptions.cs
papafe: I agree with the enum, less issues with wrong values being used. But I would avoid using `Base10` as the default value, using it is semantically different from not passing it.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Expressions/AstConvertExpression.cs
papafe: But `base` is not an expression tree, just a plain value, so we don't need to visit it. We do the same for the other parameters, `subtype`, `byteOrder`, ...

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ConvertMethodToAggregationExpressionTranslatorTests.cs
papafe: Yeah, for now I'd keep it like this and then address it in a separate ticket.

BorisDog (approved):
LGTM

sanych-sun (approved):
LGTM

PR #1947, opened 2026-04-10
CSHARP-4901 Add support for SearchMeta facet without operator

kyra-rk:


src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -174,7 +174,7 @@ internal sealed class FacetSearchDefinition<TDocument> : OperatorSearchDefinitio
         public FacetSearchDefinition(SearchDefinition<TDocument> @operator, IEnumerable<SearchFacet<TDocument>> facets)
             : base(OperatorType.Facet)
         {
-            _operator = Ensure.IsNotNull(@operator, nameof(@operator));
+            _operator = @operator;
```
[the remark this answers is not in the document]
kyra-rk: Ah, this contradicts the original purpose of the ticket, which was to to make `operator` optional as per the [Atlas Search docs](https://www.mongodb.com/docs/atlas/atlas-search/operators-collectors/facet/#syntax).

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
```diff
@@ -461,17 +461,43 @@ public void Facet_typed()
...
+        {
+            var subject = CreateSubject<Person>();
+            var facetBuilder = new SearchFacetBuilder<Person>();
+            var expected = "{ facet: { facets: { string: { type: 'string', path: 'fn', numBuckets: 100 } } } }";
+
+            AssertRendered(
+                subject.Facet( facetBuilder.String("string", x => x.FirstName, 100)),
+                expected);
+            AssertRendered(
+                subject.Facet( facetBuilder.String("string", "FirstName", 100)),
```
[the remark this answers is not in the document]
kyra-rk: Will adjust the formatting to be on the next line!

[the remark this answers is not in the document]
kyra-rk: Will adjust the formatting to be on the next line!

src/MongoDB.Driver/Search/SearchDefinitionBuilder.cs
```diff
@@ -200,6 +200,24 @@ public SearchDefinition<TDocument> Facet(
             params SearchFacet<TDocument>[] facets) =>
                 Facet(@operator, (IEnumerable<SearchFacet<TDocument>>)facets);
 
+        /// <summary>
+        /// Creates a search definition that groups results by values or ranges in the specified
```
adelinowona, line 208: The four `Facet` overloads now have near-identical doc text. A user in IntelliSense has no signal for when to pick the operator-less form. The no-operator overloads could say something like "Facets are computed over all documents in the index" to help a little.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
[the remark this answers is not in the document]
adelinowona: I think it is fine to remove the check here but we probably should then move the check up to the SearchDefinitionBuilder overloads that provide an operator. My take is if a user knowingly uses the overload that takes an operator and ends up passing a null we should throw.

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
```diff
@@ -174,7 +174,7 @@ internal sealed class FacetSearchDefinition<TDocument> : OperatorSearchDefinitio
...
             _facets = Ensure.IsNotNull(facets, nameof(facets)).ToArray();
```
adelinowona: With the new `Facet(params SearchFacet<TDocument>[] facets)` overload, we probably want to check the facets array is not null and is not empty (so use `Ensure.IsNotNullOrEmpty()`). 

if we don't then something like this, compiles and run;
`Builders.Search.Facet();`

It renders `{ facet: { facets: {} } }` and gets rejected server-side instead of at build time.

adelinowona (changes requested):

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
kyra-rk: Good callout, will add!

kyra-rk: Actually, wouldn't this change functionality a little? I just tried calling facet with an operator and no params array and there wasn't an exception:
```
[Fact]
public void Facet_should_disallow_null_facet_array()
{
    var subject = CreateSubject<Person>();

    Record.Exception(() => subject.Facet(subject.Phrase(x => x.LastName, "foo"))).Should().BeOfType<ArgumentNullException>();
}
```
Adding this check seems like the right thing to do since the empt facets array will end up getting rejected server-side. But I just want to confirm that it's okay to change the existing behaviour.

adelinowona: I think this is ok since this was something that was eventually failing server-side before so we are just moving the failure point earlier in the driver. @BorisDog ?

tests/MongoDB.Driver.Tests/Search/SearchDefinitionBuilderTests.cs
adelinowona: maintain alphabetical order of tests in this file

src/MongoDB.Driver/Search/OperatorSearchDefinitions.cs
BorisDog: It's a good question @kyra-rk. In this case it's ok for the reasons that @adelinowona  mentioned.

adelinowona (approved):
LGTM

PR #1948, opened 2026-04-10
CSHARP-5965: Add .claude folder to .gitignore

sanych-sun:


adelinowona (approved):

damieng:
The problem with this is that excludes Claude skill files.

sanych-sun:
@damieng Thank you! Updated to exclude the single file.

damieng (approved):
LGTM

PR #1950, opened 2026-04-14
CSHARP-5567: Add a LINQ equivalent for $toHashedIndexKey

kyra-rk:


kyra-rk:
(will keep as draft/open until next week, once 3.8 is released)

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+using Xunit;
+
+namespace MongoDB.Driver.Tests.Linq.Integration;
+
+public class MqlToHashedIndexKeyTests : LinqIntegrationTest<MqlToHashedIndexKeyTests.ClassFixture>
+{
+    public MqlToHashedIndexKeyTests(ClassFixture fixture)
+        : base(fixture)
+    {
+    }
```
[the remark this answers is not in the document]
kyra-rk: Based on [SERVER-49214](https://jira.mongodb.org/browse/SERVER-49214), looks like `$toHashedIndexKey` is supported 4.7 onwards and was backported to 4.4 and 4.2.24. Is it still worth adding this feature check given the latest driver supports 4.2 onwards ([this](https://www.mongodb.com/docs/drivers/compatibility/?driver-language=csharp&csharp-driver-framework=csharp-driver) is what I was using to check compatibility)? (I guess I'm curious whether the new tests I added will ever be run against a much older server).

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -2706,6 +2707,18 @@ bool IsToArrayMethod(out Expression sourceExpression)
             }
         }
 
+        void DeduceToHashedIndexKeySerializers()
```
adelinowona, line 2752: You'll need to add a serializerFinder unit test for this in `tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/MqlTests.cs`. That file has a TestCases table covering every `Mql.*` method so you can follow the convention for unit testing there.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToHashedIndexKeyMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            if (method.Is(MqlMethod.ToHashedIndexKey))
+            {
+                var valueExpression = arguments.Single();
+                var valueTranslation = ExpressionToAggregationExpressionTranslator.Translate(context, valueExpression);
+
+                return new TranslatedExpression(
+                    expression,
+                    AstExpression.Unary(AstUnaryOperator.ToHashedIndexKey, valueTranslation.Ast),
+                        Int64Serializer.Instance);
```
adelinowona: isn't this indented too far to the right? Also we shouldn't be hard coding a serializer here. You should use `context.GetSerializer(expression)` so that it uses the serializer deduced by the serializerFinder.

SerializerFinder is new so there are still places in the code that hard code serializers in translators and should be changed to use `context.GetSerializer()`  so if you see those in a file you are modifying feel free to update it or file a ticket to update it.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        var collection = Fixture.Collection;
+
+        var pipeline = new EmptyPipelineDefinition<C>()
+            .Project(d => new { d.Id, Hash = Mql.ToHashedIndexKey(d.Value) });
+
+        var result = await collection.Aggregate(pipeline).ToListAsync();
+
+        var stages = Translate(collection, pipeline, null);
+        AssertStages(stages, "{ $project : { _id : '$_id', Hash : { $toHashedIndexKey : '$Value' } } }");
```
adelinowona: Swap the order of these two. You want to assert the translation before executing the query. Currently, the query executes against the server before the MQL assertion. If translation produces invalid MQL, you get a server-side error first and never see what was actually generated.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        result.Should().HaveCount(3);
```
adelinowona: Also you probably want to assert some hash values here. Without asserting at least one expected Hash value, this test passes whether the server returns the right hashes, all zeros, all the same value, or random longs

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
adelinowona, line 1: `MqlSubtypeTests` for example has explicit `_in_filter_builder` and `_in_projection_builder` cases (lines 68-93 of that file). Those exercise the `Builders<T>.Filter.Where(...)` and `Builders<T>.Projection.Expression(...)` paths, which go through different translation entry points than IQueryable<T> does. I would suggest adding them as well for this Mql method.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        var stages = Translate(collection, queryable);
+        AssertStages(stages, $"{{ $match : {{ $expr : {{ $eq : [{{ $toHashedIndexKey : '$Value' }}, {{ $numberLong : '{hashedValue}' }}] }} }} }}");
+
+        var result = queryable.Single();
+        result.Id.Should().Be(1);
+    }
+
+    [Fact]
+    public void ToHashedIndexKey_with_int_field()
```
adelinowona: I would suggest dropping this test as it is identical to `_in_select`. What `_with_int_field` uniquely adds beyond `_in_select` is: "the MongoDB server correctly hashes an int." That's server behavior, not driver behavior. The server team owns that. The translator unit tests already check we produce the expected AST for the different field types.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        var queryable = collection.AsQueryable()
+            .Select(x => Mql.ToHashedIndexKey(x.Value));
+
+        var stages = Translate(collection, queryable);
+        AssertStages(stages, "{ $project : { _v : { $toHashedIndexKey : '$Value' }, _id : 0 } }");
+
+        var result = queryable.ToList();
+        result.Should().HaveCount(3);
+        result.Should().OnlyContain(v => v != 0);
```
adelinowona: A stronger assertion here might be to check that the hashes directly. With the current v != 0 check, any translation bug that still returned some non-zero long would pass.

Same idea applies to other similar tests in this file.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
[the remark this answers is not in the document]
adelinowona: Nah you are fine, no need for feature check. The oldest server our tests run on evergreen is 4.2

adelinowona (changes requested):
requested some changes. Also add a colon after the ticket number in your PR title.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
kyra-rk: Will do!

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToHashedIndexKeyMethodToAggregationExpressionTranslator.cs
kyra-rk: That makes sense!

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
kyra-rk: Ah, yep I see what you mean. Will remove.

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
kyra-rk: Good catch!

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
kyra-rk: Good point, will add!

tests/MongoDB.Driver.Tests/Linq/Integration/MqlToHashedIndexKeyTests.cs
kyra-rk: Will add!

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/MqlTests.cs
```diff
@@ -54,6 +54,9 @@ public void SerializerFinder_should_resolve_mql_methods(LambdaExpression express
         [TestHelpers.MakeLambda((MyModel model) => Mql.SimilarityDotProduct(model.Vector1, model.Vector2, false)), typeof(DoubleSerializer)],
         [TestHelpers.MakeLambda((MyModel model) => Mql.SimilarityEuclidean(model.Vector1, model.Vector2, false)), typeof(DoubleSerializer)],
         [TestHelpers.MakeLambda((MyModel model) => Mql.Subtype(model.Data)), typeof(NullableSerializer<BsonBinarySubType>)],
+        [TestHelpers.MakeLambda((MyModel model) => Mql.Hash(model.Data, MqlHashAlgorithm.SHA256)), typeof(BsonBinaryDataSerializer)],
+        [TestHelpers.MakeLambda((MyModel model) => Mql.ToHashedIndexKey(model.Field)), typeof(Int64Serializer)],
```
adelinowona, line 57: Move this down to maintain alphabetical order

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToHashedIndexKeyMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,45 @@
...
+    internal static class ToHashedIndexKeyMethodToAggregationExpressionTranslator
+    {
+        public static TranslatedExpression Translate(TranslationContext context, MethodCallExpression expression)
+        {
+            var method = expression.Method;
+            var arguments = expression.Arguments;
+
+            if (method.Is(MqlMethod.ToHashedIndexKey))
+            {
+
```
adelinowona: stray blank line?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/ToHashedIndexKeyMethodToAggregationExpressionTranslator.cs
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Linq;
+using System.Linq.Expressions;
+using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.Driver.Linq.Linq3Implementation.Ast.Expressions;
+using MongoDB.Driver.Linq.Linq3Implementation.Misc;
+using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
```
adelinowona: Some of these might be unused. If it's greyed out in rider IDE then you can remove it.

adelinowona (changes requested):

adelinowona (approved):
LGTM!

PR #1951, opened 2026-04-15
CSHARP-5826: Add builder support for $top, $bottom, $topN and $bottomN expressions new in 8.3

damieng:
Implements [CSHARP-5826](https://jira.mongodb.org/browse/CSHARP-5826)

## Summary

- Adds builder-level support for `$top`, `$bottom`, `$topN`, `$bottomN` as **aggregation array expressions** (MongoDB 8.3+)
- These operators were previously only available as group accumulators; MongoDB 8.3 extends them to work on array fields in `$set`, `$project`, `$addFields` etc.
- Adds `PickExpressionDefinitionBuilder` static class with factory methods for all four operators
- Adds `AggregateExpressionSetFieldDefinition` to enable using aggregation expressions as values in `$set` stages via `Builders<T>.SetFields`
- Adds `Feature.PickAccumulatorsAsExpressionsNewIn83` feature flag for server version gating

## Before and after

Before 8.3, finding the top scorer per tournament required a `$group` stage that collapses documents, losing all other fields:

```csharp
// MongoDB 5.2+ — must group, losing the rest of the tournament document
collection.Aggregate()
    .Group(x => x.TournamentId, g => new {
        Id = g.Key,
        Leader = g.Top(
            Builders<Score>.Sort.Descending(s => s.Points),
            e => e)
    });
// { $group: { _id: "$TournamentId", Leader: { $top: { sortBy: { Points: -1 }, output: "$$ROOT" } } } }
```

With 8.3 expressions the document passes through intact and `Leader` is computed directly from the embedded array in one `$set` stage:

```csharp
// MongoDB 8.3+ — operates on the embedded array, document shape preserved
collection.Aggregate()
    .Set(Builders<Tournament>.SetFields.Set(
        new StringFieldDefinition<Tournament, Player>("Leader"),
        PickExpressionDefinitionBuilder.Top<Tournament, Player>(
            x => x.Players,
            Builders<Player>.Sort.Descending(p => p.Score))));
// { $set: { Leader: { $top: { input: "$Players", sortBy: { Score: -1 } } } } }
```

### DIFF
--- a/src/MongoDB.Driver/PickExpressionDefinitionBuilder.cs
+++ b/src/MongoDB.Driver/PickExpressionDefinitionBuilder.cs
@@ -1,0 +1,275 @@
+/* Copyright 2010-present MongoDB Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// A builder for pick aggregation expression definitions ($top, $bottom, $topN, $bottomN).
+    /// These operators sort an array and return the top or bottom element(s). Requires MongoDB 8.3+.
+    /// </summary>
+    public static class PickExpressionDefinitionBuilder
+    {
+        /// <summary>
+        /// Creates a $top expression that returns the first element of the input array after applying the sort.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


src/MongoDB.Driver/PickExpressionDefinitionBuilder.cs
```diff
@@ -0,0 +1,275 @@
...
+using MongoDB.Bson.Serialization;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver
+{
+    /// <summary>
+    /// A builder for pick aggregation expression definitions ($top, $bottom, $topN, $bottomN).
+    /// These operators sort an array and return the top or bottom element(s). Requires MongoDB 8.3+.
+    /// </summary>
+    public static class PickExpressionDefinitionBuilder
```
sanych-sun, line 29: I think all our builders has a TDocument type parameter and available via Builders static class.

sanych-sun, line 29: Also it looks like a very new builder for us. Is this a builder for accumulators? If so - we probably ought to name it so, and make it available from Builders. Also it looks like we need to decide what will we do with another accumulators.

sanych-sun (changes requested):

sanych-sun:
I'm so confused now. The ticket name says it's about Expressions, but as far as I understood there is no appropriate Linq to translate from. The actual changes suggests this is about accumulators, but we do not use to have Builders for accumulators - this should be extended to support all accumulators then. This needs to be discussed.
@BorisDog @rishitb-mongodb

damieng:
Needs a rethink with regards to implementing existing non-expression MQL functions being exposed through the builders and this would ideally align with that.

PR #1956, opened 2026-04-17
CSHARP-5967 Ensure SerializerFinder proper support of Parse method

kyra-rk:
In addition to fixing support for short, long, float, byte, sbyte, I also added general `SerializerFinder` test coverage for float and sbyte.


sanych-sun (approved):
LGTM

PR #1957, opened 2026-04-17
CSHARP-5963: Fix failed search-index variants because of cluster name conflict

sanych-sun:
Atlas has a requirement to have unique first 23 characters in the group:

> Cluster name dbx-csharp-search-index-5843579d8414371e cannot have same first 23 characters as another cluster or serverless instance name in same group.

Reduce the prefix to let us run search index tests in parallel.


BorisDog (approved):
LGTM

PR #1961, opened 2026-04-22
CSHARP-5959: Replace switch by method name in SerializerFinderVisitMethodCall with lookup table with MethodInfo as a key

sanych-sun:
The idea of the PR is to replace the switch by string method's name and checking of methodInfo, with the dictionary of deducers by method info. Most of the use-cases is covered by "static" registration, however there are number of use-cases where we duck-typing the method call (see CreateDynamicSerializerDeducer method), in such case we do dynamic registration as we go: so the next translation of the method will use "previously registered deducer" and do not need to go through all checks.

### DIFF
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
@@ -628,6 +628,22 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
         public static IReadOnlyMethodInfoSet ReverseOverloads => __reverseOverloads;
 
         // public methods
+        public static bool IsContainsMethod(MethodInfo method)
+        {
+            var parameters = method.GetParameters();
+            if (method.Name == "Contains" && method.ReturnType == typeof(bool))
+            {
+                if (method.IsStatic)
+                {
+                    return parameters.Length == 2 && parameters[0].ParameterType.ImplementsIEnumerableOf(parameters[1].ParameterType);
+                }
+
+                return parameters.Length == 1 && method.DeclaringType.ImplementsIEnumerableOf(parameters[0].ParameterType);
+            }
+
+            return false;
+        }
+
         public static bool IsContainsMethod(MethodCallExpression methodCallExpression, out Expression sourceExpression, out Expression valueExpression)
         {
             var method = methodCallExpression.Method;
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
@@ -647,35 +647,12 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
         public static bool IsContainsMethod(MethodCallExpression methodCallExpression, out Expression sourceExpression, out Expression valueExpression)
         {
             var method = methodCallExpression.Method;
-            var parameters = method.GetParameters();
-            var arguments = methodCallExpression.Arguments;
-
-            if (method.Name == "Contains" && method.ReturnType == typeof(bool))
+            if (IsContainsMethod(method))
             {
-                if (method.IsStatic)
-                {
-                    if (parameters.Length == 2)
-                    {
-                        if (parameters[0].ParameterType.ImplementsIEnumerableOf(parameters[1].ParameterType))
-                        {
-                            sourceExpression = arguments[0];
-                            valueExpression = arguments[1];
-                            return true;
-                        }
-                    }
-                }
-                else
-                {
-                    if (parameters.Length == 1)
-                    {
-                        if (method.DeclaringType.ImplementsIEnumerableOf(parameters[0].ParameterType))
-                        {
-                            sourceExpression = methodCallExpression.Object;
-                            valueExpression = arguments[0];
-                            return true;
-                        }
-                    }
-                }
+                var arguments = methodCallExpression.Arguments;
+                sourceExpression = method.IsStatic ? arguments[0] : methodCallExpression.Object;
+                valueExpression = method.IsStatic ? arguments[1] : arguments[0];
+                return true;
             }
 
             sourceExpression = null;
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
@@ -652,35 +652,12 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
         public static bool IsContainsMethod(MethodCallExpression methodCallExpression, out Expression sourceExpression, out Expression valueExpression)
         {
             var method = methodCallExpression.Method;
-            var parameters = method.GetParameters();
-            var arguments = methodCallExpression.Arguments;
-
-            if (method.Name == "Contains" && method.ReturnType == typeof(bool))
+            if (IsContainsMethod(method))
             {
-                if (method.IsStatic)
-                {
-                    if (parameters.Length == 2)
-                    {
-                        if (parameters[0].ParameterType.ImplementsIEnumerableOf(parameters[1].ParameterType))
-                        {
-                            sourceExpression = arguments[0];
-                            valueExpression = arguments[1];
-                            return true;
-                        }
-                    }
-                }
-                else
-                {
-                    if (parameters.Length == 1)
-                    {
-                        if (method.DeclaringType.ImplementsIEnumerableOf(parameters[0].ParameterType))
-                        {
-                            sourceExpression = methodCallExpression.Object;
-                            valueExpression = arguments[0];
-                            return true;
-                        }
-                    }
-                }
+                var arguments = methodCallExpression.Arguments;
+                sourceExpression = method.IsStatic ? arguments[0] : methodCallExpression.Object;
+                valueExpression = method.IsStatic ? arguments[1] : arguments[0];
+                return true;
             }
 
             sourceExpression = null;
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
@@ -633,6 +633,22 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.Reflection
         public static IReadOnlyMethodInfoSet ReverseOverloads => __reverseOverloads;
 
         // public methods
+        public static bool IsContainsMethod(MethodInfo method)
+        {
+            var parameters = method.GetParameters();
+            if (method.Name == "Contains" && method.ReturnType == typeof(bool))
+            {
+                if (method.IsStatic)
+                {
+                    return parameters.Length == 2 && parameters[0].ParameterType.ImplementsIEnumerableOf(parameters[1].ParameterType);
+                }
+
+                return parameters.Length == 1 && method.DeclaringType.ImplementsIEnumerableOf(parameters[0].ParameterType);
+            }
+
+            return false;
+        }
+
         public static bool IsContainsMethod(MethodCallExpression methodCallExpression, out Expression sourceExpression, out Expression valueExpression)
         {
             var method = methodCallExpression.Method;
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
@@ -170,6 +170,7 @@ internal partial class SerializerFinderVisitor
         }
     }
 
+    // TODO: merge this and next methods.
     private void DeduceCollectionAndItemSerializers(Expression collectionExpression, Expression itemExpression)
     {
         DeduceItemAndCollectionSerializers(itemExpression, collectionExpression);
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
@@ -169,8 +169,7 @@ internal partial class SerializerFinderVisitor
             AddNodeSerializer(collectionExpression2, collectionSerializer2);
         }
     }
-
-    // TODO: merge this and next methods.
+    
     private void DeduceCollectionAndItemSerializers(Expression collectionExpression, Expression itemExpression)
     {
         DeduceItemAndCollectionSerializers(itemExpression, collectionExpression);
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
@@ -1731,9 +1731,14 @@ internal partial class SerializerFinderVisitor
                         _ when declaringType == typeof(decimal) => DecimalSerializer.Instance,
                         _ when declaringType == typeof(double) => DoubleSerializer.Instance,
                         _ when declaringType == typeof(int) => Int32Serializer.Instance,
-                        _ when declaringType == typeof(short) => Int64Serializer.Instance,
+                        _ when declaringType == typeof(short) => Int16Serializer.Instance,
+                        _ when declaringType == typeof(long) => Int64Serializer.Instance,
+                        _ when declaringType == typeof(float) => SingleSerializer.Instance,
+                        _ when declaringType == typeof(byte) => ByteSerializer.Instance,
+                        _ when declaringType == typeof(sbyte) => SByteSerializer.Instance,
                         _ => UnknowableSerializer.Create(declaringType)
                     };
+
                     visitor.AddNodeSerializer(expression, nodeSerializer);
                 }
             }
--- a/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
+++ b/src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
@@ -14,13 +14,16 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Threading;
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization;
 using MongoDB.Bson.Serialization.Options;
 using MongoDB.Bson.Serialization.Serializers;
+using MongoDB.Driver.Core.Misc;
 using MongoDB.Driver.Linq.Linq3Implementation.ExtensionMethods;
 using MongoDB.Driver.Linq.Linq3Implementation.Misc;
 using MongoDB.Driver.Linq.Linq3Implementation.Reflection;
@@ -30,6 +33,8 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
 
 internal partial class SerializerFinderVisitor
 {
+    private static readonly Dictionary<MethodInfo, Action<SerializerFinderVisitor, MethodCallExpression>> __serializerResolvers = new();
+    private static readonly ReaderWriterLockSlim __serializerResolversLock = new();
     private static readonly IBsonSerializer __binarySubTypeSerializer = NullableSerializer.Create(new EnumSerializer<BsonBinarySubType>());
 
     private static readonly IReadOnlyMethodInfoSet __averageOrMedianOrPercentileOverloads = MethodInfoSet.Create(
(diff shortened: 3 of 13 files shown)


src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
```diff
@@ -170,6 +170,7 @@ private void DeduceCollectionAndCollectionSerializers(Expression collectionExpre
         }
     }
 
+    // TODO: merge this and next methods.
```
sanych-sun: I'll create a ticket for this, as these 2 methods looks like doing the same stuff.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-                DeduceReturnsNumericSerializer();
-            }
-            else
-            {
-                DeduceUnknownMethodSerializer();
-            }
         }
-
-        void DeduceCompareOrCompareToMethodSerializers()
+        // DeduceAddToSetMethodSerializers
```
sanych-sun: I've left such comments so for now, for review purposes. So it would be easier to find where the implementation came from. I'll remove them before merging.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+        private static void DeduceAddToSetMethodSerializers(SerializerFinderVisitor visitor, MethodCallExpression expression)
```
sanych-sun, line 300: I've moved all deducers from local method to static-class level methods, because codecov is struggling to show the code coverage for local methods. As the result I can see which deducers are not covered by tests. Probably I need to create more test cases to improve code coverage.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -30,39 +33,10 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
...
-        EnumerableOrQueryableMethod.Where,
-        [MongoEnumerableMethod.WhereWithLimit]
-    ]);
-
     protected override Expression VisitMethodCall(MethodCallExpression node)
     {
-        var method = node.Method;
-        var arguments = node.Arguments;
+        var methodInfo = node.Method.IsGenericMethod && !node.Method.ContainsGenericParameters ? node.Method.GetGenericMethodDefinition() : node.Method;
+        var deducer = MethodCallSerializerDeducers.GetSerializerDeducer(methodInfo);
```
sanych-sun, line 39: I've decided to move all deducers implementation into private static class simply to clear the "entry" point, to have this method on the very top and visible.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+            RegisterSerializerDeducers(EnumerableOrQueryableMethod.DefaultIfEmptyOverloads, DeduceDefaultIfEmptyMethodSerializers);
+            RegisterSerializerDeducer(MongoQueryableMethod.DensifyWithArrayPartitionByFields, DeduceDensifyMethodSerializers);
+            RegisterSerializerDeducers(EnumerableOrQueryableMethod.Distinct, (visitor, expression) => visitor.DeduceCollectionAndCollectionSerializers(expression, expression.Arguments[0]));
+            RegisterSerializerDeducers([MongoQueryableMethod.Documents, MongoQueryableMethod.DocumentsWithSerializer], DeduceDocumentsMethodSerializers);
+            RegisterSerializerDeducers(EnumerableOrQueryableMethod.ElementAtOverloads, (visitor, expression) => visitor.DeduceItemAndCollectionSerializers(expression, expression.Arguments[0]));
+            RegisterSerializerDeducers(EnumerableOrQueryableMethod.Except, DeduceExceptMethodSerializers);
+            // TODO: Definitely wrong registration copy-pasted from prev code, investigate before merge to main
+            // RegisterSerializerResolver(EnumerableMethod.First, DeduceSetWindowFieldsMethodSerializers);
```
sanych-sun: I'm about to investigate this problem. Will fix or remove the registration before merging the PR.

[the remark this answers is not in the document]
sanych-sun: Yep, will investigate. Added comment on that.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-                    var firstItemSerializer =  ArraySerializerHelper.GetItemSerializer(firstSerializer);
-                    AddNodeSerializer(resultSelectorFirstParameter, firstItemSerializer);
+                    var declaringType = expression.Method.DeclaringType;
+                    var nodeSerializer = declaringType switch
+                    {
+                        _ when declaringType == typeof(DateTime) => DateTimeSerializer.Instance,
+                        _ when declaringType == typeof(decimal) => DecimalSerializer.Instance,
+                        _ when declaringType == typeof(double) => DoubleSerializer.Instance,
+                        _ when declaringType == typeof(int) => Int32Serializer.Instance,
+                        _ when declaringType == typeof(short) => Int64Serializer.Instance,
```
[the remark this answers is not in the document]
sanych-sun: Will fix this.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-                DeduceCollectionAndItemSerializers(node, selectorLambda.Body);
+                if (nodeSerializer is IBsonTupleSerializer tupleSerializer)
+                {
+                    for (var i = 1; i <= expression.Arguments.Count; i++)
+                    {
+                        var argumentExpression = expression.Arguments[i];
+                        if (visitor.IsNotKnown(argumentExpression))
+                        {
+                            var itemSerializer = tupleSerializer.GetItemSerializer(i);
+                            if (i == 8)
```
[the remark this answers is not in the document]
sanych-sun: Will investigate.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
adelinowona: Don't forget about this!

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/ByteTests.cs
```diff
@@ -42,13 +42,14 @@ public void SerializerFinder_should_resolve_byte_operations(LambdaExpression exp
         [TestHelpers.MakeLambda((MyModel model) => model.Value.CompareTo(model.Other)), typeof(Int32Serializer)],
         [TestHelpers.MakeLambda((MyModel model) => model.Value.Equals((byte)1)), typeof(BooleanSerializer)],
         [TestHelpers.MakeLambda((MyModel model) => model.Value.Equals(model.Other)), typeof(BooleanSerializer)],
-        [TestHelpers.MakeLambda((MyModel model) => byte.Parse("42")), typeof(ByteSerializer)],
+        [TestHelpers.MakeLambda((MyModel model) => byte.Parse(model.StringValue)), typeof(ByteSerializer)],
```
adelinowona, line 45: I am assuming this is just a refactor to not use the same constant all over right?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+            if (visitor.IsNotKnown(inputParameter) && visitor.IsKnown(partitionExpression, out var partitionSerializer))
             {
-                DeduceUnknownMethodSerializer();
+                var inputSerializer = ((ISetWindowFieldsPartitionSerializer)partitionSerializer).InputSerializer;
+                visitor.AddNodeSerializer(inputParameter, inputSerializer);
             }
         }
 
-        void DeduceTruncateSerializers()
+        public static Action<SerializerFinderVisitor,MethodCallExpression> GetSerializerDeducer(MethodInfo methodInfo)
```
adelinowona: `Action<SerializerFinderVisitor,MethodCallExpression>` is missing a space after the comma

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-                }
-
-                if (IsNotKnown(node))
-                {
-                    var nodeSerializer = StandardSerializers.GetSerializer(node.Type);
-                    AddNodeSerializer(node, nodeSerializer);
+    private static class MethodCallSerializerDeducers
+    {
+        private static readonly Dictionary<MethodInfo, Action<SerializerFinderVisitor, MethodCallExpression>> __serializerResolvers = new();
+        private static readonly ReaderWriterLockSlim __serializerResolversLock = new();
```
adelinowona: Why do we not use a ConcurrentDictionary instead of the whole read-lock / upgrade-to-write / recheck dance in GetSerializerDeducer?

ajcvickers:
Looks like a reasonable change to me.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
```diff
@@ -628,6 +628,22 @@ static EnumerableMethod()
         public static IReadOnlyMethodInfoSet ReverseOverloads => __reverseOverloads;
 
         // public methods
+        public static bool IsContainsMethod(MethodInfo method)
```
damieng, line 636: Is this going to also match String.Contains?  A lot of providers typically special case excluding String.Contains as we don't want to handle looking for chars in a string in the same way we look for types in a containable list.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/FloatTests.cs
```diff
@@ -42,13 +42,14 @@ public void SerializerFinder_should_resolve_float_operations(LambdaExpression ex
         [TestHelpers.MakeLambda((MyModel model) => model.Value.CompareTo(model.Other)), typeof(Int32Serializer)],
         [TestHelpers.MakeLambda((MyModel model) => model.Value.Equals(1f)), typeof(BooleanSerializer)],
         [TestHelpers.MakeLambda((MyModel model) => model.Value.Equals(model.Other)), typeof(BooleanSerializer)],
-        [TestHelpers.MakeLambda((MyModel model) => float.Parse("42")), typeof(SingleSerializer)],
+        [TestHelpers.MakeLambda((MyModel model) => float.Parse(model.StringValue)), typeof(SingleSerializer)],
```
damieng, line 45: 42 probably should have fractial parts for the float/decimal/double tests originally?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-                    DeduceItemAndCollectionSerializers(selectorSourceItemParameter, sourceExpression);
-                }
-
-                if (IsNotKnown(node))
-                {
-                    var nodeSerializer = StandardSerializers.GetSerializer(node.Type);
-                    AddNodeSerializer(node, nodeSerializer);
+    private static class MethodCallSerializerDeducers
+    {
+        private static readonly Dictionary<MethodInfo, Action<SerializerFinderVisitor, MethodCallExpression>> __serializerResolvers = new();
```
adelinowona: Also everything else in this code seems to be using the vocab "Deducer" so I would suggest to rename this to `serializerDeducers` for consistency. 

Also rename the lambda parameter name `resolver` in `RegisterSerializerDeducer`

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                var sourceExpression = expression.Arguments[0];
+                var selectorLambda = ExpressionHelper.UnquoteLambdaIfQueryableMethod(expression.Method, expression.Arguments[1]);
+                var selectorItemParameter = selectorLambda.Parameters.Single();
+                visitor.DeduceCollectionAndItemSerializers(sourceExpression, selectorItemParameter);
             }
+
+            visitor.DeduceStandardSerializer(expression);
+        }
+
+        private static void DeduceStringEqualsMethodSerializers(SerializerFinderVisitor visitor, MethodCallExpression expression)
```
adelinowona: This method is identical to `DeduceEqualsMethodSerializers`. I'll suggest renaming this method to `DeduceEqualsMethodSerializers`, point both registration sites at it and remove the local-only `DeduceEqualsMethodSerializers` method in `CreateDynamicSerializerDeducer`

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                {
+                    var source = expression.Method.IsStatic ? expression.Arguments[0] : expression.Object;
+                    if (visitor.IsKnown(source, out var sourceSerializer))
+                    {
+                        var sourceItemSerializer = ArraySerializerHelper.GetItemSerializer(sourceSerializer);
+                        var resultSerializer = ListSerializer.Create(sourceItemSerializer);
+                        visitor.AddNodeSerializer(expression, resultSerializer);
+                    }
+                }
             }
```
adelinowona, line 1804: Why are DeduceToArray and DeduceToList so different? Shouldn't they have similar bodies?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                var dictionaryExpression = expression.Object;
+                var keyExpression = expression.Arguments[0];
+                if (visitor.IsNotKnown(keyExpression) && visitor.IsKnown(dictionaryExpression, out var dictionarySerializer))
+                {
+                    var keySerializer = (dictionarySerializer as IBsonDictionarySerializer)?.KeySerializer;
+                    visitor.AddNodeSerializer(keyExpression, keySerializer);
+                }
+
+                visitor.DeduceSerializer(expression, BooleanSerializer.Instance);
             }
```
adelinowona: why are both of these not static?

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
adelinowona, line 636: I think this new method is also duplicating matching logic with the existing `IsContainsMethod(methodcallexpression, out, out)` method. You rewrite the old method as 
`if (IsContainsMethod(methodCallExpression.Method)) { … }` and then extract source/value

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+
+            visitor.DeduceItemAndCollectionSerializers(collectionSelectorSourceItemParameter, sourceExpression);
+            visitor.DeduceItemAndCollectionSerializers(resultSelectorSourceItemParameter, sourceExpression);
+            visitor.DeduceItemAndCollectionSerializers(resultSelectorCollectionItemParameter, collectionSelectorLambda.Body);
+            visitor.DeduceCollectionAndItemSerializers(expression, resultSelectorLambda.Body);
         }
 
-        void DeduceSetEqualsMethodSerializers()
+        // DeduceSelectManySerializers + EnumerableOrQueryableMethod.SelectManyWithSelectorTakingIndex branch
+        private static void SelectManyWithSelectorTakingIndexMethodSerializers(SerializerFinderVisitor visitor, MethodCallExpression expression)
```
adelinowona: Missing the Deduce prefix?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                "Exists" when method.Is(ArrayMethod.Exists) || ListMethod.IsExistsMethod(method) => DeduceExistsMethodSerializers,
+                "get_Item" when BsonValueMethod.IsGetItemWithIntMethod(method) || BsonValueMethod.IsGetItemWithStringMethod(method) =>
+                    (visitor, expression) => visitor.DeduceSerializer(expression, BsonValueSerializer.Instance),
+                "get_Item" when !method.IsStatic => DeduceGetItemInstanceMethodSerializers,
+                "IsSubsetOf" when IsSubsetOfMethod(method) => DeduceIsSubsetOfMethodSerializers,
+                "Parse" when IsParseMethod(method) => DeduceParseMethodSerializers,
+                "SetEquals" when ISetMethod.IsSetEqualsMethod(method) => DeduceSetEqualsMethodSerializers,
+                "ToArray" when IsToArrayMethod(method) => DeduceToArrayMethodSerializers,
+                "ToList" => DeduceToListSerializers,
+                "ToString" => (visitor, expression) => visitor.DeduceSerializer(expression, StringSerializer.Instance),
```
adelinowona, line 1598: why don't these have guards like the others?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1731 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                not null when returnType == typeof(double) => DoubleSerializer.Instance,
+                not null when returnType == typeof(int) => Int32Serializer.Instance,
+                not null when returnType == typeof(long) => Int64Serializer.Instance,
+                not null when returnType == typeof(float) => SingleSerializer.Instance,
+                not null when returnType == typeof(decimal?) => NullableSerializer.NullableDecimalInstance,
+                not null when returnType == typeof(double?) => NullableSerializer.NullableDoubleInstance,
+                not null when returnType == typeof(int?) => NullableSerializer.NullableInt32Instance,
+                not null when returnType == typeof(long?) => NullableSerializer.NullableInt64Instance,
+                not null when returnType == typeof(float?) => NullableSerializer.NullableSingleInstance,
+                _ => throw new NotImplementedException($"Cannot resolver serializer for {returnType}")
```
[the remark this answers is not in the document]
adelinowona: Is this not a typo?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -72,2948 +46,1735 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+            var fromExpression = expression.Arguments[1];
+            var fromCollection = fromExpression.GetConstantValue<IMongoCollection>(expression);
+            var foreignDocumentSerializer = fromCollection.DocumentSerializer;
+            var localFieldLambda = ExpressionHelper.UnquoteLambdaIfQueryableMethod(expression.Method, expression.Arguments[2]);
+            var localFieldLambdaParameter = localFieldLambda.Parameters.Single();
+            var foreignFieldLambda = ExpressionHelper.UnquoteLambdaIfQueryableMethod(expression.Method, expression.Arguments[3]);
+            var foreignFieldLambdaParameter = foreignFieldLambda.Parameters.Single();
+            var pipelineLambda = ExpressionHelper.UnquoteLambdaIfQueryableMethod(expression.Method, expression.Arguments[4]);
+            var pipelineLambdaLocalParameter = pipelineLambda.Parameters[0];
+            var pipelineLamdbaForeignQueryableParameter = pipelineLambda.Parameters[1];
```
adelinowona: fix typo here in variable name lamdba -> lambda

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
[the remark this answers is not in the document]
adelinowona: I second copilot on this

adelinowona (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Will do. Somehow I was thinking having `Dictionary<>` for "mostly readonly" access might be better, but apparently `ConcurrentDictionary` also handles this use case better then combination of `Dictionary<>` and `ReadWriteLockSlim`.
Thank you for the comment!

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/ByteTests.cs
sanych-sun: Not exactly. We need non-constant expression here, because otherwise `PartialEvaluator` will execute it and replace with `ConstantExpression` and we will not test the `Parse` method at all.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/SerializerFinders/FloatTests.cs
sanych-sun: We need non-constant here, so the `PartialEvaluator` will not replace this with `ConstantExpression`.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: It's an interesting question, but it's out of this PR scope. I'll double-check if the method was migrated/moved properly and can create a separate ticket for the investigation if you want.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Do you mean checking if node is known already? Because `DeduceSerializer` is doing that for us + we do not create any new serializers here, but passing pre-created one.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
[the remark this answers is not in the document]
sanych-sun: Fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Done, however instead of simply use the `GetOrAdd` method, I've decided to use `TryGetValue`, then manually call `CreateDynamicSerializerDeducer` and if it returns not-null then use `GetOrAdd`. We need this to avoid caching of `null` values in the serializer deducers dictionary.

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
sanych-sun: Nope, `String.Contains` has it's pre-registered deducer (see like 216).

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Done.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Done.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Fixed

src/MongoDB.Driver/Linq/Linq3Implementation/Reflection/EnumerableMethod.cs
sanych-sun: Done.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -30,2990 +33,1725 @@ namespace MongoDB.Driver.Linq.Linq3Implementation.SerializerFinders;
...
-                DeduceReturnsNumericSerializer();
+                return deducer;
             }
-            else
+
+            deducer = CreateDynamicSerializerDeducer(methodInfo);
+            if (deducer == null)
             {
-                DeduceUnknownMethodSerializer();
+                return null;
```
adelinowona, line 1573: I believe the old code was caching this null case. We don't need to do that anymore?

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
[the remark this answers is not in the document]
sanych-sun: I've created a ticket to investigate the problem: https://jira.mongodb.org/browse/CSHARP-6053

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderHelperMethods.cs
sanych-sun: Created: https://jira.mongodb.org/browse/CSHARP-6113

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: Investigated. `EnumerableMethod.First` - is covered as a part of `EnumerableOrQueryableMethod.FirstOrLastOrSingleOverloads` registration. Window method First is covered by `WindowMethod.First` registration. Seems like it was a wrong branch in the original code.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
sanych-sun: I think there is no needs to cache the negative detection. Because if `SerializerFinder` cannot find a serializer it means we already either on a way to throw an exception or we are about to switch to client side projection. I do not think we need to optimize those cases by cost of potentially unbound growing cache.

src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -1000,2163 +629,1246 @@ static IBsonSerializer GetResultSerializer(Expression expression, Type toType)
...
-            {
-                DeduceUnknownMethodSerializer();
-            }
-        }
+            var sourceExpression = expression.Arguments[0];
+            var fromCollection = expression.Arguments[1].GetConstantValue<IMongoCollection>(expression);
+            var foreignDocumentSerializer = fromCollection.DocumentSerializer;
+            var pipelineLambda = ExpressionHelper.UnquoteLambdaIfQueryableMethod(expression.Method, expression.Arguments[2]);
+            var pipelineLambdaLocalParameter = pipelineLambda.Parameters[0];
+            var pipelineLamdbaForeignQueryableParameter = pipelineLambda.Parameters[1];
```
adelinowona: typo in "lambda"  on pipelineLamdbaForeignQueryableParameter.

adelinowona (approved):
LGTM but noted a minor typo to fix

PR #1963, opened 2026-04-23
CSHARP-5930: [Tiny] Fix synchronization issue in ChangeStream examples

ajcvickers:


BorisDog (approved):
LGTM.
And also we are considering to remove the examples completely for this project. Docs team will be maintaining a separate set of examples.

PR #1966, opened 2026-04-23
CSHARP-5930: [Tiny] Gate ConnectivityTests.cs on environment

ajcvickers:


evergreen/run-atlas-connectivity-tests.sh
```diff
@@ -6,6 +6,8 @@ set -o errexit  # Exit the script with error if any of the commands fail
 #            Main Program                  #
 ############################################
 
+export ATLAS_SEARCH_CONNECTIVITY_TESTS_ENABLED=true
```
adelinowona: These are not Atlas search tests. Just connection tests. I'll suggest dropping the search term from the env var

adelinowona (changes requested):

adelinowona (approved):

PR #1967, opened 2026-04-23
CSHARP-5930: [Tiny] Avoid use of generic test method

ajcvickers:
This messes with test discovery in Rider.

### DIFF
--- a/tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
+++ b/tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
@@ -40,9 +40,15 @@ namespace MongoDB.Driver.Tests.Search
 
         protected override void DisposeInternal() => _mongoClient.Dispose();
 
-        [Theory]
-        [MemberData(nameof(BinaryVectorSearchTestData))]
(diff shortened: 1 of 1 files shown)


BorisDog (commented):
On my machine, I can see that the tests are discovered only at runtime. Which is not optimal. 
I am not sure we should be avoiding using generics in tests as a strict rule. But I think it's fine in this case.
Any idea if generics work better in the next xunit version?

Also this probably should be CSHARP-5930.

BorisDog (approved):
LGTM + please update the ticket # in the title.

ajcvickers:
@BorisDog XUnit test discovery is tricky. Real discovery requires executing code from the test assembly--this is what allows discovery to be customized by your own code. This means that a) graphical runners, like Rider, often do heuristic test discovery by static analysis, meaning the heuristically discovered tests may not be the same as the tests discovered for real. This can lead to inconsistencies in the UI, like in this issue.

It also means that b) if custom code fails in discovery, then it can be hard to find and debug. Usually it results in lots of Inconclusive test results--as was the case in the bug I fixed a couple of weeks ago. But sometimes it can just silently not run tests. Which is bad. :-)

PR #1969, opened 2026-04-24
CSHARP-5930: Workaround for DNS issue with Cloudflare

ajcvickers:
Root cause: Cloudflare WARP writes 4 entries to /etc/resolv.conf:

- 127.0.2.2 (works — 187ms)
- 127.0.2.3 (works — 28ms)
- ::ffff:127.0.2.2 (IPv4-mapped IPv6 form)
- ::ffff:127.0.2.3 (IPv4-mapped IPv6 form)

DnsClient reads /etc/resolv.conf directly on macOS and picks up all 4 entries. When it selects one of the ::ffff: IPv4-mapped IPv6 addresses for a UDP query, macOS doesn't route that correctly (IPv4-mapped IPv6 loopback via UDP isn't reliable on macOS), so the socket times out. The error [::ffff:127.0.2.3]:53 timed out or is a transient error is exactly that — DnsClient tried the IPv6-mapped form of the address and got no response.

The fix is that DnsClientWrapper.cs — CreateLookupClient() now calls NameServer.ResolveNameServers(), filters out any IPv4-mapped IPv6 addresses whose IPv4 equivalent is already in the list, and passes the cleaned list to LookupClientOptions. This fixes the root cause for WARP (and any other config that produces such duplicates).

This makes the tests pass, but also fixes this as a possible real issues for customers using Cloudflare.

### DIFF
--- a/src/MongoDB.Driver/Core/Misc/DnsClientWrapper.cs
+++ b/src/MongoDB.Driver/Core/Misc/DnsClientWrapper.cs
@@ -15,8 +15,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net;
+using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using DnsClient;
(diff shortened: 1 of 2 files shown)


BorisDog (changes requested):
I think such a change should be discussed in a broader forum.
Probably this logic should be similar in all driver, which requires [spec changes](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md).

ajcvickers:
@BorisDog

Me: Isn't the change in this commit working around an issue in a dependency? If so, what are the chances that the same issue happens for non .NET drivers?                                                               
                                                                                                                                                         

⏺ Yes, this is working around an issue in the DnsClient NuGet package, compounded by a macOS behavior.
                                                                                                                                                                                                                    
What causes it:

  1. Cloudflare WARP writes duplicate entries to /etc/resolv.conf — both plain IPv4 (127.0.2.3) and IPv4-mapped IPv6 (::ffff:127.0.2.3).
  2. DnsClient reads /etc/resolv.conf directly (it has its own resolver, not the OS's) and picks up all entries.
  3. When it selects ::ffff:127.0.2.3 for a UDP query, macOS silently drops it — IPv4-mapped IPv6 loopback via UDP is unreliable there.

Chances for other drivers:
                                                                                                                                                                                                                    
Relatively low, because most other drivers use the OS system resolver (getaddrinfo, mDNSResponder on macOS, or c-ares) rather than reading /etc/resolv.conf themselves. The OS resolver handles or ignores IPv4-mapped duplicates gracefully.                        
                                                                                                                                                                                                                    
The risk only materializes when three things align simultaneously:
- Cloudflare WARP is installed and writes the duplicate entries
- The driver uses a pure-implementation DNS library that reads /etc/resolv.conf directly (not the OS resolver) 
- That library sends UDP queries to IPv4-mapped IPv6 loopback addresses without filtering
                                   
The C# driver's use of DnsClient is what makes it uniquely exposed. Other drivers using dnspython, Java's dnsjava, etc. could theoretically hit the same issue if they also read /etc/resolv.conf directly, but the most common path for those ecosystems goes through the OS resolver. Worth a note to those teams if it comes up, but it's not an immediate concern.

BorisDog (changes requested):
From my quick research, I’m not confident that it is always safe to remove IPv6 duplicates. Are there specific clients or OSs that expect only IPv6? Is resolving IPv6 formatted IPv4 and IPv4 always equivalent in all possible network settings? (Side question, if it is 100% equivalent, why does Cloudflare bother to add it?)

Furthermore, this doesn't appear to be an issue with the DnsClient NuGet package, but rather a case of invalid configuration. I don’t believe our library should compensate for configuration errors, just as we don’t compensate for invalid DNS records or a malformed MongoUrl. But if we choose to do that, we should clearly state that in drivers spec.
For changes like this, we should at least consult with other drivers; ideally, we should aim for consistent behavior across the board.

Since this fix is being driven by local test failures, I’d prefer to explore a less intrusive solution to avoid unnecessary risks.

PR #1972, opened 2026-04-27
CSHARP-5930: [Tiny] Correctly lookup server type for read preference test verification

ajcvickers:
See CSHARP-6008 for details about why this change is needed.

### DIFF
--- a/tests/MongoDB.Driver.Tests/ReadPreferenceOnStandaloneTests.cs
+++ b/tests/MongoDB.Driver.Tests/ReadPreferenceOnStandaloneTests.cs
@@ -19,7 +19,6 @@ using System.Threading;
 using FluentAssertions;
 using MongoDB.Bson;
 using MongoDB.Driver.Core;
-using MongoDB.Driver.Core.Clusters;
 using MongoDB.Driver.Core.Events;
 using MongoDB.Driver.Core.Servers;
 using MongoDB.Driver.Core.TestHelpers.Logging;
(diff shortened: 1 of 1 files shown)


tests/MongoDB.Driver.Tests/ReadPreferenceOnStandaloneTests.cs
```diff
@@ -60,9 +59,9 @@ public void ReadPreference_should_not_be_sent_to_standalone_server(
                 CommandStartedEvent sentCommand = ((CommandStartedEvent)eventCapturer.Events[0]);
                 SpinWait.SpinUntil(() => client.Cluster.Description.Servers.Any(s => s.State == ServerState.Connected), TimeSpan.FromSeconds(5)).Should().BeTrue();
 
-                var clusterType = client.Cluster.Description.Type;
+                var serverType = client.Cluster.Description.Servers.First().Type;
 
-                var expectedContainsReadPreference = clusterType != ClusterType.Standalone;
+                var expectedContainsReadPreference = serverType != ServerType.Standalone;
```
sanych-sun, line 64: The suggested changes looks like a reasonable workaround for the ClusterType detection problem. Should we mention this workaround in the [CSHARP-6008](https://jira.mongodb.org/browse/CSHARP-6008) by asking to remove the workaround when/if the bug will be fixed?

ajcvickers: I'm not sure that 6008 is a bug or will be fixed. I'm following up on what appear to be two different uses of directConnection=true--the original intent being to connect to a specific node for diagnostics/admin/etc., and the use with Atlas Local where it makes it possible for Atlas to control the internal network, etc.

I think if we decide to make changes based on 6008, then part of that should be finding all uses and validating that they are doing the right thing.

papafe (approved):
LGTM

PR #1974, opened 2026-04-28
CSHARP-5930: Update replica-set discovery and reporting in tests

ajcvickers:
See CSHARP-6008 for details about why this change is needed.

There were multiple iterations getting to this point:

## Iteration 1

RequireServer.cs
- ClusterType()/ClusterTypes(): Now use GetActualClusterType() which maps a directConnect RS member to ClusterType.ReplicaSet (instead of the misleading Standalone)
- topology/topologies case: Uses actual server type — single matches only Standalone, replicaset matches IsReplicaSetMember()

DriverTestConfiguration.cs
- IsReplicaSet(): Checks server.Type.IsReplicaSetMember() instead of cluster.Description.Type == ClusterType.ReplicaSet, so a directConnect RS is correctly identified

ReadPreferenceOnStandaloneTests.cs (from earlier)
- Uses ServerType.Standalone instead of ClusterType.Standalone to decide whether $readPreference is expected

ServerDiscoveryProseTests.cs
- Missing secondary now throws SkipException instead of a plain Exception

UnifiedTestSpecRunner.cs
- Added IsDirectConnectionToReplicaSet() helper
- ServerDiscoveryAndMonitoring: Skips logging-replicaset.json, replicaset-emit-topology-changed-before-close.json, rediscover-quickly-after-step-down.json when on directConnect RS (these require full multi-server topology discovery events)
- ServerSelection: Skips replica-set.json when on directConnect RS (same reason)

## Iteration 2

- Root cause: My GetActualClusterType() change made RequireServer.ClusterType(ClusterType.ReplicaSet) pass for a directConnect RSPrimary. This caused Connection_pool_should_not_be_cleared_when_replSetStepDown_and_GetMore to run — a test that was previously always skipped on this setup. That test calls { replSetStepDown: 30, force: true }, which makes the single-node RS step down and refuse writes for up to 30 seconds. All subsequent write operations in other tests (like DropCollection in fixture setup) then fail with MongoNotPrimaryException.
- Fix: Added RequireServer.ReplicaSetDataBearingMembers(int minimum) — a new check that counts how many data-bearing RS members the cluster has

## Iteration 3

- Applied .ReplicaSetDataBearingMembers(2) to Connection_pool_should_not_be_cleared_when_replSetStepDown_and_GetMore — the step-down test only makes sense (and is safe) with at least 2 RS members so a new primary can be elected after the step-down

## Iteration 4

- CausalConsistencyExamples.cs — Causal_Consistency_Example_3: Replaced the hardcoded "mongodb://localhost/?readPreference=secondaryPreferred" with settings built from CoreTestConfiguration.ConnectionString plus ReadPreference.SecondaryPreferred. The example intent is preserved.
- ClientSideEncryption2Examples.cs — FLE2AutomaticEncryption: Added an explicit skip when CRYPT_SHARED_LIB_PATH is not set (CSFLE/auto-encryption requires it), and fixed two new MongoClient() calls to use CoreTestConfiguration.ConnectionString.ToString() instead of the default localhost:27017.

### DIFF
--- a/tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
+++ b/tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
@@ -18,7 +18,9 @@ using System.Linq;
 using MongoDB.Bson;
 using MongoDB.Driver.Core.Clusters;
 using MongoDB.Driver.Core.Misc;
+using MongoDB.Driver.Core.Servers;
 using MongoDB.Driver.Encryption;
+using MongoDB.Driver.Tests;
 using Xunit.Sdk;
 
 namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
@@ -67,7 +69,7 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
 
         public RequireServer ClusterType(ClusterType clusterType)
         {
-            var actualClusterType = CoreTestConfiguration.Cluster.Description.Type;
+            var actualClusterType = GetActualClusterType();
             if (actualClusterType == clusterType)
             {
                 return this;
@@ -77,7 +79,7 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
 
         public RequireServer ClusterTypes(params ClusterType[] clusterTypes)
         {
-            var actualClusterType = CoreTestConfiguration.Cluster.Description.Type;
+            var actualClusterType = GetActualClusterType();
             if (clusterTypes.Contains(actualClusterType))
             {
                 return this;
@@ -239,6 +241,16 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
             throw new SkipException($"Test skipped because the cluster does not have multiple mongoses.");
         }
 
+        public RequireServer ReplicaSetDataBearingMembers(int minimum)
+        {
+            var actualCount = DriverTestConfiguration.GetReplicaSetNumberOfDataBearingMembers(CoreTestConfiguration.Cluster);
+            if (actualCount >= minimum)
+            {
+                return this;
+            }
+            throw new SkipException($"Test skipped because replica set has {actualCount} data-bearing member(s) and at least {minimum} are required.");
+        }
+
         public RequireServer VersionGreaterThanOrEqualTo(SemanticVersion version)
         {
             var actualVersion = CoreTestConfiguration.ServerVersion;
@@ -285,6 +297,19 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
         }
 
         // private methods
+        private ClusterType GetActualClusterType()
+        {
+            var description = CoreTestConfiguration.Cluster.Description;
+            // For direct connections, Description.Type is always Standalone regardless of the actual
+            // server type. Use the server's reported type to get the true effective cluster type.
+            if (description.DirectConnection)
+            {
+                var serverType = description.Servers.FirstOrDefault()?.Type ?? ServerType.Unknown;
+                return serverType.ToClusterType();
+            }
+            return description.Type;
+        }
+
         private bool CanRunOn(BsonDocument requirements)
         {
             return requirements.All(IsRequirementSatisfied);
@@ -335,9 +360,8 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
                     return true;
                 case "topologies":
                 case "topology":
-                    var actualClusterType = CoreTestConfiguration.Cluster.Description.Type;
-                    var runOnClusterTypes = requirement.Value.AsBsonArray.Select(topology => MapTopologyToClusterType(topology.AsString)).ToList();
-                    return runOnClusterTypes.Contains(actualClusterType);
+                    var actualServerType = CoreTestConfiguration.Cluster.Description.Servers.FirstOrDefault()?.Type ?? ServerType.Unknown;
+                    return requirement.Value.AsBsonArray.Any(topology => IsTopologyMatch(topology.AsString, actualServerType));
                 case "csfle":
                     return IsCsfleRequirementSatisfied(requirement);
                 default:
@@ -365,15 +389,15 @@ namespace MongoDB.Driver.Core.TestHelpers.XunitExtensions
             return SemanticVersionCompareToAsReleased(actualLibmongocryptVersion, minLibmongocryptVersion) >= 0;
         }
 
-        private ClusterType MapTopologyToClusterType(string topology)
+        private bool IsTopologyMatch(string topology, ServerType serverType)
         {
             switch (topology)
             {
-                case "single": return Clusters.ClusterType.Standalone;
-                case "replicaset": return Clusters.ClusterType.ReplicaSet;
+                case "single": return serverType == ServerType.Standalone;
+                case "replicaset": return serverType.IsReplicaSetMember();
                 case "sharded-replicaset":
-                case "sharded": return Clusters.ClusterType.Sharded;
-                case "load-balanced": return Clusters.ClusterType.LoadBalanced;
+                case "sharded": return serverType == ServerType.ShardRouter;
+                case "load-balanced": return serverType == ServerType.LoadBalanced;
                 default: throw new ArgumentException($"Invalid topology: \"{topology}\".", nameof(topology));
             }
         }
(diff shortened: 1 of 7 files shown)


tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
```diff
@@ -239,6 +242,40 @@ public RequireServer MultipleMongoses(bool required)
...
+            // immediately rather than spinning the full 10s waiting for a count that cannot grow.
+            if (cluster.Description.DirectConnection && minimum > 1)
+            {
+                throw new SkipException($"Test skipped because directConnection only sees the single targeted server and at least {minimum} data-bearing member(s) are required: {cluster.Description}.");
+            }
+            // Spin directly on the data-bearing count so a transient SDAM blip on a non-data-bearing
+            // member (or a still-Connecting data-bearing member) does not cause an early skip: as
+            // soon as `minimum` data-bearing members are visible we accept the topology.
+            var hasMinimumDataBearing = SpinWait.SpinUntil(
+                () => cluster.Description.Servers.Count(s => s.IsDataBearing) >= minimum,
```
sanych-sun: I think the condition is suboptimal: if we are waiting for 3 databearing nodes, but cluster configured only to have 2 - this will spin until timeout. We probably should spin-wait for the cluster to be connected, and then simply check number of the data-bearing nodes outside of the spin-wait.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
```diff
@@ -285,6 +322,37 @@ public RequireServer VersionLessThanOrEqualTo(string version)
...
+            }
+            // For non-direct connections the cluster machinery resolves Description.Type before tests
+            // run, but spin briefly here as well so early-startup invocations don't race against
+            // the initial SDAM rounds.
+            var typeIsKnown = SpinWait.SpinUntil(
+                () => cluster.Description.Type != ClusterTypeEnum.Unknown,
+                TimeSpan.FromSeconds(10));
+            if (!typeIsKnown)
+            {
+                throw new SkipException($"Test skipped because cluster type could not be determined: {cluster.Description}.");
```
sanych-sun: We should throw something like `InvalidOperationException` here, we want to fail the test not just skip it.

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -277,6 +303,92 @@ private void Run(JsonDrivenTestCase testCase, Predicate<LogEntry> loggingFilter
...
+            var servers = cluster.Description.Servers;
+            if (servers.Count != 1)
+            {
+                return false;
+            }
+            // Matches both directConnect to a RS member and a non-directConnect single-node RS
+            return servers[0].Type.IsReplicaSetMember();
+        });
+
+        private static bool IsSingleNodeReplicaSet() => __isSingleNodeReplicaSet.Value;
```
sanych-sun: Can we move `IsSingleNodeReplicaSet` into `DriverTestConfiguration`, please?

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
```diff
@@ -239,6 +242,40 @@ public RequireServer MultipleMongoses(bool required)
             throw new SkipException($"Test skipped because the cluster does not have multiple mongoses.");
         }
 
+        public RequireServer ReplicaSetDataBearingMembers(int minimum)
+        {
```
sanych-sun, line 245: Can we reuse `DriverTestConfiguration.GetReplicaSetNumberOfDataBearingMembers` method hrere?

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
```diff
@@ -285,6 +322,37 @@ public RequireServer VersionLessThanOrEqualTo(string version)
         }
 
         // private methods
+        private ClusterType GetActualClusterType()
```
sanych-sun: Should we move this method to `DriverTestConfiguration`?

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -277,6 +303,92 @@ private void Run(JsonDrivenTestCase testCase, Predicate<LogEntry> loggingFilter
...
+        // Note: the resource-name check is intentionally global — it matches a fileName found
+        // anywhere under the assembly's embedded spec resources, not just under the calling spec
+        // method's own resource subtree. A SkipFile call from one spec method that names a JSON
+        // file owned by a different spec area will validate (no typo thrown) but the per-testcase
+        // StartsWith check below will still not match within the wrong runner's testcase stream, so
+        // it skips nothing. Callers are expected to reference files from their own spec area.
+        private static void SkipFile(JsonDrivenTestCase testCase, string fileName, string reason)
+        {
+            var suffix = "." + fileName;
+            if (!__embeddedJsonSpecResources.Value.Any(r => r.EndsWith(suffix)))
```
sanych-sun: Is this a way to validate if the test case belong to a particular test file? Should we extend `JsonDrivenTestCase` instead and enrich it with new property like "source resource name"?

sanych-sun: Quick update - we already enriching the BsonDocument with test cases with `_localPath` and `_fileName`. I think we might use that information. See UnifiedTestsDiscoverer ln 110

sanych-sun (changes requested):

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: Fixed.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: Done.

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
ajcvickers: Done.

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
ajcvickers: Done.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: Done.

tests/MongoDB.Driver.TestHelpers/Core/XunitExtensions/RequireServer.cs
ajcvickers: Fixed in method.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -277,6 +301,20 @@ private void Run(JsonDrivenTestCase testCase, Predicate<LogEntry> loggingFilter
         private static void RequireKmsMock() =>
             RequireEnvironment.Check().EnvironmentVariable("KMS_MOCK_SERVERS_ENABLED");
 
+        private static bool IsSingleNodeReplicaSet() => DriverTestConfiguration.IsSingleNodeReplicaSet(CoreTestConfiguration.Cluster);
+
+        // Skip all tests sourced from a specific JSON spec file. The source file is identified via
```
sanych-sun: _Minor_: If we want to keep this comment, let's convert it into XML-docs on the method.

tests/MongoDB.Driver.Tests/Specifications/UnifiedTestSpecRunner.cs
```diff
@@ -277,6 +301,20 @@ private void Run(JsonDrivenTestCase testCase, Predicate<LogEntry> loggingFilter
...
+        // UnifiedTestsDiscoverer), rather than by parsing the composite test-case name.
+        private static void SkipFile(JsonDrivenTestCase testCase, string fileName, string reason)
+        {
+            if (testCase.Shared.GetValue("_fileName", null)?.AsString == fileName)
+            {
+                throw new SkipException(reason);
+            }
+        }
+
+        // Skip tests whose name contains the given operation substring.
```
sanych-sun: _Minor_: same here, let's convert to XML docs

PR #1978, opened 2026-04-29
CSHARP-5930: Skip tests on Mac with Cloudflare WARP due to DnsClient bug

ajcvickers:
Original change attempting to fix the bug in the product was rejected. This change detects the condition and skips the tests instead.

- RequireEnvironment.cs — added NoDuplicateIpv4MappedNameServers(). On non-Windows it reads /etc/resolv.conf, parses nameserver lines, and throws SkipException if any IPv4-mapped IPv6 address (::ffff:x.x.x.x) is present alongside its plain IPv4 equivalent. On Windows or when /etc/resolv.conf doesn't exist it's a no-op.


papafe (approved):
LGTM

BorisDog (approved):
Nice, LGTM.

PR #1981, opened 2026-05-01
CSHARP-5658: Queries using Equals with different types throw

kyra-rk:


kyra-rk:
Other filter translators have the same `GetConstantValue<object>` and `SerializeValue` pattern and could hit the same `InvalidCastException` for mismatched types:

  - CompareComparisonExpressionToFilterTranslator
  - BitMaskComparisonExpressionToFilterTranslator
  - HasFlagMethodToFilterTranslator
  - ContainsMethodToFilterTranslator
  - GetItemMethodToFilterFieldTranslator

If this fix looks good, it might be worth extracting the conversion into a shared helper and applying it to these as well.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/EqualsMethodToFilterTranslatorTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        [Fact]
+        public void Equals_with_uint64_and_nullable_int32_should_translate()
+        {
+            var collection = Fixture.Collection;
+            ulong longPrm = 2;
+
+            var queryable = collection.AsQueryable()
+                .Where(e => e.ReportsTo.Equals(longPrm));
```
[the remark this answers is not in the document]
kyra-rk: Outdated

papafe:
Hey @adelinowona, I took over Kyra for this ticket. What do you think of the approach?

adelinowona (commented):
Sorry for the delay in review but I think it's solving the problem at the wrong layer.

  The driver already handles numeric field/value mismatches in [`FieldValueSerializerHelper.GetSerializerForValueType`](https://github.com/mongodb/mongo-csharp-driver/blob/bb89755e25a5ea519bf30d305a0140e244dd9bc1/src/MongoDB.Driver/FieldValueSerializerHelper.cs#L51-L60) (used by the builder API, `FieldDefinition`, `LinqProviderAdapter`). Its numeric branch deliberately *doesn't* convert the value — it serializes the value with a serializer for its own type, and also covers nullable/enum/string-representation.

  Coercing the value via `Convert.ChangeType` instead introduces problems that mechanism avoids:

  - Silent rounding: `Equals(2.9) → { IntField : 3 }`, matching docs the original never could.
  - Enums: `Type.GetTypeCode(enum)` is numeric, so an enum field hits `Convert.ChangeType(int, enumType)` → InvalidCastException.
  - Leaks OverflowException/InvalidCastException instead of ExpressionNotSupportedException.
  - Precision loss on decimal/double; and it only fixes `.Equals`, not ==/>/etc.

  Can we route Equals through `GetSerializerForValueType` instead? That fixes all of the above, keeps Equals consistent with other operators, and lets us drop the private `IsNumericType` (which duplicates the
  existing `TypeExtensions.IsNumeric()`).

  One decision to make explicit: the lossless path renders `Equals(2.9)` as `{ IntField : 2.9 }` (matches nothing) rather than throwing — fine, but let's choose it deliberately.

papafe:
> Sorry for the delay in review but I think it's solving the problem at the wrong layer.
> 
> The driver already handles numeric field/value mismatches in [`FieldValueSerializerHelper.GetSerializerForValueType`](https://github.com/mongodb/mongo-csharp-driver/blob/bb89755e25a5ea519bf30d305a0140e244dd9bc1/src/MongoDB.Driver/FieldValueSerializerHelper.cs#L51-L60) (used by the builder API, `FieldDefinition`, `LinqProviderAdapter`). Its numeric branch deliberately _doesn't_ convert the value — it serializes the value with a serializer for its own type, and also covers nullable/enum/string-representation.
> 
> Coercing the value via `Convert.ChangeType` instead introduces problems that mechanism avoids:
> 
> * Silent rounding: `Equals(2.9) → { IntField : 3 }`, matching docs the original never could.
> * Enums: `Type.GetTypeCode(enum)` is numeric, so an enum field hits `Convert.ChangeType(int, enumType)` → InvalidCastException.
> * Leaks OverflowException/InvalidCastException instead of ExpressionNotSupportedException.
> * Precision loss on decimal/double; and it only fixes `.Equals`, not ==/>/etc.
> 
> Can we route Equals through `GetSerializerForValueType` instead? That fixes all of the above, keeps Equals consistent with other operators, and lets us drop the private `IsNumericType` (which duplicates the existing `TypeExtensions.IsNumeric()`).
> 
> One decision to make explicit: the lossless path renders `Equals(2.9)` as `{ IntField : 2.9 }` (matches nothing) rather than throwing — fine, but let's choose it deliberately.

Makes sense, I've committed a change to use `GetSerializerForValueType`

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/EqualsMethodToFilterTranslatorTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        [Fact]
+        public void Equals_with_string_and_nullable_int_should_translate()
+        {
+            var value = "2";
+
+            var filter = Builders<TestClass>.Filter.Where(e => e.NullableIntegerProperty.Equals(value));
+
+            filter.Render(__args).Should().Be("{ NullableIntegerProperty : 2 }");
+        }
```
adelinowona: So the renamed Equals_with_string_and_nullable_int_should_translate test now asserts `e.NullableIntegerProperty.Equals("2") → {  NullableIntegerProperty : 2 }`, i.e. a string comparand silently produces a matching numeric filter.  I don't think we should allow this and should probably throw in this cases. Having .`Equals("2")` against an int field is much more likely a bug than intent, and silently matching hides it.

adelinowona (changes requested):

adelinowona:
Also I think the PR might need a rebase on main as the evergreen tests are failing because the SharpCompress warning suppression is missing.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/EqualsMethodToFilterTranslatorTests.cs
papafe: I agree that this is not the best behaviour, it would be unexpected. 
But throwing an exception here is difficult because when a filter translator throws `ExpressionNotSupportedException`, the outer `ExpressionToFilterTranslator` catches it and retries the predicate as an aggregation expression ($expr). For `field.Equals("2")` that retry succeeds and it produces `{ $expr: { $eq: ["$field", "2"] } }` so no exception reaches the user and changing this would require having a much broader change. Overall I think that the current approach of just not converting it to numerical value and using the string would be "expected". What do you think?
This also required to check if the value type and the field type are both numeric.

adelinowona: A user-facing throw would be nice but it's not very practical in this case as you mentioned. At least making sure that `e.NullableIntegerProperty.Equals("2")` doesn't produce a matching numeric filter is the behavior I was after and your changes address that so I'm ok with your current approach.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/EqualsMethodToFilterTranslator.cs
```diff
@@ -72,7 +73,25 @@ private static AstFilter Translate(TranslationContext context, Expression expres
...
+
+            var fieldType = fieldTranslation.Serializer.ValueType;
+            var valueType = value?.GetType();
+
+            IBsonSerializer valueSerializer;
+            if (value == null || fieldType.IsAssignableFrom(valueType))
+            {
+                valueSerializer = fieldTranslation.Serializer;
+            }
+            else if (fieldType.IsNumericOrNullableNumeric() && valueType.IsNumericOrNullableNumeric())
```
adelinowona: Nit: `valueType.IsNumericOrNullableNumeric()` could just be `valueType.IsNumeric()`. `valueType` comes from `value.GetType()` on a boxed constant, and a boxed `Nullable<T>` always reports its underlying type (or is plain null) — so valueType is never a `Nullable<T>` and the nullable check can't fire. Purely cosmetic though no behavior change.

adelinowona (approved):
LGTM. Kyra mentioned that there are other places where we might want to have similar changes. Do we have tickets for those?

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToFilterTranslators/MethodTranslators/EqualsMethodToFilterTranslator.cs
papafe: Reasonable! Fixed.

papafe:
> LGTM. Kyra mentioned that there are other places where we might want to have similar changes. Do we have tickets for those?

Created this ticket to track it: https://jira.mongodb.org/browse/CSHARP-6067

PR #1985, opened 2026-05-04
CSHARP-6011: Remove Session from IBinding by moving it to OperationContext

sanych-sun:
This PR is a pre-requisite for the operations refactoring. It moves Session from IBinding to OperationContext. (Bindings most likely will be removed in scope of operations refactoring)


src/MongoDB.Driver/Core/Operations/AggregateOperation.cs
```diff
@@ -392,25 +392,23 @@ private ReadCommandOperation<AggregateResult> CreateOperation(OperationContext o
...
         }
 
-        private AsyncCursor<TResult> CreateCursorFromCursorResult(IChannelSourceHandle channelSource, IChannelHandle channel, AggregateResult result)
+        private AsyncCursor<TResult> CreateCursorFromCursorResult(OperationContext operationContext, IChannelSourceHandle channelSource, IChannelHandle channel, AggregateResult result)
         {
             var cursorId = result.CursorId.GetValueOrDefault(0);
             var getMoreChannelSource = ChannelPinningHelper.CreateGetMoreChannelSource(channelSource, channel, cursorId);
             return new AsyncCursor<TResult>(
                 getMoreChannelSource,
+                operationContext.Session.Fork(),
```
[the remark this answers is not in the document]
sanych-sun: Updated the AsyncCursor class not to make any Forks on it's own.

src/MongoDB.Driver/Core/Operations/AggregateOperation.cs
```diff
@@ -392,25 +392,23 @@ private ReadCommandOperation<AggregateResult> CreateOperation(OperationContext o
...
-                return CreateCursorFromCursorResult(channelSource, channel, result);
-            }
-            else
-            {
-                // don't need connection pinning
-                return CreateCursorFromInlineResult(result);
+                return CreateCursorFromCursorResult(operationContext, channelSource, channel, result);
             }
+
+            return new SingleBatchAsyncCursor<TResult>(result.Results);
```
sanych-sun, line 402: Instead of creating `AsyncCursor` without `channelSource` we can simplify the execution by creating `SingleBatchAsyncCursor` which exists for this very reason - to create a cursor from inlined response. I can separate this to another PR if needed.

src/MongoDB.Driver/Core/Operations/FindOperation.cs
```diff
@@ -353,11 +353,12 @@ private AsyncCursor<TDocument> CreateCursor(IChannelSourceHandle channelSource,
...
             {
-                channelSource.Session.SetSnapshotTimeIfNeeded(atClusterTime.AsBsonTimestamp);
+                operationContext.Session.SetSnapshotTimeIfNeeded(atClusterTime.AsBsonTimestamp);
             }
 
             return new AsyncCursor<TDocument>(
                 getMoreChannelSource,
+                operationContext.Session.Fork(),
                 collectionNamespace,
                 _comment,
```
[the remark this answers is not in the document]
sanych-sun: AsyncCursor was updated not to create forks.

src/MongoDB.Driver/Core/Operations/ListCollectionsOperation.cs
```diff
@@ -179,13 +178,14 @@ private ReadCommandOperation<BsonDocument> CreateOperation()
...
-        private IAsyncCursor<BsonDocument> CreateCursor(IChannelSourceHandle channelSource, IChannelHandle channel, BsonDocument result)
+        private IAsyncCursor<BsonDocument> CreateCursor(OperationContext operationContext, IChannelSourceHandle channelSource, IChannelHandle channel, BsonDocument result)
         {
             var cursorDocument = result["cursor"].AsBsonDocument;
             var cursorId = cursorDocument["id"].ToInt64();
             var getMoreChannelSource = ChannelPinningHelper.CreateGetMoreChannelSource(channelSource, channel, cursorId);
             var cursor = new AsyncCursor<BsonDocument>(
                 getMoreChannelSource,
+                operationContext.Session.Fork(),
                 CollectionNamespace.FromFullName(cursorDocument["ns"].AsString),
```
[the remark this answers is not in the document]
sanych-sun: AsyncCursor does not use Fork anymore.

src/MongoDB.Driver/Core/Operations/ListIndexesUsingCommandOperation.cs
```diff
@@ -162,13 +162,14 @@ private ReadCommandOperation<BsonDocument> CreateOperation()
...
-        private IAsyncCursor<BsonDocument> CreateCursor(IChannelSourceHandle channelSource, IChannelHandle channel, BsonDocument result)
+        private IAsyncCursor<BsonDocument> CreateCursor(OperationContext operationContext, IChannelSourceHandle channelSource, IChannelHandle channel, BsonDocument result)
         {
             var cursorDocument = result["cursor"].AsBsonDocument;
             var cursorId = cursorDocument["id"].ToInt64();
             var getMoreChannelSource = ChannelPinningHelper.CreateGetMoreChannelSource(channelSource, channel, cursorId);
             var cursor = new AsyncCursor<BsonDocument>(
                 getMoreChannelSource,
+                operationContext.Session.Fork(),
                 CollectionNamespace.FromFullName(cursorDocument["ns"].AsString),
```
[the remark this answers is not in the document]
sanych-sun: AsyncCursor does not use Fork anymore.

src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -667,8 +667,8 @@ public void ThrowIfNotSupported(IMongoClient client, CancellationToken cancellat
         {
             var cluster = client.GetClusterInternal();
             // TODO: CSOT implement proper way to obtain the operationContext
-            var operationContext = new OperationContext(null, cancellationToken);
-            using (var binding = new ReadWriteBindingHandle(new WritableServerBinding(cluster, NoCoreSession.NewHandle())))
+            var operationContext = new OperationContext(NoCoreSession.NewHandle(), null, cancellationToken);
+            using (var binding = new ReadWriteBindingHandle(new WritableServerBinding(cluster)))
             using (var channelSource = binding.GetWriteChannelSource(operationContext))
```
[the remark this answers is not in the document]
sanych-sun: Fixed.

src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -686,8 +686,8 @@ public async Task ThrowIfNotSupportedAsync(IMongoClient client, CancellationToke
         {
             var cluster = client.GetClusterInternal();
             // TODO: CSOT implement proper way to obtain the operationContext
-            var operationContext = new OperationContext(null, cancellationToken);
-            using (var binding = new ReadWriteBindingHandle(new WritableServerBinding(cluster, NoCoreSession.NewHandle())))
+            var operationContext = new OperationContext(NoCoreSession.NewHandle(), null, cancellationToken);
+            using (var binding = new ReadWriteBindingHandle(new WritableServerBinding(cluster)))
             using (var channelSource = await binding.GetWriteChannelSourceAsync(operationContext).ConfigureAwait(false))
```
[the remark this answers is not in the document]
sanych-sun: Fixed.

tests/MongoDB.Driver.Tests/Specifications/client-backpressure/prose-tests/ClientBackpressureProseTests.cs
```diff
@@ -90,7 +90,7 @@ private async Task AssertBackoffBehavior<TContext>(
         Action<OperationContext, TContext> executeSync,
         Func<OperationContext, TContext, Task> executeAsync)
     {
-        var operationContext = new OperationContext(TimeSpan.FromSeconds(30), CancellationToken.None);
+        var operationContext = new OperationContext(NoCoreSession.NewHandle(), timeout: TimeSpan.FromSeconds(30));
 
         // Test with no backoff (jitter = 0)
```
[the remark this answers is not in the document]
sanych-sun: Fixed.

tests/MongoDB.Driver.Tests/OperationExecutorTests.cs
```diff
@@ -43,7 +43,7 @@ public void StartImplicitSession_should_call_cluster_StartSession()
         public async Task ExecuteReadOperation_throws_on_null_operation([Values(true, false)] bool async)
         {
             var subject = CreateSubject(out _);
-            var operationContext = new OperationContext(Timeout.InfiniteTimeSpan, CancellationToken.None);
+            var operationContext = new OperationContext(NoCoreSession.NewHandle(), Timeout.InfiniteTimeSpan, CancellationToken.None);
             var session = Mock.Of<IClientSessionHandle>();
```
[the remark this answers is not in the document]
sanych-sun: Fixed

src/MongoDB.Driver/OperationContext.cs
```diff
@@ -211,13 +197,13 @@ public OperationContext WithTimeout(TimeSpan timeout)
...
 
-            return new OperationContext(Clock, timeout, CancellationToken)
+            return new OperationContext(Session, Clock, timeout, CancellationToken)
             {
                 RootContext = RootContext,
                 OperationName = OperationName,
                 DatabaseName = DatabaseName,
                 CollectionName = CollectionName,
-                IsTracingEnabled = IsTracingEnabled
+                IsTracingEnabled = IsTracingEnabled,
```
papafe: nit: maybe no comma

src/MongoDB.Driver/MongoDB.Driver.csproj
```diff
@@ -17,6 +17,7 @@
     <Compile Include="..\MongoDB.Shared\DictionaryComparer.cs" Link="Shared\DictionaryComparer.cs" />
     <Compile Include="..\MongoDB.Shared\SequenceComparer.cs" Link="Shared\SequenceComparer.cs" />
     <Compile Include="..\MongoDB.Shared\Hasher.cs" Link="Shared\Hasher.cs" />
+    <Compile Remove="Core\Operations\DelayedEvaluationWriteConcernSerializer.cs" />
```
papafe: Do we need this?

src/MongoDB.Driver/Core/Bindings/EndTransactionReadWriteBinding.cs
```diff
@@ -31,72 +31,110 @@ internal sealed class EndTransactionReadWriteBinding : IReadWriteBinding
...
+        }
 
         public Task<IChannelSourceHandle> GetWriteChannelSourceAsync(OperationContext operationContext, IReadOnlyCollection<ServerDescription> deprioritizedServers, IMayUseSecondaryCriteria mayUseSecondary)
-            => _innerBinding.GetWriteChannelSourceAsync(operationContext, deprioritizedServers, mayUseSecondary);
+        {
+            EnsureInnerBinding(operationContext);
+            return _innerBinding.GetWriteChannelSourceAsync(operationContext, deprioritizedServers, mayUseSecondary);
+        }
+
+        private void EnsureInnerBinding(OperationContext operationContext)
```
papafe, line 131: Should we worry about thread safety for this?

tests/MongoDB.Driver.Tests/Authentication/AuthenticationHelperTests.cs
```diff
@@ -66,15 +67,16 @@ public async Task Authenticate_should_invoke_authenticators_when_they_exist(
             mockConnection.SetupGet(c => c.Description).Returns(description);
             mockConnection.SetupGet(c => c.Settings).Returns(settings);
 
+            using var operationContext = new OperationContext(NoCoreSession.NewHandle());
```
papafe, line 70: nit. Given we use this in quite a lot of places, would it make sense to move it to a static factory...?

papafe (commented):
Overall it looks good! Just left some comments.

src/MongoDB.Driver/OperationContext.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoDB.Driver.csproj
sanych-sun: Nope. Removed.

src/MongoDB.Driver/Core/Bindings/EndTransactionReadWriteBinding.cs
sanych-sun: I do not think we have any code path that uses Bindings in a concurrent way. `RebuildInnerBinding` (the previous place where it was assigned before) was also non-thread safe.

tests/MongoDB.Driver.TestHelpers/Core/MockClusterableServerFactory.cs
```diff
@@ -186,7 +186,8 @@ public void PublishDescription(ServerDescription description)
                     var maxWireVersion = description.MaxWireVersion;
                     var server = (Server)result.Server;
                     var helloResult = new HelloResult(new BsonDocument { { "compressors", new BsonArray() }, { "maxWireVersion", maxWireVersion } });
-                    var mockConnection = Mock.Get(server._connectionPool().AcquireConnection(OperationContext.NoTimeout));
+
+                    var mockConnection = Mock.Get(server._connectionPool().AcquireConnection(It.IsAny<OperationContext>()));
```
adelinowona: `It.IsAny<OperationContext>()` is being used here as a concrete method argument, not inside a Moq `Setup`/`Verify` call. Outside that context, `It.IsAny<T>()` returns `default(T)`, so this passes `null` to `AcquireConnection`. It works by accident because the mock was set up with `It.IsAny<OperationContext>()` which also matches `null` — but it's the wrong tool here. This was presumably meant to replace the deleted `OperationContext.NoTimeout` static. Should be something like `new OperationContext(NoCoreSession.NewHandle())`.

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
```diff
@@ -467,32 +456,28 @@ private bool IsMongoCursorNotFoundException(MongoCommandException exception)
...
-            // TODO: CSOT implement proper way to obtain the operationContext
-            var operationContext = new OperationContext(null, cancellationToken);
             using (EventContext.BeginOperation(_operationId))
             using (EventContext.BeginKillCursors(_collectionNamespace))
             using (var channel = _channelSource.GetChannel(operationContext.WithTimeout(TimeSpan.FromSeconds(10))))
             {
                 if (!channel.Connection.IsExpired)
                 {
-                    ExecuteKillCursorsCommand(channel, cancellationToken);
+                    ExecuteKillCursorsCommand(operationContext, channel);
```
adelinowona, line 463: The 10-second timeout applied via `operationContext.WithTimeout(TimeSpan.FromSeconds(10))` scopes the channel *acquisition* only — `ExecuteKillCursorsCommand` here receives the original uncapped `operationContext`. The kill command itself runs without any timeout guard. The timeout-scoped context should be captured in a local variable and passed through to `ExecuteKillCursorsCommand` as well (same issue on the async path at line 480).

src/MongoDB.Driver/Core/Bindings/EndTransactionReadWriteBinding.cs
adelinowona, line 131: `EnsureInnerBinding` doesn't call `ThrowIfDisposed()` before doing its work. All 12 public `Get*ChannelSource*` methods delegate here without checking disposed state first, so after `Dispose()` any of those calls silently rebuilds `_innerBinding`. Every other binding class guards its public methods — and `RebuildInnerBinding` (line 43) does correctly call `ThrowIfDisposed` — which makes the gap here inconsistent. `ThrowIfDisposed()` should be the first line of this method.

tests/MongoDB.Driver.TestHelpers/Core/CoreTestConfiguration.cs
```diff
@@ -384,47 +387,37 @@ private static bool AreSessionsSupported(ClusterDescription clusterDescription)
...
 
-        private static IReadBindingHandle CreateReadBinding(IClusterInternal cluster, ReadPreference readPreference, ICoreSessionHandle session)
+        private static IReadBindingHandle CreateReadBinding(IClusterInternal cluster, ReadPreference readPreference)
         {
-            var binding = new ReadPreferenceBinding(cluster, readPreference, session.Fork());
+            var binding = new ReadPreferenceBinding(cluster, readPreference);
             return new ReadBindingHandle(binding);
         }
 
         private static IReadWriteBindingHandle CreateReadWriteBinding(ICoreSessionHandle session)
```
adelinowona, line 406: The `session` parameter is no longer used in the body — it was needed when bindings took a session argument, but `WritableServerBinding` no longer does. Dead parameter, should be removed.

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
```diff
@@ -262,15 +267,15 @@ private IAsyncCursor<RawBsonDocument> Resume(CancellationToken cancellationToken
         {
             ReconfigureOperationResumeValues();
             // TODO: CSOT implement proper way to obtain the operationContext
```
adelinowona, line 269: The session wiring is now correct — this was fixed as part of this PR. The only remaining open item is threading a CSOT timeout through, not the entire `operationContext` construction. Worth narrowing the comment to something like `// TODO CSHARP-XXXX: pass CSOT timeout to Resume` so it's clear what's actually still missing.

adelinowona (commented):
**`CoreSession.cs` — `ExecuteEndTransactionOnPrimary[Async]` lines 562–563 and 575–576**

By the time `ExecuteEndTransactionOnPrimary` is called, `operationContext.Session` is already a `NonDisposingCoreSessionHandle` wrapping `this` — the callers (`CommitTransaction`, `AbortTransaction`, and their async variants) create one four lines before calling this method. The second `new NonDisposingCoreSessionHandle(this)` here is a redundant wrapper around the same object and can be eliminated by passing `operationContext.Session` directly to `CreateEndTransactionBinding`.

src/MongoDB.Driver/GridFS/GridFSBucket.cs
```diff
@@ -84,7 +84,7 @@ public ImmutableGridFSBucketOptions Options
         {
             Ensure.IsNotNull((object)id, nameof(id));
             // TODO: CSOT implement proper way to obtain the operationContext
-            var operationContext = new OperationContext(null, cancellationToken);
+            using var operationContext = new OperationContext(NoCoreSession.NewHandle(), null, cancellationToken);
```
adelinowona: `NoCoreSession.NewHandle()` is created inline and never disposed. `OperationContext.Dispose()` intentionally does not dispose the session, so the `CoreSessionHandle` (and its `ReferenceCountedCoreSession`) returned by `NewHandle()` is silently abandoned on every GridFS call.

Copilot already flagged this in `Feature.cs` and it was fixed there. The same fix is needed for all occurrences in this file:

```csharp
using var session = NoCoreSession.NewHandle();
using var operationContext = new OperationContext(session, null, cancellationToken);
```

src/MongoDB.Driver/Core/Servers/RoundTripTimeMonitor.cs
```diff
@@ -125,7 +126,7 @@ private void MonitorServer()
             _logger?.LogDebug(_serverId, "Monitoring started");
 
             var helloOk = false;
-            using var operationContext = new OperationContext(null, _cancellationToken);
+            using var operationContext = new OperationContext(NoCoreSession.NewHandle(), null, _cancellationToken);
```
adelinowona: Same inline `NoCoreSession.NewHandle()` leak as flagged in `Feature.cs` and `GridFSBucket.cs`. `OperationContext.Dispose()` does not dispose the session, so the handle is abandoned on every RTT measurement iteration. Fix:

```csharp
using var session = NoCoreSession.NewHandle();
using var operationContext = new OperationContext(session, null, _cancellationToken);
```

src/MongoDB.Driver/Core/Servers/ServerMonitor.cs
```diff
@@ -321,7 +322,7 @@ private HelloResult GetHelloResult(
 
         private void Heartbeat(CancellationToken cancellationToken)
         {
-            using var operationContext = new OperationContext(null, cancellationToken);
+            using var operationContext = new OperationContext(NoCoreSession.NewHandle(), null, cancellationToken);
```
adelinowona: Same inline `NoCoreSession.NewHandle()` leak as flagged in `Feature.cs` and `GridFSBucket.cs`. `OperationContext.Dispose()` does not dispose the session, so the `CoreSessionHandle` (and its `ReferenceCountedCoreSession`) is abandoned on every heartbeat. Fix:
```csharp
using var session = NoCoreSession.NewHandle();
using var operationContext = new OperationContext(session, null, cancellationToken);
```

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
```diff
@@ -508,7 +493,8 @@ public bool MoveNext(CancellationToken cancellationToken)
                 return hasMore;
             }
 
-            var batch = GetNextBatch(cancellationToken);
+            using var operationContext = new OperationContext(_session, cancellationToken: cancellationToken);
```
adelinowona, line 495: The old `GetNextBatch` had a `// TODO: CSOT implement proper way to obtain the operationContext` comment documenting that cursor get-more operations have no CSOT timeout. This refactor moved `OperationContext` creation here but did not carry the TODO forward, so the known limitation is now undocumented. `ChangeStreamCursor.Resume` still keeps the equivalent TODO, creating inconsistency. Same issue applies to line 513 (`MoveNextAsync`).

src/MongoDB.Driver/Core/Bindings/EndTransactionReadWriteBinding.cs
sanych-sun: Nice catch, fixed. Thank you!

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Core/Servers/ServerMonitor.cs
sanych-sun: Fixed for all non-tests code.

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
sanych-sun: You are right, I was thinking this was a bug introduced by CSOT changes, but I've double-checked the history - this behavior was even before CSOT. It looks suspicious, I think we need to investigate this.

tests/MongoDB.Driver.TestHelpers/Core/MockClusterableServerFactory.cs
sanych-sun: Fixed.

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
BorisDog: @sanych-sun 
So `operationContext` is the right context to be used with `ExecuteKillCursorsCommand;` ? So why do we need `killCursorGetChannelOperationContext ` as well?

BorisDog (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
sanych-sun: Because there is explicit hardcoded timeout of 10 seconds for the channel resolving.
Do not really know. Here is the version of the code before introducing of the OperationContext:
```
using (EventContext.BeginOperation(_operationId))
            using (EventContext.BeginKillCursors(_collectionNamespace))
            using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(10)))
            using (var channel = await _channelSource.GetChannelAsync(cancellationTokenSource.Token).ConfigureAwait(false))
            {
                if (!channel.Connection.IsExpired)
                {
                    await ExecuteKillCursorsCommandAsync(channel, cancellationToken).ConfigureAwait(false);
                }
            }
```

As you can see we explicitly creates a new cancellationToken from timeout for channel resolving, but command execution itself is driven by provided cancellationToken. Not sure why we had that code, it looks suspicious for me. Probably worth to investigate (in separate ticket).
Should I create a ticket for that?

BorisDog: Yes please create a ticket.

papafe (approved):
LGTM!
The idea is simple, lots of changes though 😁

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
sanych-sun: Created: https://jira.mongodb.org/browse/CSHARP-6123

BorisDog (approved):
LGTM!

PR #1987, opened 2026-05-05
CSHARP-6024: Support ReadOnlySet Serialization

papafe:


src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
```diff
@@ -153,34 +153,52 @@ protected override object CreateAccumulator()
...
-                var value = (TValue)noArgumentConstructorInfo.Invoke(new Type[] { });
+                var value = (TValue)noArgumentConstructorInfo.Invoke(null);
                 foreach (var item in (IEnumerable<TItem>)accumulator)
                 {
                     addMethodInfo.Invoke(value, new object[] { item });
                 }
                 return value;
             }
 
+            // last resort: try a constructor that takes a HashSet<TItem> (e.g. ReadOnlySet<T>(ISet<T>))
```
papafe: I've put this as the last possibility in the method so that we do not cause issues with preexisting code.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
```diff
@@ -153,34 +153,52 @@ protected override object CreateAccumulator()
...
+            // Note: collapses duplicate elements, which is correct for set-shaped targets.
+            var setCtor = FindSingleArgumentConstructor(typeof(HashSet<TItem>));
```
sanych-sun, line 178: Should it be `ISet<TItem>`?

tests/MongoDB.Bson.Tests/Serialization/Serializers/EnumerableInterfaceImplementerSerializerTests.cs
```diff
@@ -226,5 +229,86 @@ private IBsonSerializer<C> CreateSubject()
             serializerRegistry.RegisterSerializer(typeof(C), subject);
             return subject;
         }
+
+        // exercises the HashSet<TItem> ctor fallback in FinalizeResult: target type with
+        // only an ISet<TItem> constructor — same shape as ReadOnlySet<T>
+        public class SetOnlyCtorEnumerable<T> : IEnumerable<T>
```
sanych-sun: Do we really need this mock as we have `net10.0` target for tests and can exercise with the `ReadOnlySet<>` itself?

sanych-sun (changes requested):

tests/MongoDB.Bson.Tests/Serialization/Serializers/EnumerableInterfaceImplementerSerializerTests.cs
papafe: Agreed, removed this one.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
papafe: Agreed, fixed.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
```diff
@@ -153,34 +153,52 @@ protected override object CreateAccumulator()
         protected override TValue FinalizeResult(object accumulator)
         {
             // find and call a constructor that we can pass the accumulator to
-            var accumulatorType = accumulator.GetType();
-            foreach (var constructorInfo in typeof(TValue).GetTypeInfo().GetConstructors())
+            var accumulatorCtor = FindSingleArgumentConstructor(accumulator.GetType());
```
sanych-sun, line 156: Shouldn't we have HashSet<T> here as an accumulator instead? What serializer we selects to handle the ReadOnlySet? I assume it's not `HashSetSerializer`. Should we either improve HashSetSerializer to support ReadOnlySet too or create a new serializer for ReadOnlySet?

sanych-sun (changes requested):

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
papafe: So, on serialization we use the same serializer (`EnumerableInterfaceImplementerSerializer`), but that works because `ReadOnlySet` implements `IEnumerable` and we just loop through the values. 
We could add a specific serializer as we do for other types, but that would mean to add a target for .NET 9.0 (`ReadOnlySet` was added then). As an alternative, we could add a serializer specialized for types implementing `ISet` and use that, but it would be almost a copy of this one, just specialised. 
What do you think?

adelinowona: I don't think a separate serializer is warranted. Serialization already works correctly (iterate as IEnumerable<T>), and the deserialization fallback covers the construction side. A dedicated ISet<T> serializer would be essentially identical with no behavioral difference — BSON arrays carry no set semantics.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
adelinowona: I think the change from `HashSet<TItem> `to `ISet<TItem>` went in the wrong direction.

Is the intent of this fallback not to "find any single-argument constructor that a HashSet<TItem> can satisfy"? 

In which case the right probe is therefore `typeof(HashSet<TItem>)`, not `typeof(ISet<TItem>)`. The difference matters because `HashSet<T>` implements `IReadOnlySet<T> `(added in .NET 5), but `ISet<T> `does not. A type with a constructor taking `IReadOnlySet<T>` would work with `HashSet<TItem>` as the probe but silently fall through to the exception with `ISet<TItem>`. No such type exists in the BCL today, but the `HashSet<TItem>` probe is strictly more correct as the intent.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
papafe: Yep, I think that the current approach is a good compromise.

adelinowona (commented):
`FinalizeResult` currently seems to do reflection on every deserialization call to figure out how to construct `TValue`, but the answer is fixed for any given `TValue/TItem` pair. A static readonly field on the generic class could compute this once per concrete type thus saving us a reflection cost. @sanych-sun WDYT?

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
adelinowona: @papafe @sanych-sun any thoughts on my comment above?

papafe: Yep, I think we should try to find a constructor for the actual type we're using.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
sanych-sun: It already exists: `HashSetSerializer<T>`, we just need to improve it to support ReadOnlySet too or create similar for ReadOnlySet. More over we have specialized serializer for `ImmutableHashSet`, `ImmutableSortedSetSerializer`. And I think if we have a specialized serializer to work with HashSet - it's a little confusing why ReadOnlySet is not handled in a similar way. I would expect us either to process both via specialized serializer(s) or handle both with "generic one".

sanych-sun:
About `FinalizeResult`: the logic there is based on method's parameter real type, not on class generic type parameter. I'm not sure if we can assume that all instances of this class always get accumulator of this exact type. Worth extra investigation.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
sanych-sun: I think we should decide on https://github.com/mongodb/mongo-csharp-driver/pull/1987#discussion_r3321089649 comment first, because it might change the way we approach the problem.

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
papafe: So, the thing is that we have a `HashSetSerializer`, but it's not in `MongoDB.Bson`, but only in `MongoDB.Driver` and it's used only for LINQ pipeline translations, so it's not used for serialization/deserialization of POCOs. 
Right now, `HashSet` gets serialized/deserialized through `EnumerableInterfaceImplementerSerializer`, there is no specific serializer for that. So overall `ReadOnlySet` is treated in the same way that `HashSet` does. 
Also if we wanted to add a serializer for `ReadOnlySet` we'd need to add a net9.0 target.

Regarding `Immutable` types, they need a specific serializer because they don't have a public constructor or `Add`, but only builders.

adelinowona:
@sanych-sun Regarding `FinalizeResult`: I did some investigation aided by claude. Traced the call chain: `FinalizeResult` is `protected abstract `and the only caller is `EnumerableSerializerBase.Deserialize`, which always passes it the result of `CreateAccumulator()` directly. Since `EnumerableInterfaceImplementerSerializer<TValue, TItem>` is sealed and doesn't override `Deserialize`, the accumulator reaching `FinalizeResult` is always new `List<TItem>()` — so the static field assumption seems safe.

sanych-sun:
@adelinowona In our codebase - yes, but we are talking about changes in public base class, there might be another implementations out of our control.

adelinowona:
@sanych-sun yep, my proposed change is entirely contained within `EnumerableInterfaceImplementerSerializer<TValue, TItem>` — the sealed class. The static field will be a private member there, and FinalizeResult is already overridden there. Nothing in the public base classes changes. External subclasses of the base would be completely unaffected.

papafe:
@sanych-sun @adelinowona Given that it's a little bit controversial, should capture the possible improvement with the static readonly field in a new ticket? As far as I understand this PR is not making the situation worse performance wise, so we can keep the improvement for later and merge this for now.

adelinowona:
@papafe the possible optimization should definitely be a new ticket if we agree it's worth it. I'll LGTM the current PR.

adelinowona (approved):

src/MongoDB.Bson/Serialization/Serializers/EnumerableInterfaceImplementerSerializer.cs
sanych-sun: Agree with using of concrete type.

sanych-sun (approved):
LGTM

PR #1988, opened 2026-05-06
CSHARP-5930: Fix for tests that run against all KMS providers

ajcvickers:
Added __allKmsProviderNames constant to tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/EncryptionTestHelper.cs (the semicolon-delimited list of all standard providers), and changed GetKmsProviders to fall back to it when kmsProviderNames is null — exactly restoring what GetKmsProviders(null) used to do.

Tested with targeted EG patch.


sanych-sun (approved):
LGTM

PR #1990, opened 2026-05-07
CSHARP-5846: Global GuidSerializer settings ignored for object-typed properties

papafe:


tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -725,6 +727,67 @@ public void Serialize_guid_should_throw_when_guid_representation_is_unspecified(
...
+            using var reader = new BsonBinaryReader(memoryStream);
+            var context = BsonDeserializationContext.CreateRoot(reader);
+            reader.ReadStartDocument();
+            reader.ReadName("x");
+            var result = subject.Deserialize<object>(context);
+
+            result.Should().Be(guid);
+        }
+
+        private static ObjectSerializer CreateObjectSerializerWithRegisteredGuidSerializer(
```
papafe: I'm curious about what the team thinks about this. Unfortunately without having a local serialization registry, we can't test this properly otherwise.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -725,6 +727,80 @@ public void Serialize_guid_should_throw_when_guid_representation_is_unspecified(
...
+                result = subject.Deserialize<object>(context);
+            });
+
+            result.Should().Be(guid);
+        }
+
+        private static void WithRegisteredGuidSerializer(GuidRepresentation registeredGuidRepresentation, Action action)
+        {
+            var cacheField = typeof(BsonSerializerRegistry).GetField("_cache", BindingFlags.NonPublic | BindingFlags.Instance);
+            var cache = (ConcurrentDictionary<Type, IBsonSerializer>)cacheField.GetValue(BsonSerializer.SerializerRegistry);
```
BorisDog, line 830: Let's add a task to update this to use SerializationDomain once it's merged. Comment to existing ticket also works.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
```diff
@@ -725,6 +727,80 @@ public void Serialize_guid_should_throw_when_guid_representation_is_unspecified(
                 exception.Should().BeOfType<BsonSerializationException>();
             }
         }
```
BorisDog, line 729: Let's test that this doesn't introduce collisions with GuidSerializer registrations (the lazy functionality):
1. ObjectSerialzer is registered and resolved
2. GuidSerializer X is registered
3. Test that X is used by ObjectSerialzer  despite it being resolved before X registration

BorisDog (changes requested):
I guess [CSHARP-5498](https://jira.mongodb.org/browse/CSHARP-5498) needs to be updated (closed?) as well.

src/MongoDB.Bson/Serialization/Serializers/ObjectSerializer.cs
```diff
@@ -135,10 +135,17 @@ public ObjectSerializer(
...
         {
             _discriminatorConvention = discriminatorConvention ?? throw new ArgumentNullException(nameof(discriminatorConvention));
-            _guidRepresentation = guidRepresentation;
-            _guidSerializer = new GuidSerializer(_guidRepresentation);
             _allowedDeserializationTypes = allowedDeserializationTypes ?? throw new ArgumentNullException(nameof(allowedDeserializationTypes));
             _allowedSerializationTypes = allowedSerializationTypes ?? throw new ArgumentNullException(nameof(allowedSerializationTypes));
+            _guidRepresentation = guidRepresentation;
+            if (guidRepresentation != GuidRepresentation.Unspecified)
+            {
+                _guidSerializer = new Lazy<IBsonSerializer<Guid>>(() => new GuidSerializer(guidRepresentation));
```
adelinowona, line 143: Looks like the non-Unspecified path here is paying a price of using Lazy. Is this regressing this path and are there no optimizations for this?

papafe: Yes, we're paying the price of lazy for both paths now. We can definitely have a specific behaviour for the non-Unspecified path, but given that this not on the hot path (uncommon serializer / the guid serializer is even a less common case), I would prefer to keep a simpler code. What do you think?

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
papafe: Added tests for this.

tests/MongoDB.Bson.Tests/Serialization/Serializers/ObjectSerializerTests.cs
papafe: Added a comment in the ticket.

BorisDog (approved):
LGTM

PR #1996, opened 2026-05-11
update SharpCompress to fix vulnerability issue

WeihanLi:
see https://github.com/advisories/GHSA-6c8g-7p36-r338

### DIFF
--- a/src/MongoDB.Driver/MongoDB.Driver.csproj
+++ b/src/MongoDB.Driver/MongoDB.Driver.csproj
@@ -22,7 +22,7 @@
   <ItemGroup>
     <PackageReference Include="DnsClient" Version="1.6.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
-    <PackageReference Include="SharpCompress" Version="0.30.1" />
+    <PackageReference Include="SharpCompress" Version="0.48.0" />
     <PackageReference Include="Snappier" Version="1.3.1" />
     <PackageReference Include="ZstdSharp.Port" Version="0.7.3" />
     <PackageReference Include="System.Buffers" Version="4.6.1" />


tebeco:
@ajcvickers  / @sanych-sun 
I feel like the handling of transitive needs extra attention

i don't mean just this one
this specific dependency is a case study for supply chain attack
(old dependency that gets a 5years update at once)


This update would indeed remove the CVE but there's a few catch:
* the previous package is pre-dotnet core (2021)
* the package was 5 years old
* this would jump a 0.xx.yy to 0.aa.bb) (no real 1.0 ?)
* Will someone open all the release notes from 0.30.1 to 0.48.0 and asses no breaking changes ?\
* Do all the E2E test fully cover everything SharpCompress is responsible for ?
* Shouldn't there be a target framework to bet net 10.0 for `MongoDB.Driver` ?
* Why does `MongoDB.Driver` still target net6.0 ?
* Shouldn't all transitive be updated and inspected to avoid stacking up CVE all around in the next weeks ?

damieng:
The upstream vulnerability in SharpCompress specifically relates to the `WriteToDirectory` function which we do not use.

We should still upgrade and consider all the points raised here but given we're not actually affected by this particular issue we can consider what to do a bit more thoughtfully without rushing.

tebeco:
thx for confirming @damieng 
We've added this in our `Directory.Build.targets`:
```xml
  <ItemGroup>
    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-6c8g-7p36-r338" />
  </ItemGroup>
```

as you confirmed the CVE code isn't used by mongo this prevent possible regression until the Mongo Team asses entirely every ~18 version gap + fully test end to end the change.

sanych-sun:
Abandon this PR as a duplicate with https://github.com/mongodb/mongo-csharp-driver/pull/2006

PR #1997, opened 2026-05-11
CSHARP-6037: Replace SharpCompress with DeflateStream for zlib compression

damieng:
## Summary

Removes the `SharpCompress 0.30.1` dependency from `MongoDB.Driver`. This package has a known moderate severity vulnerability (GHSA-6c8g-7p36-r338) and the fixed version drops .NET targets we support. 

Zlib compression support is replaced with a thin `ZlibStream` wrapper around `System.IO.Compression.DeflateStream`, implementing the same RFC 1950 framing (2-byte header + deflate payload + Adler-32 checksum) that SharpCompress provided.

Header validation and Adler-32 checksum verification on decompress are preserved.

## Compression level mapping

`System.IO.Compression` does not expose zlib's full 0–9 integer level range. The mapping is:

| User-configured level | Behaviour |
|---|---|
| 0 (no compression) | Exact |
| -1 / 6 (default / optimal) | Exact |
| 1 (fastest) | Exact |
| 2–3 | Mapped to level 1 (marginally less compression than requested) |
| 4–5, 7–8 | Mapped to level 6 (within one or two levels of requested) |
| 9 (maximum compression) | Mapped to level 6 — users who explicitly set level 9 to maximise compression ratio will see a reduction in compression effectiveness |

Levels 2–8 are rarely configured explicitly in practice. Level 9 is the most notable change. All output remains valid RFC 1950 zlib and interoperates correctly with the server.

## Approach

This is preferrable to updating SharpCompress as we use a tiny fraction of its functionality but are flagged when a vulnerability appears even in functionality we don't use such as this latest one in archive reading. The fix is in a version that has dropped support for .NET targets we still support in this 3.x release.

Given how trivial the actual functionality we require is it makes no sense to continue with this dependency for the sake of adding a 2-byte header and Adler32 checksum over a DeflateStream.

Superceeds #1996

### DIFF
--- a/src/MongoDB.Driver/Core/Compression/ZlibStream.cs
+++ b/src/MongoDB.Driver/Core/Compression/ZlibStream.cs
@@ -1,4 +1,4 @@
-/* Copyright 2019-present MongoDB Inc.
+﻿/* Copyright 2019-present MongoDB Inc.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
--- a/src/MongoDB.Driver/Core/Compression/ZlibStream.cs
+++ b/src/MongoDB.Driver/Core/Compression/ZlibStream.cs
@@ -20,8 +20,13 @@ using System.IO.Compression;
 namespace MongoDB.Driver.Core.Compression
 {
     // Replacement for SharpCompress.Compressors.Deflate.ZlibStream.
-    // Implements RFC 1950 (zlib framing) using System.IO.Compression.DeflateStream for the
-    // inner RFC 1951 (deflate) payload, with manual Adler-32 checksum handling.
+    // Implements RFC 1950 (zlib framing) around System.IO.Compression.DeflateStream (RFC 1951),
+    // with manual Adler-32 checksum handling.
+    //
+    // Both compress and decompress initialise lazily — no I/O happens in the constructor.
+    // Decompress path: the compressed input is buffered up-front (small — wire messages are
+    // length-bounded), but the decompressed output is produced lazily as the caller reads,
+    // with Adler-32 accumulated incrementally and validated when DeflateStream signals EOF.
     internal sealed class ZlibStream : Stream
     {
         // RFC 1950 headers: CMF=0x78 (deflate, 32K window) + FLG satisfying (CMF*256+FLG)%31==0.
@@ -31,14 +36,20 @@ namespace MongoDB.Driver.Core.Compression
 
         private readonly Stream _stream;
         private readonly CompressionMode _mode;
+        private readonly CompressionLevel _level;
         private readonly bool _leaveOpen;
 
-        // Compress-mode state
-        private DeflateStream _deflateStream;
-        private uint _adler32 = 1u;
+        // Compress-mode state (initialised lazily on first Write)
+        private DeflateStream _compressDeflate;
+        private uint _compressAdler32 = 1u;
+        private bool _compressInitialized;
 
-        // Decompress-mode state
-        private MemoryStream _decompressed;
+        // Decompress-mode state (initialised lazily on first Read)
+        private DeflateStream _decompressDeflate;
+        private byte[] _trailer;
+        private uint _decompressAdler32 = 1u;
+        private bool _decompressInitialized;
+        private bool _trailerValidated;
 
         private bool _disposed;
 
@@ -46,18 +57,8 @@ namespace MongoDB.Driver.Core.Compression
         {
             _stream = stream;
             _mode = mode;
+            _level = level;
             _leaveOpen = leaveOpen;
-
-            if (mode == CompressionMode.Compress)
-            {
-                var header = level == CompressionLevel.NoCompression ? s_noCompressionHeader : s_defaultHeader;
-                stream.Write(header, 0, 2);
-                _deflateStream = new DeflateStream(stream, level, leaveOpen: true);
-            }
-            else
-            {
-                _decompressed = ReadAndDecompress(stream);
-            }
         }
 
         public override bool CanRead => _mode == CompressionMode.Decompress && !_disposed;
@@ -65,16 +66,35 @@ namespace MongoDB.Driver.Core.Compression
         public override bool CanSeek => false;
         public override long Length => throw new NotSupportedException();
         public override long Position { get => throw new NotSupportedException(); set => throw new NotSupportedException(); }
-        public override void Flush() => _deflateStream?.Flush();
+        public override void Flush() => _compressDeflate?.Flush();
         public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
         public override void SetLength(long value) => throw new NotSupportedException();
 
-        public override int Read(byte[] buffer, int offset, int count) => _decompressed.Read(buffer, offset, count);
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            EnsureDecompressInitialized();
+            var n = _decompressDeflate.Read(buffer, offset, count);
+            if (n > 0)
+            {
+                _decompressAdler32 = UpdateAdler32(_decompressAdler32, buffer, offset, n);
+            }
+            else if (!_trailerValidated)
+            {
+                _trailerValidated = true;
+                var expected = (uint)(_trailer[0] << 24 | _trailer[1] << 16 | _trailer[2] << 8 | _trailer[3]);
+                if (_decompressAdler32 != expected)
+                {
+                    throw new InvalidDataException("Zlib Adler-32 checksum mismatch.");
+                }
+            }
+            return n;
+        }
 
         public override void Write(byte[] buffer, int offset, int count)
         {
-            _adler32 = UpdateAdler32(_adler32, buffer, offset, count);
-            _deflateStream.Write(buffer, offset, count);
+            EnsureCompressInitialized();
+            _compressAdler32 = UpdateAdler32(_compressAdler32, buffer, offset, count);
+            _compressDeflate.Write(buffer, offset, count);
         }
 
         protected override void Dispose(bool disposing)
@@ -84,11 +104,17 @@ namespace MongoDB.Driver.Core.Compression
                 _disposed = true;
                 if (_mode == CompressionMode.Compress)
                 {
-                    _deflateStream.Dispose();
-                    _stream.WriteByte((byte)(_adler32 >> 24));
-                    _stream.WriteByte((byte)(_adler32 >> 16));
-                    _stream.WriteByte((byte)(_adler32 >> 8));
-                    _stream.WriteByte((byte)_adler32);
+                    // Emit a valid empty zlib stream if Write was never called (matches SharpCompress).
+                    EnsureCompressInitialized();
+                    _compressDeflate.Dispose();
+                    _stream.WriteByte((byte)(_compressAdler32 >> 24));
+                    _stream.WriteByte((byte)(_compressAdler32 >> 16));
+                    _stream.WriteByte((byte)(_compressAdler32 >> 8));
+                    _stream.WriteByte((byte)_compressAdler32);
+                }
+                else
+                {
+                    _decompressDeflate?.Dispose();
                 }
                 if (!_leaveOpen)
                     _stream.Dispose();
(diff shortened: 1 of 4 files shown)


PhilCoultard:
Thanks for addressing this. I was about to report the vulnerability with SharpCompress but great to see you have it in hand already. Noted about the backwards compatibility requirement.

tebeco:
hello @damieng 

just to be sure which PR to follow.
I have the feeling that this PR would be more beneficial in the long term than just bumping SharpCompress

i also see that it's still in draft.
Is there any idea which of the two or both is currently concidered ?

or if Mongodb.Driver can confirm or not what's being written in the other PR ?
(someone mentionned the CVE does affect portion of used code and so could be dismissed until it's addressed)

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (mode == CompressionMode.Compress)
+            {
+                var header = level == CompressionLevel.NoCompression ? s_noCompressionHeader : s_defaultHeader;
+                stream.Write(header, 0, 2);
+                _deflateStream = new DeflateStream(stream, level, leaveOpen: true);
+            }
+            else
+            {
+                _decompressed = ReadAndDecompress(stream);
```
sanych-sun: Does it mean we do all IO/decompression in ctor? I do not think this is a good idea.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+                                       compressed[compressed.Length - 2] << 8 |
+                                       compressed[compressed.Length - 1]);
+            if (UpdateAdler32(1u, decompressedBytes, 0, decompressedBytes.Length) != expectedAdler)
+                throw new InvalidDataException("Zlib Adler-32 checksum mismatch.");
+
+            decompressed.Position = 0;
+            return decompressed;
+        }
+
+        private static uint UpdateAdler32(uint adler, byte[] buf, int offset, int count)
```
sanych-sun, line 223: We need unit tests for this method.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+                if (!_leaveOpen)
+                    _stream.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        private static MemoryStream ReadAndDecompress(Stream input)
+        {
+            byte[] compressed;
+            using (var ms = new MemoryStream())
```
sanych-sun: Do we really need this copy of steam?  It seems like we can combine this one with MemoryStream instantiated on line 116.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+* limitations under the License.
+*/
+
+using System;
+using System.IO;
+using System.IO.Compression;
+
+namespace MongoDB.Driver.Core.Compression
+{
+    // Replacement for SharpCompress.Compressors.Deflate.ZlibStream.
```
sanych-sun: I doubt we have to mention dependency which we are trying to drop.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
+/* Copyright 2019-present MongoDB Inc.
```
sanych-sun: _Minor_: should be 2010.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (disposing && !_disposed)
+            {
+                _disposed = true;
+                if (_mode == CompressionMode.Compress)
+                {
+                    _deflateStream.Dispose();
+                    _stream.WriteByte((byte)(_adler32 >> 24));
+                    _stream.WriteByte((byte)(_adler32 >> 16));
+                    _stream.WriteByte((byte)(_adler32 >> 8));
+                    _stream.WriteByte((byte)_adler32);
```
sanych-sun: I would say it's surprising to write something to the stream in dispose. I would prefer to have this code in Close method.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,147 @@
...
+using System;
+using System.IO;
+using System.IO.Compression;
+
+namespace MongoDB.Driver.Core.Compression
+{
+    // Replacement for SharpCompress.Compressors.Deflate.ZlibStream.
+    // Implements RFC 1950 (zlib framing) using System.IO.Compression.DeflateStream for the
+    // inner RFC 1951 (deflate) payload, with manual Adler-32 checksum handling.
+    internal sealed class ZlibStream : Stream
```
sanych-sun, line 31: For modern targets we have built-in `ZLibStream`. Should we use it and keep this implementation for `net472` only?

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: all of this needs to be tested on a real aspnetcore app such as a WebApplicationFactory where aspnetcore will throw at any attempt to use SyncIO for stream
this method isn't async i suspect it won't work

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: you probably meant CloseAsync or even FlushAsync if this needs to write stuff ?

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: I/O in a ctor is really bad practice

all of this needs to be tested on a real aspnetcore app such as a WebApplicationFactory where aspnetcore will throw at any attempt to use SyncIO for stream
this method isn't async i suspect it won't work

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
sanych-sun: `Close` for sync code-path and `CloseAsync` for async code-path. `Flush`/`FlushAsync` might be used to clean the buffer for some reasons, but more data could be expected.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
sanych-sun: Yep, I do agree that we should rewrite the code to properly handle async code-path. But also we should start with evaluation if built-in `ZLibStream` can be used for net6+ targets, and keep this class only for net472 target (after proper implementation of async)

damieng: Yes, sorry, this PR was just a draft to prove that the wrapped deflate technique worked correctly against the the real variants of MongoDB Server we support.

tebeco:
@damieng / @sanych-sun 

can we please concider targeting explicitly net10.0 ?


and start looking at net11.0 explicit tfm ?

https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/preview4/libraries.md#span-based-deflate-zlib-and-gzip-encoderdecoder-apis

there's many changes that adds more support for Zlib so it would be cool to know how to anticipate changes ahead of time to avoid possible breaking changes that api would force/offer

its 6 months ahead

tebeco:
in a more personnal comment
i could only suggest you drop all tfm except
netstandard2.0
netstandard2.1 (only if you already explicitly support it)
net8.0 (until this november)
net10.0

netstandard2.1 imply runtime that would both support it and that are not already "modern dotnet"

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: Originally I had this implementation in ZLibCompressor.cs but decided to split it out into a class and file of its own... For some reason ZLibCompressor.cs has 2019 in it.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: Now moved out.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: The Adler values are covered in the zlib tests.  If we want to indepdentely validate the Adelr32 we'll need to make this internal or public.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: Done.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: I think we should do that in a subsequent update otherwise we have double the surface area changing with two implementations potentially in a patch release. If/when we can drop 4.7.2 I'd like it if we can go full-native with the .NET 6 one and drop this work here entirely.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: Writing the trailer on Dispose matches the BCL pattern (DeflateStream, GZipStream, CryptoStream all do the same).  DeflateStream's Dispose calls [PurgeBuffers](https://github.com/dotnet/runtime/blob/fb7020f32aacb98109893cc40e437937a01c8dd6/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs#L643).

Overriding Close() goes against [Microsoft's guidance](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.close?view=netframework-4.8.1#notes-to-inheritors) which recommends instead putting it in dispose.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: The limitation with WebApplicationFactory and async specifically relates to ASP.NET Request and Response. If we were shipping a general-purpose public ZlibStream that would definitely be a scenario we'd want to support.

In our internal case we only run this on ByteBufferStream, an in-memory buffer that gets populated by NetworkStream (which supports sync and async) on the wire side, and is then consumed by
CompressedMessageBinaryEncoder (which is already only sync today the same as all our IMessageEncoder implementations).

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,225 @@
...
+                var expected = (uint)(_trailer[0] << 24 | _trailer[1] << 16 | _trailer[2] << 8 | _trailer[3]);
+                if (_decompressAdler32 != expected)
+                {
+                    throw new InvalidDataException("Zlib Adler-32 checksum mismatch.");
+                }
+            }
+            return n;
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
```
sanych-sun, line 107: If we intentionally support only sync code path, should we override async methods and throw there?

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,225 @@
...
+    //
+    // Both compress and decompress initialise lazily — no I/O happens in the constructor.
+    // Decompress path: the compressed input is buffered up-front (small — wire messages are
+    // length-bounded), but the decompressed output is produced lazily as the caller reads,
+    // with Adler-32 accumulated incrementally and validated when DeflateStream signals EOF.
+    internal sealed class ZlibStream : Stream
+    {
+        // RFC 1950 headers: CMF=0x78 (deflate, 32K window) + FLG satisfying (CMF*256+FLG)%31==0.
+        // FLEVEL bits in FLG are informational only (RFC 1950 §2.2) and do not affect decompression.
+        private static readonly byte[] s_defaultHeader = { 0x78, 0x9C };       // FLEVEL=2
```
sanych-sun: _Minor_: rename to `__defaultHeader` and `__noCompressionHeader `

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: this would be possible to have async path and would be used in `IAsyncDisposable`

assuming we want `DisposeAsync`, but I can't really tell as i'm not knowing in depth this repo.
It could possibly be a small performance boost

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                return;
+            }
+            _decompressInitialized = true;
+
+            byte[] compressed;
+            using (var ms = new MemoryStream())
+            {
+                _stream.CopyTo(ms);
+                compressed = ms.ToArray();
```
sanych-sun: As far as I understood we need this copy only to limit size of the stream. Right? Taking into account that this stream could be relatively big, can we avoid this copy somehow or at least use some pool? Can we implement something like a size limited stream? The stream which wraps another stream but restrict it's size?

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: > If/when we can drop 4.7.2 I'd like it if we can go full-native with the .NET 6 one and drop this work here entirely.

you can already go native for anything above net6.0

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

```cs
#if !NET6_0_OR_GREATER

// ENTIRE FILE

#endif
```

I feel like this must 100% what should be done
you would also 
```cs
#if NET6_0_OR_GREATER
using System.IO.Compression
#else
using The.Other.Namespace
#endif
```

tebeco: I feel like it's becomming urgent to also stop targetting net6.0 and start targetting at least 10.0 as mentionned previously

as for support for netfx, that would be a breaking change

tebeco: I wonder if mongo starts targeting net8.0, net9.0, net10.0 explicitly
then you can release a nuget package for these TFM by removing SharpCompress there
and I wonder if consumer on net10.0 would still have the CVE warning.

In the meanwhile another PR could deal with supporting ZLib for Netfx

tebeco: pretty sure it's probably the best alternative to rely on native when native is supported, and use `#if` for legacy

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: Nothing in the driver would be calling the async methods - we would need to change a lot more than just the zlibstream to light them up. Might be worth it at some point but well beyond the scope of this piece of work.

Given this is not a general-purpose Zlib implementation but rather something we only use internally in a very specific path throwing on the Async doesn't get us anything. We completely control how this is used and there is no async path into it.

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
damieng: We'll have to leave the targetting conversations for a different place - that's not something that can be decided in a security patch release.

There is some benefit to providing .NET 6 or greater conditionals and using the built-in. I didn't initially want double the surface area but have come around enough to try it in another commit.

tebeco: cool to see the native code being used for existing modern TFM, thx a lot

(I wonder if net10.0 resolve to ns2.0 or net6.0)

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
tebeco: thx for clarifying

src/MongoDB.Driver/Core/Compression/ZlibStream.cs
0xced: For a `net10.0` project, `net6.0` will be resolved.

See [Get Nearest Framework](https://nugettools.azurewebsites.net/7.6.0/get-nearest-framework?project=net10.0&package=netstandard2.0%0D%0Anet6.0%0D%0A) in NuGet Tools.

PR #2004, opened 2026-05-19
CSHARP-5992: Add LINQ translation benchmark suite

adelinowona:
## Summary

Adds a 9-benchmark `LinqBench` suite that exercises the LINQ-to-aggregation translation layer in isolation (no queries executed at benchmark time), plus a `LinqEndToEnd` suite that compares LINQ vs hand-built `BsonDocument` query plans end-to-end. Wires `LinqBench` into the perf-job category filter and the composite-score path, giving us a defensible signal when the translator (especially `SerializerFinder`, `AstSimplifier`, and the method-call sub-translators) moves.

Cross-commit runs against pre-/post-SerializerFinder and an in-flight optimization PR (#1961) show the suite catches the regressions and improvements we'd want it to — see Validation.

## Motivation

The driver has spec-driven benchmarks for I/O-heavy patterns (`Find`, `BulkWrite`, `GridFS`, BSON encode/decode) but nothing for the LINQ translator. As internal paths shift — `SerializerFinder` overhauls, `AstSimplifier` changes, new visitor support — we have no signal on translation-cost movement. When CSHARP-5572 introduced `SerializerFinder` (#1700), we couldn't quantify what it cost. This closes that gap.

## What this adds

| File | Purpose |
|---|---|
| `benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqTranslationBenchmark.cs` | 9-benchmark translation suite (filter / field / projection / update / IQueryable entry points) |
| `benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs` | 12 LINQ-vs-raw end-to-end benchmarks (6 patterns × {LINQ, Raw}) against a mongod |
| `benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqBenchmarkDataTypes.cs` | Shared POCO models for both LINQ suites |
| `benchmarks/MongoDB.Driver.Benchmarks/Linq/README.md` | Suite docs: full benchmark inventory and how to interpret |
| `benchmarks/MongoDB.Driver.Benchmarks/DriverBenchmarkCategory.cs` | New `LinqBench` and `ExcludeFromComposite` consts; `LinqBench` and `BulkWriteBench` added to `AllCategories`. `ExcludeFromComposite` keeps the end-to-end benchmarks out of the `LinqBench` composite while still running them |
| `benchmarks/MongoDB.Driver.Benchmarks/BenchmarkResult.cs` + `Exporters/*.cs` | Scores time-throughput benchmarks as operations/second; composite derives its unit from members and skips `ExcludeFromComposite` benchmarks |
| `evergreen/run-perf-tests.sh` | Adds `LinqBench` to the `--anyCategories` filter |

Cedar/SPS auto-discovers new composite categories — no dashboard config required.

## Design decisions

- **Translation-only as the primary suite.** `LinqTranslationBenchmark` calls translator entry points directly (`LinqProviderAdapter.TranslateExpressionTo*`, `ExpressionToExecutableQueryTranslator.Translate`); no queries run at benchmark time, isolating translator regressions from network and serialization noise. (Caveat: `QueryablePipeline` and `GroupByAggregation` need a `MongoQueryProvider<T>`, obtained from `collection.AsQueryable()` on a real `MongoClient` in `[GlobalSetup]`; its background cluster monitor is the only DB-side activity, expected below the 2-4% drift floor.) End-to-end coverage lives in `LinqEndToEndBenchmark`; it shares the `LinqBench` category so it runs in the perf job, and is marked `ExcludeFromComposite` so it stays out of the composite.
- **Representative user queries, not a visitor matrix.** An earlier proposal organized benchmarks by which `SerializerFinderVisit*.cs` file they exercised; reframed to "what users actually write." Matrix coverage stays an option for targeted gaps.
- **`LinqBench` does not cross-tag with the spec categories** (`DriverBench`, `BsonBench`, etc.), keeping LINQ numbers out of the spec composite averages. It lands in `AllCategories` so its own composite is emitted.
- **`BulkWriteBench` composite bundled here.** Previously excluded with a "not part of the benchmarking spec" comment; team wanted its composite tracked, done here to avoid a tiny standalone PR.
- **`[MemoryDiagnoser]` on everything.** Allocation regressions matter independently of time and surface earlier than time regressions on noisy hardware.

## Benchmark inventory

Translation suite: 9 benchmarks across filter (4), field (1), projection (1), update (1), and IQueryable (2) entry points. End-to-end suite: 12 benchmarks — 6 patterns (`MultiFieldSearch`, `OrFilter`, `GroupBy`, `Projection`, `InFilter`, `PagedQuery`) each run LINQ vs hand-built raw. The full per-benchmark breakdown (patterns, translator paths exercised) lives in [`Linq/README.md`](benchmarks/MongoDB.Driver.Benchmarks/Linq/README.md).

The e2e suite seeds 500 documents with secondary indexes on `Status`, `CreatedAt`, and `ShippingAddress.City` in `[GlobalSetup]`; each LINQ/Raw pair renders to byte-equivalent BSON (verified via `LinqProviderAdapter`), so the LINQ−Raw delta isolates translator + provider overhead from query-shape differences.

## Validation

Five lines of evidence that the suite produces actionable signal.

### 1. Within-run noise

Across all perf-hw runs (n=10 each, multiple commits), BDN-reported within-run StdDev is <1% on most benchmarks (typically 0.2-0.9%), with the fastest micro-benchmarks at 0.3-1.4%. Each individual run is a well-converged measurement.

### 2. Selectivity — targeted regression injection

`Thread.SpinWait(300)` (~10 µs on M1) injected into four translator code paths in turn:

| Injection point | Expected affected benchmarks |
|---|---|
| `SerializerFinder.FindSerializers()` | All benchmarks (universal path) |
| `GetItemMethodToFilterFieldTranslator` | `FieldSelection` only |
| `NotExpressionToFilterTranslator` | `MultiFieldSearch` (Not), `OrFilter` (chains Comparison dispatch) |
| `GroupByMethodToPipelineTranslator` | `GroupByAggregation` only |

Each injection moved only the benchmarks that should have moved — clean per-path selectivity, so when something regresses the benchmarks that move tell you *which translator* moved.

### 3. Cross-run drift on the perf-job hardware (n=10 on `rhel90-dbx-perf-large`)

.NET 8.0, X64 RyuJit, single perf-task invocation on the same host.

| Benchmark | Min | Median | Max | Range% | CV% | Within-run StdDev% |
|---|---:|---:|---:|---:|---:|---:|
| `MultiFieldSearch` | 494.0 µs | 497.2 µs | 507.8 µs | 2.8% | 1.03% | 0.23% |
| `OrFilter` | 13.3 µs | 13.5 µs | 13.7 µs | 3.1% | 1.01% | 0.23% |
| `BatchLookup` | 144.9 µs | 145.8 µs | 148.1 µs | 2.2% | 0.74% | 0.88% |
| `ArrayElementQuery` | 177.4 µs | 179.4 µs | 181.4 µs | 2.2% | 0.77% | 0.51% |
| `FieldSelection` | 5,610 ns | 5,795 ns | 6,025 ns | 7.1% | 2.30% | 0.32% |
| `AggregationProjection` | 406.5 µs | 412.6 µs | 420.1 µs | 3.3% | 1.07% | 0.47% |
| `UpdatePipeline` | 95.5 µs | 96.3 µs | 98.4 µs | 3.0% | 1.06% | 0.59% |
| `QueryablePipeline` | 874.5 µs | 890.7 µs | 913.6 µs | 4.4% | 1.38% | 0.52% |
| `GroupByAggregation` | 489.7 µs | 495.1 µs | 506.6 µs | 3.4% | 0.94% | 0.77% |

Every benchmark but `FieldSelection` lands in 2-4.5% range, CV ≤1.5% — a ~5× compression of the M1 drift bands characterized during development. `FieldSelection` (~6µs) drifts wider (7.1%), fast enough that small absolute drift looks proportionally large. Sub-10% deltas are individually resolvable, which matters for the optimization comparison below.

### 4. Cross-commit reality check

Transplanted onto pinned commits and run on `rhel90-dbx-perf-large` (n=10 per commit), the suite caught both a known historical regression and an in-flight optimization:

- **SerializerFinder introduction (#1700)** — parent (`46640eac98`) vs the merge (`59c9d34180`): every benchmark moved well above its drift band, most 2-3× slower. `UpdatePipeline` was a +626% time / +898% allocation outlier because `TranslateExpressionToSetStage` runs `SerializerFinder` on the un-preprocessed lambda, unlike the other entry points which preprocess first (root cause in the follow-up below). Confirms the suite surfaces real translator regressions.
- **Optimization #1961** (`SerializerFinderVisitMethodCall` switch → `MethodInfo`-keyed lookup), base `66780341e7` vs head `54973d039a` — measurable allocation wins on the method-call / IQueryable benchmarks (`ArrayElementQuery` -12.4%, `QueryablePipeline` -4.5%, `BatchLookup` / `GroupByAggregation` -3.3%) with smaller time effects at the edge of the drift bands. The suite resolves what *kind* of change this is — an allocation reduction — which was invisible at M1 noise levels.

### 5. End-to-end overhead — Atlas dev cluster, perf-hardware, n=10

Six patterns run twice each (LINQ-translated vs hand-built `BsonDocument` / pipeline), 500 docs with secondary indexes, against a live Atlas dev cluster from the perf host. Each LINQ/Raw pair renders to byte-equivalent BSON, so the LINQ−Raw delta reflects translator and provider overhead, not query-shape differences.

| Pattern | LINQ median | Raw median | LINQ−Raw | LINQ/Raw | **Translator share** | LINQ alloc | Raw alloc | Alloc ratio |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `MultiFieldSearch` | 1,966 µs | 1,185 µs | +782 µs | 1.66× | **39.7%** | 74.2 KB | 43.6 KB | 1.70× |
| `OrFilter` | 7,321 µs | 7,230 µs | +91 µs | 1.01× | **1.2%** | 719.1 KB | 711.1 KB | 1.01× |
| `GroupBy` | 2,034 µs | 1,476 µs | +558 µs | 1.38× | **27.4%** | 65.4 KB | 20.6 KB | 3.18× |
| `Projection` | 2,160 µs | 1,051 µs | +1,109 µs | 2.05× | **51.3%** | 124.8 KB | 35.5 KB | 3.52× |
| `InFilter` | 1,100 µs | 784 µs | +315 µs | 1.40× | **28.7%** | 41.5 KB | 28.8 KB | 1.44× |
| `PagedQuery` | 1,645 µs | 1,213 µs | +432 µs | 1.36× | **26.3%** | 63.3 KB | 34.8 KB | 1.82× |

**Translator share** is the fraction of user-visible LINQ time that disappears if you write raw `BsonDocument` instead — an *upper bound* on translator cost (the delta also includes provider overhead like cursor construction and command serialization).

- Projection-heavy patterns (`Projection` 51%, `MultiFieldSearch` 40%): translator is ~40-50% of user-visible latency on indexed selective queries; a 10% translator regression is a ~5% user-visible one.
- Filter/aggregation patterns (`GroupBy` 27%, `InFilter` 29%, `PagedQuery` 26%): translator is ~25-30%; meaningful server-side work (`$group`, `$in`, sort-skip-limit) partially offsets it.
- Broad scans (`OrFilter` 1.2%): translator is ~1% because the 4-way OR matches many documents and serializes ~700 KB, so a 10% translator regression is invisible to users.
- Allocation ratios isolate LINQ-side overhead: `Projection` 3.5×, `GroupBy` 3.2× — projected-type-serializer and `IGroupingSerializer` construction is allocation-heavy, and catches translator-side allocation regressions even when network time masks the time delta.

Caveat: Atlas dev cluster across the internet from the perf host; per-run absolute times range ±15-30% (up to ±50% where server time dominates), but the translator-share *ratios* are more stable because LINQ and Raw on the same iteration see correlated network noise. Single-batch result, 500 docs.

## Regression-alert thresholds (perf-hardware-calibrated)

Calibrated to observed drift on `rhel90-dbx-perf-large`:

| Bucket | Threshold | Benchmarks | Rationale |
|---|---|---|---|
| Tight | 8% | `MultiFieldSearch`, `BatchLookup`, `ArrayElementQuery`, `AggregationProjection`, `UpdatePipeline`, `QueryablePipeline`, `GroupByAggregation`, `OrFilter` | Observed range 2.2–4.4%; 8% gives ~2× headroom. |
| Wider | 12% | `FieldSelection` | ~6µs benchmark; observed range 7.1%. |

Allocation thresholds should be even tighter — observed allocation drift is 0-1.2%.

## Follow-ups (not in this PR)

- File a ticket for the `TranslateExpressionToSetStage` preprocessing asymmetry surfaced above (SerializerFinder runs on the un-preprocessed tree; `UpdatePipeline` +626% time, +898% alloc). The design is intentional — dispatch pattern-matches on `NewExpression`/`MemberInitExpression`, which `PartialEvaluator` would collapse if applied at the top — so any fix must preserve that dispatch shape. Not a one-line change.
- Sub-3% time-delta detection on real changes hasn't been measured end-to-end (current experiments validate large deltas and 3-12% alloc deltas); the perf-hw drift bands suggest it's feasible.
- Coverage gaps: `Lookup`/`Join`, `Distinct`, `SelectMany`, `Cast`, `$expr` fallback paths.
- Quantify the background cluster-monitor's noise contribution to `QueryablePipeline`/`GroupByAggregation` (expected sub-1%, absorbed by the 2-4% drift floor).
- Runtime equivalence guard for the e2e suite: assert each LINQ expression's translated BSON equals its pre-built Raw doc in `[GlobalSetup]`, so a future translator shape change fails the benchmark loudly instead of silently shifting the share numbers.

### DIFF
--- a/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
+++ b/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
@@ -1,0 +1,185 @@
+/* Copyright 2010-present MongoDB Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using BenchmarkDotNet.Attributes;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using static MongoDB.Benchmarks.BenchmarkHelper;
+using static MongoDB.Benchmarks.Linq.LinqTranslationBenchmark;
+
+namespace MongoDB.Benchmarks.Linq;
+
+[MemoryDiagnoser]
+public class LinqEndToEndBenchmark
+{
+    private const string DatabaseName = "linqbench";
+    private const string CollectionName = "orders";
+    private const int SeedCount = 500;
+
+    private IMongoClient _client;
+    private IMongoCollection<OrderDocument> _collection;
+
+    private readonly string _statusFilter = "Active";
+    private readonly string _prefix = "Acme";
+    private readonly string _city = "Seattle";
+    private readonly DateTime _cutoff = new(2025, 1, 1);
+
+    private BsonDocument _multiFieldFilter;
+    private BsonDocument _orStatusFilter;
+    private BsonDocument[] _groupByPipeline;
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        _client = MongoConfiguration.CreateClient();
+        var db = _client.GetDatabase(DatabaseName);
+        _collection = db.GetCollection<OrderDocument>(CollectionName);
+
+        SeedData();
+        PreBuildQueries();
+    }
+
+    [GlobalCleanup]
+    public void Cleanup()
+    {
+        _client.DropDatabase(DatabaseName);
+        _client.Dispose();
+    }
+
+    // --- MultiFieldSearch: compound filter ---
+
+    [Benchmark]
+    public List<OrderDocument> MultiFieldSearchLinq()
+    {
+        return _collection.Find(x =>
+            x.Status == _statusFilter &&
+            x.CustomerName.StartsWith(_prefix) &&
+            x.ShippingAddress.City == _city &&
+            x.CreatedAt > _cutoff &&
+            !x.IsPaid).ToList();
+    }
+
+    [Benchmark]
+    public List<OrderDocument> MultiFieldSearchRaw()
+    {
+        return _collection.Find(_multiFieldFilter).ToList();
+    }
+
+    // --- OrStatusFilter: simple OR filter ---
+
+    [Benchmark]
+    public List<OrderDocument> OrStatusFilterLinq()
+    {
+        return _collection.Find(x =>
+            x.Status == "Active" || x.Status == "Pending" || x.Status == "Processing" || x.Status == "Shipped").ToList();
+    }
+
+    [Benchmark]
+    public List<OrderDocument> OrStatusFilterRaw()
+    {
+        return _collection.Find(_orStatusFilter).ToList();
+    }
+
+    // --- GroupBy aggregate ---
+
+    [Benchmark]
+    public List<BsonDocument> GroupByLinq()
+    {
+        return _collection.Aggregate()
+            .Group(x => x.Status, g => new { Status = g.Key, Count = g.Count(), TotalRevenue = g.Sum(x => x.Total) })
+            .As<BsonDocument>()
+            .ToList();
+    }
+
+    [Benchmark]
+    public List<BsonDocument> GroupByRaw()
+    {
+        var pipeline = PipelineDefinition<OrderDocument, BsonDocument>.Create(_groupByPipeline);
+        return _collection.Aggregate(pipeline).ToList();
+    }
+
+    private void PreBuildQueries()
+    {
+        _multiFieldFilter = new BsonDocument
+        {
+            { "Status", _statusFilter },
+            { "CustomerName", new BsonDocument("$regex", $"^{_prefix}") },
+            { "ShippingAddress.City", _city },
+            { "CreatedAt", new BsonDocument("$gt", _cutoff) },
+            { "IsPaid", new BsonDocument("$ne", true) }
+        };
+
+        _orStatusFilter = new BsonDocument("$or", new BsonArray
+        {
+            new BsonDocument("Status", "Active"),
+            new BsonDocument("Status", "Pending"),
+            new BsonDocument("Status", "Processing"),
+            new BsonDocument("Status", "Shipped")
+        });
+
+        _groupByPipeline = new[]
+        {
+            new BsonDocument("$group", new BsonDocument
+            {
+                { "_id", "$Status" },
+                { "Count", new BsonDocument("$sum", 1) },
+                { "TotalRevenue", new BsonDocument("$sum", "$Total") }
+            })
+        };
+    }
+
+    private void SeedData()
+    {
+        var statuses = new[] { "Active", "Pending", "Processing", "Shipped", "Cancelled" };
+        var cities = new[] { "Seattle", "Portland", "Denver", "Austin", "Boston" };
+        var names = new[] { "Acme Corp", "Acme Inc", "Globex", "Initech", "Umbrella" };
+        var random = new Random(42);
+
+        var documents = Enumerable.Range(0, SeedCount).Select(i => new OrderDocument
+        {
+            Id = i,
+            CustomerName = names[i % names.Length],
+            Status = statuses[i % statuses.Length],
+            Total = random.Next(10, 1000),
+            Subtotal = random.Next(10, 900),
+            Tax = random.Next(1, 100),
+            Discount = random.Next(0, 50),
+            CreatedAt = new DateTime(2024, 1, 1).AddDays(random.Next(0, 730)),
+            IsPaid = i % 3 != 0,
+            Currency = "USD",
+            ItemCount = random.Next(1, 10),
+            PaymentMethod = "Card",
+            ShippingMethod = "Standard",
+            ShippingAddress = new Address
+            {
+                Street = $"{i} Main St",
+                City = cities[i % cities.Length],
+                State = "WA",
+                Zip = "98101",
+                Country = "US"
+            },
+            Items = Enumerable.Range(0, random.Next(1, 5)).Select(j => new OrderItem
+            {
+                ProductId = $"prod-{j}",
+                Price = random.Next(5, 200)
+            }).ToList()
+        }).ToList();
+
+        _collection.InsertMany(documents);
+    }
+}
--- a/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
+++ b/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
@@ -24,13 +24,12 @@ using static MongoDB.Benchmarks.BenchmarkHelper;
 namespace MongoDB.Benchmarks.Linq;
 
 [MemoryDiagnoser]
+[BenchmarkCategory(DriverBenchmarkCategory.LinqBench, DriverBenchmarkCategory.ExcludeFromComposite)]
 public class LinqEndToEndBenchmark
 {
     private const string DatabaseName = "linqbench";
     private const string CollectionName = "orders";
-    // 500 is intentional: the broad-scan vs. selective contrast (e.g. OrFilter's ~700 KB
-    // result dominating its LINQ-Raw delta) needs enough documents to make serialization
-    // a visible cost. Translation-only timing lives in LinqTranslationBenchmark, not here.
+    // Enough documents to return a realistic result size; see the README for why.
     private const int SeedCount = 500;
 
     private IMongoClient _client;
@@ -64,9 +63,7 @@ public class LinqEndToEndBenchmark
         PreBuildQueries();
     }
 
-    // Indexes minimize server-side filter/group time so cross-benchmark deltas reflect
-    // translation cost rather than COLLSCAN variance. Without indexes, ~95% of e2e time
-    // on the heavier benchmarks goes to the server, and translator changes are invisible.
+    // Indexed so server-side filter/group time stays small; see the README for why.
     private void CreateIndexes()
     {
         _collection.Indexes.CreateMany(new[]
--- a/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
+++ b/benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
@@ -39,9 +39,16 @@ public class LinqEndToEndBenchmark
     private readonly string _city = "Seattle";
     private readonly DateTime _cutoff = new(2025, 1, 1);
 
+    private readonly int[] _lookupIds = { 1, 50, 100, 200, 400 };
+    private const int PageSize = 10;
+
     private BsonDocument _multiFieldFilter;
-    private BsonDocument _orStatusFilter;
+    private BsonDocument _orFilter;
     private BsonDocument[] _groupByPipeline;
+    private BsonDocument _projectionFilter;
+    private BsonDocument _projectionDoc;
+    private BsonDocument _inFilter;
+    private BsonDocument[] _pagedQueryPipeline;
 
     [GlobalSetup]
     public void Setup()
@@ -51,7 +58,21 @@ public class LinqEndToEndBenchmark
         _collection = db.GetCollection<OrderDocument>(CollectionName);
 
         SeedData();
+        CreateIndexes();
         PreBuildQueries();
+    }
+
+    // Indexes minimize server-side filter/group time so cross-benchmark deltas reflect
+    // translation cost rather than COLLSCAN variance. Without indexes, ~95% of e2e time
+    // on the heavier benchmarks goes to the server, and translator changes are invisible.
+    private void CreateIndexes()
+    {
+        _collection.Indexes.CreateMany(new[]
+        {
+            new CreateIndexModel<OrderDocument>(Builders<OrderDocument>.IndexKeys.Ascending(x => x.Status)),
+            new CreateIndexModel<OrderDocument>(Builders<OrderDocument>.IndexKeys.Ascending(x => x.CreatedAt)),
+            new CreateIndexModel<OrderDocument>(Builders<OrderDocument>.IndexKeys.Ascending("ShippingAddress.City")),
+        });
     }
 
     [GlobalCleanup]
@@ -80,19 +101,19 @@ public class LinqEndToEndBenchmark
         return _collection.Find(_multiFieldFilter).ToList();
     }
 
-    // --- OrStatusFilter: simple OR filter ---
+    // --- OrFilter: simple OR filter ---
 
     [Benchmark]
-    public List<OrderDocument> OrStatusFilterLinq()
+    public List<OrderDocument> OrFilterLinq()
     {
         return _collection.Find(x =>
             x.Status == "Active" || x.Status == "Pending" || x.Status == "Processing" || x.Status == "Shipped").ToList();
     }
 
     [Benchmark]
-    public List<OrderDocument> OrStatusFilterRaw()
+    public List<OrderDocument> OrFilterRaw()
     {
-        return _collection.Find(_orStatusFilter).ToList();
+        return _collection.Find(_orFilter).ToList();
     }
 
     // --- GroupBy aggregate ---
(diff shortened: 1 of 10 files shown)


benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+    [Benchmark]
+    public List<OrderDocument> MultiFieldSearchRaw()
+    {
+        return _collection.Find(_multiFieldFilter).ToList();
+    }
+
+    // --- OrStatusFilter: simple OR filter ---
+
+    [Benchmark]
+    public List<OrderDocument> OrStatusFilterLinq()
```
BorisDog: minor: I think we need more descriptive naming, not tied to a specific field name.  Something like `OrFilterLinq`, or `FindOrFilterLinq`

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+            .As<BsonDocument>()
+            .ToList();
+    }
+
+    [Benchmark]
+    public List<BsonDocument> GroupByRaw()
+    {
+        var pipeline = PipelineDefinition<OrderDocument, BsonDocument>.Create(_groupByPipeline);
+        return _collection.Aggregate(pipeline).ToList();
+    }
```
BorisDog, line 128: I would also add additional case with all the following
* Select/Pojection, 
* something with arrays, like $in, $all....
* OrderBy.ThenBy, Take

BorisDog (changes requested):
Look very good overall.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+
+    // --- MultiFieldSearch: compound filter ---
+
+    [Benchmark]
+    public List<OrderDocument> MultiFieldSearchLinq()
+    {
+        return _collection.Find(x =>
+            x.Status == _statusFilter &&
+            x.CustomerName.StartsWith(_prefix) &&
```
ajcvickers, line 89: Does StartsWith create the exact same regex as in the raw version?

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+using static MongoDB.Benchmarks.Linq.LinqTranslationBenchmark;
+
+namespace MongoDB.Benchmarks.Linq;
+
+[MemoryDiagnoser]
+public class LinqEndToEndBenchmark
+{
+    private const string DatabaseName = "linqbench";
+    private const string CollectionName = "orders";
+    private const int SeedCount = 500;
```
ajcvickers, line 33: Consider creating indexes in the database such that server time is minimized and translation time changes will be more apparent.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+        return _collection.Find(_orStatusFilter).ToList();
+    }
+
+    // --- GroupBy aggregate ---
+
+    [Benchmark]
+    public List<BsonDocument> GroupByLinq()
+    {
+        return _collection.Aggregate()
+            .Group(x => x.Status, g => new { Status = g.Key, Count = g.Count(), TotalRevenue = g.Sum(x => x.Total) })
```
ajcvickers, line 118: There is projection to an anonymous type here, followed by creation of the BSON document. The raw example does not project to an anonymous type.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,303 @@
...
+            new CreateIndexModel<OrderDocument>(Builders<OrderDocument>.IndexKeys.Ascending(x => x.CreatedAt)),
+            new CreateIndexModel<OrderDocument>(Builders<OrderDocument>.IndexKeys.Ascending("ShippingAddress.City")),
+        });
+    }
+
+    [GlobalCleanup]
+    public void Cleanup()
+    {
+        _client.DropDatabase(DatabaseName);
+        _client.Dispose();
```
sanych-sun, line 81: In v3 MongoDB.Driver this will not unregister the cluster. If we really want to clean up we should follow [the clean up guide](https://www.mongodb.com/docs/drivers/csharp/current/connect/mongoclient/#manage-mongoclient-disposables).

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
sanych-sun: It might makes sense to made a first translation somewhere in GlobalSetup and compare the produced MQL. So if we will change the translation in future the Benchmark will throw.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,303 @@
...
+using static MongoDB.Benchmarks.Linq.LinqTranslationBenchmark;
+
+namespace MongoDB.Benchmarks.Linq;
+
+[MemoryDiagnoser]
+public class LinqEndToEndBenchmark
+{
+    private const string DatabaseName = "linqbench";
+    private const string CollectionName = "orders";
+    private const int SeedCount = 500;
```
sanych-sun, line 33: Do we really need 500 documents? As far as I understood the point is to measure the LINQ translation overhead, so I would say it should be enough to have even a single document in the collection to reduce the server/network/serialization overhead.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
adelinowona: Thanks for flagging — checked the chain and the cleanup is actually already correct here.

`MongoConfiguration.CreateClient()` in `BenchmarkHelper.cs:81` explicitly sets `settings.ClusterSource = DisposingClusterSource.Instance` before constructing the client. With that:

- `MongoClient.Dispose()` calls `_settings.ClusterSource.Return(_cluster)` (`MongoClient.cs:187`)
- `DisposingClusterSource.Return(cluster)` calls `ClusterRegistry.Instance.UnregisterAndDisposeCluster(cluster)` (`IClusterSource.cs:39`)

So the cluster does get unregistered and disposed. The cleanup-guide warning applies to default-configured clients (`DefaultClusterSource`, whose `Return` is a no-op) — we opt out of that via the benchmark helper. Same pattern is used by all 15 existing benchmark classes (`FindOneBenchmark`, `BulkWriteMixedOpsBenchmark`, etc.).

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
adelinowona: Reasonable read of the suite, but the e2e benchmark is measuring something different from translation overhead — that's what `LinqTranslationBenchmark` is for (no queries executed, just translator entry points). The e2e suite measures **translator share of user-visible latency under realistic result sizes**, and the count is calibrated so the broad-scan / selective contrast stays visible.

Concrete example from §5 of the PR description: `OrFilter`'s translator share is **~1.2%**, specifically because the 4-way OR matches a large fraction of documents and the result set serializes ~700 KB — at that point network + deserialization dominate, so a 10% translator regression is invisible to users. With 1 document the contrast collapses; every benchmark behaves the same and we lose the "when does translation matter to users?" question the e2e suite is built around.

Added an inline comment on `SeedCount` (cba39e9b5a) to record this for future readers.

sanych-sun (approved):
LGTM

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,306 @@
...
+
+    [Benchmark]
+    public List<OrderSummary> ProjectionRaw()
+    {
+        return _collection.Find(_projectionFilter)
+            .Project<OrderSummary>(_projectionDoc)
+            .ToList();
+    }
+
+    // --- InFilter: array Contains → $in ---
```
BorisDog: I don't think such comments are very helpful.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,306 @@
...
+        _projectionFilter = new BsonDocument("Status", _statusFilter);
+        _projectionDoc = new BsonDocument
+        {
+            { "CustomerName", 1 },
+            { "Total", 1 },
+            { "_id", 0 }
+        };
+
+        // ids.Contains(x.Id) translates to { "_id": { "$in": [...] } } since Id maps to _id by convention.
+        _inFilter = new BsonDocument("_id", new BsonDocument("$in", new BsonArray(_lookupIds.Select(id => (BsonValue)id))));
```
BorisDog, line 233: Should we `_lookupIds.Select( ...)` be evaluated once? So it doesn't contributed to benchmark time.

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
```diff
@@ -0,0 +1,306 @@
...
+using BenchmarkDotNet.Attributes;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using static MongoDB.Benchmarks.BenchmarkHelper;
+using static MongoDB.Benchmarks.Linq.LinqTranslationBenchmark;
+
+namespace MongoDB.Benchmarks.Linq;
+
+[MemoryDiagnoser]
+public class LinqEndToEndBenchmark
```
BorisDog, line 28: Should this have a category? `[BenchmarkCategory(DriverBenchmarkCategory.LinqBench)]` ?

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqTranslationBenchmark.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        _groupByAggregationExpression = queryable
+            .GroupBy(x => x.Status)
+            .Select(g => new GroupResult { Status = g.Key, Count = g.Count(), TotalRevenue = g.Sum(x => x.Total) })
+            .Expression;
+    }
+
+    #region Models
+
+    public class OrderDocument
```
BorisDog: Let's define this in `LinqBenchmarkDataTypes` or similar. Same as done in `BsonBenchmarkDataTypes.cs`.

BorisDog (changes requested):

benchmarks/MongoDB.Driver.Benchmarks/Linq/LinqEndToEndBenchmark.cs
adelinowona: Good instinct, but it's already setup-only. That `_lookupIds.Select(...)` is inside `PreBuildQueries()`, which runs from `Setup()` under `[GlobalSetup]` — so the `$in` array is materialized into `_inFilter` once before any measured iteration, not per-invocation.

BorisDog: Right, might bad :)

BorisDog (approved):
LGTM

PR #2005, opened 2026-05-19
CSHARP-6009: Send afterClusterTime on writes in causally-consistent sessions

sanych-sun:


src/MongoDB.Driver/Core/Operations/ReadConcernHelper.cs
```diff
@@ -27,6 +27,22 @@ public static BsonDocument GetReadConcernForCommand(ICoreSession session, Connec
             return session.IsInTransaction ? null : ToBsonDocument(session, connectionDescription, readConcern);
         }
 
+        // Write commands (insert, update, delete) do not support readConcern.level; only afterClusterTime is allowed for causal consistency.
```
ajcvickers: What about read and write commands, like findAndModify? Do they also have no readConcern.level?

sanych-sun: Yes, FindAndModify are write commands. I'll update the comment to avoid the confusion.

ajcvickers (approved):

PR #2007, opened 2026-05-22
CSHARP-5840: Proper XUnit shared test fixture for Atlas tests

ajcvickers:
Replaces the suite's dependence on pre-configured Atlas sample databases and search indexes. The new AtlasSearchFixture (collection-scoped) owns a single client, an EventCapturer, and a fixture-managed atlas_search_<guid> database. Lazy seeders populate historical_documents, movies, embedded_movies, airbnb_listings, test_classes, binary_vector_items, auto_embed_movies, two synonym source collections, and the existing return-scope directors collection, and create the ten search/vector indexes required by the tests (waiting for READY). Dispose drops the whole database.

The seed data is baked in as C# literals in AtlasSearchFixtureSeedData.cs (generated by tests/Tools/AtlasSeedExtractor from Atlas Local sample data once) so the test suite has zero runtime dependency on sample databases — only on an Atlas-search-capable cluster being reachable at ATLAS_SEARCH_URI. Embeddings are preserved as real Voyage-AI float32 vectors via BinaryVectorFloat32.

AtlasSearchTests, VectorSearchTests, AutoEmbedVectorSearchTests, and AtlasSearchIndexManagementTests join via [Collection(AtlasSearchCollection.Name)] and pull their client / collections through the fixture. AtlasSearchTests loses its static __indexesCreated / __databaseUniquifier / __returnScopeTestCount machinery and the inline return-scope seed. AtlasSearchTestsUtils is obsoleted.

AutoEmbedVectorSearchTests now run for real (the previous CSHARP-5840 skip is gone): the auto-embed collection and index live in the fixture so the Voyage AI embedding work happens once per test session instead of three times, the corpus is reduced to 4 movies (2 pass the runtime+year filter, 2 don't), and the ctor's unbounded wait loop is replaced with a 15-minute deadline that surfaces a clear error mentioning VOYAGE_API_KEY. Gated on a new AUTO_EMBEDDING_TESTS_ENABLED env var so contributors without a Voyage AI key can still run the rest of the Atlas Search suite. When enabled, the Atlas Local container must have VOYAGE_API_KEY provisioned via the Atlas UI's Model API Keys page (the key must authenticate against ai.mongodb.com, not api.voyageai.com). docker-compose.atlas-local.yml is included as a convenience launcher.

Count- and ranking-dependent assertions are recalibrated for the lean corpus: phrase-match counts drop from 108 to 1, geo-result counts from 25 to membership checks, and vector-search top-K assertions switch from order-sensitive ShouldBeEquivalentTo to order-insensitive BeEquivalentTo. The Pinocchio→Oz swap in VectorSearch top-5 is documented in the test where it lives. The Rerank test guards on AtlasSearchFixture.IsRerankSupported, which probes the deployment on first use.

Validated against Atlas Local: 77/77 AtlasSearch* tests pass, 36/36 AtlasSearchIndexManagement tests pass, 3/3 AutoEmbedVectorSearchTests pass (with a valid VOYAGE_API_KEY), 74 tests skip cleanly when env vars are unset, and no atlas_search_* databases remain after the suite exits.


tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,865 @@
...
+
+            Client = new MongoClient(settings);
+            DatabaseName = "atlas_search_" + Guid.NewGuid().ToString("N");
+        }
+
+        public IMongoClient Client { get; }
+
+        public EventCapturer EventCapturer { get; }
+
+        public string DatabaseName { get; }
```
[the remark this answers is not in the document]
ajcvickers: Copilot is wrong; it is assuming NRTs are being used.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,865 @@
...
+            }
+            catch (Exception ex)
+            {
+                // best-effort cleanup; never block tear-down, but surface the failure so
+                // operators can spot leaked test databases on the cluster.
+                Console.Error.WriteLine(
+                    $"AtlasSearchFixture: failed to drop database '{DatabaseName}': {ex}");
+            }
+
+            Client.Dispose();
```
sanych-sun: In the current (v3) driver's version MongoClient dispose do not unregister cluster, which means SDAM, connection pool and everything else will be active. If we want to clean up all resources, we must follow the [MongoClient Disposables guide](https://www.mongodb.com/docs/drivers/csharp/current/connect/mongoclient/#manage-mongoclient-disposables).

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,865 @@
...
+        /// <summary>
+        /// Lazily probes for $rerank support on first access. Tests guard themselves on this so the
+        /// suite continues to work on Atlas deployments that don't yet ship the rerank model.
+        /// </summary>
+        public bool IsRerankSupported
+        {
+            get
+            {
+                EnsureMoviesInitialized();
+                if (_isRerankSupported is bool cached)
```
sanych-sun: I would suggest to use `_isRerankSupported.HasValue` instead to clearly show what exactly we are checking.

sanych-sun: BTW, it looks like `Lazy<>` could be better here.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,865 @@
...
+                              "date":     [{ "type": "date" },   { "type": "dateFacet" }],
+                              "comments": { "type": "embeddedDocuments", "dynamic": true }
+                            }
+                          }
+                        }
+                        """))
+                });
+
+                _historicalInitialized = true;
+                EventCapturer?.Clear();
```
sanych-sun: Each `EnsureXXXInitialized` method ends with `EventCapturer?.Clear();`. Is it safe to clear the event capturer in such a way? User's code does not control when `EnsureXXXInitialized` method is being executed, so if the test case will looks like:
```
var collection1 = Fixture.Collection1;
collection1.Aggregate(...) // do something useful with the collection - but these events could be lost because of the next line.
var collection2 = Fixture.Collection2;
```

Should we have a single `EnsureTestDataSet` method to initialize all collections?

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
```diff
@@ -29,16 +30,29 @@ namespace MongoDB.Driver.Tests.Search
...
     {
+        private readonly AtlasSearchFixture _fixture;
         private readonly IMongoClient _mongoClient;
 
-        public VectorSearchTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
+        public VectorSearchTests(AtlasSearchFixture fixture, ITestOutputHelper testOutputHelper) : base(testOutputHelper)
         {
-            (_mongoClient, _) = AtlasSearchTestsUtils.CreateAtlasSearchMongoClient();
+            RequireEnvironment.Check().EnvironmentVariable("ATLAS_SEARCH_TESTS_ENABLED");
+            RequireEnvironment.Check().EnvironmentVariable("ATLAS_SEARCH_URI");
```
sanych-sun, line 42: It looks like we are always check ATLAS_SEARCH_TESTS_ENABLED and ATLAS_SEARCH_URI together. Do we really need ATLAS_SEARCH_TESTS_ENABLED then?

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
```diff
@@ -29,16 +30,29 @@ namespace MongoDB.Driver.Tests.Search
...
+            _mongoClient = fixture.Client;
+
+            // The fixture's EventCapturer is shared across every test class in the
+            // collection. We don't assert on it here, but clear it so captured aggregate
+            // events from VectorSearchTests don't accumulate without bound.
+            _fixture.EventCapturer?.Clear();
         }
 
-        protected override void DisposeInternal() => _mongoClient.Dispose();
+        protected override void DisposeInternal()
```
sanych-sun: Do we need this empty override then?

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
```diff
@@ -71,13 +85,15 @@ private void BinaryVectorSearch<T>(BinaryVector<T> binaryVector, string fieldNam
         [Fact]
         public void VectorSearch()
         {
+            // Lean seed: Pinocchio's embedding loses to Oz on this query vector in our 19-doc set.
```
sanych-sun: I do not think we need this comment in source code. It makes sense to have it on the PR to explain the code difference, but in source code it makes no sense.

sanych-sun (changes requested):

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: Removed dispose, since we generally don't current dispose clients, so this matches that behavior.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: Good catch. This isn't an issue currently, since all tests are in a single collection, and XUnit doesn't run tests in the same collection in parallel. However, fixing this now means that if we move tests into separate collections and run in parallel (for perf), then things will not break.

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
ajcvickers: I don't think so, but I'm not sure why we have all the environment variables we have, so I have been cautious about consolidating. If we're okay with gating on the URI being set, then we can do this.

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
ajcvickers: Removed.

tests/MongoDB.Driver.Tests/Search/VectorSearchTests.cs
ajcvickers: Removed.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,932 @@
...
+        /// <summary>
+        /// Lazily probes for $rerank support on first access. Tests guard themselves on this so the
+        /// suite continues to work on Atlas deployments that don't yet ship the rerank model.
+        /// </summary>
+        public bool IsRerankSupported
+        {
+            get
+            {
+                EnsureMoviesInitialized();
+                if (_isRerankSupported is { } cached)
```
sanych-sun, line 117: I would say the following code would be more self-descriptive:
```
if (_isRerankSupported.HasValue)
{
    return _isRerankSupported.Value;
}
```

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,932 @@
...
+
+        private void EnsureEmbeddedMoviesInitialized()
+        {
+            if (_embeddedMoviesInitialized) return;
+            lock (_initLock)
+            {
+                if (_embeddedMoviesInitialized) return;
+
+                var collection = Database.GetCollection<BsonDocument>(EmbeddedMoviesName);
+                if (!collection.AsQueryable().Any())
```
sanych-sun, line 328: Does it makes sense trying to preserve previously existed data. What if the collection is filled with something else, or what if we decide to change the test data?

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
```diff
@@ -0,0 +1,932 @@
...
+                    // AddEventSubscriber before the first IsRerankSupported probe fires.
+                    ClearCapturedEvents();
+                    return _isRerankSupported.Value;
+                }
+            }
+        }
+
+        // ---- Typed collection accessors ----
+
+        public IMongoCollection<T> GetHistoricalDocumentsCollection<T>()
```
sanych-sun, line 140: Why this methods has type parameter? Do we expect different tests to use different models?

tests/Tools/AtlasSeedExtractor/AtlasSeedExtractor.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
sanych-sun: I do agree that this tool looks useful, but do we really need it in the Driver's repo? Are we expecting to use it frequently?

Should we have a separate repo for tools like this? Somewhere under https://github.com/mongodb-labs .

I'm not insisting, just asking.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: I disagree. It's just older.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: Will take a look.

tests/Tools/AtlasSeedExtractor/AtlasSeedExtractor.csproj
ajcvickers: I'll just remove it.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: Lazy<> is not a good choice here because:

- Exception handling is different. Lazy will keep throwing; current code does not.
- EnsureAutoEmbedInitialized needs to run an AUTO_EMBEDDING_TESTS_ENABLED fail-fast check
- IsRerankSupported depends on EnsureMoviesInitialized() running first.
- Lazy<bool> purely-for-side-effects is an awkward idiom.
- Independent per-Lazy locks vs. the shared _initLock makes the sharing logic easier to grok.

However, I did some refactoring to reduce some of the duplication.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: It's so that the database name can be pinned during debugging, and the test can be run multiple times without needing to re-seed every time, which slows down the inter-loop.

tests/MongoDB.Driver.Tests/Search/AtlasSearchFixture.cs
ajcvickers: It's currently used with HistoricalDocument and HistoricalDocumentWithCommentsOnly.

sanych-sun (approved):
LGTM

PR #2012, opened 2026-05-26
CSHARP-6043: NullReferenceException in IEnumerableSerializerBase.Serialize when projecting nullable array with conditional Select

papafe:
The NRE happened during LINQ3 query translation when the translator serialized the null constant from the ternary's true branch through a `IEnumerableSerializerBase.Serialize` that didn't expect null.

The BSON-layer `EnumerableSerializerBase` already handles null but the LINQ path uses its own dedicated `IEnumerableSerializerBase` and that one was missing the null guards. So the fix is to bring the LINQ path serializer in par with the BSON layer one.


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Serializers/IEnumerableSerializerBaseTests.cs
```diff
@@ -101,6 +102,37 @@ public void GetHashCode_should_return_zero()
...
+            var document = new BsonDocument("x", BsonNull.Value);
+            using var reader = new BsonDocumentReader(document);
+            reader.ReadStartDocument();
+            reader.ReadName("x");
+            var context = BsonDeserializationContext.CreateRoot(reader);
+
+            var result = serializer.Deserialize(context, new BsonDeserializationArgs { NominalType = typeof(IEnumerable<int>) });
+
+            result.Should().BeNull();
+        }
```
BorisDog, line 134: Would be good to have some limited end-to-end testing. By "e2e" here I mean serializing and deserializing a real POCO, using a real expression (without DB).
But not sure where we place such tests now, any ideas @sanych-sun?

BorisDog (approved):
LGTM + question about testing.

sanych-sun (approved):
LGTM

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Serializers/IEnumerableSerializerBaseTests.cs
sanych-sun: It probably should be a Theory with bytes on one side and expected value on the another. We might need such for all serializers.

PR #2013, opened 2026-05-26
CSHARP-6040: Fix StackOverflowException in convention CouldApply on self-referencing IEnumerable<T> types

papafe:


tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
```diff
@@ -508,6 +515,17 @@ public void Convention_should_work_with_recursive_type()
             ConventionRegistry.Remove("objectRecursive");
         }
 
+        [Fact]
+        public void Convention_should_not_stack_overflow_with_self_referencing_enumerable_type()
+        {
+            var subject = new ObjectSerializerAllowedTypesConvention();
+            var memberMap = CreateMemberMap(c => c.SelfEnumerableProp);
+
+            var exception = Record.Exception(() => subject.Apply(memberMap));
```
BorisDog: No need to explicitly test for the lack of exception ( `exception.Should().BeNull()`), 
`subject.Apply(memberMap)` should be sufficient.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -90,14 +90,27 @@ IBsonSerializer Reconfigure(IBsonSerializer s)
                 => s.ValueType.IsEnum ? (s as IRepresentationConfigurable)?.WithRepresentation(_representation) : null;
 
             bool CouldApply(Type type)
```
BorisDog, line 92: nit: consider renaming to `IsApplicable` or 'IsTypeEnumConfigurable"  or similar

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
```diff
@@ -90,14 +90,27 @@ IBsonSerializer Reconfigure(IBsonSerializer s)
...
-                => type.IsEnum ||
-                   type.IsNullableEnum() ||
-                   (type.IsArray && CouldApply(type.GetElementType())) || // This is covered by IEnumerable<T>, but it's a good short-circuit
-                   type.GetInterfaces().Prepend(type).Any(
-                       i =>
-                           i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IEnumerable<>) && CouldApply(i.GetGenericArguments()[0]) || // IEnumerable<T>
-                           i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IDictionary<,>) && (CouldApply(i.GetGenericArguments()[0]) || CouldApply(i.GetGenericArguments()[1])) // IDictionary<TKey, TValue>
-                   );
+            {
+                var visited = new HashSet<Type>();
```
BorisDog, line 94: Double checking: this is not hot path, each type is resolved once?
So we are not introducing new allocation on a hot path.

BorisDog (changes requested):
Minor changes requested

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: Yes, this is done once per type.

src/MongoDB.Bson/Serialization/Conventions/EnumRepresentationConvention.cs
papafe: I think the name "Could" here is more descriptive, because this is mostly a pre-filter on types. If this returns true, it doesn't necessarily mean we're going to change the serializer.

tests/MongoDB.Bson.Tests/Serialization/Conventions/ObjectSerializerAllowedTypesConventionTests.cs
papafe: Corrected.

BorisDog (approved):
LGTM

PR #2014, opened 2026-05-27
CSHARP-5495: Support $concatArrays and $setUnion aggregation accumulators

papafe:


src/MongoDB.Driver/Linq/Linq3Implementation/SerializerFinders/SerializerFinderVisitMethodCall.cs
```diff
@@ -94,6 +94,7 @@ void DeduceMethodCallSerializers()
                 case "As": DeduceAsMethodSerializers(); break;
                 case "AsQueryable": DeduceAsQueryableMethodSerializers(); break;
                 case "Concat": DeduceConcatMethodSerializers(); break;
+                case "ConcatArrays": DeduceConcatArraysMethodSerializers(); break;
```
sanych-sun, line 105: Need unit-tests for these new supported methods.

papafe: Added SerializerFinder unit tests for those.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -470,6 +470,19 @@ public override AstNode VisitPickExpression(AstPickExpression node)
...
 
+            public override AstNode VisitReduceExpression(AstReduceExpression node)
+            {
+                // { $reduce : { input : { $map : { input : "$_elements", as : "x", in : f(x) } }, initialValue : [], in : { $concatArrays : ["$$value", "$$this"] } } }
+                // => { __agg0 : { $concatArrays : f(x => element) } } + "$__agg0"
+                if (IsSelectManyShapeOverElements(node, out var rewrittenArg))
+                {
+                    var accumulatorExpression = AstExpression.UnaryAccumulator(AstUnaryAccumulatorOperator.ConcatArrays, rewrittenArg);
+                    return CreateGetAccumulatorFieldExpression(accumulatorExpression);
+                }
```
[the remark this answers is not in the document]
papafe: It's done conditionally now

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -483,6 +496,16 @@ public override AstNode VisitUnaryExpression(AstUnaryExpression node)
...
 
+                // { $setUnion : { $reduce : { input : { $map : { input : "$_elements", as : "x", in : f(x) } }, initialValue : [], in : { $concatArrays : ["$$value", "$$this"] } } } }
+                // => { __agg0 : { $setUnion : f(x => element) } } + "$__agg0"
+                if (node.Operator == AstUnaryOperator.SetUnion &&
+                    node.Arg is AstReduceExpression setUnionReduceArg &&
+                    IsSelectManyShapeOverElements(setUnionReduceArg, out var setUnionArg))
+                {
+                    var accumulatorExpression = AstExpression.UnaryAccumulator(AstUnaryAccumulatorOperator.SetUnion, setUnionArg);
+                    return CreateGetAccumulatorFieldExpression(accumulatorExpression);
+                }
```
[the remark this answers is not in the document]
papafe: There is gating now.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -522,6 +559,43 @@ private bool IsMappedElementsField(AstExpression expression, out AstExpression r
...
+                    IsEmptyArrayConstant(node.InitialValue) &&
+                    IsConcatArraysOfValueAndThis(node.In))
+                {
+                    rewrittenArg = mappedArg;
+                    return true;
+                }
+
+                rewrittenArg = null;
+                return false;
+            }
```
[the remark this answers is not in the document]
papafe: Fixed.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -313,23 +320,28 @@ public string AddAccumulatorExpression(AstAccumulatorExpression value)
...
-            private AccumulatorMover(AccumulatorSet accumulator, AstExpression element)
+            private AccumulatorMover(AccumulatorSet accumulator, AstExpression element, ExpressionTranslationOptions translationOptions)
             {
                 _accumulators = accumulator;
                 _element = element;
+                _translationOptions = translationOptions;
             }
 
+            private bool ArrayAccumulatorsSupported =>
+                Feature.ConcatArraysAndSetUnionAccumulators.IsSupported((_translationOptions?.CompatibilityLevel).ToWireVersion());
```
sanych-sun, line 343: I know it is probably outside of this PR scope, but should we consider to change the `ToWireVersion`, to return min supported wire version for the `null` parameter? Because otherwise it's hardcoded to 4.0, which is 2 versions under our lowest supported version.

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
```diff
@@ -313,23 +320,28 @@ public string AddAccumulatorExpression(AstAccumulatorExpression value)
...
 
-            private AccumulatorMover(AccumulatorSet accumulator, AstExpression element)
+            private AccumulatorMover(AccumulatorSet accumulator, AstExpression element, ExpressionTranslationOptions translationOptions)
             {
                 _accumulators = accumulator;
                 _element = element;
+                _translationOptions = translationOptions;
             }
 
+            private bool ArrayAccumulatorsSupported =>
```
sanych-sun: Property name seems to be misleading a little. Especially when we are using it to decide if we can use setUnion.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -81,26 +89,292 @@ public void Queryable_SelectMany_with_index_should_work()
...
+    public void SelectMany_in_GroupBy_should_work()
+    {
+        var collection = Fixture.Collection;
+
+        var queryable = collection.AsQueryable()
+            .GroupBy(x => x.Cat)
+            .Select(g => new { Cat = g.Key, AllTags = g.SelectMany(x => x.Tags).ToList() });
+
+        var stages = Translate(collection, queryable);
+        if (__accumulatorsSupported)
```
sanych-sun: I'm confused now: as far as I understood we hide the new behavior under CompatibilityLevel, which user has to specify manually via TranslationOptions. I do not see how we set CompatibilityLevel in this test. But `__accumulatorsSupported` variable is initialized based on the actual wire version... So this test (and some other in this file) should fail on all servers that supports new features, because we expect the generated MQL to use modern features.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -81,26 +89,292 @@ public void Queryable_SelectMany_with_index_should_work()
...
+        results.Should().HaveCount(2);
+        results[0].Cat.Should().Be("A");
+        results[0].UniqueTags.Should().BeEquivalentTo(new[] { "x", "y", "z" });
+        results[1].Cat.Should().Be("B");
+        results[1].UniqueTags.Should().BeEquivalentTo(new[] { "y", "z" });
+    }
+
+    [Theory]
+    [InlineData(ServerVersion.Server80)]
+    [InlineData(ServerVersion.Server81)]
```
sanych-sun, line 167: And here I'm confused again: If we run the test on MongoDB 5.0 and set compatibility level to 8.0 - should this test fail with some server side error? Should we skip the test on older versions?

sanych-sun (changes requested):

src/MongoDB.Driver/Linq/Linq3Implementation/Ast/Optimizers/AstGroupingPipelineOptimizer.cs
papafe: I agree, changed it.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
papafe: The `CompatibilityLevel` is set globally by the test harness. `DriverTestConfiguration.GetTranslationOptions()` sets the client's default `TranslationOptions.CompatibilityLevel` to the connected server's version (`CoreTestConfiguration.MaxWireVersion.ToServerVersion()`). 
So in this test we're essentially verifying that independently from the server we're connected to we create the right MQL and get the right results.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
papafe: This test is more of a regression check. In this case we're testing that if we have a `SelectMany` with index, we do not convert it to an accumulator, otherwise the index information will be unbound and rejected by the server. 
In this case we want to be sure that independently from the compatibility level, we do not use the new accumulators (that were added in version 8.1). Since we use the "old" MQL, this is compatible on all versions.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -81,26 +89,292 @@ public void Queryable_SelectMany_with_index_should_work()
...
+        var results = aggregate.ToList();
+        results.Should().HaveCount(2);
+        results[0].UniqueTags.Should().BeEquivalentTo(new[] { "x", "y", "z" });
+        results[1].UniqueTags.Should().BeEquivalentTo(new[] { "y", "z" });
+    }
+
+    [Theory]
+    [InlineData(ServerVersion.Server81, true)]
+    [InlineData(ServerVersion.Server80, false)]
+    public void SelectMany_in_GroupBy_should_emit_accumulator_only_when_CompatibilityLevel_supports_it(ServerVersion compatibilityLevel, bool expectAccumulator)
```
papafe, line 316: Regarding this test and the following... These could be possibly considered superfluous, but they test that the MQL is actually independent from the server version but only depends on the compatibility leve.

sanych-sun: So this test actually does not make all the way to the database? I would say it's unusual... I would prefer to actually execute the request and assert MQL based on the expected version, and skip test if we are trying to set the compatibility level to the version which is above the actual server version.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
sanych-sun: If we set compatibility level to ServerVersion.Server81 shouldn't we use the new functionality?

papafe: No, if we have `SelectMany`  with index (`g.SelectMany((x, i) =>  ...`) we can't use the new accumulator, this is what the test is verifying. If we have a `SelectMany` without index (`g.SelectMany( x => ...`), then we can use the new accumulator.

tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/SelectManyMethodToAggregationExpressionTranslatorTests.cs
papafe: Yes, they do not go all the way to the DB, because setting up the compatibility level to an arbitrary value could generate an MQL that is not valid (for example compatibility level 8.1 on a 5.0 server). 
If you think about a matrix of server version (less than 8.1, from 8.1) and compatibility level (server80, server81), all the other tests here are checking only the "diagonal", so when the compatibility level is the same as the server version. This test here instead is checking all the possible combinations, as we can also have compatibility level equal to server81 and the actual server version is 5.0. 
I would say this test is more for completeness, but I can understand if we want to remove it and keep the more direct behaviour. What do you think?

sanych-sun: We have a variants matrix and we run tests against many server versions. As I said: I would prefer to convert it into the integration test as other tests in this file. But guard it if we are trying to execute it against the server lower then compatibility level.

papafe: I've converted them to be integration tests with gating.

sanych-sun (approved):
LGTM

PR #2015, opened 2026-05-27
CSHARP-5746: Update $lookup prose test for MONGOCRYPT-793

adelinowona:
## Summary
- Adds Case 10 (`db.qe` joins `db.non_csfle_schema`) from the client-side-encryption prose tests; server 8.2 + libmongocrypt 1.17 (MONGOCRYPT-793 / SERVER-100260) now permit mixing a QE schema with a non-CSFLE JSON Schema validator in `$lookup`.
- Case 8's assertion already accepts both the old (`not supported`) and the new (`Cannot specify both encryptionInformation and csfleEncryptionSchemas...`) error substrings, so no change there.

JIRA: https://jira.mongodb.org/browse/CSHARP-5746
Spec: https://github.com/mongodb/specifications/commit/d41d48b90ef22aedae934dd22f3a21c65b5a13bc

### DIFF
--- a/specifications/client-side-encryption/prose-tests/etc/data/lookup/schema-non-csfle.json
+++ b/specifications/client-side-encryption/prose-tests/etc/data/lookup/schema-non-csfle.json
@@ -1,0 +1,3 @@
+{
+   "bsonType": "object"
+}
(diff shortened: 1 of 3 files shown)


papafe (approved):
LGTM

PR #2023, opened 2026-06-02
CSHARP-5895: Deprecate mongo-orchestration in favor of mongodb-runner - follow up fix

sanych-sun:
`mongodb-runner` produces MONGO_URL with the trailing slash, what collides with the url manipulation logic in `run-tests.sh`. Fix removes that slash if present.


BorisDog (approved):
LGTM

PR #2026, opened 2026-06-03
CSHARP-6066: FilteredMongoCollectionBase.FindOneAndUpdate throws NullReferenceException when options is null

flibustier7seas:
## Summary

Fixes https://jira.mongodb.org/browse/CSHARP-6066

All four `FindOneAndUpdate` / `FindOneAndUpdateAsync` overloads in `FilteredMongoCollectionBase` access `options.IsUpsert` without a null check, despite the options parameter defaulting to null. This causes a `NullReferenceException` when callers rely on the default value, e.g. `collection.OfType<Derived>().FindOneAndUpdate(filter, update).`

## Fix

Replace `options.IsUpsert` with `options?.IsUpsert ?? false` in all four overloads. This is consistent with how `IsUpsert` defaults to false in `FindOneAndUpdateOptions`, and matches the null-tolerant pattern used elsewhere in the same file (e.g. MapReduce does options = options ?? new ...).

## Testing

Added `FindOneAndUpdate_should_not_throw_when_options_is_null` unit test covering both sync and async paths.

### DIFF
--- a/tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
+++ b/tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
@@ -757,19 +757,14 @@ namespace MongoDB.Driver.Tests
             var subject = CreateSubject();
             var update = new BsonDocument("$set", new BsonDocument("x", 5));
 
-            var exception = Record.Exception(() =>
+            if (async)
             {
-                if (async)
-                {
-                    subject.FindOneAndUpdateAsync(_providedFilter, update, null, CancellationToken.None);
-                }
-                else
-                {
-                    subject.FindOneAndUpdate(_providedFilter, update, null, CancellationToken.None);
-                }
-            });
-
-            exception.Should().BeNull();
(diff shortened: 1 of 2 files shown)


tests/MongoDB.Driver.Tests/OfTypeMongoCollectionTests.cs
```diff
@@ -757,6 +757,29 @@ public void FindOneAndUpdate_should_include_the_filter(
...
+                {
+                    subject.FindOneAndUpdateAsync(_providedFilter, update, null, CancellationToken.None);
+                }
+                else
+                {
+                    subject.FindOneAndUpdate(_providedFilter, update, null, CancellationToken.None);
+                }
+            });
+
+            exception.Should().BeNull();
```
BorisDog: We shouldn't check for null exception.
@papafe do you think we can add this as a rule for review agents?

flibustier7seas: > We shouldn't check for null exception.

Fixed

papafe (approved):

papafe:
@flibustier7seas Thanks a lot for this PR!

PR #2033, opened 2026-06-10
CSHARP-5826: Add $top, $bottom, $topN and $bottomN expression operators

papafe:


tests/MongoDB.Driver.Tests/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslatorTests.cs
```diff
@@ -25,16 +25,23 @@
 using MongoDB.TestHelpers.XunitExtensions;
 using Xunit;
 
-namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Jira
+namespace MongoDB.Driver.Tests.Linq.Linq3Implementation.Translators.ExpressionToAggregationExpressionTranslators.MethodTranslators
```
papafe, line 28: I had already moved this test file here and converted it to use the new "classFixture" style, so I've decided to keep it. 
Also this is more efficient since the collection is not created for each test.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -36,94 +36,143 @@ public static TranslatedExpression Translate(TranslationContext context, MethodC
...
+                return TranslateExpressionOperator(context, expression, method, arguments);
+            }
+
             if (method.IsOneOf(EnumerableMethod.PickOverloads))
             {
-                var sourceExpression = arguments[0];
-                var sourceTranslation = ExpressionToAggregationExpressionTranslator.TranslateEnumerable(context, sourceExpression);
-                NestedAsQueryableHelper.EnsureQueryableMethodHasNestedAsQueryableSource(expression, sourceTranslation);
-                var itemSerializer = ArraySerializerHelper.GetItemSerializer(sourceTranslation.Serializer);
+                return TranslateAccumulator(context, expression, method, arguments);
```
papafe: It's the same code as before, just moved to its own method.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/MethodTranslators/PickMethodToAggregationExpressionTranslator.cs
```diff
@@ -36,94 +36,143 @@ public static TranslatedExpression Translate(TranslationContext context, MethodC
...
+            var itemSerializer = ArraySerializerHelper.GetItemSerializer(sourceTranslation.Serializer);
 
-                var ast = AstExpression.PickExpression(
-                    @operator,
-                    sourceAst,
-                    sortBy,
-                    @as,
-                    selectorAst,
-                    nTranslation?.Ast);
+            var sortByExpression = arguments[1];
```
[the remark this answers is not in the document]
papafe: Solved

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -114,6 +114,36 @@ public static IEnumerable<TResult> BottomN<TSource, TKey, TResult>(
...
         }
 
+        /// <summary>
+        /// Returns the bottom element of an array according to a sort order.
+        /// </summary>
+        /// <typeparam name="TSource">The source type.</typeparam>
+        /// <param name="source">The source array.</param>
+        /// <param name="sortBy">The sort order.</param>
+        /// <returns>The bottom element.</returns>
+        public static TSource Bottom<TSource>(
```
sanych-sun: I do not think we need this method. Most of LINQ users will use `Last` instead, in case sort needs to be changed - they will use `OrderBy` followed by `Last`.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -114,6 +114,36 @@ public static IEnumerable<TResult> BottomN<TSource, TKey, TResult>(
...
+
+        /// <summary>
+        /// Returns the bottom n elements of an array according to a sort order.
+        /// </summary>
+        /// <typeparam name="TSource">The source type.</typeparam>
+        /// <param name="source">The source array.</param>
+        /// <param name="sortBy">The sort order.</param>
+        /// <param name="n">The number of elements to return.</param>
+        /// <returns>The bottom n elements.</returns>
+        public static IEnumerable<TSource> BottomN<TSource>(
```
sanych-sun: Same here: `TakeLast`.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -1311,6 +1341,36 @@ public static IEnumerable<TResult> TopN<TSource, TKey, TResult>(
...
         }
 
+        /// <summary>
+        /// Returns the top element of an array according to a sort order.
+        /// </summary>
+        /// <typeparam name="TSource">The source type.</typeparam>
+        /// <param name="source">The source array.</param>
+        /// <param name="sortBy">The sort order.</param>
+        /// <returns>The top element.</returns>
+        public static TSource Top<TSource>(
```
sanych-sun: `Take` here is more discoverable for LINQ users.

src/MongoDB.Driver/Linq/MongoEnumerable.cs
```diff
@@ -1311,6 +1341,36 @@ public static IEnumerable<TResult> TopN<TSource, TKey, TResult>(
...
+        }
+
+        /// <summary>
+        /// Returns the top n elements of an array according to a sort order.
+        /// </summary>
+        /// <typeparam name="TSource">The source type.</typeparam>
+        /// <param name="source">The source array.</param>
+        /// <param name="sortBy">The sort order.</param>
+        /// <param name="n">The number of elements to return.</param>
+        /// <returns>The top n elements.</returns>
```
sanych-sun: `OrderBy` + `Take`

sanych-sun (commented):
In general I do not think we need the proposed methods for LINQ, as there are built-in methods with same or very close semantic.

We might need to implement TopN, BottomN as a [window accumulator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/topN/).

papafe:
@sanych-sun I agree with your comments, don't know why it didn't came to mind 🤦 
I looked at the tickets and it seems the new methods are mostly convenience, so we can keep using the more supported MQL for those cases.
I modified the PR to just add support for `TakeLast` since we did not support it before, and added tests in the way we had them before for the already supported `Take` methods. I wonder if it makes sense to add tests that also use `OrderBy` to show that we actually support the $top/$bottom methods implicitly, even though with a different MQL or it's enough to test the various parts separately.

sanych-sun (approved):
LGTM

PR #2042, opened 2026-06-18
CSHARP-6080: Pool clear with closeInUseConnections leaks raw ObjectDisposedException to the application instead of a retryable connection error

sanych-sun:


src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -430,6 +437,10 @@ public void Open(OperationContext operationContext)
                     _connection.Open(operationContext);
                     SetEffectiveGenerationIfRequired(_connection.Description);
                 }
+                catch (ObjectDisposedException ex) when (_closedByPoolClear)
```
BorisDog: What about `Reauthenticate`? If it's not needed, add a comment why?

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -762,7 +800,7 @@ void RemoveExpiredConnections(List<PooledConnection> connections, int? generatio
                                 continue;
                             }
 
-                            RemoveConnection(connection);
+                            RemoveConnection(connection, true);
```
BorisDog: Should `true` be only set by `RemoveExpiredConnections(_connectionsInUse, ..)` call?

Or alternatively, you could avoid `isCloseInUse` and inspect the (hypothetical) result of `UntrackInUseConnection`.
Also if `isCloseInUse` stays, `UntrackInUseConnection` should be affected  by that.

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
```diff
@@ -414,6 +415,12 @@ public ConnectionSettings Settings
                 get { return _connection.Settings; }
             }
 
+            public void CloseForPoolClear()
```
BorisDog: Let's have a more consistent naming, thought all the call chain. Maybe `CloseWhileInUse`?
And `_closedWhileInUse`. Also in other places.

BorisDog (changes requested):

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: There is no `ThrowIfDisposed` check in `Reauthenticate` in `BinaryConnection`. We can add the same exception handling there too for sameness, but it will be unreachable (at least for now).

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
BorisDog: I think we should add same ODE handling. Or tests for it. It would be easy to introduce a dispose in the future and miss this update.

BorisDog (changes requested):

src/MongoDB.Driver/Core/ConnectionPools/ExclusiveConnectionPool.Helpers.cs
sanych-sun: Done

BorisDog (approved):
LGTM

PR #2043, opened 2026-06-19
CSHARP-6092: Avoid buffer allocations in ObjectId.Parse

Kloizdena:
- Eliminate byte array allocation in ObjectId(string) ctor and ObjectId.TryParse(string, out ObjectId)
- Added new Span-based TryParse method

### DIFF
--- a/src/MongoDB.Bson/BsonUtils.cs
+++ b/src/MongoDB.Bson/BsonUtils.cs
@@ -73,18 +73,18 @@ namespace MongoDB.Bson
         }
 
         /// <summary>
-        /// Parses a hex char span into its equivalent byte array.
+        /// Parses a hex char span into its equivalent byte span.
         /// </summary>
         /// <param name="s">The hex char span to parse.</param>
-        /// <param name="bytes">The result span to fill with the byte equivalent of the hex string.</param>
-        public static void ParseHexString(ReadOnlySpan<char> s, Span<byte> bytes)
+        /// <param name="destination">The destination span to fill with the byte equivalent of the hex string.</param>
+        public static void ParseHexString(ReadOnlySpan<char> s, Span<byte> destination)
         {
             int expectedLength = GetByteLength(s.Length);
-            if (bytes.Length != expectedLength)
+            if (destination.Length != expectedLength)
             {
                 throw new FormatException($"Target should be {expectedLength} bytes long");
             }
-            if (!TryParseHexString(s, bytes))
+            if (!TryParseHexString(s, destination))
             {
                 throw new FormatException("String should contain only hexadecimal digits.");
             }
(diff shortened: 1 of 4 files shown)


src/MongoDB.Bson/BsonUtils.cs
```diff
@@ -212,14 +230,43 @@ public static DateTime ToUniversalTime(DateTime dateTime)
...
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Calculate the result byte length for the hex string length
+        /// </summary>
+        /// <param name="hexStringLength">The length of the hex string</param>
+        /// <returns>The required length to convert the hex string to bytes</returns>
+        internal static int GetByteLength(int hexStringLength)
```
BorisDog: styling:
`GetByteLength` can be private (and placed in alphabetical order within other private methods at the end of the class). 
Also no need for comment for non public methods.

src/MongoDB.Bson/BsonUtils.cs
```diff
@@ -212,14 +230,43 @@ public static DateTime ToUniversalTime(DateTime dateTime)
...
+            => (hexStringLength + 1) / 2;
+
+        /// <summary>
+        /// Tries to parse a hex char span to a byte span.
+        /// </summary>
+        /// <param name="s">The hex chars.</param>
+        /// <param name="bytes">A byte span.</param>
+        /// <returns>True if the hex char span was successfully parsed.</returns>
+        public static bool TryParseHexString(ReadOnlySpan<char> s, Span<byte> bytes)
```
BorisDog: We could have more descriptive naming for `bytes`, like `destination`.
I know that in other overloads the naming is suboptimal.

src/MongoDB.Bson/BsonUtils.cs
```diff
@@ -72,6 +72,24 @@ public static byte[] ParseHexString(string s)
             return bytes;
         }
 
+        /// <summary>
+        /// Parses a hex char span into its equivalent byte array.
+        /// </summary>
+        /// <param name="s">The hex char span to parse.</param>
+        /// <param name="bytes">The result span to fill with the byte equivalent of the hex string.</param>
+        public static void ParseHexString(ReadOnlySpan<char> s, Span<byte> bytes)
```
BorisDog: We need positive and negative tests for all new public methods (BsonUtilsTests.cs).

BorisDog (changes requested):
Looks good overall, minor requests + tests.

src/MongoDB.Bson/BsonUtils.cs
adelinowona: Yeah and if we rename here, we should also rename in the other new span overloads for consistency

src/MongoDB.Bson/BsonUtils.cs
```diff
@@ -72,6 +72,24 @@ public static byte[] ParseHexString(string s)
             return bytes;
         }
 
+        /// <summary>
+        /// Parses a hex char span into its equivalent byte array.
```
adelinowona: says "Parses a hex char span into its equivalent byte array" — but it writes into a caller-provided span, not a byte array. Small wording mismatch worth tidying.

adelinowona (approved):

BorisDog (approved):
LGTM, thanks @Kloizdena!

PR #2044, opened 2026-06-22
CSHARP-5693: Allow Instantiated MongoClients to Send Client Metadata On-Demand

adelinowona:


src/MongoDB.Driver/Core/Clusters/Cluster.cs
```diff
@@ -43,6 +44,7 @@ internal abstract class Cluster : IClusterInternal
         #endregion
 
         private readonly TimeSpan _minHeartbeatInterval = __minHeartbeatIntervalDefault;
+        private readonly ClientMetadata _clientMetadata;
```
sanych-sun, line 47: As far as I understood the cluster is owning ClientMetadata and it is cached inside the cluster only. In scope of the v4 we are going to work on mongoClient/cluster disposability - which means that disposing of all mongoClients that relates to the cluster should automatically unregister and dispose the cluster. What means that all metadata that was registered will be lost. We probably should take this into account now, when we are about to introduce the functionality.
Currently we could consider the following manual workflow:
- new client from a url was created
- a metadata was registered
- another client with the same url was created
- at this point both clients are reporting the same metadata
- dispose both clients
- unregister the cluster (this is the manual step which will be happening automatically in 4.0)
- create a new client with the same url.

At the current state - I believe the 3rd client will not send the metadata registered in step 2. For now it makes sense, or at least easy to explain. But when we will implement reference counting disposable cluster source - it will be hard to understand and hard to explain behavior.

Should we store a metadata in some static dictionary and use  ClusterKey as a key instead?

BorisDog: I think for now we can postpone solving this, and return to it when working on disposability.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public void Append(LibraryInfo libraryInfo)
+    {
+        Ensure.IsNotNull(libraryInfo, nameof(libraryInfo));
+        libraryInfo = Normalize(libraryInfo);
+
+        lock (_lock)
+        {
+            if (_libraryInfos.Contains(libraryInfo))
```
sanych-sun: Should we use `ConcurrentDictionary<LibraryInfo, bool>` instead to avoid exclusive lock? As per discussion with @BorisDog most likely EF will need to `AppendMetadata` on each `DbContext` creation, which most likely is per-request. We might want to reduce on locks here.

src/MongoDB.Driver/Core/Configuration/LibraryInfo.cs
```diff
@@ -33,6 +34,11 @@ public sealed class LibraryInfo : IEquatable<LibraryInfo>
         /// </summary>
         public string Version { get; }
 
+        /// <summary>
+        /// Gets the library platform.
```
sanych-sun: What do we expect as a platform here? Is this OS?

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        if (_libraryInfos.Count == 0)
+        {
+            return null;
+        }
+
+        // entries are normalized on insertion, so null is the only unset form
+        var name = string.Join("|", _libraryInfos.Select(libraryInfo => libraryInfo.Name));
+        var version = string.Join("|", _libraryInfos.Where(libraryInfo => libraryInfo.Version != null).Select(libraryInfo => libraryInfo.Version));
+        var platform = string.Join("|", _libraryInfos.Where(libraryInfo => libraryInfo.Platform != null).Select(libraryInfo => libraryInfo.Platform));
```
sanych-sun: If we filtering out nulls, will it make the result readable?:
"A|B|C;5.1|1.2;"

How do we know if 1.2 it's a version for `B` or `C`?
Is this prescribed by spec?

src/MongoDB.Driver/Core/Connections/ConnectionInitializer.cs
sanych-sun: I think we should remove this overload if possible, having such `ConnectionInitializer` we could end-up sending not all available metadata. Do we have a usage for this?

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -130,6 +131,25 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
...
+        /// <summary>
+        /// Appends the specified library information to the metadata that is sent to the server in the connection handshake.
+        /// The information is included in the handshake of connections opened after this call; existing connections are not affected.
+        /// </summary>
+        /// <param name="libraryInfo">The library information to append.</param>
+        public void AppendMetadata(LibraryInfo libraryInfo)
+        {
+            ThrowIfDisposed();
+            Ensure.IsNotNull(libraryInfo, nameof(libraryInfo));
+            if (ContainsSeparator(libraryInfo.Name) || ContainsSeparator(libraryInfo.Version) || ContainsSeparator(libraryInfo.Platform))
```
sanych-sun: I would expect such checks inside the `LibraryInfo` class instead.

src/MongoDB.Driver/MongoClient.cs
```diff
@@ -130,6 +131,25 @@ internal void ConfigureAutoEncryptionMessageEncoderSettings(MessageEncoderSettin
         }
 
         // public methods
+        /// <summary>
+        /// Appends the specified library information to the metadata that is sent to the server in the connection handshake.
+        /// The information is included in the handshake of connections opened after this call; existing connections are not affected.
+        /// </summary>
+        /// <param name="libraryInfo">The library information to append.</param>
+        public void AppendMetadata(LibraryInfo libraryInfo)
```
sanych-sun, line 139: I know we cannot change the interface just now, but we probably should create a jira ticket to add this method to `IMongoClient` interface in 4.0

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Clusters/Cluster.cs
adelinowona: Appended metadata is intentionally tied to the live cluster, not persisted across teardown. A rebuilt cluster reports its settings-derived metadata, and callers that append at runtime (e.g. EF per-DbContext) re-append on the new client — so nothing relies on resurrecting a disposed client's runtime appends. A static ClusterKey-keyed store would leak unboundedly and would "resurrect" metadata onto unrelated new clients.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
adelinowona: I believe the hot path here will be `GetClientDocument()`, which runs on every connection handshake — and it's already lock-free: it does a volatile read of the cached document and returns without taking the lock once warmed. The lock only guards `Append`, which the EF per-DbContext pattern will probably call with the same `LibraryInfo` each time, so after the first call every subsequent one is a dedup no-op over a tiny `List.Contains`. The critical section is a few nanoseconds and uncontended in that pattern, so I don't think the per-request lock is a real cost.

`ConcurrentDictionary<LibraryInfo, bool>` also wouldn't be a solution: its enumeration order is unspecified, and we rely on insertion order to build the `name|version|platform` lists deterministically (covered by `Append_should_accumulate_library_infos_in_order`). Switching to it would make the handshake string nondeterministic.

sanych-sun: `_libraryInfos.Contains(libraryInfo)` is under the lock as well. Should we switch to use `ReaderWriterLockSlim` then? So the `Contains` part could be done under shared read lock?

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
```diff
@@ -0,0 +1,102 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Configuration;
+using MongoDB.Driver.Core.Misc;
+
+namespace MongoDB.Driver.Core.Connections;
+
+/// <summary>
```
BorisDog: No need for such comments.

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -66,6 +66,11 @@ internal static BsonDocument CreateClientDocument(string applicationName, BsonDo
         {
             driverDocument = AppendLibraryInfoToDriverDocument(driverDocument, libraryInfo);
 
+            if (libraryInfo != null && !string.IsNullOrWhiteSpace(libraryInfo.Platform))
```
BorisDog: nit: `if (!string.IsNullOrWhiteSpace(libraryInfo?.Platform))`

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
adelinowona: The handshake spec only requires that appended values go to their respective fields delimited by "|" . It doesn't prescribe positional alignment or empty placeholders for libraries missing a version. Omitting unset versions also matches our original single-append behavior (AppendLibraryInfoToDriverDocument already appends name unconditionally and version only when present), so this keeps the two paths consistent.

I am not sure how the analytics side parses this info accurately but I'll check

src/MongoDB.Driver/MongoClient.cs
adelinowona: Will move it. It is also technically a behavior change for the existing `LibraryInfo(name, version)` but this is like a bug fix since `LibraryInfo` values containing "|" was malformed metadata anyways.

src/MongoDB.Driver/MongoClient.cs
BorisDog: And probably add an extension method for now?

src/MongoDB.Driver/Core/Configuration/LibraryInfo.cs
adelinowona: Not the OS — the handshake reports the operating system separately in its own `os` document. Platform is a free-form description of the runtime the wrapping library targets; it gets appended to the handshake's top-level platform field, which by default holds the .NET runtime description (`RuntimeInformation.FrameworkDescription`). Clarified the XML doc to say this.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
BorisDog: IMO, as Adelin mentioned, it's not crucial optimization, as it's not a hot path, happens once on startup. So we can prefer a simpler code in this case. But not harm in introducing a quick optimization, if we can agree on one that doesn't complicate things.

src/MongoDB.Driver/MongoClient.cs
adelinowona: Filed CSHARP-6099 to add AppendMetadata to IMongoClient in 4.0

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
BorisDog: Yeah that's a good question.
I'd expect "|" to be omitted for null values as well. 

It's discussed [here](https://docs.google.com/document/d/1iU-Wzed0KlsuFhIPnhY3B59dQ0bYV8lYdmhFGUYv8fw/edit?disco=AAAB9061ihs). I agree with Valentin that this should be clarified in the spec, and have proper tests.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
adelinowona: I think `ReaderWriterLockSlim` has materially higher per-acquire overhead than a plain Monitor/lock here. It only pays off when read critical sections are long and highly contended. Our read critical section is an O(n) Contains over a list of ~1–3 entries — a handful of nanoseconds. Wrapping that in RW-lock bookkeeping would likely make the common path slower, not faster but not sure on how to test this exactly. WDYT? @sanych-sun

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
adelinowona: I already refactored this area of the code.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
sanych-sun: > it's not a hot path, happens once on startup

I do not think it's actually true. If EF for example has to set this on `DbContext` creation (probably need input from @damieng here) it means this code will be executed once per request if we are talking about Web API application:

> A DbContext instance is designed to be used for a single unit-of-work. This means that the lifetime of a DbContext instance is usually very short.

https://learn.microsoft.com/en-us/ef/core/dbcontext-configuration/

I still think that `Contains` should be done outside of the **exclusive lock**. I suppose we can do it outside of lock at all, because underlying collection is append only. Even if a concurrent thread will change the collection it could be false-negative check, which we will mitigate by double-checking inside the exclusive lock while adding the library info.

sanych-sun: > a handful of nanoseconds.

It's true that Contains for a small lists are extremally fact, but having the **exclusive** lock - it makes all requests go through the single exclusive really short lock - which makes it a bottleneck for system performance.

src/MongoDB.Driver/Core/Configuration/LibraryInfo.cs
```diff
@@ -85,16 +105,33 @@ public bool Equals(LibraryInfo rhs)
...
+                .Hash(Version)
+                .Hash(Platform)
+                .GetHashCode();
 
         /// <inheritdoc/>
-        public override string ToString() => $"{Name}-{Version}";
+        public override string ToString() => Platform == null ? $"{Name}-{Version}" : $"{Name}-{Version}-{Platform}";
+
+        // private methods
+        private static string EnsureNoSeparator(string value, string paramName)
```
BorisDog, line 127: minor: can use `CallerArgumentExpression` here

src/MongoDB.Driver/Core/Clusters/ClusterFactory.cs
```diff
@@ -27,18 +28,20 @@ namespace MongoDB.Driver.Core.Clusters
...
         // fields
+        private readonly ClientMetadata _clientMetadata;
         private readonly IEventSubscriber _eventSubscriber;
         private readonly ILoggerFactory _loggerFactory;
         private readonly IClusterableServerFactory _serverFactory;
         private readonly ClusterSettings _settings;
 
         // constructors
-        public ClusterFactory(ClusterSettings settings, IClusterableServerFactory serverFactory, IEventSubscriber eventSubscriber, ILoggerFactory loggerFactory)
+        public ClusterFactory(ClusterSettings settings, IClusterableServerFactory serverFactory, IEventSubscriber eventSubscriber, ILoggerFactory loggerFactory, ClientMetadata clientMetadata = null)
```
BorisDog: In what cases `clientMetadata` can be null?

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -412,28 +414,51 @@ internal static BsonDocument RemoveOptionalFieldsUntilDocumentIsLessThan512Bytes
...
+                driverName = $"{driverName}|{libraryInfo.Name}";
+
+                if (!string.IsNullOrWhiteSpace(libraryInfo.Version))
+                {
+                    driverVersion = $"{driverVersion}|{libraryInfo.Version}";
+                }
             }
 
-            return new()
+            return new BsonDocument
```
BorisDog: minor: revert this change?

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -412,28 +414,51 @@ internal static BsonDocument RemoveOptionalFieldsUntilDocumentIsLessThan512Bytes
...
 
-            if (!string.IsNullOrWhiteSpace(libraryInfo.Version))
+            foreach (var libraryInfo in libraryInfos)
             {
-                driverVersion = $"{driverVersion}|{libraryInfo.Version}";
+                driverName = $"{driverName}|{libraryInfo.Name}";
+
+                if (!string.IsNullOrWhiteSpace(libraryInfo.Version))
+                {
+                    driverVersion = $"{driverVersion}|{libraryInfo.Version}";
```
BorisDog, line 433: Following up on this [comment](https://github.com/mongodb/mongo-csharp-driver/pull/2044#discussion_r3469376815)

Has a decision to not append a '|' for empty value been made?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedAppendMetadataOperation.cs
```diff
@@ -0,0 +1,100 @@
...
+* limitations under the License.
+*/
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using MongoDB.Bson;
+using MongoDB.Driver.Core.Configuration;
+
+namespace MongoDB.Driver.Tests.UnifiedTestOperations
```
BorisDog: nit: filescoped namespaces in new files?

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
BorisDog: Yes, I missed this scenario, append with same metadata can actually be on a hot path.
So we should try to avoid lock for duplicates check.

BorisDog (changes requested):

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
adelinowona: I don't know if a decision has been made, I can't access the document you shared in the thread.

src/MongoDB.Driver/Core/Configuration/LibraryInfo.cs
adelinowona: Nice idea, but I think it's not really minor change. `CallerArgumentExpressionAttribute` only exists in the BCL from .NET 6 — so on netstandard2.1 and net472 driver targets this wouldn't compile without adding an internal polyfill of the attribute (similar to `MongoDB.Shared/IsExternalInit.cs`), conditioned to the non-net6.0 TFMs to avoid colliding with the framework type. I'd rather not do this unless you really think it's necessary.

src/MongoDB.Driver/Core/Clusters/ClusterFactory.cs
adelinowona: Only in tests. In production, `ClusterFactory` always receives a non-null `clientMetadata` instance from `ClusterBuilder` and forwards it to the created cluster. The null default exists purely so test code that constructs factories/clusters directly — for SDAM/topology tests that have nothing to do with metadata — doesn't have to supply one. I guess a nullable that "can't be null in prod"  is a code smell? Lmk if you're not ok with that.

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
```diff
@@ -57,14 +58,15 @@ internal static void SetFileSystemProvider(IFileSystemProvider fileSystemProvide
...
         // private static methods
-        internal static BsonDocument CreateClientDocument(string applicationName, LibraryInfo libraryInfo)
+        internal static BsonDocument CreateClientDocument(string applicationName, IReadOnlyList<LibraryInfo> libraryInfos)
         {
-            return CreateClientDocument(applicationName, __driverDocument.Value, __osDocument.Value, __platformString.Value, __envDocument.Value, libraryInfo);
+            return CreateClientDocument(applicationName, __driverDocument.Value, __osDocument.Value, __platformString.Value, __envDocument.Value, libraryInfos);
         }
 
-        internal static BsonDocument CreateClientDocument(string applicationName, BsonDocument driverDocument, BsonDocument osDocument, string platformString, BsonDocument envDocument, LibraryInfo libraryInfo)
+        internal static BsonDocument CreateClientDocument(string applicationName, BsonDocument driverDocument, BsonDocument osDocument, string platformString, BsonDocument envDocument, IReadOnlyList<LibraryInfo> libraryInfos)
```
sanych-sun, line 66: It looks like this overload is only being used from the previous method + tests. Should we merge this 2 overloads into the single one that will be actually in use?

src/MongoDB.Driver/Core/Clusters/ClusterFactory.cs
BorisDog: I'd prefer not to have a code tailored to our testing here.

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
BorisDog: You should have access now.
Intuitively I am leaning towards emitting an pipe for null values, but let's align with that doc.

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            var current = _libraryInfos;
+            if (Array.IndexOf(current, libraryInfo) >= 0)
+            {
+                return;
+            }
+
+            var updated = new LibraryInfo[current.Length + 1];
+            Array.Copy(current, updated, current.Length);
+            updated[current.Length] = libraryInfo;
```
sanych-sun: _Minor_: I believe we can use fancy C# 12 feature here:
```
var updated = [..current, libraryInfo]
```

src/MongoDB.Driver/Core/Connections/ClientMetadata.cs
```diff
@@ -0,0 +1,90 @@
...
+            _libraryInfos = updated;
+            _clientDocument = null;
+        }
+    }
+
+    // empty strings are considered unset, so normalize them to null for deduplication
+    private static LibraryInfo Normalize(LibraryInfo libraryInfo)
```
sanych-sun: Should we move this to LibraryInfo too? Same way as we validate for | symbol, we can validate and adjust the values, instead of instantiating a new instance. Can save a little on allocations.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Connections/ClientDocumentHelper.cs
adelinowona: The 6-arg overload is a test seam rather than dead surface: `CreateClientDocument_with_args_should_return_expected_result` and `CreateClientDocument_should_append_library_info_platform_to_platform_field` use it to inject specific `driverDocument/osDocument/platformString/envDocument` so they're deterministic instead of depending on the real OS/driver/env values. I'd lean to leave it as-is unless you feel strongly.

BorisDog (changes requested):

src/MongoDB.Driver/Core/Configuration/LibraryInfo.cs
BorisDog: Ah, I always forget that.
I think we'll start using it in 4x (with polyfill). But can skip doing that in 3.x

sanych-sun (approved):
LGTM

BorisDog (changes requested):
Looks good, there is a remaining question about empty values.

BorisDog (approved):
LGTM!

PR #2050, opened 2026-06-23
CSHARP-5984: Add QE prefix+suffix GA and rename API to string

adelinowona:
## Summary

Adds the GA Queryable Encryption `prefix`/`suffix` string query types and renames the QE "text" API to "string", while keeping the experimental preview query types working on servers before 9.0.

- Introduce `StringOptions` and `EncryptionAlgorithm.String` as the GA names for the QE string API.
- Keep `TextOptions` and `EncryptionAlgorithm.TextPreview` as `[Obsolete]` deprecated aliases. `TextPreview` is translated to the `String` algorithm, so existing callers continue to work.
- Add the stable `prefix`/`suffix` query types (server 9.0+).
- Keep `prefixPreview`/`suffixPreview` (and `substringPreview`) working on servers before 9.0 — to be dropped later in DRIVERS-3548.
- Bump libmongocrypt to **1.19.1** (restores the preview query types that 1.19.0 removed, via MONGOCRYPT-937).

## Notes

- The `String` enum member is declared after the existing `TextPreview` alias so `TextPreview` keeps its released ordinal.
- The text explicit-encryption prose test is split by server-version requirement into GA, substring (preview), and prefix/suffix-preview variants. Cases 7-9 are GA-only and intentionally use the GA query types.

https://jira.mongodb.org/browse/CSHARP-5984 (DRIVERS-3321)

### DIFF
--- a/src/MongoDB.Driver.Encryption/EncryptOptions.cs
+++ b/src/MongoDB.Driver.Encryption/EncryptOptions.cs
@@ -96,8 +96,7 @@ namespace MongoDB.Driver.Encryption
     /// Prefix options.
     /// </summary>
     /// <remarks>
-    /// PrefixOptions is used with TextOptions and provides further options to support "prefixPreview" queries.
-    /// PrefixOptions is currently unstable API and subject to backwards breaking changes.
+    /// PrefixOptions is used with StringOptions (or the deprecated TextOptions) and provides further options to support "prefix" and "prefixPreview" queries.
     /// </remarks>
     public sealed class PrefixOptions
     {
@@ -141,7 +140,7 @@ namespace MongoDB.Driver.Encryption
     /// Substring options.
     /// </summary>
     /// <remarks>
-    /// SubstringOptions is used with TextOptions and provides further options to support "substringPreview" queries.
+    /// SubstringOptions is used with StringOptions (or the deprecated TextOptions) and provides further options to support "substringPreview" queries.
     /// SubstringOptions is currently unstable API and subject to backwards breaking changes.
     /// </remarks>
     public sealed class SubstringOptions
@@ -201,8 +200,7 @@ namespace MongoDB.Driver.Encryption
     /// Suffix options.
     /// </summary>
     /// <remarks>
-    /// SuffixOptions is used with TextOptions and provides further options to support "suffixPreview" queries.
-    /// SuffixOptions is currently unstable API and subject to backwards breaking changes.
+    /// SuffixOptions is used with StringOptions (or the deprecated TextOptions) and provides further options to support "suffix" and "suffixPreview" queries.
     /// </remarks>
     public sealed class SuffixOptions
     {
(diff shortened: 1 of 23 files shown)


src/MongoDB.Driver.Encryption/EncryptionAlgorithm.cs
```diff
@@ -59,10 +61,20 @@ public enum EncryptionAlgorithm
         /// TextPreview algorithm.
         /// </summary>
         /// <remarks>
-        /// The TextPreview algorithm is in preview and should be used for experimental workloads only.
-        /// To insert or query with a "TextPreview" encrypted payload, use a MongoClient configured with AutoEncryptionOptions.
+        /// This is a deprecated alias for <see cref="String"/> and is translated to "String". Use <see cref="String"/> instead.
+        /// </remarks>
+        [Obsolete("Use String instead.")]
```
BorisDog, line 66: I think we need to add a server version in the obsolete warnings.
Is String algorithm available on pre 9.0 servers? If not, we still need to support textPreview until `String` is released. Should we introduce obsolete warnings then?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -317,10 +377,13 @@ private static string ConvertEnumAlgorithmToString(EncryptionAlgorithm encryptio
             {
                 EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Deterministic => "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic",
                 EncryptionAlgorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Random => "AEAD_AES_256_CBC_HMAC_SHA_512-Random",
+#pragma warning disable CS0618 // TextPreview is a deprecated alias for String and is translated to "String".
+                EncryptionAlgorithm.TextPreview => EncryptionAlgorithm.String.ToString(),
```
BorisDog, line 381: Will this work with old API?

src/MongoDB.Driver.Encryption/EncryptOptions.cs
```diff
@@ -372,11 +438,48 @@ public EncryptOptions(
...
+            string algorithm,
+            StringOptions stringOptions,
+            Optional<string> alternateKeyName = default,
+            Optional<long?> contentionFactor = default,
+            Optional<Guid?> keyId = default,
+            Optional<string> queryType = default)
+        {
+            Ensure.IsNotNull(algorithm, nameof(algorithm));
+            Ensure.IsNotNull(stringOptions, nameof(stringOptions));
+            if (Enum.TryParse<EncryptionAlgorithm>(algorithm, out var @enumStr))
```
BorisDog: minor: is this `enumStr` (string) or actually an enum?

BorisDog (changes requested):

src/MongoDB.Driver.Encryption/EncryptionAlgorithm.cs
papafe: As far as I have understood this is not a problem for the server, as this value is just interpreted by libmongocrypt, so it shouldn't be an issue.

src/MongoDB.Driver/Core/Misc/Feature.cs
```diff
@@ -204,9 +204,20 @@ public class Feature
...
+
+    /// <summary>
+    /// Gets the csfle2 string (preview) query types feature (prefixPreview, suffixPreview, substringPreview).
+    /// </summary>
+    public static Feature Csfle2QEv2StringPreviewAlgorithm { get; } = new("csfle2Qev2StringPreviewAlgorithm", WireVersion.Server82);
+
     /// <summary>
     /// Gets the csfle2 textPreview algorithm feature.
     /// </summary>
+    [Obsolete("Use Csfle2QEv2StringAlgorithm instead.")]
```
papafe: Shouldn't this refer to `Csfle2QEv2StringPreviewAlgorithm` (the preview one)?

papafe (changes requested):

src/MongoDB.Driver.Encryption/EncryptionAlgorithm.cs
adelinowona: The algorithm name here is interpreted by libmongocrypt, not the server — and the bundled libmongocrypt (1.19.1, which this PR bumps to) speaks "String", not "textPreview". So passing the literal "textPreview" to our new bundled libmongocrypt will result in an error. 

So deprecating `TextPreview` in favor of `String` doesn't strand pre-9.0 users — `TextPreview` is just the old alias for the same thing (and as a literal it no longer works at all after the 1.19.1 bump). Not sure if it really matters to add a server version in the obsolete warnings here.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: Yes — this line is the backward-compat bridge for exactly that.  This allows old code that passes `EncryptionAlgorithm.TextPreview` — or the string "TextPreview" via the string overload — to be translated to "String" and keep working against the new bundled libmongocrypt; without this mapping it would fail with unsupported algorithm string error.

src/MongoDB.Driver.Encryption/EncryptOptions.cs
adelinowona: nope it's actually an enum, renamed the var

src/MongoDB.Driver/Core/Misc/Feature.cs
adelinowona: yep, fixed. Though this feature isn't being used anywhere in our internal code anyway. Ideally I will just remove it but since this is a public class, apicompat will complain. I created [CSHARP-6098](https://jira.mongodb.org/browse/CSHARP-6098) so we don't have to do this deprecation dance for this class anymore.

papafe: I totally agree with the ticket 🙏

src/MongoDB.Driver.Encryption/EncryptOptions.cs
BorisDog: Thanks!

BorisDog (approved):
LGTM

papafe (approved):
LGTM

PR #2053, opened 2026-06-30
CSHARP-6100: Clarify status and semver treatment for shipped IWM driver APIs

papafe:


src/MongoDB.Driver/MongoUrl.cs
```diff
@@ -249,6 +249,7 @@ public bool DirectConnection
         /// <summary>
         /// Gets a value indicating whether overload retargeting is enabled.
         /// </summary>
+        /// <remarks>This option works with MongoDB Server Version 9.0 and above.</remarks>
```
damieng: We should perhaps use "requires" rather than "works with" for clarity that it is a hard requirement.

papafe: Agree, changed wording.

damieng (approved):
It might be nice to indicate if these options are ignored prior to 9.0 or whether they will throw but otherwise LGTM.

PR #2057, opened 2026-07-02
CSHARP-6084: Enable and modernize the style tooling

papafe:
[CSHARP-6084](https://jira.mongodb.org/browse/CSHARP-6084)


src/MongoDB.Bson/MongoDB.Bson.csproj
```diff
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
```
sanych-sun: This PropertyGroup is empty now and can be removed.

src/MongoDB.Driver.Authentication.AWS/MongoDB.Driver.Authentication.AWS.csproj
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
```
sanych-sun: This PropertyGroup is empty now and can be removed.

src/MongoDB.Bson/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.cs]
+dotnet_diagnostic.CA1033.severity = none
+dotnet_diagnostic.CA1065.severity = none
```
sanych-sun: Do we have lots of violation of CA1065? Can we update the code instead? As the rule itself looks useful.

src/MongoDB.Driver.Encryption/.editorconfig
```diff
@@ -0,0 +1,6 @@
+[*.cs]
+dotnet_diagnostic.CA1061.severity = warning
+dotnet_diagnostic.CA1063.severity = warning
+dotnet_diagnostic.CA1065.severity = warning
```
sanych-sun: Do we have many violations of CA1061, CA1063 and CA1065? Should we try to fix the code and keep this as error?

src/MongoDB.Driver/Authentication/Oidc/OidcObtainCredentialsSaslStep.cs
```diff
@@ -43,7 +43,7 @@ public OidcObtainCredentialsSaslStep(IOidcCallbackAdapter oidcCallback, string p
             byte[] bytesReceivedFromServer,
             CancellationToken cancellationToken)
         {
-            var credentials = await _oidcCallback.GetCredentialsAsync(new OidcCallbackParameters(1, _principalName), cancellationToken);
+            var credentials = await _oidcCallback.GetCredentialsAsync(new OidcCallbackParameters(1, _principalName), cancellationToken).ConfigureAwait(false);
```
sanych-sun, line 46: +1

src/MongoDB.Driver/GridFS/MongoDB.Driver.GridFS.csproj
```diff
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
```
sanych-sun: This PropertyGroup is empty now and can be removed.

src/MongoDB.Driver/MongoDB.Driver.csproj
```diff
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
```
sanych-sun: This PropertyGroup is empty now and can be removed.

tests/AtlasConnectivity.Tests/AtlasConnectivity.Tests.csproj
```diff
@@ -2,7 +2,6 @@
   <Import Project="..\BuildProps\Tests.Build.props" />
 
   <PropertyGroup>
```
sanych-sun: PropertyGroup can be removed.

tests/MongoDB.Bson.TestHelpers/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.cs]
+dotnet_diagnostic.CA1033.severity = none
+dotnet_diagnostic.CA2237.severity = none
```
sanych-sun: Do we need CA2237 suppression? As far as I remember we removed ISerializable in 3.0

tests/MongoDB.Bson.Tests/MongoDB.Bson.Tests.csproj
```diff
@@ -2,7 +2,6 @@
   <Import Project="..\BuildProps\Tests.Build.props" />
 
   <PropertyGroup>
```
sanych-sun: PropertyGroup can be removed

tests/MongoDB.Bson.TestHelpers/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.cs]
```
sanych-sun: Why do we need .editorconfig in some tests project? Can we have a single file in the `tests` folder instead?

tests/MongoDB.Driver.Examples/MongoDB.Driver.Examples.csproj
```diff
@@ -2,7 +2,6 @@
   <Import Project="..\BuildProps\Tests.Build.props" />
 
   <PropertyGroup>
```
sanych-sun: Empty PropertyGroup

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
```diff
@@ -2,7 +2,6 @@
   <Import Project="..\BuildProps\Tests.Build.props" />
 
   <PropertyGroup>
```
sanych-sun: Empty PropertyGroup

.editorconfig
```diff
@@ -157,8 +157,27 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
 csharp_space_between_parentheses = false
 csharp_space_between_square_brackets = false
 
-# Analyzers
-dotnet_code_quality.ca1802.api_surface = private, internal
+# Code quality (CA) analyzer severities
+dotnet_diagnostic.CA1001.severity = warning
```
sanych-sun, line 161: CA1001 - I believe it should be an error, with suppression in each case with explanatorily comment why it is suppressed. Do we have many violations in our code?

.editorconfig
```diff
@@ -157,8 +157,27 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
+dotnet_diagnostic.CA1016.severity = warning
+dotnet_diagnostic.CA1033.severity = warning
+dotnet_diagnostic.CA1060.severity = warning
+dotnet_diagnostic.CA1065.severity = suggestion
```
sanych-sun: CA1065 - better to keep as error.

.editorconfig
```diff
@@ -157,8 +157,27 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
+dotnet_diagnostic.CA1401.severity = warning
+dotnet_diagnostic.CA1821.severity = warning
+dotnet_diagnostic.CA2002.severity = warning
+dotnet_diagnostic.CA2007.severity = warning
+dotnet_diagnostic.CA2100.severity = warning
```
sanych-sun: CA2100 looks irrelevant to our project

.editorconfig
```diff
@@ -157,8 +157,27 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
+dotnet_diagnostic.CA2101.severity = warning
+dotnet_diagnostic.CA2200.severity = warning
+dotnet_diagnostic.CA2207.severity = warning
+dotnet_diagnostic.CA2213.severity = warning
```
sanych-sun, line 173: CA2213 - I believe should be error, with suppression and comment in each violation

.editorconfig
```diff
@@ -157,8 +157,27 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
+dotnet_diagnostic.CA2214.severity = warning
+dotnet_diagnostic.CA2216.severity = warning
+dotnet_diagnostic.CA2231.severity = warning
+dotnet_diagnostic.CA2237.severity = warning
```
sanych-sun: I believe CA2237 is irrelevant for our project.

sanych-sun (changes requested):
As far as I understood we decided to explicitly specify which rules we want to run for the code analysis. I believe we'd better to use `<AnalysisMode>Recommended</AnalysisMode>` and disable the rules we do not like instead. So if there is a new code analysis rule, it will be applied to our codebase without us manually enable it.

src/MongoDB.Bson/MongoDB.Bson.csproj
papafe: Removed all of them.

src/MongoDB.Driver.Authentication.AWS/MongoDB.Driver.Authentication.AWS.csproj
papafe: Removed.

src/MongoDB.Driver/GridFS/MongoDB.Driver.GridFS.csproj
papafe: Removed.

src/MongoDB.Driver/MongoDB.Driver.csproj
papafe: Removed.

tests/AtlasConnectivity.Tests/AtlasConnectivity.Tests.csproj
papafe: Removed.

tests/MongoDB.Driver.Examples/MongoDB.Driver.Examples.csproj
papafe: Removed.

tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj
papafe: Removed.

tests/MongoDB.Bson.TestHelpers/.editorconfig
papafe: Up to now this is mostly a "1-to-1" conversion with what we had so far. These 3 test project were using a different ruleset (`MongoDBLegacyTest.ruleset`) than the other test projects, so they needed a different `editorConfig`.
I've simplified this now by adding a scoped section to the tests `editorconfig`

papafe:
@sanych-sun I've fixed the empty `PropertyGroup`s.

Regarding your other questions. For now in this PR I tried only to move exactly the current rules that we have at the moment from ruleset/FxCop to editorConfig/default analyzers, without trying to reason about what should be done. 
This is also the reason why for now I put `AnalysisMode` to `None` instead of `Recommended`, so we get exactly the rules we had before. 
In my idea the target of this PR is just to modernize the tooling, without touching the code itself (as much as possible at least). We can definitely remove some irrelevant/obsolete rules as we go though, as these do not force us to change the code. With the move from `ruleset` to `editorconfig` I've already removed all the rules that do not exists anymore since they are obsolete/are covered by other modern rules (for example all the CAS rules, as CAS is a deprecated legacy technology now).

By the way, as a note. Regarding setting some rules as errors, that's something we've never done before, and instead preferred to set them as warnings and then enable `TreatWarningsAsError`. So far it's the same, given that we enable it everywhere, but it's worth taking a decision if we want to be more modular. 


For the fun of it, I've set `AnalysisMode` to `Recommended` and we got about 55 CA rules violations, with 16k warnings in `src` folder alone, that are treated as build errors. Most of them are due to:

Rule | Count | Rule content
-- | -- | --
CA1305 | 4,142 | Pass IFormatProvider/CultureInfo to every format/parse
CA1507 | 3,714 | Use nameof instead of string literals
CA1510 | 3,126 | Use ArgumentNullException.ThrowIfNull throw-helpers
CA1822 | 1,654 | Mark members static where they don't touch instance state
CA1859 | 768 | Use concrete types over interfaces for perf
CA1805 | 264 | Don't redundantly initialize to default

Besides that, this PR is still in draft, so there is still some stuff to do 😁

.editorconfig
papafe: Removed, together with CA2237, CA2229, CA2235

tests/MongoDB.Bson.TestHelpers/.editorconfig
papafe: Removed.

.editorconfig
papafe: Removed.

sanych-sun:
> For the fun of it, I've set AnalysisMode to Recommended and we got about 55 CA rules violations, with 16k warnings in src folder alone, that are treated as build errors.

Honestly I would suggest to go this way anyway: It's easy to add the global suppression just now (by changing the severity to `none` or `suggestion`), and create a Jira ticket to address the violations and enforce important rules (not necessarily all of the in the same task, we can do that based on impact/pririty). I believe it's more reliable way to evolve the product, rather then stick to the decade old rule set. Because code standards are evolving and we either should stick to some industry standards, or we should inspect the ruleset periodically.

@BorisDog WDYT?

src/MongoDB.Driver/Authentication/Oidc/OidcObtainCredentialsSaslStep.cs
BorisDog: How did this pass compilation beforehand?

BorisDog:
> > For the fun of it, I've set AnalysisMode to Recommended and we got about 55 CA rules violations, with 16k warnings in src folder alone, that are treated as build errors.
> 
> Honestly I would suggest to go this way anyway: It's easy to add the global suppression just now (by changing the severity to `none` or `suggestion`), and create a Jira ticket to address the violations and enforce important rules (not necessarily all of the in the same task, we can do that based on impact/pririty). I believe it's more reliable way to evolve the product, rather then stick to the decade old rule set. Because code standards are evolving and we either should stick to some industry standards, or we should inspect the ruleset periodically.
> 
> @BorisDog WDYT?

I am leaning towards trying working with `Recommended`.  We can always switch back to `None` if we don't like it. I don't have a preference if it's done in this PR or later.

papafe:
@sanych-sun I think I was not clear in my comment, I completely agree that we should not stick to decade old rulesets. 
My only point here is more about when we do it, because I agree we should adapt the modern set of rules. 

The plan here would be to have one PR where we apply the whole set of rules, both the `Recommended` and the rules we agree in our codestyle document. This PR is mostly for removing the old `FxCop` analyzers and `ruleset`, without touching the rules we currently follow. 
I already have a [ticket](https://jira.mongodb.org/browse/CSHARP-6090) for applying the new rules, and I can add a statement about enabling the `Recommended` analysis mode to make it clearer. I'd prefer to everything in one PR so we can try to reduce our work to only one commit, and it's easier to do a single "cleanup" of the rules in case there are inconsistencies/conflicts.

src/MongoDB.Driver/Authentication/Oidc/OidcObtainCredentialsSaslStep.cs
papafe: `FxCop` is quite old, so it didn't consider `ValueTask` (that we use here), only `Task`.

src/MongoDB.Driver.Encryption/.editorconfig
papafe: We have `TreatWarningsAsError` enabled, so this will be an error.

.editorconfig
papafe: We have `TreatWarningsAsError` enabled, so this will be an error in any case.

.editorconfig
papafe: Added as warning

src/MongoDB.Bson/.editorconfig
papafe: Removed, now it's a warning (inherited)

papafe:
@BorisDog @sanych-sun I think the PR is ready to review/merge now. 
I've tried to keep the same rules we had before as much as possible, minus some minor changes, including the ones suggested by Alex. 
One thing worth discussing in the team is when we decide for severity `warning` and when for `error`. At the moment we never use `error`, but we get the same effect with `warning` as we use `TreatWarningsAsErrors`. I suppose this is more of a "semantic" decision, more than a practical one. 
  Any direction we decide to go, I don't think it matters for this PR necessarily.

BorisDog (approved):
LGTM

src/MongoDB.Driver/Linq/Linq3Implementation/Misc/ReferenceEqualityComparer.cs
```diff
@@ -26,6 +26,7 @@ internal class ReferenceEqualityComparer<T> : IEqualityComparer<T>
 
         public bool Equals(T x, T y) => object.ReferenceEquals(x, y);
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1065:Do not raise exceptions in unexpected locations", Justification = "This comparer supports reference equality only and is never used for hashing.")]
         public int GetHashCode(T obj) => throw new InvalidOperationException();
```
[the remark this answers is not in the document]
papafe: Not necessary

sanych-sun (approved):
LGTM

PR #2061, opened 2026-07-06
CSHARP-3985: Support multiple SerializerRegistries 

papafe:
[JIRA ticket](https://jira.mongodb.org/browse/CSHARP-3985)

Introduces `IBsonSerializationDomain`, an interface bundling everything the static BsonSerializer class does today (registry, class maps, conventions, discriminators, defaults), so that callers can create non-global serialization configurations instead of one shared global state. `BsonSerializationDomain.Default` preserves today's behavior for backward compatibility.

This work does not touch the public API and it's only an initial step in getting away from global serialization. This work will need to be completed in following tickets/PRs once we decide the shape of the public API. There are some spots in the code tagged with `DOMAIN-API` that will need to be rivisited at a later time, and that is captured in [this ticket ](https://jira.mongodb.org/browse/CSHARP-6110).


ajcvickers:
I have the same concern here as I brought up in the meeting. `IBsonSerializationDomain` bakes in a mutable, register-at-any-time contract. That rules out a future "configure once, freeze, then read" model, where reads are barrier-free in stead of going through the concurrent cache's acquire reads on every lookup. This kind of build-then-freeze design makes sense to me because most serializers are discovered up-front. There can be a concurrent dictionary-based fallback to find serializers for unknown types, while leaving most accesses completely lock and barrier free.

ajcvickers:
I did an analysis of test coverage. I think there is quite a lot missing here. Specifically, the default path is well-covered, but the isolation invariant that justifies the whole PR is only tested very lightly. 

- **Negative isolation**: a registration in one domain is absent from `Default` and from a second domain (serializers, discriminators, conventions, id-generators).
- **Discriminator / known-type isolation**: polymorphic cross-talk across domains.
- **Concurrent domains**: parallel registration + lookup on two domains.
- **Bread-and-butter LINQ under a custom domain**: `Where`/`Select`/`GroupBy`/`OrderBy`/`OfType`, not just exotic operators.
- **Provider fallback path**: non-domain-aware providers/serializers resolving nested types.
- **Formatter byte-level test**: `ClientBulkWriteOpsSectionFormatter` wire output. 
- **Lifecycle**: dispose behavior of a custom domain.

<details>
<summary>Details</summary>
                                                                                                                                                                                                                                                                                                                   
### 1. Negative isolation assertions (highest priority)                                                                                                                                                                                                                                                          
  Flagged by **bson** and **spec-conformance**. Existing new tests (`BsonSerializationDomainTests`, `MultipleRegistriesProviderPathTests`) assert the *positive* direction — "a serializer/convention registered in domain X resolves within domain X" — but never the negative: that the same registration is     
  **absent** from `BsonSerializationDomain.Default` and from an independent second domain. Negative assertions are what actually prove isolation; the positive direction would pass even if everything still shared global state.                                                                                  
                                                                                                                                                                                                                                                                                                                   
  - `BsonSerializationDomainTests.cs:27` — add: register a serializer / discriminator / convention / id-generator in a custom domain, then assert `Default.LookupSerializer(...)` (etc.) does **not** return it, and vice-versa.                                                                                   
  - `MultipleRegistriesProviderPathTests.cs:307` — `RegisterDiscriminatorConvention_positive_path_under_custom_domain` should gain a paired assertion that `Default.LookupDiscriminatorConvention(...)` differs.                                                                                                   
                                                                                                                                                                                                                                                                                                                   
  All three reviewers who raised this **manually verified isolation currently holds** — so this is a regression-safety gap, not a known bug. Without it, a future change that reintroduces shared state would pass CI.                                                                                             
                                                                                                                                                                                                                                                                                                                   
  ### 2. Discriminator / known-type isolation                                                                                                                                                                                                                                                                      
  Flagged by **bson** (`BsonSerializationDomain.cs:389`). `IsTypeDiscriminated` / `RegisterDiscriminator` / `LookupActualType` use per-domain dictionaries, but nothing tests the classic polymorphism cross-talk failure: a discriminator or known-type registered in one domain leaking into another's           
  `LookupActualType`. This is the deserialization path most prone to silent `Default` fallback and the most security-adjacent (type resolution).                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                   
  ### 3. Concurrent domains                                                                                                                                                                                                                                                                                        
  Flagged by **bson** (`BsonSerializationDomain.cs:67`) and **spec-conformance**. Every domain owns its own `ConfigLock` and registries, and "parallel testing / concurrent deployments" is the stated design goal — yet all domain tests are single-threaded. Add a smoke test that mutates and reads two domains 
  from parallel threads and asserts no cross-contamination and no lock fault. (The async reviewer confirmed lock discipline was migrated faithfully, so this is about proving the *design goal*, not chasing a suspected race.)                                                                                    
                                                                                                                                                                                                                                                                                                                   
  ### 4. Bread-and-butter LINQ under a custom domain                                                                                                                                                                                                                                                               
  Flagged by **linq** (`MultipleRegistriesLinqTests.cs:33`). The new LINQ tests cover Inject, Ceiling/Floor, Pick/BottomN, `Mql.Constant`, and Concat-throw — but not the highest-traffic surfaces:                                                                                                                
                                                                                                                                                                                                                                                                                                                   
  - `Where(p => p.Name == "x")` / `Select` where the **constant** is serialized through a value-transforming domain serializer (the `SerializerFinderVisitConstant` path).                                                                                                                                         
  - `GroupBy`-key and `OrderBy` **field-name** resolution under a domain with a custom element-name convention.                                                                                                                                                                                                    
  - Nested-document member resolution.                                                                                                                                                                                                                                                                             
  - `OfType` / discriminator resolution under a non-default domain.                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                   
  The reviewer verified the constant path works via an offline repro (`"Mario"` → `"Mariotest"` in `$match` and `$project`), so again: coverage gap, not a defect.                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                   
  ### 5. Provider fallback path                                                                                                                                                                                                                                                                                    
  Flagged by **bson** (`BsonSerializationDomain.cs:892`) and touched by **spec-conformance**. Several providers (`DiscriminatedInterface…`, `Collections…`, `Primitive…`, `Attributed…`, `BsonObjectModel…`) are constructed without a domain reference; correctness depends on the registry (`IRegistryAware`) or 
  class-map path threading the domain in. Spec-conformance noted the parameterless `CustomAttributeSerializer` exercises *a* non-domain-aware path, but there is no test targeting a provider that itself **resolves nested types** without being registry-aware — the exact shape that would silently fall back to
  `Default`. Add a test with such a provider and assert nested resolution stays in-domain.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                   
  ### 6. Formatter byte-level test                                                                                                                                                                                                                                                                                 
  Flagged by **transport** (`ClientBulkWriteOpsSectionFormatter.cs`). No unit test asserts the formatter's wire-byte output for either the default or a custom-domain path — coverage is only indirect via integration/spec tests. Given this formatter writes the OP_MSG bulk-write payload (and the change here  
  fixed a latent static-registry bug), a direct byte-level regression test is warranted.                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                   
  ### 7. Domain lifecycle / dispose                                                                                                                                                                                                                                                                                
  Flagged by **async** (`BsonSerializationDomain.cs:874`) and **spec-conformance**. `BsonSerializationDomain : IDisposable` disposes `_configLock`, but nothing invokes `Dispose()` and no test covers dispose / use-after-dispose / the guarantee that the `Default` singleton is never disposed. If the          
  `IDisposable` stays, add a lifecycle test; if it's dropped (per the async reviewer's suggestion), no test is needed.                                                                                                                                                                                             

</details>

src/MongoDB.Bson/Serialization/Serializers/KeyValuePairSerializer.cs
ajcvickers, line 103: I don't think any of the constructors here will be used by the factory, meaning this will fall through to using the parameterless constructor. We should check all types are handled correctly by the factory, and add tests that will confirm this and make sure it doesn't break when constructor signatures are changed.

src/MongoDB.Driver/FieldValueSerializerHelper.cs
```diff
@@ -51,7 +52,7 @@ public static IBsonSerializer GetSerializerForValueType(IBsonSerializer fieldSer
             // serialize numeric values without converting them
             if (fieldType.IsNumeric() && valueType.IsNumeric())
             {
-                var valueSerializer = BsonSerializer.SerializerRegistry.GetSerializer(valueType);
+                var valueSerializer = StandardSerializers.GetSerializer(valueType);
```
ajcvickers: Two things about StandardSerializers. First, when do we decide it's ok to ignore the registered serializers and just use the standard regardless of configuration? Second, does StandardSerializers support all the types directed at it. For example, decimal128? There should be tests for all this to ensure the correct serialization is being used in each case.

ajcvickers (dismissed):

src/MongoDB.Driver/FieldValueSerializerHelper.cs
BorisDog: Yeah our test coverage needs some work.

It's not clear why in the current main branch `BsonSerializer.SerializerRegistry` is used instead of `serializerRegistry` parameter.  cc @sanych-sun maybe you have some context?

BorisDog:
> 3. Concurrent domains

I want to specifically highlight the Concurrent domains part in @ajcvickers testing suggestions.
I think even before this change, we lacked concurrency and stress testing in the serialization area. This change makes that coverage even more critical.

But I am not against adding the missing coverage in a follow up PR, so we can unblock other dependent tickets.

> have the same concern here as I brought up in the meeting. `IBsonSerializationDomain` bakes in a mutable, register-at-any-time contract. That rules out a future "configure once, freeze, then read" model, where reads are barrier-free in stead of going through the concurrent cache's acquire reads on every lookup. This kind of build-then-freeze design makes sense to me because most serializers are discovered up-front. There can be a concurrent dictionary-based fallback to find serializers for unknown types, while leaving most accesses completely lock and barrier free.

"configure once, freeze, then read" is our future goal, which should taken into account in [CSHARP-4924](https://jira.mongodb.org/browse/CSHARP-4924). IMO `IBsonSerializationDomain` is an interim step, and will split in the future to multiple interfaces/classes. That's why it's internal, and AFAIK there is not intention for it to be a public API. We decided on this gradual approach to have more digestible PRs, and minimize the risks in this highly sensitive area. @papafe please correct me if needed.

BorisDog (changes requested):
Looks good overall

src/MongoDB.Driver/FieldValueSerializerHelper.cs
sanych-sun: We have number of cases when we do not resolve serializer from the registry, when we work with something that has predefined data type and is not configurable in any way. For example if we have `model.StringField.Length` expression which is translated to `$strLenCP` operator, which always returns int32 response - so we skip on lookup and use `Int32Serializer` (see [here](https://github.com/mongodb/mongo-csharp-driver/blob/6e33dce5faea0cbd2f655df97308545fd6c6bfe8/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/PropertyTranslators/LengthPropertyToAggregationExpressionTranslator.cs#L34)).

So the `StandardSerializers` helper class was invented to support the similar cases, but when we have runtime type resolution. For example: if we translating arithmetic operators - the result serializer is one of the "primitive" types serializers, but which one we know only in runtime ([here](https://github.com/mongodb/mongo-csharp-driver/blob/6e33dce5faea0cbd2f655df97308545fd6c6bfe8/src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToAggregationExpressionTranslators/BinaryExpressionToAggregationExpressionTranslator.cs#L280)).

About test coverage, you do not have to ask me twice: I've always been asking for more unit-tests on this project.

sanych-sun:
I doubt if we can really make the serializer registry lock/barrier free. Because MongoDB Driver DOES NOT require to register a models upfront, and I do not think we are ready to make our users to register everything. We always were trying to make "zero configuration with sensible defaults" approach. We do have ClassMap which is a way to register the model upfront, but I'm not sure if it's the most used way to configure the serialization.

My understanding of "configure once, freeze, then read" - assumes that we let user do all necessary registration/configuration BEFORE we resolve any serializer. And once the configuration is done, the configuration is immutable (no new class maps could be registered, or conventions, or something else), but internal structures are not readonly, serializer registry will be discovering and caching new serializers as we go.

src/MongoDB.Driver/FieldValueSerializerHelper.cs
sanych-sun: However I'm not saying it's OK to use it in this particular case, need to investigate it deeper.

ajcvickers:
> assumes that we let user do all necessary registration/configuration BEFORE we resolve any serializer.

It assumes that the known serializer finder will have found most of the serializers needed, and so all those accesses can be lock-free. Only if we don't find the serializer to we go to a more expensive dynamic cache for models we don't know before hand.

However, this does bring up the question of how we discover models in the first place, how that is impacted by other configuration, and how we decide whether we can re-use the existing serializers or not. This was another thing I brought up in the meeting. Has anybody already investigated this, or should I start figuring out what currently happens and what will happen in the future?

src/MongoDB.Driver/FieldValueSerializerHelper.cs
BorisDog: The simpler version of this question is:
Previous code resolved the serializers from `BsonSerializer.SerializerRegistry`, which could be user configured. And have not used the provided `serializerRegistry`, which was equivalent to `BsonSerializer.SerializerRegistry`.
Is the `right` thing here to use the provided `serializerRegistry` instead of `StandardSerializers `? Note that current code is behavior change, and it's hard to access its  full impact.

src/MongoDB.Bson/Serialization/Serializers/ImpliedImplementationInterfaceSerializer.cs
```diff
@@ -74,15 +81,20 @@ public ImpliedImplementationInterfaceSerializer(IBsonSerializer<TImplementation>
         /// </summary>
         /// <param name="serializerRegistry">The serializer registry.</param>
         public ImpliedImplementationInterfaceSerializer(IBsonSerializerRegistry serializerRegistry)
-            : this(new Lazy<IBsonSerializer<TImplementation>>(() => serializerRegistry.GetSerializer<TImplementation>()))
+            : this((serializerRegistry as BsonSerializerRegistry)?.SerializationDomain)
         {
-            if (serializerRegistry == null)
+        }
```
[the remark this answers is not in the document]
papafe: Fixed

src/MongoDB.Driver/FieldValueSerializerHelper.cs
papafe: I looked into this. `StandardSerializers` has no `Decimal128` entry, but `IsNumeric()` allows it, so it would throw and this is a problem. 

On why main used `BsonSerializer.SerializerRegistry` instead of the `serializerRegistry` parameter... I suppose it could be due just to legacy.  That line was added in 2016, before the `serializerRegistry` was passed as a method and used in some of the others if/else branches. 

I think the fix here would be to pass `serializerRegistry.GetSerializer(valueType)` as it preserves main's behavior (including resolving Decimal128). 
As a side note, and that's something that we discussed long time ago probably, we should probably think about having a "standard domain", that can't be modified externally, and that can be used to retrieve the default serialization of primitives and such when needed.

src/MongoDB.Bson/Serialization/Attributes/BsonSerializerAttribute.cs
```diff
@@ -86,16 +98,28 @@ internal IBsonSerializer CreateSerializer(Type type)
...
-                return (IBsonSerializer)Activator.CreateInstance(closedSerializerType);
+                closedSerializerType = _serializerType.MakeGenericType(genericArguments);
             }
             else
             {
-                return (IBsonSerializer)Activator.CreateInstance(_serializerType);
+                closedSerializerType = _serializerType;
             }
+
+            var domainCtor = closedSerializerType.GetConstructor(
```
sanych-sun: Variable name is confusing (this is not Domain's ctor, but Serializer's), should it be something like "closedSerializerCtorWithDomain"

src/MongoDB.Bson/Serialization/Conventions/ConventionRegistryInstance.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <param name="type">The type.</param>
+        /// <returns>The conventions for that type.</returns>
+        public IConventionPack Lookup(Type type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException("type");
+            }
+
+            lock (_lock)
```
sanych-sun, line 52: Not sure if it in scope of this PR, but we might want to do a lookup outside of exclusive lock (we can use one of Concurrent collection, or do a trick when we have plain array which is never changed, but replaced with the new copy when need to be adjusted.)

src/MongoDB.Bson/Serialization/Conventions/IConventionRegistry.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace MongoDB.Bson.Serialization.Conventions
+{
+    internal interface IConventionRegistry
+    {
+        /// <summary>
+        /// Looks up the effective set of conventions that apply to a type.
+        /// </summary>
+        /// <param name="type">The type.</param>
+        /// <returns>The conventions for that type.</returns>
+        IConventionPack Lookup(Type type);
```
sanych-sun, line 27: `IConventionPack` has the only member - IEnumerable of IConvention. Do we really need this extra wrapper? May be returning list itself is enough?

src/MongoDB.Bson/Serialization/Conventions/IConventionRegistry.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <param name="name">The name.</param>
+        /// <param name="conventions">The conventions.</param>
+        /// <param name="filter">The filter.</param>
+        void Register(string name, IConventionPack conventions, Func<Type, bool> filter);
+
+        /// <summary>
+        /// Removes the conventions specified by the given name.
+        /// </summary>
+        /// <param name="name">The name.</param>
+        void Remove(string name);
```
sanych-sun, line 41: I would suggest not to have this method in the interface, or mark it as Obsolete if we do need it for now. The problem is: in the initial phase (when we do not separate configuration and usage read-only domain) removal of the convention could lead to very unpredictable serialization depending on order of operations. In the "ultimate goal" stage (when we will separate configuration and usage API) this method makes no sense at all (if somebody register the convention, and then remove it BEFORE the configuration is done - it means that convention was never used.)

src/MongoDB.Bson/Serialization/Conventions/StandardDiscriminatorConvention.cs
```diff
@@ -31,14 +31,24 @@ public abstract class StandardDiscriminatorConvention : IDiscriminatorConvention
...
+            : this(BsonSerializationDomain.Default, elementName)
         {
+        }
+
+        private protected StandardDiscriminatorConvention(IBsonSerializationDomain serializationDomain, string elementName)
+        {
+            if (serializationDomain == null)
+            {
+                throw new ArgumentException("Serialization domain cannot be null.", nameof(serializationDomain));
+            }
```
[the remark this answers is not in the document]
sanych-sun: +1, for `ArgumentNullException`

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
```diff
@@ -50,6 +50,18 @@ public static TValue Deserialize<TValue>(this IBsonSerializer<TValue> serializer
             return serializer.Deserialize(context, args);
         }
 
+        internal static IBsonSerializer GetSerializerForDerivedType(this IBsonSerializer serializer, Type derivedType)
```
sanych-sun: Method name is not correspond to what the body does. Using this method I can obtain ANY serializer from the same domain if serializer is domain-aware. We should either enforce derivedType is inherited from `serializer.ValueType` or rename the method to something like `GetSerializerForSameDomain`

src/MongoDB.Bson/Serialization/IBsonSerializationDomain.cs
```diff
@@ -0,0 +1,332 @@
...
+using System.Threading;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization.Conventions;
+
+namespace MongoDB.Bson.Serialization
+{
+    /// <summary>
+    /// Represents a BSON serialization domain.
+    /// </summary>
+    internal interface IBsonSerializationDomain
```
sanych-sun, line 27: (no action required in scope of this PR) This interface is huge and as it was discussed will need to be reworked in future.

BorisDog (approved):
LGTM!!!

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/ConcatMethodToPipelineTranslator.cs
```diff
@@ -43,6 +43,13 @@ secondQueryable.Provider is IMongoQueryProviderInternal secondProvider &&
                     secondProvider.CollectionNamespace is var secondCollectionNamespace &&
                     secondCollectionNamespace != null)
                 {
+                    if (secondProvider.SerializationDomain != context.SerializationDomain)
```
sanych-sun, line 46: I think such checks could be harmful, as I saw in code we sometimes fallback to SerializationDomain.Default - so it might be serializers from SerializationDomain.Default domain.

sanych-sun: Here and in other places too we might want to check if the serializers belong to the same domain, or "other" belong to Default domain.

src/MongoDB.Driver/MongoCollectionSettings.cs
```diff
@@ -116,9 +118,23 @@ public ReadPreference ReadPreference
...
+        public IBsonSerializerRegistry SerializerRegistry => SerializationDomain.SerializerRegistry;
+
+        /// <summary>
+        /// Gets or sets the serialization domain.
+        /// </summary>
+        internal IBsonSerializationDomain SerializationDomain
         {
-            get { return BsonSerializer.SerializerRegistry; }
+            get => _serializationDomain.Value ?? BsonSerializer.DefaultSerializationDomain;
+            set
```
sanych-sun: If we can set SerializationDomain per collection, how do we expected to make a Joins? I saw we enforcing that serializers should belong to the same domain, but if we configure non-default domains for 2 collections - does it mean we cannot make a Join?

sanych-sun (commented):
I've reviewed most of the changes (I've skipped tests for now). Posted some questions/comments. I do not block the PR on that comments, because I'll be away next week. Please investigate the comments and decide what the best way to move forward: either address them in this PR, or create a separate PR.

src/MongoDB.Bson/Serialization/Conventions/ConventionRegistryInstance.cs
BorisDog: Good suggestion, 
I'd recommend not to complicate this PR and a create follow up ticket.

src/MongoDB.Bson/Serialization/Conventions/IConventionRegistry.cs
BorisDog: IMO this PR should focus more on copying things as they are and fixing critical bugs, while further improvements would be easier to review in following PRs. If I am not mistaken, this interface is same as existing static `ConventionRegistry`.

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/ConcatMethodToPipelineTranslator.cs
BorisDog: Good point.
My 2 cents: 
We can mark this as a known limitation of custom serialization settings. We can also require same builder/settings for such collections in the future. Meaning that we better have this check and adapt it later, rather not to have it at all.

Also @papafe 
1. Do we need similar checks in Join/GroupJoin?
2. We need tests for these checks for all cases

src/MongoDB.Driver/MongoCollectionSettings.cs
BorisDog: Looks like it's for tests only? If so let's be clarify this in code.

src/MongoDB.Bson/Serialization/Conventions/IConventionRegistry.cs
papafe: Yes, all the interfaces are derived from existing API, so all modifications should be probably done in other PRs.

src/MongoDB.Bson/Serialization/Conventions/IConventionRegistry.cs
papafe: All the interfaces are derived from existing API, so all modifications should be probably done in other PRs. The work in this PR has been mostly mechanical, trying to not change the public API in any shape or form.

src/MongoDB.Bson/Serialization/IBsonSerializationDomain.cs
papafe: (yep)

src/MongoDB.Driver/Linq/Linq3Implementation/Translators/ExpressionToPipelineTranslators/ConcatMethodToPipelineTranslator.cs
papafe: So, I think that the check is worth keeping, as it would raise errors instead of having unexpected effects. I also think that we should remove the possibility of setting specific domains on db/collections and keep that only on client. This would avoid probably a long list of complications, while removing a feature of which we don't know the utility. At a later time we could go back to this if we want. 

Another thing that we should do is probably having a "Standard" domain. An internal domain that contains standard configuration and can't be modified. The "Default" domain is the one exposed to developers that can still be modified. 
The idea is that the standard domain will be compatible with all other domains, so there won't be any exception.

src/MongoDB.Driver/MongoCollectionSettings.cs
papafe: As said in the other comment, I think it makes sense that the domain can be set per-client, and not at the db/collection level.

src/MongoDB.Bson/Serialization/Attributes/BsonSerializerAttribute.cs
papafe: Fixed

src/MongoDB.Bson/Serialization/IBsonSerializerExtensions.cs
papafe: Fixed

src/MongoDB.Bson/Serialization/Serializers/KeyValuePairSerializer.cs
papafe: Fixed by adding an internal (`IBsonSerializationDomain`) ctor + a regression test. I also audited the other composite serializers reachable via the factory (Nullable, Tuple/ValueTuple, arrays, dictionaries, enumerables, Object), and they all have a domain or single-arg registry ctor, so KeyValuePair was the only one affected.

ajcvickers (commented):
I'm not blocking on this. But I honestly still don't see how this gets us much closer to the ultimate goal. Also, I still don't think this design works well for that goal. But I think this is further evidence that I just don't understand this vision that you are all working to. It feels like it's missing extremely important parts like the caching and re-use goals (i.e. the layering) which will fundamentally impact the way the code in this PR should work, and what should be in this layer.

I'm not in favor of waterfall development, but for me, without understanding what the final design is at least intended to be, I don't know how to review the building blocks.

ajcvickers:
Test coverage now looks good.

PR #2062, opened 2026-07-06
CSHARP-6100: Clarify status and semver treatment for shipped IWM driver APIs

papafe:


sanych-sun (approved):
LGTM

ajcvickers:
@papafe  Is this only ever going to be available on Atlas? Or will it also be shipped in community and Enterprise?

PR #2064, opened 2026-07-07
CSHARP-5696: Integrate AWS SDK v4 in MongoDB.Driver.Authentication.AWS

papafe:
https://jira.mongodb.org/browse/CSHARP-5696

Migrates `MongoDB.Driver.Authentication.AWS` to AWS SDK for .NET v4 (`AWSSDK.SecurityToken` 4.0.100.2), replacing the deprecated `FallbackCredentialsFactory` with `DefaultAWSCredentialsIdentityResolver`. Targeted for the next major release.

Supersedes #2054 (from @BossDarkReaper).


tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
```diff
@@ -64,105 +62,5 @@ public void Ecs_should_fill_AWS_CONTAINER_CREDENTIALS_RELATIVE_URI()
             var awsContainerUri = Environment.GetEnvironmentVariable("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI") ?? Environment.GetEnvironmentVariable("AWS_CONTAINER_CREDENTIALS_FULL_URI");
```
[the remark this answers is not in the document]
papafe: So... I'm not sure we need this test. For a couple of reasons:
1. We don't have those methods used here anymore. `FallbackCredentialsFactory.CredentialsGenerators` is obsolete and its closest candidate `AWSConfigs.AWSCredentialsGenerators` is write only, so we can't read the default handlers
2. This test was essentially verifying that we get the "expected" chain of handlers to retrieve AWS credentials. Should we worry  that the official SDK actually uses our "expected" chain of handlers?
3. We actually verify that the correct credentials are chosen in our CI matrix with our `run-aws-auth-test-with-...` tasks.

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
```diff
@@ -33,8 +35,8 @@ public AWSCredentials GetCredentials(CancellationToken cancellationToken)
             _lock.Wait(cancellationToken);
             try
             {
-                // returns cached credentials source immediately. Only if cached source unavailable, makes quite heavy steps
-                credentialsSource = FallbackCredentialsFactory.GetCredentials();
+                // resolving the credentials source walks the full provider chain, so cache it and reuse until reset
+                credentialsSource = _cachedCredentialsSource ??= DefaultAWSCredentialsIdentityResolver.GetCredentials(null);
```
papafe: Just FYI, this is very similar to what `FallbackCredentialsFactory.GetCredentials()` was doing internally ([link](https://github.com/aws/aws-sdk-net/blob/d72ac7c2ae061080e3f28c96f8b6f644e5a0d1b1/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs#L171))

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
```diff
@@ -64,105 +62,5 @@ public void Ecs_should_fill_AWS_CONTAINER_CREDENTIALS_RELATIVE_URI()
...
             (awsContainerUri != null).Should().Be(isEcs);
         }
-
```
papafe, line 67: I added the following as an answer to a Copilot comment, but I think it's worth to add it here as well.

So... I'm not sure we need this test. For a couple of reasons:
1. We don't have those methods used here anymore. `FallbackCredentialsFactory.CredentialsGenerators` is obsolete and its closest candidate `AWSConfigs.AWSCredentialsGenerators` is write only, so we can't read the default handlers.
2. This test was essentially verifying that we get the "expected" chain of handlers to retrieve AWS credentials. Should we worry  that the official SDK actually uses our "expected" chain of handlers?
3. We actually verify that the correct credentials are chosen in our CI matrix with our `run-aws-auth-test-with-...` tasks.

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
adelinowona: So `DefaultAWSCredentialsIdentityResolver` doesn't internally cache resolved credentials source?

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
adelinowona: It seemed we were testing third-party library internals so I am supporting dropping the test. Unless there is a very good reason for doing so. @sanych-sun thoughts?

adelinowona (changes requested):
You should also update the AGENTS.md files in the project for any reference to AWS SDK v3 stuff

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
papafe: Good catch! It actually does cache internally. `FallbackCredentialsFactory` never invalidated the result, while v4's `DefaultAWSCredentialsIdentityResolver` auto-invalidates on env-var/credentials-file changes and the source objects self-refresh on expiry. This means that the caching was actually redundant and I've made `ResetCache()` a no-op (v4 exposes no API to clear the resolver's cache anyway, and `AWSInstanceCredentialsSource.ResetCache() was already a no-op`). Also updated the AGENTS.md references

tests/MongoDB.Driver.Tests/Communication/Security/AwsAuthenticationTests.cs
papafe: @sanych-sun Do you agree with removing this test?

sanych-sun: I'm OK with dropping the test as long as there is no our code involved.

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
```diff
@@ -16,65 +16,38 @@
...
-                _lock.Release();
-            }
-
+            cancellationToken.ThrowIfCancellationRequested();
+            var credentialsSource = await DefaultAWSCredentialsIdentityResolver.GetCredentialsAsync(null).ConfigureAwait(false);
             var immutableCredentials = await credentialsSource.GetCredentialsAsync().ConfigureAwait(false);
             return CreateAWSCredentials(immutableCredentials);
         }
 
         public void ResetCache()
```
adelinowona: I suppose ResetCache is a no-op now and can be removed from the IAWSCredentialsSource interface.

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
```diff
@@ -16,65 +16,38 @@
...
-                // returns cached credentials source immediately. Only if cached source unavailable, makes quite heavy steps
-                credentialsSource = FallbackCredentialsFactory.GetCredentials();
-            }
-            finally
-            {
-                _lock.Release();
-            }
-
+            cancellationToken.ThrowIfCancellationRequested();
+            var credentialsSource = await DefaultAWSCredentialsIdentityResolver.GetCredentialsAsync(null).ConfigureAwait(false);
```
adelinowona, line 38: Also this probably doesn't have to block this PR any further (can be investigated in a separate ticket) but copilot had a comment about how our cancellationToken isn't passed to the AWS SDK. Upon inspecting the SDK code, it will use its own default cancellationToken of 35 seconds if we use the `DefaultAWSCredentialsIdentityResolver.GetCredentials/GetCredentialsAsync`  methods. There is also `ResolveIdentity(IClientConfig, CancellationToken) / ResolveIdentityAsync(...)` methods which are instance methods that do take a cancellationToken so if we really wanted to fix this and honor our cancellationToken, we could create our own `DefaultAWSCredentialsIdentityResolver` instance and call `ResolveIdentity(IClientConfig, CancellationToken) / ResolveIdentityAsync(...)` so we can pass our cancellationToken.

Also worth investigating if IAWSCredentialsSource needs to be disposable since its dispose currently does nothing.

adelinowona (changes requested):

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
papafe: Yep, this is what I was thinking about. This and the `Dispose` method.

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
papafe: I agree with this, I'll create a follow up ticket.

papafe: Created: https://jira.mongodb.org/browse/CSHARP-6145

src/MongoDB.Driver.Authentication.AWS/CredentialsSources/AWSFallbackCredentialsSource.cs
papafe: Done

adelinowona (approved):

PR #2068, opened 2026-07-09
CSHARP-6119: Disable apicompat validation for v4.0 development

sanych-sun:
validate-apicompat task was disabled for v4.0 development where we are expecting breaking changes. It's not scheduled for PR build and should not be scheduled for main branch build too.

@BorisDog we need to remove the check from PR's required checks.


BorisDog (approved):
LGTM

PR #2074, opened 2026-07-14
CSHARP-5630: Add IAsyncDisposable support to cursors

adelinowona:
Add DisposeAsync to IAsyncCursor and its implementers so await using / await foreach / async terminal operators release the server-side cursor through an async killCursors round-trip instead of blocking on the synchronous one.

- IAsyncCursor now implements IAsyncDisposable
- DisposeAsync on AsyncCursor, ChangeStreamCursor, BatchTransformingAsyncCursor, SingleBatchAsyncCursor, DeferredAsyncCursor (written leak-free), and a real AsyncCursorEnumerator.DisposeAsync
- IAsyncCursorExtensions async terminal operators use await using

### DIFF
--- a/src/MongoDB.Driver/Core/BatchTransformingAsyncCursor.cs
+++ b/src/MongoDB.Driver/Core/BatchTransformingAsyncCursor.cs
@@ -97,17 +97,23 @@ namespace MongoDB.Driver
         /// <inheritdoc/>
         public void Dispose()
         {
-            _disposed = true;
-            _current = null;
-            _wrapped.Dispose();
+            if (!_disposed)
+            {
+                _disposed = true;
+                _current = null;
+                _wrapped.Dispose();
+            }
         }
 
         /// <inheritdoc/>
         public async ValueTask DisposeAsync()
         {
-            _disposed = true;
-            _current = null;
-            await _wrapped.DisposeAsync().ConfigureAwait(false);
+            if (!_disposed)
+            {
+                _disposed = true;
+                _current = null;
+                await _wrapped.DisposeAsync().ConfigureAwait(false);
+            }
         }
 
         private void ThrowIfDisposed()
--- a/src/MongoDB.Driver/Core/BatchTransformingAsyncCursor.cs
+++ b/src/MongoDB.Driver/Core/BatchTransformingAsyncCursor.cs
@@ -16,7 +16,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Driver.Core.Misc;
@@ -101,6 +100,14 @@ namespace MongoDB.Driver
             _disposed = true;
             _current = null;
             _wrapped.Dispose();
+        }
+
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
+        {
+            _disposed = true;
+            _current = null;
+            await _wrapped.DisposeAsync().ConfigureAwait(false);
         }
 
         private void ThrowIfDisposed()
--- a/src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
+++ b/src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
@@ -117,13 +117,19 @@ namespace MongoDB.Driver
         {
             if (!_disposed)
             {
-                _disposeAction();
-                if (_cursor != null)
-                {
-                    await _cursor.DisposeAsync().ConfigureAwait(false);
-                    _cursor = null;
-                }
                 _disposed = true;
+                try
+                {
+                    _disposeAction();
+                }
+                finally
+                {
+                    if (_cursor != null)
+                    {
+                        await _cursor.DisposeAsync().ConfigureAwait(false);
+                        _cursor = null;
+                    }
+                }
             }
         }
 
--- a/src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
+++ b/src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
@@ -112,6 +112,21 @@ namespace MongoDB.Driver
             }
         }
 
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
+        {
+            if (!_disposed)
+            {
+                _disposeAction();
+                if (_cursor != null)
+                {
+                    await _cursor.DisposeAsync().ConfigureAwait(false);
+                    _cursor = null;
+                }
+                _disposed = true;
+            }
+        }
+
         private void ThrowIfDisposed()
         {
             if (_disposed)
--- a/src/MongoDB.Driver/Core/IAsyncCursor.cs
+++ b/src/MongoDB.Driver/Core/IAsyncCursor.cs
@@ -28,7 +28,7 @@ namespace MongoDB.Driver
     /// Represents an asynchronous cursor.
     /// </summary>
     /// <typeparam name="TDocument">The type of the document.</typeparam>
-    public interface IAsyncCursor<out TDocument> : IDisposable
+    public interface IAsyncCursor<out TDocument> : IDisposable, IAsyncDisposable
     {
         /// <summary>
         /// Gets the current batch of documents.
@@ -117,7 +117,7 @@ namespace MongoDB.Driver
         /// <returns>A Task whose result is true if the cursor contains any documents.</returns>
         public static async Task<bool> AnyAsync<TDocument>(this IAsyncCursor<TDocument> cursor, CancellationToken cancellationToken = default(CancellationToken))
         {
-            using (cursor)
+            await using (cursor.ConfigureAwait(false))
             {
                 var batch = await GetFirstBatchAsync(cursor, cancellationToken).ConfigureAwait(false);
                 return batch.Any();
@@ -149,7 +149,7 @@ namespace MongoDB.Driver
         /// <returns>A Task whose result is the first document.</returns>
         public static async Task<TDocument> FirstAsync<TDocument>(this IAsyncCursor<TDocument> cursor, CancellationToken cancellationToken = default(CancellationToken))
         {
-            using (cursor)
+            await using (cursor.ConfigureAwait(false))
             {
                 var batch = await GetFirstBatchAsync(cursor, cancellationToken).ConfigureAwait(false);
                 return batch.First();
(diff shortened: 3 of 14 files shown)


src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
```diff
@@ -106,6 +106,21 @@ public void Dispose()
             }
         }
 
+        /// <inheritdoc/>
```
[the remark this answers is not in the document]
adelinowona: Correct, and this is a deliberate, known divergence. The new DisposeAsync is intentionally written leak-free, while the pre-existing sync Dispose() bug (the dispose action is skipped when the cursor was never iterated) is fixed separately in CSHARP-6124 so it can be backported to the 3.x line — this PR's IAsyncDisposable addition is a 4.0-only break and can't be backported.

src/MongoDB.Driver/Core/BatchTransformingAsyncCursor.cs
```diff
@@ -103,6 +102,14 @@ public void Dispose()
             _wrapped.Dispose();
         }
 
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
+        {
+            _disposed = true;
+            _current = null;
+            await _wrapped.DisposeAsync().ConfigureAwait(false);
```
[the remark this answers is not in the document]
adelinowona: fixed

src/MongoDB.Driver/Core/IAsyncCursor.cs
```diff
@@ -28,7 +28,7 @@ namespace MongoDB.Driver
     /// Represents an asynchronous cursor.
     /// </summary>
     /// <typeparam name="TDocument">The type of the document.</typeparam>
-    public interface IAsyncCursor<out TDocument> : IDisposable
+    public interface IAsyncCursor<out TDocument> : IDisposable, IAsyncDisposable
```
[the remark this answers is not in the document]
adelinowona: Incorrect. This compiles on net472 as-is. `Microsoft.Bcl.AsyncInterfaces` (which supplies `IAsyncDisposable` for net472) is already resolved transitively — the driver has been consuming `IAsyncEnumerator<T>` and `ValueTask DisposeAsync()` from that same package on net472 for a while (e.g. `AsyncCursorEnumerator`). No .csproj change or polyfill is needed; the net472 leg of CI confirms the build.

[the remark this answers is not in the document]
adelinowona: Intentional — this ticket targets the 4.0 major release version

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
```diff
@@ -342,13 +347,28 @@ protected virtual void Dispose(bool disposing)
...
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            if (!_disposed)
+            {
+                await CloseIfNotAlreadyClosedFromDisposeAsync().ConfigureAwait(false);
+
+                _channelSource?.Dispose();
+                _disposed = true;
```
damieng: Do you want to move this up to before the await to minimize the changes this gate is being crossed more than once as per your previous PR and the other dispose methods in this PR?

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
```diff
@@ -342,13 +347,28 @@ protected virtual void Dispose(bool disposing)
...
+            }
+            GC.SuppressFinalize(this);
+        }
+
         private void CloseIfNotAlreadyClosed(CancellationToken cancellationToken)
```
damieng: You can make this accept a default cancellation token if one is not supplied so you don't have to pass one on 438.

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
```diff
@@ -342,13 +347,28 @@ protected virtual void Dispose(bool disposing)
...
             }
         }
 
+        public async ValueTask DisposeAsync()
+        {
+            if (!_disposed)
+            {
+                await CloseIfNotAlreadyClosedFromDisposeAsync().ConfigureAwait(false);
+
+                _channelSource?.Dispose();
```
damieng, line 359: Why is _channelSource null-safe called if we don't null it out after dispose?

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -73,13 +73,14 @@ public void Dispose()
...
         {
-            // TODO: implement true async disposal (CSHARP-5630)
-            Dispose();
-
-            // TODO: convert to ValueTask.CompletedTask once we stop supporting older target frameworks
-            return default; // Equivalent to ValueTask.CompletedTask which is not available on older target frameworks.
+            if (!_disposed)
+            {
+                _disposed = true;
+                _batchEnumerator?.Dispose();
```
damieng, line 82: Again, do we need the null-safe operation and if so do we want to null out our ref as well?

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
```diff
@@ -73,13 +73,14 @@ public void Dispose()
...
+                await _cursor.DisposeAsync().ConfigureAwait(false);
```
damieng, line 84: Why don't we null-safe on this one? It feels like we should be consistent on how we handle disposable resources on instance variables.

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
```diff
@@ -103,7 +99,17 @@ public void Dispose()
...
+        }
+
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
+        {
+            if (!_disposed)
+            {
+                _disposed = true;
+                await _cursor.DisposeAsync().ConfigureAwait(false);
+                _binding.Dispose();
```
damieng, line 117: As per other comments.

src/MongoDB.Driver/Core/DeferredAsyncCursor.cs
```diff
@@ -106,6 +106,21 @@ public void Dispose()
...
+        {
+            if (!_disposed)
+            {
+                _disposeAction();
+                if (_cursor != null)
+                {
+                    await _cursor.DisposeAsync().ConfigureAwait(false);
+                    _cursor = null;
+                }
+                _disposed = true;
```
damieng: Should be at the top.

src/MongoDB.Driver/Core/IAsyncCursor.cs
```diff
@@ -341,7 +341,7 @@ public static class IAsyncCursorExtensions
         /// <returns>A Task whose result is the only document of a cursor, or a default value if the cursor contains no documents.</returns>
         public static async Task<TDocument> SingleOrDefaultAsync<TDocument>(this IAsyncCursor<TDocument> cursor, CancellationToken cancellationToken = default(CancellationToken))
         {
-            using (cursor)
+            await using (cursor.ConfigureAwait(false))
```
damieng, line 344: There's a behavioural change here you may not have intended.

If cursor is null here previously the `using (cursor)` will be safe, it will just do nothing.

If cursor is null with `using (cursor.ConfigureAwait(false))` now it will throw.

Might want to safeguard this.

damieng (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
sanych-sun: We could create an `AsyncCursor` without providing `_channelSource` for the inline results (when all documents are returned together with the operation results). It makes sense to return `SingleBatchAsyncCursor` instead (and I did that in some parts of our codebase), but it might be outside of this ticket scope.

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
sanych-sun: We should use `DisposeAsync` here too,

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/AsyncCursor.cs
adelinowona: `_channelSource` does get nulled out — in `DisposeChannelSourceIfNoLongerNeeded()`, which runs during iteration and nulls both `_channelSource` and `_session` once the cursor drains (`_cursorId == 0`). So by the time `Dispose`/`DisposeAsync` runs it can already be null, hence the `?.`

adelinowona: @sanych-sun  Agreed — for inline results a `SingleBatchAsyncCursor` can be returned and it matches what `AggregateOperation/DistinctOperation/etc.` already do. As you noted it's outside this PR's scope, so I've filed CSHARP-6134 for it.

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
adelinowona: The null-safe operation is needed here: `_batchEnumerator` is genuinely nullable — it stays null until the first `MoveNext`. Nulling the ref isn't necessary since `_disposed` guards re-entry, but `MoveNext` already nulls it on exhaustion, so I can null it in `Dispose`/`DisposeAsync` too for consistency

src/MongoDB.Driver/Core/Operations/AsyncCursorEnumerator.cs
adelinowona: `_cursor` is `Ensure.IsNotNull` in the constructor, so it can never be null

src/MongoDB.Driver/Core/Operations/ChangeStreamCursor.cs
adelinowona: `_cursor` and `_binding` are checked for null in constructor so they can never be null.

src/MongoDB.Driver/Core/IAsyncCursor.cs
adelinowona: I don't think there's an actual change here. `cursor.ConfigureAwait(false)` doesn't throw on null, it just wraps the reference in a struct. So the behavior remains the same as before where it will throw on the first dereference of a null cursor inside the using block.

adelinowona: You did flag a real inconsistency though: this `Any/First/Single` family doesn't validate its cursor argument, unlike `ForEachAsync/ToList` in the same class. Pre-existing and unrelated to this PR, so I filed CSHARP-6133 to add the guards consistently.

sanych-sun (approved):
LGTM

damieng (approved):
LGTM

PR #2077, opened 2026-07-17
CSHARP-6005: CSFLE/QE support for HTTP Proxies

adelinowona:
https://jira.mongodb.org/browse/CSHARP-6005

Adds support for routing CSFLE/Queryable Encryption KMS traffic through an HTTP proxy (HTTP `CONNECT` tunnel), implementing the spec proposed in [specifications#1956](https://github.com/mongodb/specifications/pull/1956) (DRIVERS-2920).

## Public API

A new optional `IKmsConnector` can be set on `ClientEncryptionOptions` and `AutoEncryptionOptions`:

```csharp
public interface IKmsConnector
{
    Stream Connect(string host, int port, CancellationToken cancellationToken);
    Task<Stream> ConnectAsync(string host, int port, CancellationToken cancellationToken);
}
```

When supplied, the driver invokes the connector to obtain the transport stream to a KMS host (`host`/`port`) instead of opening a direct socket, then performs the KMS TLS handshake over the returned stream using the provider's configured TLS options. The typical implementation opens an `HTTP CONNECT` tunnel to a proxy and returns the tunnel stream. When no connector is set, behavior is unchanged (direct connection).

## How it works

The KMS state machine already builds `new SslStreamFactory(tlsOptions, baseStreamFactory)` and runs TLS against the KMS endpoint. This change swaps the inner `baseStreamFactory` for a connector-backed `KmsConnectorStreamFactory` when a connector is configured. Because the endpoint is unchanged, SNI and certificate/hostname verification still target the **real KMS host**, not the proxy — so the driver verifies KMS's identity end-to-end through the tunnel. This applies to both explicit encryption (`ClientEncryption`) and auto-encryption, and the existing KMS retry re-invokes the connector on transient network errors.

No libmongocrypt change is required — this is entirely driver-side.

## Why a new `IKmsConnector` rather than reusing `IStreamFactory`?

`IStreamFactory` is public and already models "create a stream to an endpoint," so it was a candidate for the callback type. We introduced a dedicated interface instead, for three reasons:

- **Contract shape.** `IStreamFactory` operates on `EndPoint`, which would force users to unpack a `DnsEndPoint` and leak transport plumbing into their callback. The KMS-connect contract is conceptually `(host, port) -> stream`; `IKmsConnector` exposes exactly that, matching the cross-driver spec's `kmsConnectCallback(host, port)` and the shapes used by the Node and C drivers.
- **Scope and coupling.** `IStreamFactory` is the general-purpose transport abstraction used throughout SDAM, connection pools, TLS, and SOCKS5. Reusing it as the KMS-proxy hook would conflate a narrow, KMS-only extension point with a broad internal-transport contract, and would couple this public API to an interface that evolves for unrelated transport reasons. A purpose-built interface keeps the public surface minimal and decoupled.
- **Intent and discoverability.** `IKmsConnector.Connect(host, port)` communicates what an implementer is expected to do; `IStreamFactory.CreateStream(EndPoint)` does not.

Internally we still reuse the existing machinery rather than reimplementing it: `KmsConnectorStreamFactory` adapts `IKmsConnector` to `IStreamFactory`, so the TLS wrapping (`SslStreamFactory`) and the KMS state machine are unchanged — we only swap the base stream source.

## Notes

- **Sync/async interface rather than a single callback.** The driver mandates paired sync/async I/O (no sync-over-async), and the KMS path has both `SendKmsRequest` and `SendKmsRequestAsync`, so the connector is a two-method interface. Consumers that only use async encryption APIs may throw from the sync method.
- **Timeout (spec Case 5) not implemented.** The C# KMS path threads only a `CancellationToken`, not a deadline; CSOT is not available. The connect-callback timeout case is therefore not supported and its prose test is skipped.

## Testing

Adds prose test 28 (`KMS Connect Callback`) — cases 1-4 and 6; case 5 (timeout) is skipped as above. The tests exercise the plain and TLS HTTP proxies, the full auto-encryption pipeline, error propagation, and retry-after-network-error. They require real AWS KMS credentials and the `kms_http_proxy.py` servers (ports 9004/9005) started by drivers-evergreen-tools, so they gate on `KMS_MOCK_SERVERS_ENABLED` and run in the mocked-KMS-TLS Evergreen variants.

### DIFF
--- a/src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
+++ b/src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
@@ -26,6 +26,7 @@ namespace MongoDB.Driver.Encryption
     {
         // private fields
         private TimeSpan? _keyExpiration;
+        private readonly IKmsConnector _kmsConnector;
         private readonly IMongoClient _keyVaultClient;
         private readonly CollectionNamespace _keyVaultNamespace;
         private readonly IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> _kmsProviders;
@@ -39,12 +40,14 @@ namespace MongoDB.Driver.Encryption
         /// <param name="keyVaultNamespace">The key vault namespace.</param>
         /// <param name="kmsProviders">The KMS providers.</param>
         /// <param name="tlsOptions">The tls options.</param>
+        /// <param name="kmsConnector">The KMS connector used to open connections to KMS hosts.</param>
         public ClientEncryptionOptions(
             IMongoClient keyVaultClient,
             CollectionNamespace keyVaultNamespace,
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> kmsProviders,
-            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default)
-            : this(keyVaultClient, keyVaultNamespace, kmsProviders, tlsOptions, keyExpiration: null)
+            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default,
+            Optional<IKmsConnector> kmsConnector = default)
+            : this(keyVaultClient, keyVaultNamespace, kmsProviders, tlsOptions, kmsConnector, keyExpiration: null)
         {
         }
 
@@ -53,12 +56,14 @@ namespace MongoDB.Driver.Encryption
             CollectionNamespace keyVaultNamespace,
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> kmsProviders,
             Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default,
+            Optional<IKmsConnector> kmsConnector = default,
             Optional<TimeSpan?> keyExpiration = default)
         {
             _keyVaultClient = Ensure.IsNotNull(keyVaultClient, nameof(keyVaultClient));
             _keyVaultNamespace = Ensure.IsNotNull(keyVaultNamespace, nameof(keyVaultNamespace));
             _kmsProviders = Ensure.IsNotNull(kmsProviders, nameof(kmsProviders));
             _tlsOptions = tlsOptions.WithDefault(new Dictionary<string, SslSettings>());
+            _kmsConnector = kmsConnector.WithDefault(null);
             _keyExpiration = keyExpiration.WithDefault(null);
 
             EnsureKmsProvidersAreValid(_kmsProviders);
@@ -71,6 +76,14 @@ namespace MongoDB.Driver.Encryption
         /// Gets the data encryption key cache expiration time.
         /// </summary>
         public TimeSpan? KeyExpiration => _keyExpiration;
+
+        /// <summary>
+        /// Gets the KMS connector used to open connections to KMS hosts.
+        /// </summary>
+        /// <value>
+        /// The KMS connector to connect directly to KMS hosts.
+        /// </value>
+        public IKmsConnector KmsConnector => _kmsConnector;
 
         /// <summary>
         /// Gets the key vault client.
@@ -111,18 +124,21 @@ namespace MongoDB.Driver.Encryption
         /// <param name="keyVaultNamespace">The key vault namespace.</param>
         /// <param name="kmsProviders">The KMS providers.</param>
         /// <param name="tlsOptions">The tls options.</param>
+        /// <param name="kmsConnector">The KMS connector used to open connections to KMS hosts.</param>
         /// <returns>A new ClientEncryptionOptions instance.</returns>
         public ClientEncryptionOptions With(
             Optional<IMongoClient> keyVaultClient = default,
             Optional<CollectionNamespace> keyVaultNamespace = default,
             Optional<IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>>> kmsProviders = default,
-            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default)
+            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default,
+            Optional<IKmsConnector> kmsConnector = default)
         {
             return new ClientEncryptionOptions(
                 keyVaultClient: keyVaultClient.WithDefault(_keyVaultClient),
                 keyVaultNamespace: keyVaultNamespace.WithDefault(_keyVaultNamespace),
                 kmsProviders: kmsProviders.WithDefault(_kmsProviders),
                 tlsOptions: Optional.Create(tlsOptions.WithDefault(_tlsOptions)),
+                kmsConnector: Optional.Create(kmsConnector.WithDefault(_kmsConnector)),
                 keyExpiration: _keyExpiration);
         }
 
--- a/src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
+++ b/src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
@@ -81,7 +81,7 @@ namespace MongoDB.Driver.Encryption
         /// Gets the KMS connector used to open connections to KMS hosts.
         /// </summary>
         /// <value>
-        /// The KMS connector to connect directly to KMS hosts.
+        /// The KMS connector used to open connections to KMS hosts.
         /// </value>
         public IKmsConnector KmsConnector => _kmsConnector;
 
--- a/src/MongoDB.Driver.Encryption/KmsConnectorStreamFactory.cs
+++ b/src/MongoDB.Driver.Encryption/KmsConnectorStreamFactory.cs
@@ -1,4 +1,4 @@
-/* Copyright 2019-present MongoDB Inc.
+/* Copyright 2010-present MongoDB Inc.
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


src/MongoDB.Driver/Encryption/IKmsConnector.cs
```diff
@@ -0,0 +1,48 @@
...
+public interface IKmsConnector
+{
+    /// <summary>
+    /// Opens a connection to the specified KMS host.
+    /// </summary>
+    /// <param name="host">The KMS hostname (for example, <c>kms.us-east-1.amazonaws.com</c>).</param>
+    /// <param name="port">The KMS port (typically 443).</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A stream connected to the KMS host. The driver wraps this stream in TLS.</returns>
+    Stream Connect(string host, int port, CancellationToken cancellationToken);
```
sanych-sun: Should we use Endpoint instead of host and port?

src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -390,5 +396,33 @@ public async Task<Stream> CreateStreamAsync(EndPoint endPoint, CancellationToken
             // private methods
             private static Socket CreateSocket() => new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
         }
+
+        private class KmsConnectorStreamFactory : IStreamFactory
```
sanych-sun: Can we place this class into a separate file?

src/MongoDB.Driver/Encryption/IKmsConnector.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace MongoDB.Driver.Encryption;
+
+/// <summary>
+/// Opens the transport connection used to reach a KMS host. When supplied via
+/// ClientEncryptionOptions or <see cref="AutoEncryptionOptions"/>, the driver
+/// invokes this instead of opening a direct TCP connection to the KMS host, then wraps the
+/// returned stream in TLS using the KMS provider's configured TLS options.
+/// The primary use case is routing KMS traffic through an HTTP proxy via HTTPS CONNECT.
+/// </summary>
+public interface IKmsConnector
```
sanych-sun, line 35: Why do we need this new abstraction? `IStreamFactory` looks very similar and it's public too. Can we let users provide the stream factory and do not introduce a new interface?

sanych-sun (changes requested):

src/MongoDB.Driver/Encryption/IKmsConnector.cs
adelinowona: Discussed offline but I added some reasoning in the PR description.

src/MongoDB.Driver/Encryption/IKmsConnector.cs
adelinowona: Discussed offline as well but for visibility: 
I'd keep host/port. `EndPoint` looks tidier but the base class doesn't actually expose Host/Port — they only exist on `DnsEndPoint`, so a `Connect(EndPoint, ...)` signature forces every user to downcast and just know it's always a `DnsEndPoint`. host/port is self-documenting, needs no cast, and maps 1:1 to the spec's callback.

src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
```diff
@@ -39,12 +40,14 @@ public sealed class ClientEncryptionOptions
...
+        /// <param name="kmsConnector">The KMS connector used to open connections to KMS hosts.</param>
         public ClientEncryptionOptions(
             IMongoClient keyVaultClient,
             CollectionNamespace keyVaultNamespace,
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, object>> kmsProviders,
-            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default)
-            : this(keyVaultClient, keyVaultNamespace, kmsProviders, tlsOptions, keyExpiration: null)
+            Optional<IReadOnlyDictionary<string, SslSettings>> tlsOptions = default,
+            Optional<IKmsConnector> kmsConnector = default)
+            : this(keyVaultClient, keyVaultNamespace, kmsProviders, tlsOptions, kmsConnector, keyExpiration: null)
```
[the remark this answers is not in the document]
BorisDog: @adelinowona I guess we'll address this during backporting?

src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
```diff
@@ -199,6 +202,9 @@ private static DnsEndPoint CreateKmsEndPoint(string value)
             return new DnsEndPoint(host, port);
         }
 
+        private IStreamFactory GetBaseStreamFactory() =>
```
BorisDog: It doesn't seem that `_networkStreamFactory` is used independently from `GetBaseStreamFactory`. Why don't we just set `_networkStreamFactory  = new KmsConnectorStreamFactory(_kmsConnector) `in ctor?
This also helps avoids allocation per request.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -4049,6 +4192,302 @@ private class Patient
...
+            }
+
+            public async Task<Stream> ConnectAsync(string host, int port, CancellationToken cancellationToken)
+            {
+                if (Interlocked.Increment(ref _invocationCount) == 1)
+                {
+                    throw new IOException("Simulated transient network error.");
+                }
+
+                return await _inner.ConnectAsync(host, port, cancellationToken).ConfigureAwait(false);
```
BorisDog: nit: no need for `configureawait` in tests.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -4049,6 +4192,302 @@ private class Patient
...
+            var handler = new HttpClientHandler
+            {
+                ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => ValidateAgainstCa(cert, caCertificate)
+            };
+            return new HttpClient(handler);
+        }
+
+        private static X509Certificate2 LoadProxyCaCertificateOrSkip()
+        {
+            var caFile = Environment.GetEnvironmentVariable("CSFLE_TLS_CA_FILE");
```
BorisDog, line 4261: Where is CSFLE_TLS_CA_FILE set?

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
```diff
@@ -909,6 +910,148 @@ void TestCase(ClientEncryption testCaseClientEncryption, BsonDocument masterKey,
...
+        // CONNECT tunnel (verified against the real KMS host). Success confirms the KMS host - not the
+        // proxy - was verified.
+        // https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#case-2-https-proxy
+        [Theory]
+        [ParameterAttributeData]
+        public void KmsConnectCallback_via_https_proxy([Values(false, true)] bool async)
+        {
+            RequireEnvironment.Check().KmsProvider("aws");
+            RequireEnvironment.Check().EnvironmentVariable("KMS_MOCK_SERVERS_ENABLED", isDefined: true);
+            var caCertificate = LoadProxyCaCertificateOrSkip();
```
BorisDog: using?

src/MongoDB.Driver/Encryption/IKmsConnector.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MongoDB.Driver.Encryption;
+
+/// <summary>
+/// Opens the transport connection used to reach a KMS host. When supplied via
+/// ClientEncryptionOptions or <see cref="AutoEncryptionOptions"/>, the driver
```
BorisDog: <c>ClientEncryptionOptions</c> ?

BorisDog (changes requested):

src/MongoDB.Driver.Encryption/ClientEncryptionOptions.cs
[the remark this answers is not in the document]
adelinowona: yep

src/MongoDB.Driver.Encryption/LibMongoCryptControllerBase.cs
adelinowona: Thanks for the catch! fixed it.

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
adelinowona: This is set by the driver-evergreen-tools scripts that run in the test variants CSFLE tests

tests/MongoDB.Driver.Tests/Specifications/client-side-encryption/prose-tests/ClientEncryptionProseTests.cs
adelinowona: Done

src/MongoDB.Driver/Encryption/IKmsConnector.cs
adelinowona: ClientEncryptionOptions type lives in the encryption package which isn't referenced by the driver package so the see cref won't resolve. I am assuming that's your question here.

BorisDog: Sorry, I meant \<c>ClientEncryptionOptions \<c>

src/MongoDB.Driver.Encryption/KmsConnectorStreamFactory.cs
```diff
@@ -0,0 +1,53 @@
...
+    public KmsConnectorStreamFactory(IKmsConnector kmsConnector)
+    {
+        _kmsConnector = Ensure.IsNotNull(kmsConnector, nameof(kmsConnector));
+    }
+
+    public Stream CreateStream(EndPoint endPoint, CancellationToken cancellationToken)
+    {
+        var (host, port) = GetHostAndPort(endPoint);
+        var stream = _kmsConnector.Connect(host, port, cancellationToken);
+        return Ensure.IsNotNull(stream, $"{nameof(IKmsConnector)}.{nameof(IKmsConnector.Connect)}");
```
BorisDog: minor: Consider a clearer message, like  "connector returned null" or similar.

src/MongoDB.Driver/AutoEncryptionOptions.cs
```diff
@@ -137,6 +142,14 @@ private AutoEncryptionOptions(
         /// </remarks>
         public IReadOnlyDictionary<string, object> ExtraOptions => _extraOptions;
 
+        /// <summary>
+        /// Gets the KMS connector used to open connections to KMS hosts.
+        /// </summary>
+        /// <value>
+        /// The KMS connector to connect directly to KMS hosts.
```
BorisDog: I think technically it's not "directly to KMS hosts"?

BorisDog (approved):
LGTM + minor comments

src/MongoDB.Driver/AutoEncryptionOptions.cs
adelinowona: Fixed

src/MongoDB.Driver.Encryption/KmsConnectorStreamFactory.cs
adelinowona: fixed

sanych-sun (approved):
LGTM

PR #2078, opened 2026-07-20
CSHARP-5994: Remove ReplaceOptions and UpdateOptions in favor of ReplaceOptions<T> and UpdateOptions<T>

sanych-sun:


src/MongoDB.Driver/ReplaceOptions.cs
```diff
@@ -44,7 +44,7 @@ internal static ReplaceOptions From(UpdateOptions updateOptions)
                     throw new ArgumentException("ArrayFilters cannot be used with ReplaceOne.", nameof(updateOptions));
                 }
 
-                return new ReplaceOptions
+                return new ReplaceOptions<T>
```
adelinowona: `UpdateOptions<T>` exposes `Sort`, but it is not being copied in the constructor. 

Also the `Comment` option is not copied but that's pre-existing

src/MongoDB.Driver/MongoCollectionBase.cs
adelinowona: this type check is redundant now. Similar dead type checks in other replaceOne methods in this file

adelinowona, line 1: Could we also just remove the Obsolete ReplaceOne overloads for 4.0?

src/MongoDB.Driver/MongoCollectionBase.cs
```diff
@@ -677,22 +677,22 @@ private async Task<ReplaceOneResult> ReplaceOneAsync(FilterDefinition<TDocument>
...
         }
 
-        private UpdateResult UpdateMany(FilterDefinition<TDocument> filter, UpdateDefinition<TDocument> update, UpdateOptions options, Func<IEnumerable<WriteModel<TDocument>>, BulkWriteOptions, BulkWriteResult<TDocument>> bulkWrite)
+        private UpdateResult UpdateMany(FilterDefinition<TDocument> filter, UpdateDefinition<TDocument> update, UpdateOptions<TDocument> options, Func<IEnumerable<WriteModel<TDocument>>, BulkWriteOptions, BulkWriteResult<TDocument>> bulkWrite)
         {
             Ensure.IsNotNull(filter, nameof(filter));
             Ensure.IsNotNull(update, nameof(update));
 
-            options = options ?? new UpdateOptions();
+            options ??= new UpdateOptions<TDocument>();
```
adelinowona, line 667: `UpdateMany` has no use for a sort but `UpdateOptions<T>` exposes a sort for users to set. `UpdateMany` will ignore it but this seems like a design smell on the API. Arguably acceptable as the cost of consolidating onto one type but just wanted to flag it for visibility if we should do something about it.

adelinowona (changes requested):

src/MongoDB.Driver/ReplaceOptions.cs
sanych-sun: Good catch! We had this bug before this changes, will fix.

src/MongoDB.Driver/MongoCollectionBase.cs
sanych-sun: Fixed

src/MongoDB.Driver/ReplaceOptions.cs
sanych-sun: OK, this code was removed together with obsolete APIs.

src/MongoDB.Driver/MongoCollectionBase.cs
sanych-sun: Done.

src/MongoDB.Driver/MongoCollectionBase.cs
sanych-sun: Good observation. Added validation for UpdateMany(Async) method.

adelinowona (approved):

PR #2085, opened 2026-07-28
CSHARP-5998: RawBsonDocument refactoring/removal

sanych-sun:
Part 2 of the RawBsonDocument deprecation work: remove RawBsonDocument from ChangeStreams.


BorisDog (approved):
LGTM, just need to double check that PostBatchResumeToken  is always present.

PR #2086, opened 2026-07-30
CSHARP-6147: Add a dotnet format check

papafe:
https://jira.mongodb.org/browse/CSHARP-6147

We also looked at running the check as a commit/push git hook, so problems would be caught before CI. Unfortunately `dotnet format` is quite slow on projects of this size, so it is not included here.


sanych-sun (approved):
LGTM

PR #2087, opened 2026-07-30
CSHARP-5998: RawBsonDocument refactoring/removal

sanych-sun:
RawBsonDocument refactoring/removal - part 3 out of 4.

Although this change materializes each filter/command document up front via BsonSerializer.Deserialize<BsonDocument>, it does not add any materialization that wasn't already happening: because these documents were embedded as nested values inside a command, they were serialized to the wire through BsonValueSerializer → BsonDocumentSerializer, which dispatches by BsonType and walks elements one at a time — the zero-copy RawBsonDocument fast path (WriteRawBsonDocument) was never reached. In fact RawBsonDocument caches nothing and re-parses its byte buffer from the start on every GetElement/ElementCount call, so the old code re-parsed the slice repeatedly during serialization; the new code parses once into a reusable tree. The result is functionally identical BSON with slightly less work, and it removes the IDisposable/unmanaged-buffer lifetime concern as a bonus.

### DIFF
--- a/src/MongoDB.Driver.Encryption/AutoEncryptionLibMongoController.cs
+++ b/src/MongoDB.Driver.Encryption/AutoEncryptionLibMongoController.cs
@@ -19,6 +19,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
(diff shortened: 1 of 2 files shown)


BorisDog (approved):
LGTM

PR #2089, opened 2026-07-31
CSHARP-6117: Consolidate target frameworks to net472, net6.0, net8.0 and net10.0

ajcvickers:
Replace netstandard2.1;net472;net6.0 with net472;net6.0;net8.0;net10.0. Drops netstandard2.1, whose only remaining unique consumers were out of support (.NET Core 3.1, .NET 5, Xamarin, standalone Mono) plus current Mono/IL2CPP Unity, and adds first-class net8.0 and net10.0 assets. net6.0 is retained.

Test projects move to net6.0;net8.0;net10.0 (+net472 on Windows), dropping the netcoreapp3.1 target that existed only to exercise the netstandard2.1 asset. The SmokeTests consumer targets netcoreapp3.1 and net5.0 are removed, as they can no longer resolve a compatible asset.

Polyfill package references are scoped to net472; all NETSTANDARD2_1 branches are removed and the remaining conditionals normalised to NET6_0_OR_GREATER and NET472, so the three modern assets expose an identical public surface.

Fixes three latent issues surfaced by the new targets:
- BsonClassMap used FormatterServices.GetUninitializedObject (obsolete as SYSLIB0050 from .NET 8), now RuntimeHelpers.GetUninitializedObject.
- ScramSha1/ScramSha256 used Rfc2898DeriveBytes constructors (obsolete as SYSLIB0060 from .NET 10), now the static Rfc2898DeriveBytes.Pbkdf2.
- The X509CertificateLoader test shim was guarded !NET8_0_OR_GREATER, but the BCL type arrived in .NET 9; corrected to !NET9_0_OR_GREATER.

Also gives net10.0 the full integration matrix in Evergreen, and fixes the gcp KMS test case, which an #if NETCOREAPP3_0 that no target framework defines had silently disabled.

NOTE: the test-analyzer task now passes TARGET_FRAMEWORK=net8.0 and requires a matching change in mongodb/mongo-csharp-analyzer before it can pass.


tests/MongoDB.Driver.Encryption.Tests/BasicTests.cs
```diff
@@ -418,7 +418,10 @@ public void TestLocalKeyCreationStepwise()
         [Theory]
         [InlineData("aws")]
         [InlineData("azure")]
-#if NETCOREAPP3_0
+        // GCP requires RSA signing, which SigningRSAESPKCSCallback does not support on .NET Framework
+        // (it throws PlatformNotSupportedException there). Was previously guarded by an #if NETCOREAPP3_0
+        // that no target framework ever defined, so this case never ran at all.
```
sanych-sun: Could you please remove "Was previously guarded by an #if NETCOREAPP3_0 that no target framework ever defined, so this case never ran at all." part of the comment? Looks like it's an intermediate comment that slipped into the final PR.

sanych-sun (approved):
LGTM + minor changes requested. Also please populate the related Jira ticket with updated list of TFMs.

PR #2091, opened 2026-07-31
CSHARP-5996: Remove obsolete APIs

adelinowona:
## Summary

Removes the driver's accumulated `[Obsolete]` public surface ahead of 4.0. Each commit covers one coherent group and is independently reviewable — going commit-by-commit will be much easier than reading the combined diff.

## Notes

**1. GeoHaystack is the one removal that drops working functionality.** Everything else here removes either dead code or a member with a live replacement. GeoHaystack is different: the driver's minimum server is 4.4, geoHaystack was deprecated in 4.4 and removed by the server in 5.0. So it still works against the oldest server we support and fails on everything newer. I think it's ok to remove now.

**2. `count` and `mapReduce` are now skipped in both spec-test factories, with a reason.** The suites drive `count` in 44 places. Every one of those paths is also covered by `countDocuments` or `estimatedDocumentCount`, and the spec's own test names call them "Deprecated count" — so I believe that no unique conformance coverage is lost.

**3. In `LibmongocryptTests`.** A SERVER-106469 workaround gated on `ServerDescription.Version`, which this PR removes. Rather than port it the whole `catch` is deleted: the ticket is Closed/Fixed in server 8.2.1, the smoke matrix runs `5.0/6.0/7.0/8.0/latest`, and left in place the block would swallow any `MongoEncryptionException` on Windows *and* skip the `expectedLogs` assertions below it — turning a genuine failure green.

**4. `GridFSFileInfo`'s `"IdAsBsonValue"` string literal stays** even though the property is gone. It's the registered serializer member name that the live `Id` property reads `_id` through.

## Deliberately not removed

21 `[Obsolete]` members remain in `src/`, all intentional:

- **The `waitQueueSize` / `waitQueueMultiple` family (14 members)** — split out to CSHARP-6153. These aren't dead: they parse from the connection string and read through to `ExclusiveConnectionPool._waitQueueFreeSlots`, which throws `MongoWaitQueueFullException` when exhausted. Removal is a behaviour decision, plus there's a coupling to untangle — the obsolete `MongoClientSettings.WaitQueueSize` also feeds the *non*-obsolete `ClusterSettings.MaxServerSelectionWaitQueueSize`.
- **Hedged reads (3 members)** — Deprecated by server 8.0 but still rendered into `$readPreference` and honoured by 4.4–7.x, i.e. most of the supported range. That makes the case for removal *weaker* than GeoHaystack's, which this PR does remove.
- **`BsonBinarySubType.OldBinary`, `MqlHashAlgorithm.MD5`, `DelegatingStream.CreateWaitHandle`, `OperationContext.CombinedCancellationToken`** — wire-format compatibility, deliberate legacy support, a BCL-obsoleted override, and an `internal` member that bridges to public interfaces taking a bare `CancellationToken`.


src/MongoDB.Driver/IAggregateFluent.cs
```diff
@@ -524,18 +524,6 @@ IAggregateFluent<TResult> UnionWith<TWith>(
...
-        /// <summary>
-        /// Appends an unwind stage to the pipeline.
-        /// </summary>
-        /// <typeparam name="TNewResult">The type of the result of the stage.</typeparam>
-        /// <param name="field">The field.</param>
-        /// <param name="newResultSerializer">The new result serializer.</param>
-        /// <returns>
-        /// The fluent aggregate interface.
-        /// </returns>
-        [Obsolete("Use the Unwind overload which takes an options parameter.")]
```
papafe, line 536: `AggregateFluentBase` still has this  `Unwind` overload. Should we remove it?

papafe (approved):
Looks good!

src/MongoDB.Driver/Core/Operations/MapReduceOperation.cs
```diff
@@ -1,149 +0,0 @@
-/* Copyright 2010-present MongoDB Inc.
```
sanych-sun, line 1: MapReduce operation was deprecated by the server a while ago, but server still support it. I'm not sure if other drivers keep this functionality or removed. I'm OK with removal, but probably worth to understand what other drivers did, and it deserves separate line in release notes/breaking changes list (for all top level operations that were removed).

src/MongoDB.Driver/Core/Servers/ServerType.cs
```diff
@@ -48,12 +48,6 @@ public enum ServerType
         /// </summary>
         ReplicaSetSecondary,
 
-        /// <summary>
-        /// Use ReplicaSetSecondary instead.
-        /// </summary>
-        [Obsolete("Passives are treated the same as secondaries.")]
-        ReplicaSetPassive,
```
sanych-sun, line 55: We should be careful with enum members removal, because the underlying values for all next enum members will be changed. It could be a really hard breaking change if it was persisted as int. Should we explicitly specify int values for all members, so the removal will not affect them?

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestOperationFactory.cs
```diff
@@ -90,7 +91,7 @@ public IUnifiedTestOperation CreateOperation(string operationName, string target
                 {
                     "aggregate" => new UnifiedAggregateOperationBuilder(_entityMap).BuildCollectionOperation(targetEntityId, operationArguments),
                     "bulkWrite" => new UnifiedBulkWriteOperationBuilder(_entityMap).Build(targetEntityId, operationArguments),
-                    "count" => new UnifiedCountOperationBuilder(_entityMap).Build(targetEntityId, operationArguments),
+                    "count" => throw new SkipException(".NET/C# driver does not implement a Count helper; use CountDocuments or EstimatedDocumentCount."),
```
sanych-sun: I would not silently skip here, because if somebody will use such operation in the unified test we will simply skip the test. As example, some tests were using `findOne` as the "dummy" operation (basically to run any server interaction), which we do not support. As the result such tests failing loudly and we can investigate and adjust tests if needed.

Let's just remove this case so we will throw `FormatException`, if there is any tests that test this exact operation(s), let's skip them by test case/file name.

sanych-sun (changes requested):

src/MongoDB.Driver/Core/Operations/MapReduceOperation.cs
BorisDog: @adelinowona we can extract the open issues to a separate PR, to unblock the whole process.

src/MongoDB.Driver/Core/Servers/ServerType.cs
BorisDog: I agree, we should try to keep enum numeric values unchanged.
Maybe it's easier, and more clearer not remove the value at all, just leave it obsolete?

src/MongoDB.Driver/IAggregateFluent.cs
adelinowona: yep done

src/MongoDB.Driver/Core/Operations/MapReduceOperation.cs
adelinowona: Some drivers have it removed and some have it deprecated. I would rather just remove it so we can just make sure it is mentioned in the release notes

src/MongoDB.Driver/Core/Servers/ServerType.cs
adelinowona: Fair, I think we can we just leave it obsolete then

tests/MongoDB.Driver.Tests/UnifiedTestOperations/UnifiedTestOperationFactory.cs
adelinowona: Fair, updated to skip by test case/name

src/MongoDB.Driver/Core/Operations/MapReduceOperation.cs
sanych-sun: I'm OK with removal, as far as it's aligned with what other drivers.

src/MongoDB.Driver/Core/Servers/ServerType.cs
sanych-sun: Is there any API that accept this value as parameter? Should we treat it the same way as ReplicaSetSecondary?

sanych-sun (approved):
LGTM

src/MongoDB.Driver/Core/Operations/MapReduceOperation.cs
adelinowona: Im actually going to revert this change for now since the MapReduce was deprecated in 5.0 and the driver minimum supported version is still at 4.4. We can reconsider removing in the next major version.

src/MongoDB.Driver/Core/Servers/ServerType.cs
adelinowona: Any api that takes a `ServerType` will accept this value as a parameter. Two places:

`ServerTypeExtensions` methods but those methods call `ToClusterType` which never had `ReplicaSetPassive` as one of the cases, so it always threw an exception. We could make `ReplicaSetPassive` behave the same as `ReplicaSetSecondary` instead of throwing just for cleanup (behavior breaking change). @BorisDog thoughts? no particular preference from me.

`ServerDescription` class also takes a `ServerType` but that's a type mostly used internally for SDAM. Sure users can hand build a server description but I see no production reason for such a use case.

BorisDog: I think for clarity we should throw on `ReplicaSetPassive`, as it's basically not supported and shouldn't be used.

BorisDog (approved):
LGTM

PR #2094, opened 2026-08-06
CSHARP-6005: CSFLE/QE support for HTTP Proxies

adelinowona:
CSHARP-6005: CSFLE/QE support for HTTP Proxies

Backport of #2077 to `v3.x`.

patch: https://spruce.corp.mongodb.com/version/6a74aca41d612400077c27e3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

The internal machinery is identical to the `main` PR: `IKmsConnector` / `KmsConnectionContext` public types, an internal `KmsConnectorStreamFactory` that `LibMongoCryptControllerBase` uses in place of `NetworkStreamFactory` when a connector is supplied, and prose test 28 (cases 1, 2, 3, 4, 6).

## Difference from the `main` PR: how the option is exposed

On `main`, `kmsConnector` is a new optional parameter on the public `ClientEncryptionOptions` / `AutoEncryptionOptions` constructors and on their `With` methods. Adding an optional parameter changes the public signature, which is a binary break, so that isn't available on `v3.x`.

Instead this PR follows the existing `SetKeyExpiration` precedent on these same two classes — a mutating setter added for exactly this reason:

```csharp
var options = new ClientEncryptionOptions(keyVaultClient, keyVaultNamespace, kmsProviders);
options.SetKmsConnector(myConnector);
```

So the public surface changes here are purely additive:

- new `IKmsConnector` and `KmsConnectionContext` types
- `ClientEncryptionOptions.KmsConnector` property + `ClientEncryptionOptions.SetKmsConnector(IKmsConnector)`
- `AutoEncryptionOptions.KmsConnector` property + `AutoEncryptionOptions.SetKmsConnector(IKmsConnector)`

No existing constructor, `With` overload, or property signature is touched. `With` propagates the configured connector, and `AutoEncryptionOptions.Equals` / `GetHashCode` account for it (by reference, as they do for `KeyVaultClient`).

## Other deltas

- Test helpers use `Uri` rather than `string` overloads of `HttpClient.PostAsync` / `GetStringAsync`; the `v3.x` test ruleset treats CA2234 as an error.
- The `IKmsConnector` doc comment points at the setters rather than the constructor parameters.

### DIFF
--- a/src/MongoDB.Driver.Encryption/AutoEncryptionLibMongoController.cs
+++ b/src/MongoDB.Driver.Encryption/AutoEncryptionLibMongoController.cs
@@ -65,7 +65,7 @@ namespace MongoDB.Driver.Encryption
             IMongoClient metadataClient,
             CryptClient cryptClient,
             AutoEncryptionOptions autoEncryptionOptions)
-            : base(cryptClient, keyVaultClient, autoEncryptionOptions.KeyVaultNamespace, autoEncryptionOptions.KmsProviders, autoEncryptionOptions.TlsOptions)
+            : base(cryptClient, keyVaultClient, autoEncryptionOptions.KeyVaultNamespace, autoEncryptionOptions.KmsProviders, autoEncryptionOptions.TlsOptions, autoEncryptionOptions.KmsConnector)
         {
             _internalClient = internalClient; // can be null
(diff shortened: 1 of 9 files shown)


sanych-sun (approved):
LGTM

PR #2096, opened 2026-08-11
CSHARP-6189: Make the license headers, encoding and line endings uniform and enforce them in CI

papafe:
https://jira.mongodb.org/browse/CSHARP-6189

This normalizes every header to a single banner with a uniform 2010 year, updates the Rider file templates to match, and adds a CI check so they cannot drift apart again. It also sets `charset = utf-8` and `end_of_line = lf` for `.cs` files and normalizes the tree to match, so encoding and line endings are enforced by the `dotnet format` check we already run.

Nothing changes beyond license headers, the BOM and line endings. `src/MongoDB.Driver/Authentication/SaslPrepHelper.cs` is excluded, as Apache-2.0 section 4 requires retaining its dual-license attribution.


BorisDog (commented):
Look good overall.
`dotnet format --verify-no-changes` fails on windows because of the line endings.
We might want to switch to LF endings on windows as well, and to add something `*.cs text eol=lf` to  `.gitattributes`. But let's discuss.

PR #2097, opened 2026-08-11
CSHARP-6133: Validate cursor argument in IAsyncCursor/IAsyncCursorSource Extensions terminal operators

adelinowona:


src/MongoDB.Driver/Core/IAsyncCursor.cs
```diff
@@ -212,12 +224,13 @@ public static class IAsyncCursorExtensions
         public static async Task ForEachAsync<TDocument>(this IAsyncCursor<TDocument> source, Func<TDocument, int, Task> processor, CancellationToken cancellationToken = default(CancellationToken))
         {
             Ensure.IsNotNull(source, nameof(source));
-            Ensure.IsNotNull(processor, nameof(processor));
 
             // yes, we are taking ownership... assumption being that they've
             // exhausted the thing and don't need it anymore.
             await using (source.ConfigureAwait(false))
             {
+                Ensure.IsNotNull(processor, nameof(processor));
```
sanych-sun, line 233: Please add explanation why this validation is done not at the very beginning of the method. I got it, because we want to make sure the `source` will get disposed, but it's better to have comment so future us do not "standardize" the code.

sanych-sun (approved):
LGTM + minor comment

PR #2109, opened 2026-09-01
CSHARP-6214: Implement srvAllowedHostsSuffix URI option

adelinowona:
Implements the `srvAllowedHostsSuffix` URI option and the validation that bounds how broad the configured suffix may be.

Spec: [Initial DNS Seedlist Discovery](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md). Vendored spec tests are in sync with the merged spec.

### The option

`srvAllowedHostsSuffix` lets a user state the parent domain that hosts returned by an SRV lookup are validated against, instead of having it inferred from the SRV hostname. The value is normalized in spec order — strip leading/trailing dots, convert to Punycode, ASCII lowercase, apply the label rule, prepend a dot — and is rejected on a non-SRV scheme, mirroring `srvServiceName`.

`DnsMonitor` previously used `lookupDomainName` both to build the SRV query name and to validate returned hosts. Those are now separate: the option must not affect which records are queried, only which parent domain returned hosts are checked against.

### Validation

The spec bounds how broad the suffix may be, since the option relaxes a DNS-spoofing safeguard. This implements the label-count branch of the spec's rule 4: at least two `.` separated labels, or a single label naming a private or special-use namespace — `test`, `localhost`, `invalid`, `example` (RFC 6761), `local` (RFC 6762), `internal` (ICANN-reserved), `corp`, `home`, `mail`.

Worth stating plainly: **this does not catch multi-label public suffixes.** `co.uk`, `github.io` and `com.au` are all accepted. That is the known cost of the label-count branch. Users wanting PSL-grade validation can layer it on through the `srvHostValidator` callback in [DRIVERS-3632](https://jira.mongodb.org/browse/DRIVERS-3632).

### Also here: a pre-existing `DnsMonitor` bug

`DnsMonitor` required the SRV hostname to have at least three `.` separated labels, while `ConnectionString` accepts fewer and has tests saying it should. So `mongodb+srv://mongo.local/` parsed and then threw `ArgumentException` at cluster initialization.

Not introduced by this feature, but the single-label rule above makes it reachable through a configuration the spec now says MUST be accepted: `mongodb+srv://mongo.corp/?srvAllowedHostsSuffix=corp` is exactly the split-horizon case the carve-out serves, and it could not start. Fixing it separately would have meant shipping a supported option value into a path that throws.


src/MongoDB.Driver/Core/Configuration/ClusterSettings.cs
```diff
@@ -273,7 +285,8 @@ public ClusterSettings With(
             Optional<IServerSelector> postServerSelector = default(Optional<IServerSelector>),
             Optional<ConnectionStringScheme> scheme = default(Optional<ConnectionStringScheme>),
             Optional<int> srvMaxHosts = default,
-            Optional<string> srvServiceName = default(Optional<string>))
+            Optional<string> srvServiceName = default(Optional<string>),
+            Optional<string> srvAllowedHostsSuffix = default(Optional<string>))
```
sanych-sun, line 289: Isn't adding new optional parameter considered to be a breaking change?

src/MongoDB.Driver/Core/Misc/PublicSuffixList.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Resources;
+using System.Text;
+using System.Threading;
+
+namespace MongoDB.Driver.Core.Misc;
+
+internal static class PublicSuffixList
```
sanych-sun: Have you considered usage of some 3rd party package to handle the public suffix list? The problem with the embedded list that there is no way to populate it other then change the file and make a new version.
Should we use https://github.com/nager/Nager.PublicSuffix instead? This package can download the latest official list from [publicsuffix.org](https://publicsuffix.org/)

src/MongoDB.Driver/MongoUrlBuilder.cs
```diff
@@ -1091,6 +1111,10 @@ public override string ToString()
             {
                 query.AppendFormat("srvServiceName={0}&", _srvServiceName);
             }
+            if (_srvAllowedHostsSuffix != null)
+            {
+                query.AppendFormat("srvAllowedHostsSuffix={0}&", _srvAllowedHostsSuffix);
```
sanych-sun: Should we encode `_srvAllowedHostsSuffix`?

src/MongoDB.Driver/Core/Configuration/ClusterSettings.cs
adelinowona: this is for the main branch where this is fine (4.0 work), I will change it for the 3.x branch PR

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -1260,6 +1304,88 @@ private void ParseOption(string name, string value)
...
+        {
+            normalizedSuffix = null;
+            errorMessage = null;
+
+            // IdnMapping rejects an empty label, so the dots have to come off before the Punycode
+            // step rather than after it
+            var suffix = value.Trim('.');
+            if (suffix.Length == 0)
+            {
+                errorMessage = "srvAllowedHostsSuffix must name at least one domain label.";
```
sanych-sun, line 1329: This error message is misleading: error message on line 1348 asks for 2 labels. Should we use the same message here too?

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -1481,17 +1616,27 @@ private void ValidateResolvedHosts(string original, List<string> resolved)
                 var dnsEndPoint = (DnsEndPoint)endPoint;
 
                 var host = ((DnsEndPoint)endPoint).Host;
```
sanych-sun, line 1618: It looks like unused variable

sanych-sun (commented):
LGTM + minor suggestions (Also as far as I understood the callback will be implemented in another PR, right?)

sanych-sun (approved):

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
adelinowona: done

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
adelinowona: thanks for the catch!

sanych-sun (approved):
LGTM

src/MongoDB.Driver/Core/Configuration/ConnectionString.cs
```diff
@@ -1376,16 +1371,49 @@ internal static bool TryNormalizeHostName(string host, out string normalizedHost
...
-            catch (ArgumentException)
+            catch (ArgumentException exception)
             {
-                return false;
+                errorDetail = exception.Message;
+                return null;
             }
 
-            return true;
+            // .NET Core rejects a label that starts with "xn--" and does not decode as Punycode,
```
sanych-sun, line 1396: Should we hide this code under `#if net472`?

adelinowona: I don't think it's necessary since if netcore implementation fails then we just return from the method early and we don't run this code.

PR #2112, opened 2026-09-01
CSHARP-6213: Normalize SRV target host names to lower case during SRV polling

adelinowona:
## Problem

During SRV polling, `DnsMonitor` compared the host names returned by the resolver against the parent domain case-sensitively. DNS names are case-insensitive ([RFC 4343](https://www.ietf.org/rfc/rfc4343.txt)), and SDAM requires that [hostnames are normalized to lower-case](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md#hostnames-are-normalized-to-lower-case).

The initial seedlist path and the polling path reach the same comparison by different routes, and only one of them folded case:

- **Initial discovery** — `ConnectionString.ValidateResolvedHosts` parses each resolved host through `EndPointHelper.TryParse`, which lower-cases. The seed side was lower-cased by the same call during connection string parsing. Both operands are folded, so the ordinal comparison in `HasValidParentDomain` is correct.
- **Polling** — `DnsMonitor.GetValidEndPoints` took `srvRecord.EndPoint` straight from the resolver and stripped the trailing dot by hand, never going through `EndPointHelper`. `_lookupDomainName` *was* lower-cased (it comes from `ClusterSettings.EndPoints`), so a raw, possibly mixed-case target was compared ordinally against a lower-cased anchor.

### DIFF
--- a/src/MongoDB.Driver/Core/Clusters/DnsMonitor.cs
+++ b/src/MongoDB.Driver/Core/Clusters/DnsMonitor.cs
@@ -133,14 +133,14 @@ namespace MongoDB.Driver.Core.Clusters
 
             foreach (var srvRecord in srvRecords)
             {
-                var endPoint = srvRecord.EndPoint;
-                var host = endPoint.Host;
+                // DNS names are case-insensitive, and SDAM requires host names to be normalized to lower-case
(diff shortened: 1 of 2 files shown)


sanych-sun (approved):
LGTM

PR #2115, opened 2026-09-02
CSHARP-6213: Normalize SRV target host names to lower case during SRV polling

adelinowona:
Backport of #2112 to `v3.x`.


sanych-sun (approved):
LGTM

PR #2116, opened 2026-09-02
CSHARP-6234: Fix typos in documentation

sstockdev:
When working with MongoDB.Driver I noticed multiple typos in the comments. This merge request exists to fix those typos.


papafe (changes requested):
@sstockdev Thanks for opening the PR. The files coming under the specification folder are actually coming from another repo, so it would not be useful to fix them here. The fix in the other two files looks good though.

sstockdev:
Thank you for your input @papafe! I have removed the changes under the specification folder.

papafe (approved):

PR #2124, opened 2026-09-12
CSHARP-6220: Prevent cancellation-token retention in WaitAsync

mohammadshahsavari:
## Summary

- Prevent the pre-.NET 6 `TaskExtensions.WaitAsync` implementation from retaining registrations on long-lived caller cancellation tokens.
- Cancel and dispose a linked timeout cancellation source after `Task.WhenAny` completes.
- Add regression coverage for both generic and non-generic tasks.

## Testing

- `dotnet build CSharpDriver.sln`
- `dotnet test tests/MongoDB.Driver.Tests/MongoDB.Driver.Tests.csproj -f net472 --filter "FullyQualifiedName~TaskExtensionsTests"`

JIRA: https://jira.mongodb.org/browse/CSHARP-6220


src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
```diff
@@ -66,8 +66,18 @@ static async Task WaitAsyncCore(Task task, TimeSpan timeout, CancellationToken c
             {
                 if (!task.IsCompleted)
                 {
-                    var timeoutTask = Task.Delay(timeout, cancellationToken);
-                    await Task.WhenAny(task, timeoutTask).ConfigureAwait(false);
+                    using (var timeoutCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
```
sanych-sun, line 69: Should we check `cancellationToken.CanBeCanceled` before creating the `CancellationTokenSource` to optimize the case when the CancellationToken.None is being used?

mohammadshahsavari: Thanks, good point. I updated both overloads to use a regular CTS when the token can’t be canceled.

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
```diff
@@ -66,8 +66,20 @@ static async Task WaitAsyncCore(Task task, TimeSpan timeout, CancellationToken c
             {
                 if (!task.IsCompleted)
                 {
-                    var timeoutTask = Task.Delay(timeout, cancellationToken);
-                    await Task.WhenAny(task, timeoutTask).ConfigureAwait(false);
+                    using (var timeoutCancellationTokenSource = cancellationToken.CanBeCanceled ?
```
papafe, line 69: Do we need to have a special case from non-cancelable token sources?

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
```diff
@@ -66,8 +66,20 @@ static async Task WaitAsyncCore(Task task, TimeSpan timeout, CancellationToken c
...
+                        CancellationTokenSource.CreateLinkedTokenSource(cancellationToken) :
+                        new CancellationTokenSource())
+                    {
+                        var timeoutTask = Task.Delay(timeout, timeoutCancellationTokenSource.Token);
+                        try
```
papafe: I don't think we should have the try-finally here. `Task.WhenAny` does not throw when the tasks themselves throw (this is why we have the pre-existing checks afterwards).

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
sanych-sun: Do we really need the `CancellationTokenSource` at all in case when `CancellationToken` is not cancellable? As far as i understood for non-cancellable tokens registration for callback is no-op, so it will not hold a reference to anything. I suppose we can save on creating a `CancellationTokenSource` in such cases.

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
mohammadshahsavari: The previous special case created a regular CTS, which wasn’t necessary. I’ve updated it so non-cancelable tokens don’t create a CTS at all. the conditional now only avoids that unnecessary allocation.

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
mohammadshahsavari: I’ve removed the try-finally and cancel the timeout task directly after Task.WhenAny.

src/MongoDB.Driver/Core/Misc/TaskExtensions.cs
mohammadshahsavari: You’re right. The registration issue only applies to cancellable tokens so I’ve removed the CTS allocation for non-cancellable tokens.

PR #2128, opened 2026-09-17
CSHARP-6223: Fix client-side projection crash on member-init projections

damieng:
- Port of CSHARP-6223 (#2123) from `main` to `v3.x`


sanych-sun (approved):
LGTM

PR #2129, opened 2026-09-22
CSHARP-5030: Allow TCP send/receive buffer sizes to be left at the OS default

ChadNedzlek:
**Fixes:** CSHARP-5030

**tl;dr:** validated against a real `mongod` with `tc netem` injecting real-world cross-region RTTs — reading the same collection, this change measures **19–25x higher throughput**, and the benefit grows with RTT rather than shrinking. Full curve below.

## Problem

`TcpStreamSettings.ReceiveBufferSize` / `SendBufferSize` default to 64 KiB and are unconditionally passed to `Socket.ReceiveBufferSize` / `SendBufferSize`. On Linux this calls `setsockopt(SO_RCVBUF/SO_SNDBUF)`, which permanently disables the kernel's receive-window autotuning for that socket (see `tcp(7)`). On a low-latency connection this is invisible. On a higher-latency connection — cross-region, for example — it becomes a hard ceiling: for a fixed window, `throughput = window / RTT`, independent of the link's actual capacity.

Small historical note, for context: 65536 is one byte past 65535, the largest window the original 16-bit TCP window field could express before [RFC 1323](https://www.rfc-editor.org/rfc/rfc1323) ("TCP Extensions for High Performance", May 1992) introduced window scaling — motivated by exactly this problem, large bandwidth-delay-product paths needing a bigger window than a fixed small ceiling allows.

The ticket description already identifies the mechanism from a different angle, comparing OS default buffer sizes across platforms (Windows 65536, macOS 408300, Ubuntu 131072). What follows adds the RTT dimension: how large the effect gets as latency grows, and a fix that avoids the breaking-change concern raised in [Ferdinando's comment](https://jira.mongodb.org/browse/CSHARP-5030?focusedCommentId=6911336#comment-6911336):

> The easy fix for this would be to change the nullability of `TcpStreamSettings.ReceiveBufferSize` and `TcpStreamSettings.SendBufferSize`, but it would be a breaking change, so we should do that for next major.

## Fix

Adds `TcpStreamSettings.OperatingSystemDefaultBufferSize` (`-1`) as an explicit sentinel for `ReceiveBufferSize`/`SendBufferSize`. When specified, the socket option is left unset entirely, so the OS default — including Linux's autotuning — stays in effect. This matches the Java driver's `SocketSettings`, which defaults to the OS rather than a fixed value.

`0` and every non-positive value other than `-1` continue to be rejected by the existing validation, so no currently-valid caller's behavior changes, and both properties stay non-nullable `int`. The nullable-property version can still happen in a future major; this unblocks the fix without waiting for it.

## Measured impact

Reproduced against a real `mongod` in Docker, reading a ~46 MB / 180,000-document collection, with `tc netem` adding latency on the container's own interface. Each configuration reads against a server whose cache was pre-warmed on the full collection, so neither number benefits from running second. RTTs beyond same-host are real, sourced figures: 4.71 ms is an RTT measured in production between an AKS pod and an in-region Atlas replica; the rest are Azure's own [published P50s](https://learn.microsoft.com/en-us/azure/networking/azure-network-latency) between real region pairs.

| RTT | route | default (64 KiB pinned) | `OperatingSystemDefaultBufferSize` | ratio |
|---|---|---|---|---|
| ~0 ms | same host | ~845 Mbps | ~845 Mbps | no measurable difference |
| 4.71 ms | realistic in-region | 110.5 Mbps | 724.3 Mbps | 6.6x |
| 68 ms | East US 2 ↔ West US 2 | 7.8 Mbps | 169.0 Mbps | 21.7x |
| 89 ms | East US 2 ↔ West Europe | 5.9 Mbps | 140.6 Mbps | 23.7x |
| 228 ms | East US 2 ↔ Southeast Asia | 2.3 Mbps | 57.9 Mbps | 25.4x |

The same-host row is a genuine tie — checked in both run orders, since it's small enough to be sensitive to measurement noise, and neither configuration wins. Everywhere a real network sits between client and server, the ratio is large and it grows with RTT rather than shrinking, including at realistic in-region distance, not just cross-continent. Backing out the effective window from each row (`throughput × RTT ÷ 8`) shows the *default* case holding nearly flat at 65–70 KB across the whole range — exactly what a fixed window predicts — while the *fix*'s effective window grows with RTT (1.3 MB → 1.7 MB), consistent with autotuning actually adapting rather than just being a bigger constant.

## Tests

- `TcpStreamSettingsTests`: constructor accepts `OperatingSystemDefaultBufferSize` for both buffer settings without throwing; the existing negative/zero validation tests now use `-2` instead of `-1`, since `-1` is a valid sentinel.
- `TcpStreamFactoryTests`: `ConfigureConnectedSocket` sets the requested buffer sizes when given explicit values, and leaves the socket at its OS-default buffer sizes when given `OperatingSystemDefaultBufferSize`. Full existing suite, including the server-dependent tests, passes against a live `mongod`.

Open to bikeshedding `OperatingSystemDefaultBufferSize` if there's a naming convention elsewhere in the driver it should match instead.

## Recommendation for the next major

This PR is scoped to what's safe in a minor: additive, opt-in, zero behavior change for existing callers. For whenever a breaking change is on the table, though, I'd suggest going further than making the opt-out available — flip the *default* to `OperatingSystemDefaultBufferSize`, and deprecate `ReceiveBufferSize`/`SendBufferSize` as settable properties.

This isn't just an artifact of the RTTs I happened to test — it's structural. Linux's own autotuning starts at 131072 bytes (`tcp_rmem`'s default entry), already double the driver's 64 KiB pin, before it ever grows in response to the connection's actual bandwidth-delay product. So autotuning can't do worse than the fixed value once any real network sits between client and server: it either matches it or beats it, and only ever beats it by more as RTT or bandwidth grows. The measured table confirms exactly that shape — same-host is a statistical tie (no network in the loop for either config's window to bind against), and every case with a real network between client and server shows the fixed buffer losing, by a margin that gets worse the further apart the two ends are.

It's not just the growth direction, either. `setsockopt(SO_RCVBUF)` sets `SOCK_RCVBUF_LOCK`, which disables `tcp_moderate_rcvbuf` for that socket entirely — the kernel can't shrink the buffer under memory pressure any more than it can grow it under a large bandwidth-delay product. (This is separate from congestion control, which still adapts to loss normally regardless of `SO_RCVBUF` — pinning only removes the kernel's control over buffer *size*, not `cwnd`.) Real-world receive-buffer memory problems do happen — see [Cloudflare's writeup](https://blog.cloudflare.com/unbounded-memory-usage-by-tcp-for-receive-buffers-and-how-we-fixed-it/) on an autotuning-related memory bug they hit — but even there, the fix was correcting the kernel's own moderation logic, not pinning the buffer. Surrendering adaptivity in both directions isn't the standard fix for a growth-direction problem.

In every scenario I measured, and every one I'm aware of, pinning the buffer to a static value is either a wash or a real loss — never a win. That makes the properties themselves worth reconsidering: they invite a caller to reintroduce this exact regression by setting an explicit number, for an upside I haven't been able to identify.

I recognize that's a bigger ask than this PR, and it's not blocking anything here — just flagging it now so it's on the table for whenever the next major is being scoped.


papafe:
Thanks a lot for your PR, but we decided to fix this in next major (more details in JIRA ticket).

